PR #16, opened 2020-09-17
Allow to reset job.queue to None

pmrv:
This sets the run_mode to modal, not sure if that's the best default.
I also haven't tested this yet on the cluster, but it seems straightforward enough.

### DIFF
--- a/pyiron_base/server/generic.py
+++ b/pyiron_base/server/generic.py
@@ -162,6 +162,10 @@ class Server(
         Args:
             new_scheduler (str): scheduler name
         """
+        if new_scheduler == None:
+            self._active_queue = None
+            self._run_mode.mode = "modal"
+            return
         if s.queue_adapter is not None:
             cores, run_time_max, memory_max = s.queue_adapter.check_queue_parameters(
                 queue=new_scheduler,
--- a/pyiron_base/server/generic.py
+++ b/pyiron_base/server/generic.py
@@ -169,7 +169,7 @@ class Server(
             self.run_mode.modal = True
             self.cores = 1
             self.threads = 1
-            self.run_time = None
+            self._run_time = None
             self.memory_limit = None
         else:
             if s.queue_adapter is not None:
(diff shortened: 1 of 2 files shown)


jan-janssen:
Please also reset runtime, cores and memory.

pyiron_base/server/generic.py
```diff
@@ -162,6 +162,10 @@ def queue(self, new_scheduler):
         Args:
             new_scheduler (str): scheduler name
         """
+        if new_scheduler == None:
```
jan-janssen: use is None rather than == None 
https://stackoverflow.com/a/3257957/6996746

pyiron_base/server/generic.py
```diff
@@ -162,6 +162,10 @@ def queue(self, new_scheduler):
...
+            self._active_queue = None
+            self._run_mode.mode = "modal"
```
jan-janssen: I guess we should use the same API as below: 
```
self.run_mode = "modal"
```

jan-janssen: Or even better `self.run_mode.modal = True` to prevent the user from mistyping.

pyiron_base/server/generic.py
```diff
@@ -162,6 +162,10 @@ def queue(self, new_scheduler):
...
+            return
```
jan-janssen: I guess we can just use `if else` here to make it a bit more readable.

liamhuber:
This is looking nice, thanks for taking care of this Marvin!

jan-janssen (approved):
Looks good - still it would be nice to add a test

pmrv:
Is there a queue adapter defined in the settings on the test environment?

jan-janssen:
Not yet, but I assume this should also work if no queue was defined previously, just validating that `None` is now accepted and does not raise an error.

pmrv:
Force push to rebase on master. I had to use `server._run_time`, instead of `server.run_time` to side step an issue there and I didn't want to relax the typecheck in the setter as it's very useful. Also ran the tests on the cluster with defined queue adapter  and everything works.

PR #20, opened 2020-09-22
Add list_nodes/list_groups to InputList

pmrv:
Adds generic methods to distinguish groups and nodes in the list as we just discussed in the meeting.  I defined groups as nested `InputList`s only not other containers otherwise it might get confusing.


liamhuber:
Naive question: don't we still want InputList to inherit from the PyironObject template?

pmrv:
Mid-term I think yes, but that object also wants database ids and that's something we have to think more about how to do.

PR #26, opened 2020-09-30
Implement Python template job 

jan-janssen:


pmrv:
Think a separate class for python jobs makes a lot of sense.  But if we're already subclassing we might as just override `_check_if_input_should_be_written()` to always return `False` in `PythonTemplateJob`.  That way the base class doesn't need to know about python jobs and its `_check_if_input_should_be_written()` method is also easier to read.  As far as I can tell this is the only method where we use `_python_only_job`.

jan-janssen:
I added the `_check_if_input_should_be_written()` function, but there are quite a few derived classes which use the `_python_only_job` flag: 

- Hessian https://github.com/pyiron/pyiron/blob/master/pyiron/thermodynamics/hessian.py#L24
- SQS https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/job/sqs.py#L57 
- ASE based jobs https://github.com/pyiron/pyiron/master/pyiron/gpaw/pyiron_ase.py#L50

pmrv:
Shouldn't we update those to derive from PythonTemplateJob, though?  Or is the TemplateJob doing anything we don't need there?

EDIT: Ah, I see because those are AtomisticGenericJobs, we'd need some multiple inheritance stuff.  I could look into that, but maybe we discuss on Friday first.

jan-janssen:
The class hierarchy is a bit more complex: 

* HessianJob -> GenericInteractive -> (AtomisticGenericJob, InteractiveBase)
* SQSJob -> AtomisticGenericJob
* AtomisticGenericJob -> GenericJob
* AseJob -> InteractiveBase -> GenericJob
* PythonTemplateJob -> TemplateJob -> GenericJob

Therefore I currently see the PythonTemplateJob primarily for experimenting with experimental analysis and I do not recommend deriving the atomistic classes from those as atomistic jobs like LAMMPS support both Python Bindings and Communication via input and output files.

PR #28, opened 2020-09-30
Changed wait_for_jobs() to ignore non pyiron jobs that are queued

Leimeroth:
The lambda function lambda x: int(x["jobname"].split("pi_")[1]) lead to errors if non pyiron jobs were in the queue,

### DIFF
--- a/pyiron_base/server/queuestatus.py
+++ b/pyiron_base/server/queuestatus.py
@@ -234,6 +234,7 @@ def update_from_remote(project, recursive=True):
         df_submitted = df_project[df_project.status == "submitted"]
         df_combined = df_project[df_project.status.isin(["running", "submitted"])]
         df_queue = s.queue_adapter.get_status_of_my_jobs()
+        df_queue = df_queue[df_queue.jobname.str.startswith("pi_")]
         if len(df_queue) > 0:
             df_queue["pyiron_id"] = df_queue.apply(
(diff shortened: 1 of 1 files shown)


pyiron_base/server/queuestatus.py
```diff
@@ -234,6 +234,7 @@ def update_from_remote(project, recursive=True):
         df_submitted = df_project[df_project.status == "submitted"]
         df_combined = df_project[df_project.status.isin(["running", "submitted"])]
         df_queue = s.queue_adapter.get_status_of_my_jobs()
+        df_queue = df_queue[df_queue.jobname.str.startswith("pi_")]
```
jan-janssen: I do not think this works with an empty data frame - if no jobs are listed in the queue.

pyiron_base/server/queuestatus.py
```diff
@@ -234,6 +234,7 @@ def update_from_remote(project, recursive=True):
...
         if len(df_queue) > 0:
```
jan-janssen:
```suggestion
        if len(df_queue) > 0 and len(df_queue[df_queue.jobname.str.startswith("pi_")]) > 0:
```

pyiron_base/server/queuestatus.py
jan-janssen:
```suggestion
```

jan-janssen (approved):
Looks good to me

PR #34, opened 2020-10-01
Use InputList in ScriptJob + Backwards compatibility

pmrv:
Tests for compatibility starting from version 0.1.19


pmrv:
I removed one test case that looked at the groups in HDF5 inside a GenericParameters/InputList.  Ideally this would be a test on InputList, but I haven't got there yet.

PR #37, opened 2020-10-02
Add friendly error message when setting queue w/o adapter

pmrv:
The error was already thrown in the correct place, but I added a small sentence to hint where to fix it.

### DIFF
--- a/pyiron_base/server/generic.py
+++ b/pyiron_base/server/generic.py
@@ -197,7 +197,8 @@ class Server(
                 self._active_queue = new_scheduler
                 self.run_mode = "queue"
             else:
-                raise TypeError("No queue adapter defined.")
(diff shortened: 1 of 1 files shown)


pyiron_base/server/generic.py
```diff
@@ -522,4 +523,4 @@ def _init_host(host):
         if host is None:
             return socket.gethostname()
         else:
-            return host
+            return hos
```
jan-janssen:
```suggestion
            return host
```

jan-janssen (approved):
Looks good to me

pyiron_base/server/generic.py
pmrv: Whoops, no tolerance for tiny typos.

PR #43, opened 2020-10-08
Fixes problem with non pyiron jobs in queue

Leimeroth:
Seems like I actually had no other jobs in the queue when I tried the last time. Applyin the lambda function still lead to a non existent list element for non pyiron jobs.


jan-janssen (approved):
Looks good to me

PR #46, opened 2020-10-09
Remove python 2.7 support and add python 3.8

jan-janssen:


pmrv:
After this is merged we can also remove the dependency on `six`, quick grepping seems to imply we only use it very sparingly anyway.

Issue #44, opened 2020-10-08
data_dict not found


samwaseda:
The following job doesn't work (anymore). It's certainly related to the change from `GenericParameters` to `InputList`, but I don't know how it should look now.

submitter.ipynb:
```
from pyiron import Project

pr = Project('JOB')

job = pr.create_job('ScriptJob', 'job')
job.input['value'] = 300
job.script_path = 'job.ipynb'
job.run()
```

job.ipynb:
```
from pyiron import Notebook as nb
coeff_tot = nb.get_custom_dict()
print(coeff_tot['value'])`
```

pmrv:
The problem is here
https://github.com/pyiron/pyiron_base/blob/909b815f7cd4b7a952be64367ac1c93037ee1049/pyiron_base/job/external.py#L43
where the Notebook class parses the job hdf manually.  Since the HDF changed with InputList this fails.  Is this `Notebook.get_custom_dict()` supposed to work for ScriptJobs only or is this a more general mechanism?

I am very reluctant to allow any other function poke around in the HDF internals of jobs, as I consider those implementation details.  Maybe I could add a similar method to the ScriptJob.

samwaseda:
> The problem is here
> 
> https://github.com/pyiron/pyiron_base/blob/909b815f7cd4b7a952be64367ac1c93037ee1049/pyiron_base/job/external.py#L43
> 
> 
> where the Notebook class parses the job hdf manually. Since the HDF changed with InputList this fails. Is this `Notebook.get_custom_dict()` supposed to work for ScriptJobs only or is this a more general mechanism?
> I am very reluctant to allow any other function poke around in the HDF internals of jobs, as I consider those implementation details. Maybe I could add a similar method to the ScriptJob.

I'm fine with anything. It's just I can't launch simulations right now because of this, which is causing a lot of trouble to me.

jan-janssen:
Thx @samwaseda for fixing this, we had unit tests to check for this but they have been removed in the last pull request https://github.com/pyiron/pyiron_base/pull/34 so I guess we still need to improve the reviewing of the pull requests.

PR #47, opened 2020-10-09
Add tests for external script jobs

pmrv:
This also uses the generic ProjectHDFio.to_object() machinery to load
the input dictionary from the ScriptJob hdf without depending on the
internals.

This should make sure #44 doesn't happen again.

### DIFF
--- a/tests/job/test_external.py
+++ b/tests/job/test_external.py
@@ -12,9 +12,7 @@ print(coeff_tot['value'])
 class TestScriptJob(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        cls.file_location = os.path.dirname(os.path.abspath(__file__)).replace(
-            "\\", "/"
-        )
+        cls.file_location = os.path.dirname(os.path.abspath(__file__))
         cls.job_location = os.path.join(cls.file_location, "job.py")
         cls.project = Project(os.path.join(cls.file_location, "test_notebook"))
         cls.job = cls.project.create_job(cls.project.job_type.ScriptJob,
--- a/tests/job/test_external.py
+++ b/tests/job/test_external.py
@@ -1,0 +1,40 @@
+import os
+import unittest
+
+from pyiron_base import Project
+
+job_py_source = """
+from pyiron_base import Notebook as nb
+coeff_tot = nb.get_custom_dict()
+print(coeff_tot['value'])
+"""
+
+class TestScriptJob(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.file_location = os.path.dirname(os.path.abspath(__file__)).replace(
+            "\\", "/"
+        )
+        cls.job_location = os.path.join(cls.file_location, "job.py")
+        cls.project = Project(os.path.join(cls.file_location, "test_notebook"))
+        cls.job = cls.project.create_job(cls.project.job_type.ScriptJob,
+                                               "test")
+        with open(cls.job_location, 'w') as f:
+            f.write(job_py_source)
+
+    @classmethod
+    def tearDownClass(cls):
+        os.remove(cls.job_location)
+        cls.project.remove(enable=True)
+
+
+    def test_notebook_input(self):
+        """
+        Test that input is readable from external scripts.
+        """
+        self.job.input['value'] = 300
+        self.job.script_path = self.job_location
+        try:
+            self.job.run()
+        except Exception as e:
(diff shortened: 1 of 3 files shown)


samwaseda:
How can I switch to `pmrv:notebook`? `git checkout pmrv:notebook` didn't work...

tests/job/test_external.py
```diff
@@ -0,0 +1,40 @@
...
+coeff_tot = nb.get_custom_dict()
+print(coeff_tot['value'])
+"""
+
+class TestScriptJob(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.file_location = os.path.dirname(os.path.abspath(__file__)).replace(
+            "\\", "/"
+        )
```
samwaseda: Why do you need `replace`?

pmrv:
I pulled this one from my [fork](https://github.com/pmrv/pyiron_base/tree/notebook), so you'd have to add `git@github.com:pmrv/pyiron_base.git` to your git remotes.  I'll push this branch to this repo too in a minute.

pmrv:
Done, you can check it out from [here](https://github.com/pyiron/pyiron_base/tree/notebook) now.

tests/job/test_external.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_notebook_input(self):
+        """
+        Test that input is readable from external scripts.
+        """
+        self.job.input['value'] = 300
+        self.job.script_path = self.job_location
+        try:
+            self.job.run()
+        except Exception as e:
+            self.fail("Running script job failed with {}".format(e))
```
samwaseda, line 38: I guess we don't really need `try` and `except` in unit tests

tests/job/test_scriptjob.py
```diff
@@ -0,0 +1,30 @@
...
+    def test_notebook_input(self):
+        """
+        Makes sure that the ScriptJob saves its input class in
+        hdf["input/custom_group"] as this is needed when running external
+        Notebook jobs c.f. `Notebook.get_custom_dict()`.
+        """
+        self.job.input['value'] = 300
+        self.job.save()
+        self.assertTrue("custom_dict" in self.job["input"].list_groups(),
+                        "Input not saved in the 'custom_dict' group in HDF")
```
samwaseda, line 30: I guess we should also check whether `value` is 300

tests/job/test_external.py
pmrv: I copied this from [GenericJob](https://github.com/pyiron/pyiron_base/blob/master/tests/job/test_genericJob.py), my guess is it related to handling of windows paths.

tests/job/test_scriptjob.py
pmrv: I'd prefer a test that checks that the external script running reads the correct value from the `get_custom_dict` function.  Can the script write to HDF somewhere so that the calling code can read it?

tests/job/test_external.py
samwaseda: Isn't it also the reason why appveyor is failing?

tests/job/test_external.py
pmrv: Raised exceptions and test failure are treated/shown differently by `unittest`, so I'd rather keep it this way.  I can check for a more specific exception to catch though.

tests/job/test_external.py
pmrv: Could be, will remove it.  I thought that it was needed when calling `Project`, @jan-janssen do you know if it is necessary?

tests/job/test_external.py
samwaseda: ok

tests/job/test_scriptjob.py
samwaseda: Then write instead of `print(coeff_tot['value'])`:

```
if coeff_tot['value'] != 300:
    raise AssertionError('value not passed correctly')
```

in `job.py`

tests/job/test_external.py
samwaseda: `test_scriptjob.py` has the same issue

samwaseda: Ah but wait I see that `test_external.py` is already failing. So it was probably not the reason.

jan-janssen: The replace part is used because pyiron internally also replaces windows path separation `\\` with `/`. So if you want to compare the path then it is useful to keep it, that is why I added it as default to most tests.

PR #49, opened 2020-10-12
Update external.py

prince-mathews:
`'input/custom_dict/data'` does not exist and gives an `"Unknown item: data"` error for script jobs. I found the input data to be stored in `'/input/custom_dict/data_dict'`


prince-mathews:
Nevermind, I just saw that there's another thread dealing with the problem already.

PR #50, opened 2020-10-13
Added output based on Inputlist and use save_output to hide hdf5 usage

niklassiemer:


notebooks/pythonjobtemplate.ipynb
```diff
@@ -59,7 +58,8 @@
     "job = pr.create_job(job_type=ToyJob, job_name=\"toy\")\n",
     "print(job.input)\n",
     "job.run()\n",
-    "job['output/generic/energy_tot']"
+    "#job['output/generic/energy_tot']\n",
+    "job[\"output/data\"][\"energy_tot\"]"
```
jan-janssen, line 63: The output should be accessible using: `job["output/generic/energy_tot"]` as that is the unique path we commonly use, in particular it should not be separated in two parts like `job["output/data"]["energy_tot"]`

jan-janssen (changes requested):
We should extend the input list to support the pyiron path like `job["output/generic/energy_tot"]` it is like a URL to offer direct access to the individual data objects directly from the HDF5 file.

niklassiemer:
Thanks for having a look already!
Yes, I really would like to use the usual `job["output/generic/energy_tot"]` nomenclature. 
Also I will look into unittests to also test the stuff I added.

pyiron_base/job/template.py
```diff
@@ -24,39 +24,57 @@
...
+        self._input.to_hdf(hdf=hdf, group_name=None)
 
     def from_hdf(self, hdf=None, group_name=None):
         super().from_hdf(
             hdf=hdf,
             group_name=group_name
         )
-        with self.project_hdf5.open("input") as h5in:
-            self.input.from_hdf(h5in)
+        if hdf is None:
```
jan-janssen: This part should be handled in the `super()` call.

pyiron_base/job/template.py
```diff
@@ -24,39 +24,57 @@
...
+        self._input = InputList(input, table_name="input")
 
     def to_hdf(self, hdf=None, group_name=None):
         super().to_hdf(
             hdf=hdf,
             group_name=group_name
         )
-        with self.project_hdf5.open("input") as h5in:
-            self.input.to_hdf(h5in)
+        if hdf is None:
```
jan-janssen: This part should be handled in the `super()` call.

pyiron_base/job/template.py
niklassiemer: So, I should be able to just use `self._hdf5` , since in the super call, this is set to hdf if `hdf is not None`. Therefore, it has to be impossible to call this when `self._hdf` is not set - is this guaranteed to be always be the case?

pyiron_base/job/template.py
niklassiemer: Here, I should also be able to just use `self._hdf5` then.

niklassiemer: In the PythonTemplateJob class, the same lines are for sure redundant - I will take them out

pyiron_base/job/template.py
jan-janssen: Yes, you should be fine using just `self._hdf5`.

pyiron_base/job/template.py
jan-janssen: Yes, you should be fine using just `self._hdf5`.

pmrv:
Just a drive-by comment without reading the details yet.  How is `save_output` different from `save()` and `to_hdf`?  If I am a new user/dev and I want to make sure my output is saved, I search the API docs I'm bound to find at least these two plus this new one.  I think we also had examples at some point where this was done directly in the `run()` method.  Even for me it's not really clear right now where this should be done best.

I'm not against per se, just want to make sure we keep our API surface minimal.

niklassiemer:
The idea of `save_output` is just to hide the hdf5 file access and the internal self.status business. We could also rename it to 
`finalize_run()` . The `save` actually saves an initialized new job and sets `status.created = True` and `to_hdf` should be a more internal command hidden to users and to devs wanting to run a python application with pyiron_base.

jan-janssen:
Another suggestion would be to call: 
```
super().run_static()
```
At the end of the user defined run static function, then we can handle the setting of the status and the storing to hdf there. Or we could implement a `run_static` decorator, which adds this functionality automatically. I personally like the concept of decorators and the flexibility it offers but I also see that it increases the complexity: https://github.com/pyiron/pyiron/pull/815

pmrv:
> The idea of `save_output` is just to hide the hdf5 file access and the internal self.status business. We could also rename it to
> `finalize_run()` . The `save` actually saves an initialized new job and sets `status.created = True` and `to_hdf` should be a more internal command hidden to users and to devs wanting to run a python application with pyiron_base.

I just feel it doesn't really simplify anything from a dev perspective if I have call one method (save_output) instead of another one (to_hdf).  If we want to make default implementations easier, I'd instead suggest we have the implementation of `PythonTemplateJob.to_hdf` automatically call `self.output.to_hdf`.  Then all I need to worry about as a developer is shoveling my stuff from `self.input` to my code to `self.output` and the template job properly abstracts from all the infrastructure around it.

pyiron_base/job/template.py
```diff
@@ -23,29 +24,62 @@
 class TemplateJob(GenericJob):
     def __init__(self, project, job_name):
         super().__init__(project, job_name)
-        self.input = GenericParameters(table_name="input")
+        self._input = InputList(table_name="input")
```
liamhuber, line 27: I like this. Also the associated infrastructure of the property and setter, and making sure this field is automatically taken care of in `to_hdf` and `from_hdf`. I'm similarly happy to see output treated this way. 

What I would really like to see is for all of this to be taking place in `GenericJob`. Other than the fact we'd need to do the leg work of updating all the downstream classes, is there a reason not to do this? I feel `GenericJob` should by default have `input` and `output` fields, and this is the most up-to-date way of handling these I've seen.

pyiron_base/job/template.py
```diff
@@ -23,29 +24,62 @@
...
+        self._output.from_hdf(hdf=self._hdf5, group_name=None)
+
+    def to_hdf(self, hdf=None, group_name=None):
+        super().to_hdf(
+            hdf=hdf,
+            group_name=group_name
+        )
+        self._output.to_hdf(hdf=self._hdf5, group_name=None)
+
+    def save_output(self):
```
liamhuber: I really share Marvin's concern about this.

What is the use case? Who is calling this when?

Overall I am a bit concerned about the job flow. Like, is `run_static` ever called by the end user, or from another class as opposed to internally by the job itself? It should, perhaps, be private? 

And if there's stuff that should *always* be happening after "user defined" `run_static` definitions, is it not also the case that it should always be happening, it would be nice to tuck this away elsewhere in the flow so that both users and developers only have to define a `run_static` without worrying about book keeping.

These issues might be beyond the scope of this PR, but it really looks to me like we start pushing up against these issues now.

liamhuber (commented):
I like it. I want changes, but most of them stem from the fact that I like these changes so much that I want them farther upstream!!!

pyiron_base/job/template.py
niklassiemer: Yes, a generic way of handling input and output would be great! Furthermore, I am not even sure things would break immediately if these objects are given by default on the `GenericJob level`?

liamhuber: Do you have all the repos and test environments nicely set up on your machine? I would suggest simply trying pushing the change upstream to `GenericJob`, and running the tests to see whether you're right about that!

It's 100% a direction I want to go in, but if it's going to break things then I understand if we don't go there in this PR.

pyiron_base/job/template.py
niklassiemer: AFAIK `run_static` is never called by the user. It is always handled by the `run()` call, which calls all the `_run_if...` functions depending on the job status. This is the reason we have to specify  `self.status.finished = True` in the `run_static` method...
Actually we could also set `self.status.collect = True` in order to trigger  `_run_if_collect`  where we could handle the `_output.to_hdf`. However, we would be left with this `self.status.collect = True` poping up 'randomly' for a new developer.

pyiron_base/job/template.py
niklassiemer: I will test if my test suite is up to date and try to break thinks! However, I am not sure how rigorous the tests are checking what is actually written to the hdf file. With the table_name="input" we would write stuff to the same group, where we usually write to. This might play well, if the fields are orthogonal, but does not have to.

pyiron_base/job/template.py
liamhuber: Yeah, I'm wondering if we should have something like

```python

def run_static(self):
    raise NotImplementedError

def _run_static(self):
    self.run_static()
    self.status.collect = True

def collect_output(self):
    pass

def _collect_output(self):
    self.collect_output()
    self.status.finished = True
```

As I write this, I immediately see why @jan-janssen was talking elsewhere about introducing a decorator... like

```python
@status_method('collect')
def calc_static:
    raise NotImplementedError
```

I share his concern about decorator complexity. Namely, I worry about forcing new devs to remember to add the decorator when they're doing the definition. But maybe it's just a fair cognitive load to add for the beauty we get in return.

Finally, if you're right that `run_static` and `collect_output` are never called from outside their class, then we should really make them private. Right now they crowd the tab completion list on job objects. This is a general problem we need to work on.

pyiron_base/job/template.py
niklassiemer: Just a quick update at this point: Tests are failing at least for the output field if I just shift these changes upstream. I could of course rename this like io_output and io_input to avoid name clashes with usually used fields.
Furthermore, I have to update and enhance my test environment. The unittests will work at some point, however, I probably have to run the notebooks on the cluster - for some reason the Laptop runs out of memory - maybe due to the windows subsystem setup....

pyiron_base/job/template.py
niklassiemer: I just realized, we have to call `self.run()` in the `_run_static` method to trigger the `collect_output()` method. I think it would be nicer to leave the `collect_output` method for file I/O where needed and to handle everything in `run_static()` as it is done right now.

jan-janssen: It is the other way around, the `run()` method calls `collect_output` if the status is set to collect: 
https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L697
More precisely it calls `_run_if_collect()`: 
https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L1483
Which then calls the collect output.

liamhuber: > I just realized, we have to call `self.run()` in the `_run_static` method to trigger the `collect_output()` method.

@niklassiemer yes, I missed re-calling `run`. This is actually kind of my point; my understanding of the current setup is that the simulation should always flow through a series of steps controlled by its status, but as it stands right now developers still need to actively manage this flow in each subclass of jobs. The code is functional as is, so I don't want to come down too hard on it, I just feel like there is an opportunity here to really make strong gains in improving maintainability and extensibility by restructuring this flow process.

>  I think it would be nicer to leave the `collect_output` method for file I/O where needed and to handle everything in `run_static()` as it is done right now.

I still disagree here; if we have a `collect` phase and a `collect_output` method, I find it very strange to do data collection and post-processing in the `run` phase. It also shows up here a little bit that we have a whole extra method `save_output`. I worry that we are getting many similarly named methods, and still sticking functionality in methods with names that don't describe that functionality. Again, there's nothing wrong with how it operates, I just feel like it makes maintenance and further extension a headache.

liamhuber: > It is the other way around, the `run()` method calls `collect_output` if the status is set to collect:
> https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L697
> More precisely it calls `_run_if_collect()`:
> https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L1483
> Which then calls the collect output.

@jan-janssen but is it not correct that even if we set `self.status.collect = True` in `self.run_static`, that without *also* calling `self.run()` inside `run_static`, we never proceed to the collection?

jan-janssen: @liamhuber Sorry for the confusing yes you should call run, just like it is also done in the main `run_static` method for external executables https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L787

liamhuber: @jan-janssen right, so this is a little bit my point; basically every class that inherits from `GenericJob` and is overriding `run_static` then needs to be adding these lines to set the status and re-call run. I think it's fine that this is the way we *have done it so far*, but now that we have people who are explicitly hired as software developers I'd like to take our existing "reasonable solutions" and level them up to "excellent solutions".

So in this case, the `run` method should have some sort of `while self.status not in [${list of terminating statuses}]: ${keep running}` and the downstream exposed `run_static` and `collect_output` methods should be wrapped up so that we don't need to keep re-implementing the status updates and run continuation. 

I'm totally open to new ideas for the details of this implementation, and also open to the timing of when we get around to doing it, but I hope we can come to (or are already at) a consensus that some sort of better packaging of the execution flow is important.

jan-janssen: @liamhuber The pyiron jobs were originally designed to call external codes and `run_static()` is used in the job wrapper and the GernericMasters so I am a bit hesitant to change everything. As far as I understand only the python based jobs need these changes so if we implement a function like `run_python_job(self)` which is called inside the `run_static()` function we should be able to maintain both functionality and smoothen the transition.

jan-janssen: A current use case is the current hessian job https://github.com/pyiron/pyiron/blob/master/pyiron/thermodynamics/hessian.py#L44 or the jobs derived from ASE https://github.com/pyiron/pyiron/blob/master/pyiron/gpaw/pyiron_ase.py#L125 these implement the `run_static()` function directly as the input and output handling is primarily using the interactive interface.

niklassiemer:
For now, I got rid of the extra `save_output()` function and shifted the `output` field to the `TemplateJob`. I will try to have a look at the `run()` flow and think about how we could proceed there.

pyiron_base/job/template.py
liamhuber: > @liamhuber The pyiron jobs were originally designed to call external codes and `run_static()` is used in the job wrapper and the GernericMasters so I am a bit hesitant to change everything. As far as I understand only the python based jobs need these changes so if we implement a function like `run_python_job(self)` which is called inside the `run_static()` function we should be able to maintain both functionality and smoothen the transition.

@jan-janssen I am still struggling here. I think this is just proving my point -- historically we used pyiron jobs to wrap external executables, but now many of the newly written classes that are children of `GenericJob` *don't do that*. That includes the two examples you gave, the Generic Masters you mentioned, as well as [several](https://github.com/pyiron/pyiron_contrib/blob/569e059a28c1e68c58a68c65c4f8d1486c65bcc7/pyiron_contrib/atomistic/atomistics/master/sqselastic.py#L321) [examples](https://github.com/pyiron/pyiron_contrib/blob/71ba4eea9db6acebffd5fb9876b54e4268cd7aba/pyiron_contrib/protocol/generic.py#L639) [in](https://github.com/pyiron/pyiron_contrib/blob/71ba4eea9db6acebffd5fb9876b54e4268cd7aba/pyiron_contrib/image/job.py#L149) pyiron_contrib. So `GenericJob.run_static` is getting overwritten all the time, and in many of those cases the logic of `GenericJob.run_static` -- e.g. status setting and recursively calling `run()` -- needs to be re-implemented.

So `GenericJob` probably isn't the perfect parent for all these classes! Or at least this implementation of it isn't. What I'm asking for is to re-write the execution flow of a base-level `Job` class which other classes can inherit from and not need to re-do this logic all the time. This seems to be a bit the direction you're going with `run_python_job`, but also your reply sounded like you were disagreeing with me.

We don't have to do this all at once, we can make these changes incrementally. E.g. with taking a step to add `run_python_job` as you suggest, or to make `run_static` and `collect_output` private since they're never(?) called from outside themselves, etc. 

I get that overall this might not be a trivial change, but my point is that we now have a whole team of people who are explicitly software developers. I think we have the time to start gunning for harder problems so that our code can go from being 'historical' to 'optimal'. I'm not trying to nail down details, I am just hoping to get an agreement in principle.

niklassiemer:
I think the `run` framework with the recalling of `run` at the end of each step is not that easy to change, since some of the run types have the same status, however, depending on other stuff like the run_mode different methods are called which do not necessarily call `run` at the end. Also the current definition of `run_static` is used in several derived classes.
One could think about adding an additional layer in-between `GenericJob` and `JobCore` where the large amount is dealt with and `GenericJob` 'just' provides the routines for potentially file based external executables. 
Currently, I used the very low level implementation to overwrite `run_static` in the `PythonTemplateJob` and ask the downstream developer to provide a `run_python_job` function in which the actual computation would be perfomed.

niklassiemer:
I will close this for now and come back to it once we have nice hdf paths to access.

PR #52, opened 2020-10-17
Fix static jobs in interactive wrapper

jan-janssen:


niklassiemer:
The pull request enables the `remove` method of the main `run` command to also remove child jobs on request in a later stage.

- Why is this pull request necessary? Which issue is addressed?
This pull request changes the default behavior of `remove`  to allow handling and removal of static jobs in an interactive wrapper.
- Is this the right place to fix this issue? Would it be better to address the issues in a different class ?
As far as I can see it is the topmost layer to change the default behavior. Since the error occurs in the interactive wrappers, I wonder if it would be better to change the behavior only in case of running interactively.
- Does this change affect other functionality? Is there a risk that this change causes bugs later on? Is there a simpler solution for the same problem?
I do not see problems from the code here, however, for unfamiliar users it could be dangerous to allow for deleting of whole projects with all it's jobs without the need to explicitly stating this.

Please let me know if I understood something wrong, here.

PR #62, opened 2020-10-28
Change queue_delete to call _queue_delete internally

Leimeroth:
Changed the queue_delete method from calling job.remove internally to prevent people from accidentally deleting jobs permanently. Instead they are only deleted from the queuing system


Leimeroth (commented):
I should have checked better, I just saw that _queue_delete_job was used in some other place and was worried that deleting it completely would break stuff, but if it is used only once I completely agree

jan-janssen (approved):
Looks good to me

Issue #42, opened 2020-10-07
InputList asymmetry: instantiated vs loaded


liamhuber:
Accessing `InputList` elements functions differently if the input list is instantiated or loaded from HDF5. This is because saving to HDF5 wraps complex sub-elements also as `InputList` objects, and this wrapping isn't undone when we load from HDF5. 

This can create an asymmetry in how we access the data depending whether the input list is "fresh" or loaded, which is awkward.

- Is this wrapping before saving necessary?
- Can we unwrap in `from_hdf`?
  - If we do, how do we distinguish the case where the user is legitimately *trying* to nest `InputList` objects?
- I haven't dug into it yet, but I assume that the unique iterator behaviour of `InputList` is pretty important. However, I guess changing this behaviour would also be a potential place for resolving this compaint.

# Example

## Demonstrate behaviour

`InputList.to_hdf` wraps complex children as new `InputList`s.

```python
from pyiron_base import InputList, Project
pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)

inputlist = InputList(table_name='inputlist')
inputlist.dict_ = [{'foo': [1], 'bar': 'baz'}, {'foo': [1, 2], 'bar': 'bazz'}]
inputlist.list_ = ['a', 'b', 'c']
inputlist.int_ = 42

hdf = pr.create_hdf(pr.open('inputlist_hdf5').path, 'inputlist')
inputlist.to_hdf(hdf=hdf)

loaded = InputList(table_name='inputlist')
loaded.from_hdf(hdf)

print(inputlist)
# list and dict children get wrapped as InputList!
print(loaded)

> InputList({'dict_': [{'foo': [1], 'bar': 'baz'}, {'foo': [1, 2], 'bar': 'bazz'}], 'list_': ['a', 'b', 'c'], 'int_': 42})
> InputList({'dict_': InputList([InputList({'foo': InputList([1]), 'bar': 'baz'}), InputList({'foo': InputList([1, 2]), 'bar': 'bazz'})]), 'list_': InputList(['a', 'b', 'c']), 'int_': 42})
```

## Problem case

If we want to access the data, we can't use the same operations on the loaded instance that we would naively use on the fresh instance.

```python
print([d['foo'] for d in inputlist['dict_']])

> [[1], [1, 2]]
```

```python
print([d['foo'] for d in loaded['dict_']])

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[...]

PR #70, opened 2020-11-10
Update InputList recursively in HDF5

pmrv:
With this InputList supports saving arbitrary types that define a
`to_hdf`.  This also creates HDF5 groups for each of these types and
such allows to access them directly from the job HDF5 via `to_object`.
Loading InputLists from HDF5 remains backwards compatible.

Sadly, for now I had to use node and group names of the form `{key}__index_{i}` to ensure I can always reconstruct the order, even after #60 (because the items are distributed among nodes & groups and that tracks the order only within each, so I will revert it). 

I did some reading into HDF and am sure now that it can be made better and without any name mangling using "Attributes", but that would require a big rewrite of the hdfio.py and so I'd delegate that to a separate work. 

I will also add test cases for @liamhuber & @samwaseda issues tomorrow, but I'm reasonable sure these will be solved.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -23,6 +23,15 @@ __status__ = "production"
 __date__ = "Jun 17, 2020"
 
 
+_internal_hdf_nodes = [
+        "NAME",
+        "TYPE",
+        "OBJECT",
+        "VERSION",
+        "HDF_VERSION",
+        "READ_ONLY"
+]
+
 def _normalize(key):
     if isinstance(key, str):
         if key.isdecimal():
@@ -146,6 +155,7 @@ class DataContainer(MutableMapping):
     """
 
     __version__ = "0.1.0"
+    __hdf_version__ = "0.2.0"
 
     def __new__(cls, *args, **kwargs):
 
@@ -427,7 +437,7 @@ class DataContainer(MutableMapping):
         Args:
             val: new element
         """
-        self._store.append(self._wrap_val(val))
+        self._store.append(val)
 
     def extend(self, vals):
         """
@@ -566,10 +576,31 @@ class DataContainer(MutableMapping):
         group_name = group_name or self.table_name
         if group_name:
             hdf = hdf.create_group(group_name)
+        elif len(hdf.list_dirs()) > 0:
+            raise ValueError(
+                    "HDF group must be empty when neither table_name nor "
+                    "group_name are set."
+            )
+
 
         self._type_to_hdf(hdf)
-        hdf["data"] = self.to_builtin()
-        hdf["read_only"] = self.read_only
+        hdf["READ_ONLY"] = self.read_only
+        for i, (k, v) in enumerate(self.items()):
+            if isinstance(k, str) and "__index_" in k:
+                raise ValueError("Key {} clashes with internal use!".format(k))
+
+            k = "{}__index_{}".format(k if isinstance(k, str) else "", i)
+
+            if hasattr(v, "to_hdf"):
+                v.to_hdf(hdf=hdf, group_name=k)
+            else:
+                # if the value doesn't know how to serialize itself, assume
+                # that h5py knows how to
+                try:
+                    hdf[k] = v
+                except TypeError:
+                    raise TypeError("Error saving {} (key {}): InputList doesn't support saving elements "
+                                    "of type \"{}\" to HDF!".format(v, k, type(v))) from None
 
     def _type_to_hdf(self, hdf):
         """
@@ -581,6 +612,7 @@ class DataContainer(MutableMapping):
         hdf["NAME"] = self.__class__.__name__
         hdf["TYPE"] = str(type(self))
         hdf["VERSION"] = self.__version__
+        hdf["HDF_VERSION"] = self.__hdf_version__
         hdf["OBJECT"] = "DataContainer"
 
     def from_hdf(self, hdf, group_name=None):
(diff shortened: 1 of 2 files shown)


niklassiemer:
This does not support writing and reading of empty `InputList` to and from hdf any longer:
```
from pyiron_base import InputList
p = Project('test')
h = p.create_hdf(p.path, 'subgroup')
k=InputList()
k.to_hdf(h)
k.from_hdf(h)
```
results in a value error.

pmrv:
> This does not support writing and reading of empty `InputList` to and from hdf any longer:
> 
> ```
> from pyiron_base import InputList
> from pyiron_base import Project
> p = Project('test')
> h = p.create_hdf(p.path, 'subgroup')
> k=InputList()
> k.to_hdf(h)
> k.from_hdf(h)
> ```
> 
> results in a value error.

I can reproduce this issue only if the given HDF5 is not empty and no table_name is set on the InputList.  I will add an exception that catches this in `to_hdf` since serializing into a non-empty group is weird.

pmrv:
@liamhuber I added a test case for #42, but if you still have the code that caused the original issue, it might be nice to double check with this branch.

niklassiemer:
Finally, I could reproduce the failure of the test notebook I use to implement a python-function in PR #50 with a minimal setup:
For really large numbers (the 100th Fibonacci number turns out to be to large) the old implementation with `to_buildin` worked fine, however, the new version fails:
```
from pyiron_base import InputList
p = Project('test')
h = p.create_hdf(p.path, 'subgroup')
k=InputList({E: int(2E19)}, table_name="input")
k.to_hdf(h)
```
with 
`TypeError: Object dtype dtype('O') has no native HDF5 equivalent`

pmrv:
> Finally, I could reproduce the failure of the test notebook I use to implement a python-function in PR #50 with a minimal setup:
> For really large numbers (the 100th Fibonacci number turns out to be to large) the old implementation with `to_buildin` worked fine, however, the new version fails:
> 
> ```
> from pyiron_base import InputList
> from pyiron_base import Project
> p = Project('test')
> h = p.create_hdf(p.path, 'subgroup')
> k=InputList({E: int(2E19)}, table_name="input")
> k.to_hdf(h)
> ```
> 
> with
> `TypeError: Object dtype dtype('O') has no native HDF5 equivalent`

I'll go out on a limb and say this has probably something to do with h5io or h5py, but I'll have a look.

pmrv:
> Finally, I could reproduce the failure of the test notebook I use to implement a python-function in PR #50 with a minimal setup:
> For really large numbers (the 100th Fibonacci number turns out to be to large) the old implementation with `to_buildin` worked fine, however, the new version fails:
> 
> ```
> from pyiron_base import InputList
> from pyiron_base import Project
> p = Project('test')
> h = p.create_hdf(p.path, 'subgroup')
> k=InputList({E: int(2E19)}, table_name="input")
> k.to_hdf(h)
> ```
> 
> with
> `TypeError: Object dtype dtype('O') has no native HDF5 equivalent`

2e19 doesn't fit in a 64bit integer and HDF5 doesn't seem to do big ints natively, so this is an unrelated issue.

niklassiemer:
I fully agree, I just stumbled across this behavior and since this worked in the old implementation I just pointed it out.

pmrv:
It worked for the same reasons I gave [here](https://github.com/pyiron/pyiron/issues/1053#issuecomment-725329845).  JSON doesn't know about integers, so int(2e19) is coerced to 2e19 and saved as text.  I don't feel strongly about edge-cases like this, so I'd prefer to keep the implementation simple.

niklassiemer:
For me the question is how many of such edge cases would be solved by a JSON representation, however, I am fine with the current situation for now. As you already pointed out, the main missing point remains the hdf paths to access the different fields in a nice way, which we need to use the InputList as IOList in the future.

pmrv:
What we could do is collect the "non-complex" members into dictionaries and save those as json again.  But then you're trading edge cases.  What if an `InputList` has a `dict` inside with a job inside?  I would think we make this work for now and then design a refactoring for the HDF storage based on all the things we want it to do.

liamhuber:
> @liamhuber I added a test case for #42, but if you still have the code that caused the original issue, it might be nice to double check with this branch.

I don't have it trivially accessible. I'm willing to accept the high probability that this resolves the issue though, and optimistically removed its "unsolved" tag.

tests/generic/test_inputlist.py
```diff
@@ -297,79 +301,162 @@ def rec(m):
...
+    def test_to_hdf_name(self):
+        """
+        Should raise error if clashing names are given.
+        """
+        with self.assertRaises(ValueError,
+                msg="Cannot have names clashing with index mangling."):
+            InputList({'__index_0': 42}).to_hdf(hdf=self.hdf)
 
     def test_to_hdf_group(self):
+        """
```
liamhuber: Nit: short docstrings can be one line instead of three according to [PEP257](https://www.python.org/dev/peps/pep-0257/)

tests/generic/test_inputlist.py
```diff
@@ -297,79 +301,162 @@ def rec(m):
...
+        Reading from HDF should give back the same list as written.
+        """
         self.pl.to_hdf(hdf=self.hdf)
         l = InputList(table_name = "input")
         l.from_hdf(hdf=self.hdf)
         self.assertEqual(self.pl, l)
 
     def test_from_hdf_group(self):
+        """
+        Reading from HDF should give back the same list as written even with
```
liamhuber: Nit: @samwaseda and I concluded over in [PR 1163](https://github.com/pyiron/pyiron/pull/1163#event-3982966405) that we would intentionally violate PEP8's 79 character line limit in favour of the pycharm default of 120 characters.

liamhuber (approved):
With the caveat that I was reviewing only for style and structure, not actual functionality which I assume you're doing correctly, the only thing from my end is a couple of nits regarding some formatting.

liamhuber:
Thanks for taking care of this, @pmrv!

liamhuber:
@pmrv can you check @jan-janssen's concern about where we break up the HDF hierarchy?

pmrv:
> @pmrv can you check @jan-janssen's concern about where we break up the HDF hierarchy?

It works as expected.  Elements that know how serialize themselves get their own HDF groups, elements that are saved as json become HDF nodes on the same level.

```python
>>> h = pr.create_hdf(h.path, 'foo')
>>> i = InputList()
>>> i.append(InputList([1,2]))
>>> i.append({"b": 44})
>>> i.to_hdf(h, group_name='bar')
>>> h['bar']
{'groups': ['__index_0'], 'nodes': ['HDF_VERSION', 'NAME', 'OBJECT', 'READ_ONLY', 'TYPE', 'VERSION', '__index_1']}
>>> h['bar/__index_0'].to_object()
InputList([1,2])
>>> h['bar/__index_1']
{'b': 44}
```

From my side this is ready to be merged, probably best after the DataContainer rename.

liamhuber:
The only thing that makes me nervous about that plan is that the rename is still marked as draft--is it closer than that suggests?

pmrv:
> The only thing that makes me nervous about that plan is that the rename is still marked as draft--is it closer than that suggests?

Yes, I'll give it a go tonight, but Monday latest.

pmrv:
Force-push to update to new master.

PR #75, opened 2020-11-12
Fix Project imports in cli tools

pmrv:
Some of the utilities still imported `Project` from `pyiron`.


jan-janssen (approved):
looks good to me

Issue #51, opened 2020-10-15
copy_to ignores new job name if a new project is used


liamhuber:
Observed behaviour:

```python
from pyiron_base import Project
pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)
ref_job = pr.create_job(pr.job_type.ScriptJob, 'ref')
parent_job = pr.create_job(pr.job_type.ScriptJob, 'parent')
new_name = ref_job.copy_template(project=parent_job._hdf5, new_job_name='parents_ref')
new_name._hdf5.path, new_name.job_name
> ('/Users/huber/Liam/pyiron/projects/tmp/scratch/parent/ref', 'ref')
```

Desired behaviour
```python
new_name._hdf5.path, new_name.job_name
> ('/Users/huber/Liam/pyiron/projects/tmp/scratch/parent/parents_ref', 'parents_ref')
```

I've been fighting for a large chunk of the day with HDF management over in the [contrib SQS elastic branch](https://github.com/pyiron/pyiron_contrib/blob/elastic_constants/pyiron_contrib/atomistic/atomistics/master/sqselastic.py). I realized this evening that `copy_to` was simply ignoring some of it's input. 

With at least part of the problem nailed down, I've been trying to correct the `pyiron_base/job/generic/GenericJob.copy_to` behaviour, but am clearly still missing some serious facts because all my attempts end with thrown exceptions.

jan-janssen:
Inside the jobs please use the `append()` and `pop()` function:
https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/master/generic.py#L182

But maybe it is easier to discuss the details tomorrow in the pyiron meeting. If you create a quick example inside a notebook which is using the ParallelMaster to calculate the SQS structures then I can show how to use the FlexibleMaster to connect the SQS Jobs and the list of structures like we do it in the Pipeline: 
https://github.com/pyiron/pyiron/blob/master/notebooks/bandstructure.ipynb
and with the band structure:
https://github.com/pyiron/pyiron-meeting/blob/master/2020-08-28/jan-janssen/band_structures_flexible.ipynb

From there we can discuss how to make it a single class. I just have the feeling we mix up two versions of copy() one which is used to store a child job of the HDF5 file of the master before it is executed and one which is used to copy finished jobs from one project to the next. I agree they are not clearly documented and the difference is not well explained, so do not hesitate to discuss these parts with me directly.

liamhuber:
[...]

PR #80, opened 2020-11-16
Fix copy to

### LINKED ISSUE #79: Copy to does not reload job properly
The `copy_to` function which creates a new job should load the job if a job `new_job_name` is stored in the given project path. Consider the code below
```
from pyiron import Project
pr = Project("copy_to_debug")
al_fcc = pr.create_ase_bulk("Al", cubic=True)
job_lammp_template = pr.create_job(
    job_type=pr.job_type.Lammps,
    job_name="lmp"
)
job_lammp_template.structure = al_fcc
job_bulk = job_lammp_template.copy_to(
    new_job_name="lmp_bulk", 
    new_database_entry=False
)
job_bulk.calc_static()
job_bulk.run()
print(job_bulk["output/generic/energy_tot"], job_bulk.status)
```

If you run the following lines for the first time, you get an output like so : `[-13.64243201] finished`. From then on, you get the output `None initialized` which IMO shouldn't be the case.

sudarsan-surendralal:
Debugging the `copy_to` function to load the already existing job with the same name. This is necessary for many workflows (including the ones in the recent phase diagrams workshop) to work seamlessly.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -514,7 +514,7 @@ class GenericJob(JobCore):
         new_job_name = new_job_name or self.job_name
         job_table = project.job_table(recursive=False)
         if len(job_table) > 0 and new_job_name in job_table.job.values:
-                return project[new_job_name]
+                return project.load(new_job_name)
         if in_same_project and len(self.project_hdf5.h5_path.split("/")) > 2:
             new_location = self.project_hdf5.open("../" + new_job_name)
         else:
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -519,7 +519,6 @@ class GenericJob(JobCore):
             if not delete_existing_job:
                 return project.load(new_job_name)
             else:
-                print("checkpoint 2")
                 project.remove_job(new_job_name)
         if in_same_project and len(self.project_hdf5.h5_path.split("/")) > 2:
             new_location = self.project_hdf5.open("../" + new_job_name)
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -513,8 +513,7 @@ class GenericJob(JobCore):
         project = project or self.project
         new_job_name = new_job_name or self.job_name
         job_table = project.job_table(recursive=False)
-        if len(job_table) > 0:
-            if new_job_name in job_table.job.tolist():
+        if len(job_table) > 0 and new_job_name in job_table.job.values:
                 return project[new_job_name]
         if in_same_project and len(self.project_hdf5.h5_path.split("/")) > 2:
             new_location = self.project_hdf5.open("../" + new_job_name)
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -516,7 +516,6 @@ class GenericJob(JobCore):
         new_job_name = new_job_name or self.job_name
         job_table = project.job_table(recursive=False)
         if len(job_table) > 0 and new_job_name in job_table.job.values:
-            print("checkpoint 1", delete_existing_job)
             if not delete_existing_job:
                 return project.load(new_job_name)
             else:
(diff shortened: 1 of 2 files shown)


pyiron_base/job/generic.py
```diff
@@ -512,6 +512,10 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
         # Get new hdf location
         project = project or self.project
         new_job_name = new_job_name or self.job_name
+        job_table = project.job_table(recursive=False)
+        if len(job_table) > 0:
+            if new_job_name in job_table.job.tolist():
```
jan-janssen:
```suggestion
        if len(job_table) > 0 and new_job_name in job_table.job.values:
```

pyiron_base/job/generic.py
```diff
@@ -512,6 +512,10 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
...
+                return project[new_job_name]
```
jan-janssen:
```suggestion
                return project.load(new_job_name)
```
To be a bit more explicit.

jan-janssen:
It would be great to add a test for this functionality as it is at the core of pyiron and we only have a very limited number of tests for `pyiron_base` so far.

pyiron_base/job/generic.py
```diff
@@ -512,6 +512,9 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
         # Get new hdf location
         project = project or self.project
         new_job_name = new_job_name or self.job_name
+        job_table = project.job_table(recursive=False)
+        if len(job_table) > 0 and new_job_name in job_table.job.values:
+                return project.load(new_job_name)
```
jan-janssen, line 520:
```suggestion
            return project.load(new_job_name)
```

liamhuber (changes requested):
It would be nice to have something like the `delete_existing_job` that exists for `create_job`, so that you can overwrite instead of loading. I'm sure, but I think this should be as easy as putting the old code and the new code (516+517 here) inside an `if/else` clause based on the input flag?

Other than that feature request, I'm happy. Great to see more tests!

pyiron_base/job/generic.py
```diff
@@ -512,6 +514,14 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
         # Get new hdf location
         project = project or self.project
         new_job_name = new_job_name or self.job_name
+        job_table = project.job_table(recursive=False)
+        if len(job_table) > 0 and new_job_name in job_table.job.values:
+            print("checkpoint 1", delete_existing_job)
+            if not delete_existing_job:
+                return project.load(new_job_name)
+            else:
+                print("checkpoint 2")
```
jan-janssen:
```suggestion
```

pyiron_base/job/generic.py
```diff
@@ -512,6 +514,14 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
         # Get new hdf location
         project = project or self.project
         new_job_name = new_job_name or self.job_name
+        job_table = project.job_table(recursive=False)
+        if len(job_table) > 0 and new_job_name in job_table.job.values:
+            print("checkpoint 1", delete_existing_job)
```
jan-janssen:
```suggestion
```

jan-janssen:
@liamhuber I guess all the requested changes are now implemented. So I would merge this with the master and then create a new release.

liamhuber (approved):

liamhuber:
> @liamhuber I guess all the requested changes are now implemented. So I would merge this with the master and then create a new release.

@jan-janssen before the release I would really like to see #51 fixed. This function literally doesn't do what it claims in one quadrant of the (new project, new name) map.

jan-janssen:
@liamhuber I agree but I do not see anybody working on it, so I am going to create a release now so we can use the creator for your pyiron pull request.

PR #81, opened 2020-11-17
Implement job creator

jan-janssen:
Now it is possible to create jobs like this: 
```
job = pr.create.job.Lammps(job_name="lap")
```
In addition to the previous implementation: 
```
job = pr.create_job(job_type=pr.job_type.Lammps, job_name="lap")
```


pyiron_base/project/factory.py
```diff
@@ -0,0 +1,80 @@
...
+__email__ = "janssen@mpie.de"
+__status__ = "production"
+__date__ = "Nov 17, 2020"
+
+
+class Creator(object):
+    def __init__(self, project):
+        self._job_creator = JobCreator(project=project)
+
+    @property
```
liamhuber: 👍 I like the idea to hide these behind properties so they don't accidentally get overwritten.

pyiron_base/project/factory.py
```diff
@@ -0,0 +1,80 @@
...
+    def job(self):
+        return self._job_creator
+
+
+class JobCreator(object):
+    def __init__(self, project):
+        self._job_type = JobTypeChoice()
+        self._project = project
+
+    def __dir__(self):
```
liamhuber: We discussed in the last pyiron meeting the dangers of messing around with `__dir__`. I do not fully grasp all of these and don't have any hard objection to what you do here, I just wanted to flag it for your own double-check.

pyiron_base/project/generic.py
```diff
@@ -15,6 +15,7 @@
 from git import Repo, InvalidGitRepositoryError
 
 from pyiron_base.project.path import ProjectPath
+from pyiron_base.project.factory import Creator
```
liamhuber: Rather than having `generic/factory.py`, I feel like it makes more sense to put the factory files next to the definitions of their objects. This will prevent having a million factories being crammed into one generic factory file in modules where there's lots of factories. It also makes them easier to find, but that is maybe just my opinion.

With this, I think `Creator` can simply be implemented right in `project.py` as it's quite short and directly related.

liamhuber (changes requested):
I'm excited to see this change so fast!

I don't see any content/architecture changes that need to be made, but I'd like to see the factories (in this case just `JobFactory` (and `JobClass`) moved off into a file by the other job files, and `Creator` shifted right into `project.py`.

Since it's for base, it would also be extremely valuable to get tests over it.

pyiron_base/project/factory.py
jan-janssen: We did this before for `JobTypeChoice` and I do the same here, it is not the most elegant way to restrict the user interface but as the users only use this object to browse the different job types I guess it is ok.

pyiron_base/project/generic.py
jan-janssen: Done

liamhuber (approved):
Tests would still be great, but the code itself is looking good to me.

PR #89, opened 2020-11-26
Fix DocString for copy_to()

### LINKED ISSUE #88: copy_to depends on type not specified in docstring
The `GenericJob.copy_to` docstrings claim that the `project` argument is of type `ProjectHDFio`, but when passed an argument of this time raises an `AttributeError` because this class does not have access to the `job_table` method.

Minimum working example:
```python
pr = Project('scratch')
from pyiron_base.job.generic import GenericJob
job = pr.create_job(GenericJob, 'job')
print(type(job._hdf5))

> <class 'pyiron_base.generic.hdfio.ProjectHDFio'>

print(job.copy_to.__doc__.split('\n')[4])

> project (ProjectHDFio): The project to copy the job to. (Default is None, use the same project.)

copied = job.copy_to(project=job._hdf5)
AttributeError                            Traceback (most recent call last)
<ipython-input-7-fa95e13baac0> in <module>
----> 1 copied = job.copy_to(project=job._hdf5)

~/Liam/pyiron/pyiron_repos/pyiron_base/pyiron_base/job/generic.py in copy_to(self, project, new_job_name, input_only, new_database_entry, delete_existing_job)
    515         project = project or self.project
    516         new_job_name = new_job_name or self.job_name
--> 517         job_table = project.job_table(recursive=False)
    518         if len(job_table) > 0 and new_job_name in job_table.job.values:
    519             if not delete_existing_job:

AttributeError: 'ProjectHDFio' object has no attribute 'job_table'
```

jan-janssen:


pyiron_base/job/generic.py
```diff
@@ -500,7 +501,20 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
         if project is None and new_job_name is None:
             raise ValueError("copy_to requires either a new project or a new_job_name.")
 
-        in_same_project = project is None or project.path == self.project.path
+        if isinstance(project, self.project.__class__):
+            in_same_project = project.path == self.project.path
+            project = project or self.project
```
liamhuber: Once we're inside one of the `isinstance(project, X)` clauses, what is the purpose of `or` in `project = project or self.project`? Won't it always evaluate to the first term?

If I understand that point correctly, then `project = project or self.project` can safely be extracted up out of the if clauses.

pyiron_base/job/generic.py
```diff
@@ -500,7 +501,20 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
...
+            job_table = project.job_table(recursive=False)
+        elif isinstance(project, self.project_hdf5.__class__):
+            in_same_project = project.path == self.project_hdf5.path
```
liamhuber, line 509: Is there a guideline for when to use `._hdf5` and when to use its `.project_hdf5` wrapper?

liamhuber (commented):
Nice! I have a couple questions, the answer to one of which may allow for a very minor simplification.

pyiron_base/job/generic.py
jan-janssen: The suggestion is always use `project_hdf5`.

pyiron_base/job/generic.py
jan-janssen: Done

liamhuber (approved):

PR #92, opened 2020-11-27
Add new deprecate decorator

pmrv:
As we discussed in the pyiron meeting today, I made a minimal decorator.  You can pass a message and a version after which the function is expected to disappear.   I thought about making it raise an exception automatically when the given version is lower than the package version, but it is surprisingly difficult to come up with a solution that works for pyiron_base & pyiron.

### DIFF
--- a/pyiron_base/generic/util.py
+++ b/pyiron_base/generic/util.py
@@ -112,8 +112,7 @@ class Deprecator:
         """
         Initialize default values for deprecation message and version.
 
-        Args
-        ----
+        Args:
             message (str): default deprecation message
             version (str): default version after which the function might be removed
             pending (bool): only warn about future deprecation, warning category will be PendingDeprecationWarning
@@ -136,11 +135,10 @@ class Deprecator:
         Wrap the given function to emit a DeprecationWarning at call time.  The warning message is constructed from the
         given message and version.
 
-        Args
-        ----
+        Args:
             function (function): function to mark as deprecated
-        Return
-        ------
+
+        Return:
             function: raises DeprecationWarning when given function is called
         """
         if self.category == PendingDeprecationWarning:
--- a/pyiron_base/generic/util.py
+++ b/pyiron_base/generic/util.py
@@ -115,7 +115,7 @@ class Deprecator:
         Args
         ----
             message (str): default deprecation message
-            version (str): default version after which the function will be removed
+            version (str): default version after which the function might be removed
             pending (bool): only warn about future deprecation, warning category will be PendingDeprecationWarning
                 instead of DeprecationWarning
         """
--- a/pyiron_base/generic/util.py
+++ b/pyiron_base/generic/util.py
@@ -72,6 +72,10 @@ class Deprecator:
         {function_name} is deprecated: {message}. It is not guaranteed to be in
         service after {version}.
 
+    unless `pending=True` was given.  Then the message will be
+
+        {function_name} will be deprecated in a future version: {message}.
+
     If message and version are not initialized or given during the decorating
     call the respective parts are left out from the message.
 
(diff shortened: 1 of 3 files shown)


pmrv:
Sometimes we want to deprecate only certain combinations of arguments to function.  This is not handled yet, because I'd first want to see how well this works in pyiron and maybe tweak the design a bit before.

pyiron_base/generic/util.py
```diff
@@ -57,3 +61,119 @@ def static_isinstance(obj, obj_type):
...
+    >>> foo(1, 2)
+    DeprecationWarning: __main__.foo is deprecated: pyiron says no!  It is not
+    guaranteed to be in service in vers. 0.5.0
+    """
+
+    def __init__(self, message=None, version=None, pending=False):
+        """
+        Initialize default values for deprecation message and version.
+
+        Args
```
liamhuber: Formatting is weird here. Should be more like:

```python
Args:
    message (str): Default deprecation message.
    ...
```

Also, I don't have the link with me, but @samwaseda and I decided to thumb our noses at PEP8 and go for the PyCharm default of 120 characters/line. It looks like some of the breaks here might be to accommodate a shorter line.

pyiron_base/generic/util.py
```diff
@@ -57,3 +61,119 @@ def static_isinstance(obj, obj_type):
...
+        ------
+            function: raises DeprecationWarning when given function is called
+        """
+        if self.category == PendingDeprecationWarning:
+            message_format =  "{}.{} is deprecated"
+        else:
+            message_format =  "{}.{} will be deprecated"
+        message = message_format.format(function.__module__, function.__name__)
+
+        if self.message:
```
liamhuber: If this is a None test, it should be explicit...?

tests/generic/test_util.py
```diff
@@ -23,6 +25,44 @@ def test_static_isinstance(self):
...
+        def foo(a):
+            return 2*a
+
+        @deprecate("use baz instead", version="0.2.0")
+        def bar(a):
+            return 4*a
+
+        with warnings.catch_warnings(record=True) as w:
+            self.assertEqual(foo(1), 2,
+                             "Decorated function does not return original "
```
liamhuber, line 40: It took me a minute to deduce the third argument, but I like this a lot. I didn't realize unittests was set up this way (it makes sense though). I will need to start exploiting this!

liamhuber (commented):
Thanks for getting to this not only quickly but very thoroughly!

pyiron_base/generic/util.py
samwaseda: NB: Actually I agreed that 80 was too short but I'm now sticking to 100 characters because my vim default has 200+ characters and with a vertically split window it nicely accommodates an entire line... This being said, I'm perfectly comfortable with other people using 120 characters as long as they accept my default of 100 characters.

pmrv: I generally prefer shorter lines for vertical splits, but will accommodate anything that we agree on.  I changed my project config to 120 now, if that's what we're going with.

pyiron_base/generic/util.py
```diff
@@ -57,3 +61,119 @@ def static_isinstance(obj, obj_type):
...
+    def __init__(self, message=None, version=None, pending=False):
+        """
+        Initialize default values for deprecation message and version.
+
+        Args
+        ----
+            message (str):
+                default deprecation message
+            version (str):
+                default version after which the function will be removed
```
samwaseda: The function **will** be removed or **might** be removed? I would indeed prefer to systematically remove all deprecated functions (in which case this DocString is correct), but since I'm a rather weak programmer and frankly have no idea how to automatise it, I could only suggest to change it to "might be"...

Either way, "will be removed" and "not guaranteed" (which appears in the warning) are not compatible.

tests/generic/test_util.py
pmrv: I am a big fan of the test message, it helps fixing tests later on a lot

pyiron_base/generic/util.py
pmrv: I took the warning message to be close to what our manual message say right now, so I'll just change the docstring.  As I said I'd want to explore automating it, but it's not so easy on the software side I think.   Worst case would be to grep "deprecate" + version number before each bump, which isn't terribly much work.

pyiron_base/generic/util.py
liamhuber: But I still want the extra newlines and horizontal bars removed 😉

pmrv: Ah, I got my markup confused, I thought the '----' is necessary for the sphinx argument list, but it's ':'.

liamhuber (approved):

PR #93, opened 2020-11-27
Debug nested masters -related to copy_to()

jan-janssen:


liamhuber:
I still need to 100% nail down the guilt, and maybe I could find a workaround maintaining this merge... but it appears to have broken my nested master.

Is there somewhere a such a nested master that this was tested on which I can use as a reference?

liamhuber:
Indeed, somewhere between here and HEAD (inclusive), `copy_to` stops copying the `ref_job` attribute (at least when `project` is provided, `new_job_name` is provided, and `new_database_entry=False`. I'll try to nail it down further and patch.

liamhuber:
Yarp, in `pyiron_base/master/generic.py` lines 273 and 274, there is now an early return:

```python
        if job_return is not None:
             return job_return
```

This happens before the copying of children.

I'll double check that this is the issue and make a patch pr.

liamhuber:
Ok, I'm no longer convinced that the error comes from this PR (still possible I guess), and will make a proper issue.

PR #96, opened 2020-12-02
implement __repr__ in error

samwaseda:


jan-janssen (changes requested):
You overwrite: 
https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/core.py#L912

Which is used to see which groups and nodes are available in the HDF5 file. So please add a super() call to __repr__().

samwaseda:
> You overwrite:
> https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/core.py#L912
> 
> Which is used to see which groups and nodes are available in the HDF5 file. So please add a super() call to **repr**().

Wait this one affects only `GenericError` (which is not derived of anything). Do you want me to create an HDF connection?

jan-janssen:
> > You overwrite:
> > https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/core.py#L912
> > Which is used to see which groups and nodes are available in the HDF5 file. So please add a super() call to **repr**().
> 
> Wait this one affects only `GenericError` (which is not derived of anything). Do you want me to create an HDF connection?

No - sorry - I should have checked more carefully - then it is fine.

jan-janssen (approved):

samwaseda:
> No - sorry - I should have checked more carefully - then it is fine.

But you're right in the sense that this should be at some point connected to HDF5 at some point, because I'm very sure that things like [number of empty states convergence check](https://github.com/pyiron/pyiron/pull/1080) should be stored in the HDF and this class should be able to print those errors/warnings.

PR #102, opened 2020-12-05
Correction from `if None` to `if None is not None`

samwaseda:


jan-janssen (approved):
Looks good to me

jan-janssen:
Still I would assume there are even more of these cases in the code.

samwaseda:
> Still I would assume there are even more of these cases in the code.

Yeah true we will have to go step by step

PR #104, opened 2020-12-07
New docs + refactorings in JobGenerator

pmrv:
Nothing changes about the interface, but `job_name` is a predefined method now instead of special case.


jan-janssen (approved):
Looks good to me

PR #107, opened 2020-12-09
Fix GenericMaster from_hdf()

### LINKED ISSUE #106: Nested masters don't work because copy_to misses a dict
A couple weeks ago I had a nested master subclass (inheriting from `ParallelMaster` and taking as a reference job an `InteractiveWrapper` which in turn has its own `ref_job`). Now this class fails.

I tracked the problem down that `copy_to` fails to copy the `_job_object_dict` attribute of the job. Then, when we try to access the child it needs to be able to reconstruct it from this dict, can't, and throws a `ValueError`.

It works back as of merging #91 but I haven't git bisected to find the trouble yet.

MWE:

```python
from pyiron_base import Project, GenericJob, GenericMaster
pr = Project('scratch')
child = pr.create_job(GenericJob, 'child')
master = pr.create_job(GenericMaster, 'master')
master.append(child)
copy_ = master.copy_to(project=pr, new_job_name='copy', new_database_entry=False)
print(master._job_object_dict)  # Looks great
>>> {'child': {'groups': [], 'nodes': []}}
print(copy_._job_object_dict)  # Empty!!!
>>> {}
print(master[0])  # Looks great
>>> {'groups': [], 'nodes': []}
print(copy_[0])  # Throws ValueError because it needs to be able to build it from _job_object_dict
```
<details><summary>Stack trace</summary>
<p>

```python
<ipython-input-13-8f29bd3da06f> in <module>
----> 1 print(copy_[0])

~/Liam/pyiron/pyiron_repos/pyiron_base/pyiron_base/master/generic.py in __getitem__(self, item)
    441             item = self._job_name_lst[item]
    442         return self._get_item_when_str(
--> 443             item=item, child_id_lst=child_id_lst, child_name_lst=child_name_lst
    444         )
    445 

~/Liam/pyiron/pyiron_repos/pyiron_base/pyiron_base/master/generic.py in _get_item_when_str(self, item, child_id_lst, child_name_lst)
    553                 return self.project.load(child_id, convert_to_object=True)
    554         elif item_obj in self._job_name_lst:
--> 555             child = self._load_job_from_cache(job_name=item_obj)
    556             if len(name_lst) == 1:
    557                 return child

~/Liam/pyiron/pyiron_repos/pyiron_base/pyiron_base/master/generic.py in _load_job_from_cache(self, job_name)
    493             return self._job_object_dict[job_name]
    494         else:
--> 495             ham_obj = self.project_hdf5[job_name].to_object(
    496                     project=self.project_hdf5,
    497                     job_name=job_name,

~/Liam/pyiron/pyiron_repos/pyiron_base/pyiron_base/generic/hdfio.py in __getitem__(self, item)
    827                         obj = hdf_item.copy()
    828                         return obj
--> 829                 raise ValueError("Unknown item: {}".format(item))
    830             else:
    831                 if item_lst[0] == "":  # absoute HDF5 path

ValueError: Unknown item: child
```

</p>
</details>

jan-janssen:


jan-janssen:
@liamhuber The tests passed for this change and from my perspective it solves your issue. So if you agree I would merge these changes and then update the pyiron_base version so we can make sure it does not break anything inside pyiron.

liamhuber:
I'm in a meeting, but I'll check it out the moment I'm done.

pyiron_base/master/generic.py
```diff
@@ -332,6 +332,10 @@ def from_hdf(self, hdf=None, group_name=None):
             job_list_tmp = hdf5_input["job_list"]
             self._from_hdf_child_function(hdf=hdf5_input)
             self._job_name_lst = job_list_tmp
+            self._job_object_dict = {
+                job_name:self._load_job_from_cache(job_name=job_name)
```
liamhuber:
```suggestion
                job_name: self._load_job_from_cache(job_name=job_name)
```

liamhuber: PEP8 E231

liamhuber (approved):
Very nice thanks Jan.

There is still a bit of ugliness with orphaned H5 files when you have `IterativeWrapper` which really runs its reference job and thus needs to be copying it. But that's an extremely niche case and can be dealt with separately.

For my usages this is now perfect, but do we not still run into extreme slowness loading large masters now? My understanding is that `_load_job_from_cache` is reverting to reading from the transient HDF5 file. For copying this is fine, but I guess now whenever a `GenericMaster` is loaded it reinstantiates all its children immediately, or? Again, not a big deal for me personally, but you rightly mentioned it as a concern earlier.

jan-janssen:
No, the `_job_name_lst` only contains the jobs stored in the HDF5 file not the child jobs which have already been executed, so it should only be one job which is loaded.

PR #109, opened 2020-12-09
Include master name in generated job name

liamhuber:
When multiple parallel masters use the same reference job, everything runs because the project hdf5 is the parallel master's hdf5, but at the parent project level you can't load the children because their names are not unique. This fixes that, at the cost of longer child names.

### DIFF
--- a/pyiron_base/master/parallel.py
+++ b/pyiron_base/master/parallel.py
@@ -940,7 +940,7 @@ class JobGenerator(object):
         Returns:
             str: job name for the next child job
         """
-        return self._master.ref_job.job_name + "_" + str(self._childcounter)
+        return '_'.join([self._master.job_name, self._master.ref_job.job_name, str(self._childcounter)])
 
     def __iter__(self):
         return self


jan-janssen:
Is not this clear already given the path of the job?

liamhuber:
> Is not this clear already given the path of the job?

I don't understand what you mean.

```python
ref = pr.create.job.Child('ref')

parent1 = pr.create.job.Parent('p1')
parent1.ref_job = ref
parent1.input.x = 1

parent2 = pr.create.job.Parent('p2')
parent2.ref_job = ref
parent2.input.x = 2

parent1.run()
parent2.run()
```

Works fine. Then the project has `'p1', 'ref_0', 'ref_1', ..., 'p2', 'ref_0', 'ref_1', ...`. And `pr.load('ref_0')` complains. I

I took your comment to mean that I could do `pr.load('p1/ref_0')`, but this also does not work.

Can you explain further?

jan-janssen:
What about? 
```
pr['p1'].load('ref_0')
```
in case this works, then also the following should work: 
```
pr['p1/ref_0'].to_object()
```

liamhuber:
> What about?
> 
> ```
> pr['p1'].load('ref_0')
> ```

Tried this in the interim myself. It fails.

> in case this works, then also the following should work:
> 
> ```
> pr['p1/ref_0'].to_object()
> ```

Works. This is pretty obtuse though.

I would like to be able to look at `pr.job_table().job` and `pr.load` any items I see, which I can't currently do. We could fix this in `Project`, but it's exceedingly simple to fix it here. I'd rather fix it here like this, and change it back to the shorter name *after* some other fix gives the same functionality.

pmrv:
I'm slightly leaner towards @jan-janssen suggestions, but I think @samwaseda opened a similar PR recently (can't find it right now), so if the demand is there we should rethink it. 

I'm a bit worried that this might cause the job names to bloat beyond the character limit more easily, especially with nested masters.  On the conceptual side I also think that job names can't always be unique, so it's always safer to load jobs via the ID. 

My preferred solution is
```python
pr['p1'].load('ref_0')
```
if that doesn't work it's a bug in its own right.

pyiron_base/master/parallel.py
```diff
@@ -940,7 +940,7 @@ def job_name(self, parameter):
         Returns:
             str: job name for the next child job
         """
-        return self._master.ref_job.job_name + "_" + str(self._childcounter)
+        return '_'.join([self._master.job_name, self._master.ref_job.job_name, str(self._childcounter)])
```
pmrv, line 943: On the implementation side, if we adopt this, it should rather be changed below, in `.next:964`, otherwise sub classes of `JobGenerator` need to reimplement that logic.

pmrv (changes requested):

liamhuber:
@jan-janssen @pmrv no worries. If there is serious pushback then I'll just close it -- my own subclassed parallel master can work this way without imposing it on the master branch ;)

I would be perfectly happy if `pr.load('p1/child_0')` worked, and content with `pr['p1'].load('child_0')`, but I don't have the time debug/implement those at the moment.

PR #110, opened 2020-12-09
Add property for backwards compatibility

pmrv:
Other job generator sub classes in pyiron still use the `_job` attribute.  I added two getters and a deprecation warning.

### DIFF
--- a/pyiron_base/master/parallel.py
+++ b/pyiron_base/master/parallel.py
@@ -18,6 +18,7 @@ from pyiron_base.generic.parameters import GenericParameters
 from pyiron_base.job.jobstatus import JobStatus
 from pyiron_base.settings.generic import Settings
 from pyiron_base.job.wrapper import job_wrapper_function
+from pyiron_base.generic.util import deprecate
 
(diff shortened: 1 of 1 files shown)


jan-janssen:
Did it break anywhere? If not I do not think we need the property here.

pmrv:
It did not yet, is it even deployed yet? Definitely existing job generators reference `self._job` and should break if anyone runs them, e.g.

https://github.com/pyiron/pyiron/blob/5f4a6ddab12c6dd2106b949b2a5c4e4bdeaf6cc4/pyiron/atomistics/master/parallel.py#L147

https://github.com/pyiron/pyiron/blob/5f4a6ddab12c6dd2106b949b2a5c4e4bdeaf6cc4/pyiron/atomistics/master/murnaghan.py#L309

jan-janssen (approved):
Looks good to me

jan-janssen:
Then I guess I was a bit quick to trigger the new release. I guess we should still fix it inside pyiron once the new pyiron_base version is available.

pmrv:
I should have checked it before removing it in the other PR anyway.

PR #112, opened 2020-12-09
Refactor project handling for children in master jobs

pmrv:
This adds two properties that make the project in which the children live available.  I'll like to have this to combine pyiron tables with master jobs, otherwise I'll have to fiddle around a lot to find the project to set on the table.

This might also smooth out @liamhuber's problems in #109.  Loading a child job should (no tests on my side) work with
```python
master.child_project.load(job_name)
```

Btw. there's a possibilitiy @jan-janssen's suggestions didn't work.  If @liamhuber's master class sets `job.server.new_hdf=True`, then the project name is "{master_job_name}_hdf5".  I believe this is the default, but I just learned that this exists.  So what should work interim is
```python
pr["master_hdf5"].load(job_name)
```

### DIFF
--- a/pyiron_base/master/generic.py
+++ b/pyiron_base/master/generic.py
@@ -153,6 +153,17 @@ class GenericMaster(GenericJob):
             return super(GenericMaster, self).child_ids
 
     @property
+    def child_project(self):
+        """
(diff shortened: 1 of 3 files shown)


jan-janssen (approved):
Looks good to me

PR #114, opened 2020-12-12
Delete dead code in GenericMaster

pmrv:
Going on a limb here and assuming that this is left from ages long past and not needed anymore.

### DIFF
--- a/pyiron_base/master/generic.py
+++ b/pyiron_base/master/generic.py
@@ -406,39 +406,6 @@ class GenericMaster(GenericJob):
             "read procedure must be defined for derived Hamilton!"
         )
 
-    def run_if_interactive(self):
-        """
-        For jobs which executables are available as Python library, those can also be executed with a library call
-        instead of calling an external executable. This is usually faster than a single core python job.
-        """
-        raise NotImplementedError(
-            "This function needs to be implemented in the specific class."
-        )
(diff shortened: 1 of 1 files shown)


liamhuber (commented):
The problem is deeper than this -- `GenericMaster`'s parent, `GenericJob`, *also* has all these functions. And most of them that are `pass` here raise a `NotImplementedError`. Note that `GenericJob` does not inherit from `ABC` and these are not `@abstrachmethod`s, although the implementation is the same(?) (minus the IDE nag).

So (a) it's not 100% clear to me that these removals are side-effect free, and (b) what I'd really like to see is a change in the inheritance structure so that these pseudo-abcs are not in `GenericJob` to begin with, but, if they're interactive-related, sit in `InteractiveBase`...

pmrv:
Agree that this is not the perfect solution, there's a few master related attributes that are even defined on `PyironObject`.  We should definitely look into ABC's for these classes and I am happy to work on this, but I felt this a deeper change with more discussion needed.  I'll also be happy to drop this PR and do a proper solution later.

liamhuber:
Yeah, since this is really easy to recreate (or reapply -- only one commit), let's just shelve it until we can deal with it in the wider context of fixing the inheritance structure.

PR #116, opened 2020-12-14
Add a util class for import warnings

liamhuber:
As part of resolving [pyiron/pyiron issue #1230](https://github.com/pyiron/pyiron/pull/1230/files), which tackles the problem of not letting users know they're missing imports until they call `run`.

Going forward we could think about automating parts of the message as in the deprecation warning class.

### DIFF
--- a/pyiron_base/generic/util.py
+++ b/pyiron_base/generic/util.py
@@ -226,3 +226,39 @@ class Deprecator:
 
 deprecate = Deprecator()
 deprecate_soon = Deprecator(pending=True)
+
+
+class ImportAlarm:
+    """
+    In many places we have try/except loops around imports. This class is meant to accompany that code so that users
+    get an early warning when they instantiate a job that won't work when run.
+
+    Example:
+    >>> try:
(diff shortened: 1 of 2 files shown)


liamhuber:
I just saw @pmrv 's tests for the deprecator class. I'll do something similar, but not tonight. Marking as Draft until that's done.

pmrv:
Like it and added a bit of code to make it usable as a context manager.  Examples and tests are updated.

Only problem, at least on my system, `ImportWarning` are by default filtered in the warnings module.  Is it different for you?  If not we should remove that filter on pyiron import.

pmrv (approved):

liamhuber:
> Like it and added a bit of code to make it usable as a context manager. Examples and tests are updated.

Very nice!

> Only problem, at least on my system, `ImportWarning` are by default filtered in the warnings module. Is it different for you? If not we should remove that filter on pyiron import.

I manually activated all warnings in my .ipython config file, so I'm not sure. This is a very good point and I like the proposed solution, although I don't know how to implement that.

Anyhow, I think we're both satisfied with this 😄

PR #118, opened 2020-12-14
Safe script path

samwaseda:
Raised [here](https://github.com/pyiron/pyiron/issues/1261)

### DIFF
--- a/tests/job/test_script.py
+++ b/tests/job/test_script.py
@@ -17,7 +17,6 @@ class TestScriptJob(unittest.TestCase):
 
     @classmethod
     def tearDownClass(cls):
-        file_location = os.path.dirname(os.path.abspath(__file__))
         cls.project.remove(enable=True)
 
     def test_script_path(self):
--- a/tests/job/test_script.py
+++ b/tests/job/test_script.py
@@ -1,0 +1,29 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+import unittest
+import os
+from pyiron_base.project.generic import Project
+
+
+class TestGenericJob(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.file_location = os.path.dirname(os.path.abspath(__file__)).replace(
+            "\\", "/"
+        )
+        cls.project = Project(os.path.join(cls.file_location, "test_sriptjob"))
+
+    @classmethod
+    def tearDownClass(cls):
+        file_location = os.path.dirname(os.path.abspath(__file__))
+        cls.project.remove(enable=True)
+
+    def test_script_path(self):
+        job = self.project.create_job('ScriptJob', 'script')
+        with self.assertRaises(AssertionError):
+            job.run()
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/job/test_script.py
+++ b/tests/job/test_script.py
@@ -7,7 +7,7 @@ import os
 from pyiron_base.project.generic import Project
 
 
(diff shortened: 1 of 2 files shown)


pmrv:
This is not a specific issue with this PR, but in some of our `validate_ready_to_run` we raise `ValueError` and in some we raise `AssertionError`.  I'm not really partial either way, but I think we should agree on one type of exception to run or specify the semantics when to raise which.

samwaseda:
According to [python doc](https://docs.python.org/3/library/exceptions.html):

- `ValueError`: Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.
- `AssertionError`: Raised when an assert statement fails.

Since in this case nothing got an argument (let alone the right type), I would claim `ValueError` is not the right error. I'm gladly overruled if you disagree.

liamhuber:
If you expect a path and get `None`, I think `TypeError` is best.

samwaseda:
> If you expect a path and get `None`, I think `TypeError` is best.

I'm also fine with `TypeError`, but that sounds to me like the user actively put a wrong type, while in reality it's from developers' perspective that not setting the script path results in a different type than `str`.

A slightly different question: Would there be a practical problem if we set a potentially disputable warning (other than the fact that it could be misleading for the user)?

liamhuber:
> I'm also fine with `TypeError`, but that sounds to me like the user actively put a wrong type, while in reality it's from developers' perspective that not setting the script path results in a different type than `str`.

`TypeError: ScriptJob.input.script_path [or whatever the arg name is] expects a path [or whatever it is you're expecting] but got None. Please provide a path before running.`

Anyone who reads this will understand why it's a `TypeError`, and the people who still don't get it, frankly, are probably not the type of people who read stack traces to begin with 😝 

> A slightly different question: Would there be a practical problem if we set a potentially disputable warning (other than the fact that it could be misleading for the user)?

Let's rather be technically correct -- the best type of correct there is 😉

samwaseda:
Speaking of the distinction between `ValueError` and `TypeError`, there's quite a lot of work to be done, because as far as I know we mostly use `ValueError`, e.g. [here](https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/atoms.py#L157), [here](https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/atoms.py#L257), [here](https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/atoms.py#L635), [here](https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/atoms.py#L2336) or [here](https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/atoms.py#L2392) (and I'm only naming those found in `atoms.py`)

liamhuber:
Yeah, valid point. But this is definitely a niche enough issue that I'm content to "start doing it right now" rather than worry about perfect adherence throughout the codebase

tests/job/test_script.py
```diff
@@ -0,0 +1,29 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.file_location = os.path.dirname(os.path.abspath(__file__)).replace(
+            "\\", "/"
+        )
+        cls.project = Project(os.path.join(cls.file_location, "test_sriptjob"))
+
+    @classmethod
+    def tearDownClass(cls):
+        file_location = os.path.dirname(os.path.abspath(__file__))
```
liamhuber: Codacy appears correct to complain about this, indeed it doesn't appear to be used.

liamhuber (changes requested):
Please fix the codacy "unused" complaint. Fixing the duplication complaint seems like overkill though -- IMO it's just fine to duplicate code *once*.

liamhuber (approved):

PR #119, opened 2020-12-16
enable delete_existing_job when job created via ref_job

samwaseda:
What was (and still is) possible:

```
master_job = pr.create_job('SomeJob', 'some_job_name', delete_existing_job=True)
master_job.ref_job = child_job
```

What was **not** possible:

```
master_job = child_job.create_job('SomeJob', 'some_job_name', delete_existing_job=True)
```

And this is also what is going to be possible with this PR.


pmrv (approved):
lgtm

PR #124, opened 2020-12-18
Copy deprecator instance on decorating

pmrv:
Decorating a function with the Deprecator can change its instance
attributes (if you pass arguments when decorating).  That means that
when decorating different functions the warning shown used only the
arguments from the last call to the deprecator instance.

This is fixed now by copying the deprecator instance before wrapping the
given function.


jan-janssen (approved):
Looks good to me

PR #132, opened 2021-01-11
Renaming packing and unpacking

max-hassani:
As @pmrv suggested, the `packing` and `unpacking` functions are renamed to `pack` and `unpack`.


pmrv (approved):
Looking good, thanks for taking the time!

PR #133, opened 2021-01-11
Poly

### LINKED ISSUE #131: Unsightly polymorphism in job.create_job()
When creating jobs from other jobs, some job types require special treatment as shown here

https://github.com/pyiron/pyiron_base/blob/6b78c6fc3899b397f0a1735e4fe272ca9bf7e4cf/pyiron_base/job/generic.py#L1032

It's a bit weird, because `pyiron_base` needs to know about job types not defined there.  This could be done more cleanly by adding an method to `GenericJob` that is called only after is created as a child job and then overriding this method in the classes currently special cased.

If there's no objections, I'll make PR for that, but I might have missed something.

pmrv:
As discussed #131, this adds a method to handle the hardcoded special cases, when creating a job from a another job.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -975,6 +975,38 @@ class GenericJob(JobCore):
         if self.server.send_to_db:
             pass
 
+    def _init_child_job(self, parent):
+        """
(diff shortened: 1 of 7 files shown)


jan-janssen (approved):
Looks good to me

PR #137, opened 2021-01-13
Use starmap

pmrv:
The multiprocessing pool has a nice method that does exactly what the `job_wrap_function` does, so I threw it out.  The method is only available since python 3.3, but we're supporting >3.4 anyway.


jan-janssen (approved):
Looks good to me

PR #142, opened 2021-01-13
Add None check when copying reference jobs

pmrv:
Without this you get an error trying to copy a master without a reference job set.  I also suspect this `copy()` implementation should live on `GenericMaster`, any opinions?


jan-janssen (approved):
Looks good to me

PR #143, opened 2021-01-14
Introduce _new_instance method.

niklassiemer:
The _new_instance method returns a new instance of the current class via the __class__ attribute.
This allows to 'just' inherit methods like copy without the need to rewrite the same code again for each subclass.

PR #144, opened 2021-01-15
Blacklist functons in __dir__()

jan-janssen:

PR #146, opened 2021-01-15
Add decorators to hide functions from __dir__

pmrv:
This adds a method decorator and a class decorator to mark methods of a as "user facing" which will then be shown in `__dir__`.  The class decorator is also needed to make the method decorators work.  It has a toggle on each job instance to restore the default `__dir__`, maybe we rather want that to live in the `Settings` object.

Short example of usage (modulo naming)
```python
from pyiron_base import GenericJob, mangle_dir, user_facing

@mangle_dir
class A(GenericJob):

    def plumbing(self):
        return 42

    @user_facing
    def ui(self):
        return 23
```

Alternatives are #143, #147.


pmrv:
I think this is mostly compatible with #144, so if we decide on this, I'll synthesize something from both.

PR #135, opened 2021-01-13
Hdfio fix getitem

niklassiemer:
Closes #134

PR #149, opened 2021-01-18
Fix getitem hdfio

### LINKED ISSUE #148: FileHDFio should not return HDF object at /
With my current changes I introduced a new, maybe more severe, bug.
```
from pyiron import Project
pr = 'test'
job = pr['some_job_node']
job['..'] 
```
returns  `{'groups': ['some_job_node'], 'nodes': []}` of a ProjectHDFio.

My fault was to assume that
```
hdf=FileHDFio(file_name='test')
new_hdf = hdf['some_group']
old_hdf = new_hdf['..']
isinstance(old_hdf,hdf.__class__)
```
should be True.

niklassiemer:
closes #148

### DIFF
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -838,11 +838,11 @@ class FileHDFio(object):
                     )
                 # print('h5_path=', self.h5_path, 'item=', item, 'item_abs_lst=', item_abs_lst)
                 if (
-                        (item_abs_lst[0] == ".." or item_abs_lst[0] == ".") and len(item_abs_lst) == 1
+                        item_abs_lst[0] == "." and len(item_abs_lst) == 1
                 ):  # leaving the HDF5 file
                     return self._create_project_from_hdf5()
                 elif item_abs_lst[0] == "..":
-                    return self._create_project_from_hdf5()["/".join(item_abs_lst[1:])]
+                    return self._create_project_from_hdf5()["/".join(item_abs_lst)]
                 else:
                     hdf_object = self.copy()
                     hdf_object.h5_path = "/".join(item_abs_lst[:-1])
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -828,9 +828,13 @@ class FileHDFio(object):
                         return obj
                 raise ValueError("Unknown item: {}".format(item))
             else:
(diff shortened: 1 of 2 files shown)


jan-janssen:
As far as I understand this reverts some of the changes in https://github.com/pyiron/pyiron_base/pull/135 . I know that our current unit tests are not sufficient, but I want to make sure that these are the last changes and that we do not have to update it again next week. So I am wondering can we make it more explizit to make this section easier to understand for new user?

niklassiemer:
Absolutely true. I will try to add a few comments in the source to make clear what is going on and why a `FileHDFio` at the root path is nothing which should not be returned by `__getitem__`. Also the new unit tests should be clearer.

niklassiemer:
Actually, I just saw, that `rewrite_hdf5` does not work at the level of a `FileHDFio` level and has to be shifted to `ProjectHDFio`. Also `nodes` and `groups` do not work at the `FileHDFio` level , since they use a `_filter` which is not defined at this level.

niklassiemer:
I added some comments to document what the code does and some additional tests. However, the `test_copy_to` fails (commented out) and I am not sure if I made some mistake in usage or if there is a bug.

pmrv:
It's on my radar, will review after Journal Club.

pmrv:
I traced the error in the test to this

https://github.com/pyiron/pyiron_base/blob/467dee0850505675b3a9defab3442f8aca28c5a0/pyiron_base/generic/hdfio.py#L337

If the `destination.h5_path` is `'/'` then `dest_path` will be `''`, which in turn causes h5py to throw the error.  However the `File` class from h5py doesn't have a documented `copy()` method, so I'm not sure what to properly pass in instead.

jan-janssen:
@niklassiemer - Can you update me on the status of this pull request? I would like to release a new version of `pyiron_base` with the fixed dependencies and I am wondering if it makes sense to wait for this pull request or not.

niklassiemer:
@jan-janssen It fixes the bug and improves the tests in the `FileHDFio` class. Also, the new comments should explain what is done in the `getitem` method (At least I hope so). IMO this can be merged and the 'rest' handled in a follow-up PR, since I did not (yet) managed the `test_copy_to` test to pass. This, however, was also not tested before...

jan-janssen (approved):
Looks good to me - then I would suggest to add the fix in a separate pull request.

PR #151, opened 2021-01-21
Rename InputList to DataContainer

pmrv:
As we discussed in the software dev meeting, the name `InputList` is getting a bit outdated, so I changed it.  I'll leave this as a draft until @muh-hassani and I finish the work on reading/writing text files from the container.


max-hassani:
> 
> 
> As we discussed in the software dev meeting, the name `InputList` is getting a bit outdated, so I changed it. I'll leave this as a draft until @muh-hassani and I finish the work on reading/writing text files from the container.

The Last two days, I was busy with presentations I had to make for a PMD project meeting. If you give me an hour, I will commit and push my changes to this branch.

pmrv:
> > As we discussed in the software dev meeting, the name `InputList` is getting a bit outdated, so I changed it. I'll leave this as a draft until @muh-hassani and I finish the work on reading/writing text files from the container.
> 
> The Last two days, I was busy with presentations I had to make for a PMD project meeting. If you give me an hour, I will commit and push my changes to this branch.

Don't worry, no hurry on my side.

max-hassani:
I will also add a xml parser which returns a dictionary as the yaml parser does.

.ci_support/environment.yml
```diff
@@ -19,3 +19,4 @@ dependencies:
   - six >=1.14.0
   - sqlalchemy >=1.3.14
   - tqdm >=4.35.0
+  - xmltodict >=0.11.0
```
jan-janssen: Can you try using `defusedxml` because we have made good experience in parsing the VASP xml file. In addition we should add `pyyaml` if we want to support `yaml` import.

max-hassani: I understand that `defusedxml` introduces security over `xml` library, but casting the content of the parser to a dictionary is not straight forward in both cases. On the other hand `xmltodict`, [while it imports from `defusedexpat`](https://github.com/martinblech/xmltodict/blob/ae19c452ca000bf243bfc16274c060bf3bf7cf51/xmltodict.py#L5), make this casting really easy.

max-hassani: Nevertheless, using `defusedxml`, the following `dictify()` function can also turn the content of the parsed xml file to a dictionary:
```
from defusedxml import ElementTree
from copy import copy

def dictify(r,root=True):
    if root:
        return {r.tag : dictify(r, False)}
    d=copy(r.attrib)
    if str(r.text).find('\n') == -1:
        d[r.tag]=r.text
    for x in r.findall("./*"):
        if len(x) >= 1:
            if x.tag not in d:
                d[x.tag]=[]
            d[x.tag].append(dictify(x,False))
        else:
            d[x.tag] = x.text
    return d

root = ElementTree.parse('sample.xml').getroot()
final_dict = dictify(root)
```

jan-janssen: @sudarsan-surendralal Can you comment on the choice of `defusedxml` vs `xmltodict` ? Maybe it is only related to the broken xml format in VASP.

sudarsan-surendralal: `defusedxml` is more secure for parsing xml files and was suggested by codeacy which complained when we used `xmltodict` earlier. We used it to parse `vasprun.xml` files for every single VASP calculation and it works pretty hassle-free. But yes, you need to tweak it a little bit to get the contents into a dictionary form

max-hassani: @sudarsan-surendralal, thanks for your response. I guess in the `xml` files for vasp, you have expected keys, and you look for some values. Am I right? 
But in our case, we need to cast the parsed xml into a dictionary or some sort of data containers to be saved as Datacontainer object.   
Another point, here, we have used `xmltodict` and so far Codacy has not complained about it. It might be that `xmltodict` has been updated.

max-hassani: > `vasprun`

@sudarsan-surendralal, I could not find the code where you parsed `vasprun.xml` under `pyiron_atomistic/sphinx/`. Can you send me a link where `defusedxml` is used?

sudarsan-surendralal: > 
> 
> @sudarsan-surendralal, thanks for your response. I guess in the `xml` files for vasp, you have expected keys, and you look for some values. Am I right?

Yeah, that's true.

> But in our case, we need to cast the parsed xml into a dictionary or some sort of data containers to be saved as Datacontainer object.
> Another point, here, we have used `xmltodict` and so far Codacy has not complained about it. It might be that `xmltodict` has been updated.

Oh ok I was not aware of this. In that case maybe `xmltodict` is more appropriate for you. 

> @sudarsan-surendralal, I could not find the code where you parsed `vasprun.xml` under `pyiron_atomistic/sphinx/`. Can you send me a link where `defusedxml` is used?

It's in this file in the vasp directory https://github.com/pyiron/pyiron_atomistics/blob/master/pyiron_atomistics/vasp/vasprun.py

pyiron_base/generic/inputlist.py
```diff
@@ -1,656 +1,5 @@
...
-
-    def list_groups(self):
-        """
-        Return a list of keys to nested lists.
-
-        Returns:
-            :class:`list`: list of all keys to elements of :class:`InputList`.
-        """
-        return list(self.groups())
+from .datacontainer import DataContainer as InputList
```
jan-janssen: I would derive a class named `InputList` from the `DataContainer`, then once this class is used we can print a warning.

jan-janssen:
I am fine with using `xmltodict` we should just make sure that we merge with the master branch before to guarantee the dependencies are strict to work with dependabot.

jan-janssen:
@pmrv Do you think we can merge this on Friday? I guess it makes sense to release a new `pyiron_base` version and then update all classes which depend on it.

pmrv:
> @pmrv Do you think we can merge this on Friday? I guess it makes sense to release a new `pyiron_base` version and then update all classes which depend on it.

I guess so, biggest thing would be to update the outstanding PR, but only the atomicrex is the hard one.  @muh-hassani What's the status on the xml writing?

max-hassani:
> 
> 
> > @pmrv Do you think we can merge this on Friday? I guess it makes sense to release a new `pyiron_base` version and then update all classes which depend on it.
> 
> I guess so, biggest thing would be to update the outstanding PR, but only the atomicrex is the hard one. @muh-hassani What's the status on the xml writing?

I am working on it. write to yml was easy, but the xml one is a bit tricky. I will push it in less than an hour.

pmrv:
> > > @pmrv Do you think we can merge this on Friday? I guess it makes sense to release a new `pyiron_base` version and then update all classes which depend on it.
> > 
> > 
> > I guess so, biggest thing would be to update the outstanding PR, but only the atomicrex is the hard one. @muh-hassani What's the status on the xml writing?
> 
> I am working on it. I will push it in less than an hour.

No worries, we will want to discuss on Friday how to update the other repositories & PRs anyway.

max-hassani:
@pmrv, I added the two functions for writing to yml/xml file. One issue that I have is that datacontainer can have other format than a dictionary, but these two function requires that the datacontainer has some keys and some values in order to be stored in a yml/xml file. I think via `isinstance`, I can not check this, since it is an instance of datacontainer. But maybe checking whether it has keys as attribute could verify that the datacontainer has some keys and values.

pmrv:
> @pmrv, I added the two functions for writing to yml/xml file. One issue that I have is that datacontainer can have other format than a dictionary, but these two function requires that the datacontainer has some keys and some values in order to be stored in a yml/xml file. I think via `isinstance`, I can not check this, since it is an instance of datacontainer. But maybe checking whether it has keys as attribute could verify that the datacontainer has some keys and values.

You can check whether the `DataContainer` has non-integer keys with `.has_keys()`, but you should also be able to just convert integer keys to strings and then use them in the xml.

pyiron_base/generic/datacontainer.py
```diff
@@ -0,0 +1,730 @@
...
+
+    def write_yml(self, file_name):
+        with open(file_name, 'w') as output:
+            yaml.dump(self, output, default_flow_style=False)
+
+    def write_xml(self, file_name, attr_flag=False):
+        xml_data = dicttoxml(self, attr_type=attr_flag)
+        xmlfile = open(file_name, "w")
+        xmlfile.write(parseString(xml_data).toprettyxml())
+        xmlfile.close()
```
jan-janssen: Does this work with `with`?
```suggestion
        with open(file_name, "w") as xmlfile:
            xmlfile.write(parseString(xml_data).toprettyxml())
```

pmrv:
There should also be some tests with an simple example list and yaml/xml file to check that writing/parsing is lossless for all combinations of having only integer keys, some string keys and all string keys.

pyiron_base/generic/datacontainer.py
```diff
@@ -56,6 +57,7 @@ def _parse_yaml(file_name):
             warnings.warn(exc)
             return {}
 
+
 def _parse_xml(self, file_name, wrap=False):
```
jan-janssen:
```suggestion
def _parse_xml(file_name, wrap=False):
```
Is there are particular reason why the parse functions are not static functions of the class.

pmrv: I didn't want to clutter the interface too much.  If you don't want the function outside the class, I can define them inside the `read()`. 

The `self` slipped through when copying, will fix.

muh-hassani:
@pmrv, I have modified `datacontainer.py` to fix some issues with reading from xml files. But I have faced a problem, which you might have an idea how to fix it. I have tried to picture the issue in the following code snippet:
```
from dicttoxml import dicttoxml
from defusedxml.minidom import parseString
from pyiron_base.generic.datacontainer import _conv_OrderedDict_to_dict
import xmltodict
import warnings
dict_data = {"a":2, "b":{"c":4, "d":{"e":5,"f":6}}, 'g':[1,2,3], 'n':'None'}
sample_container = DataContainer(dict_data)
sample_container.write('output.xml')
container_read_data_xml = DataContainer(table_name="input_data")
container_read_data_xml.read(file_name="output.xml")
```
The content of the two datacontainers, `container_read_data_xml` and `sample_container ` are different with regards to the type of data stored, but when I try to convert them to their original type: `container_read_data_xml.to_builtin()` and `sample_container.to_builtin()`, the results  are the same! Do you have an idea how to fix this?

max-hassani:
@pmrv, by adding another function, I fixed the issue with the "item" key in the case of a nested list. If you find another solution, feel free to remove it.

pmrv:
> @pmrv, by adding another function, I fixed the issue with the "item" key in the case of a nested list. If you find another solution, feel free to remove it.

Thanks for fixing this!  I'm ok leaving this as is, I refactored the functions into their own module and will update the tests tomorrow.

jan-janssen:
I fixed the tests, so from my perspective it is now ready to merge. Once this is merged then we should release `pyiron_base` version `0.2.0`.

jan-janssen (approved):

pmrv:
> I fixed the tests, so from my perspective it is now ready to merge. Once this is merged then we should release `pyiron_base` version `0.2.0`.

I did some more refactors, will merge after I pushed them.

pmrv:
I had to force push, because the fixed didn't line up anymore of the refactored `fileio`.  I also found one more bug, when writing dicts to xml with "numeric key" like `"0"` and added a test case for it.

From my side this is ready to go.

jan-janssen:
> From my side this is ready to go.

The I would suggest we merge it now and release a new `pyiron_base` version to push the migration for `pyiron_atomistics` and `pyiron_contrib`.

pmrv:
> > From my side this is ready to go.
> 
> The I would suggest we merge it now and release a new `pyiron_base` version to push the migration for `pyiron_atomistics` and `pyiron_contrib`.

I'll update and merge #70 first and then bump the tag.

PR #156, opened 2021-01-28
Add ProjectBrowser feature

niklassiemer:
This adds a GUI element to any pyiron Project, a browser to access the data system of the Project via it's list_* functions.
In addition the display of different file types is enhanced via a new load_file function.

### DIFF
--- a/pyiron_base/generic/filedata.py
+++ b/pyiron_base/generic/filedata.py
@@ -1,0 +1,166 @@
+import json
+import os
+import pandas
+
+from pyiron_base import ImportAlarm, FileHDFio
+
+_has_imported = {}
+import_alarm = ImportAlarm()
+_not_imported = ''
+try:
+    from PIL import Image
+    _has_imported['PIL'] = True
+except ImportError:
+    _has_imported['PIL'] = False
+try:
+    from IPython.core.display import display
+    _has_imported['IPython'] = True
+except ImportError:
+    _has_imported['IPython'] = False
+if all(_has_imported.values()):
+    pass
+else:
+    for k, j in _has_imported.items():
+        if j and len(_not_imported) > 0:
+            _not_imported += ', '
+        if j:
+            _not_imported += k
+    import_alarm = ImportAlarm(
+        "Reduced functionality, since " + _not_imported + " could not be imported."
+    )
+
+
+def load_file(filename):
+    """
+        Load the file and return an appropriate object containing the data.
+
+        Args:
+            filename (str): path to the file to be displayed.
+    """
+    _, filetype = os.path.splitext(filename)
+    if filetype.lower() in ['.h5', '.hdf']:
+        return FileHDFio(file_name=filename)
+    if filetype.lower() in ['.json']:
+        return _load_json(filename)
+    elif filetype.lower() in ['.txt']:
+        return _load_txt(filename)
+    elif filetype.lower() in ['.csv']:
+        return _load_csv(filename)
+    elif _has_imported['PIL'] and filetype.lower() in Image.registered_extensions():
+        return _load_img(filename)
+    else:
+        return _load_default(filename)
+
+
+def _load_txt(file):
+    with open(file) as f:
+        return f.readlines()
+
+
+def _load_json(file):
+    with open(file) as f:
+        return json.load(f)
+
+
+def _load_csv(file):
+    return pandas.read_csv(file)
+
+
+def _load_img(file):
+    return Image.open(file)
+
+
+def _load_default(file):
+    try:
+        return _load_txt(file)
+    except UnicodeDecodeError:
+        return file
+
+
+class FileData:
+
+    """ FileData stores an instance of a data file, e.g. a single Image from a measurement """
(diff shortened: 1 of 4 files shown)


niklassiemer:
Does anyone have an idea why my test suite is not happy with the `self.assertRaises`, however, excepts the `try`-`except`?
```
        # self.assertRaises(ValueError, FileData())
        try:
            FileData()
        except ValueError:
            pass
        else:
            raise
```
----EDIT---- 
Ok found it! I was using the wrong syntax... One has to use either
```
self.assertRaises(ValueError, FileData)  # Not calling the function
```
or
```
with self.assertRaises(ValueError):
    FileData()
```

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,177 @@
+import json
```
jan-janssen, line 1: Please add the copyright we use in other classes.

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,177 @@
...
+__email__ = "siemer@mpie.de"
+__status__ = "development"
+__date__ = "Feb 02, 2021"
+
+
+_has_imported = {}
+import_alarm = ImportAlarm()
+_not_imported = ''
+try:
+    from PIL import Image
```
jan-janssen, line 23: Should we add `pillow` to the requirements?

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,177 @@
...
+
+def _load_img(file):
+    return Image.open(file)
+
+
+def _load_default(file):
+    try:
+        return _load_txt(file)
+    except UnicodeDecodeError:
+        return file
```
jan-janssen: I guess you can include the functions in the main function 

```suggestion
def load_file(filename):
      """
          Load the file and return an appropriate object containing the data.
 
          Args:
              filename (str): path to the file to be displayed.
      """
      def _load_txt(file):
          with open(file) as f:
              return f.readlines()
      
      def _load_json(file):
          with open(file) as f:
              return json.load(f)
      
      def _load_csv(file):
          return pandas.read_csv(file)
 
      def _load_img(file):
          return Image.open(file)
 
 
      def _load_default(file):
          try:
              return _load_txt(file)
          except UnicodeDecodeError:
              return file
     
      _, filetype = os.path.splitext(filename)
      if filetype.lower() in ['.h5', '.hdf']:
          return FileHDFio(file_name=filename)
      if filetype.lower() in ['.json']:
          return _load_json(filename)
      elif filetype.lower() in ['.txt']:
          return _load_txt(filename)
      elif filetype.lower() in ['.csv']:
          return _load_csv(filename)
      elif _has_imported['PIL'] and filetype.lower() in Image.registered_extensions():
          return _load_img(filename)
      else:
          return _load_default(filename)
```

pyiron_base/generic/filedata.py
niklassiemer: I don't know. For a the display of display and so on this is of course very nice. However, since we 'just' convert a image file to an image object and return this object and don't (yet) provide functionality to _do_ anything with this object I was trying to keep our dependency list shorter.

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,177 @@
...
+    @property
+    def data(self):
+        """Return the associated data."""
+        if self._hasdata:
+            return self._data
+        else:
+            return load_file(self.source)
+
+
+class DisplayItem:
```
jan-janssen, line 136: Maybe I missed it, but where is this class used?

pyiron_base/project/project_browser.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def _on_click_file(self, filename):
+        filepath = os.path.join(self.path, filename)
+        self.output.clear_output(True)
+        try:
+            data = self.project[filename]
+        except(KeyError):
+            data = None
+        with self.output:
+            if data is not None and str(type(data)) == "<class 'boto3.resources.factory.s3.Object'>":
```
jan-janssen, line 256: This logic should be in a separate class and should not be hidden in the GUI logic

pyiron_base/generic/filedata.py
niklassiemer: Not at all any more, I'll throw it out.

jan-janssen:
I have the feeling here are two things combined. One it would be nice to convert any file to a pyiron and this could be done with a very general object like `FileData`. The visualisation is then the second part.

niklassiemer:
> I have the feeling here are two things combined. One it would be nice to convert any file to a pyiron and this could be done with a very general object like `FileData`. The visualisation is then the second part.

I am fine with splitting the PR into two.

pyiron_base/project/project_browser.py
niklassiemer: I will throw it out and try to come up with a nice self-representation of s3 objects. However, this probably requires a wrapped s3-object and I was not sure if this is a good idea.

PR #171, opened 2021-02-01
Scriptjob example

### LINKED ISSUE #170: ScriptJob example?
I was hoping to show @vijjubhuva how to submit entire notebooks, but am having trouble finding documentation for this. Critically, when I try to load the `submit_job.ipynb` from our [mpie handbook page](https://github.com/pyiron/pyiron-meeting-mpie/tree/03a465ca112b8fed0436565bf4c38d2dd86b5c7a/handbook/6-file-and-job-management/6-6-submit-jupyter-notebook) I get an error loading the file. I'm not able to find an example of this elsewhere, am I simply missing it?

Would someone who is familiar with the operation of `ScriptJob` be willing to go into [its docstring](https://github.com/pyiron/pyiron_base/blob/fe177f23926d8405a00590d77a60b76cf54ddbf1/pyiron_base/job/script.py#L27) and give it some modernization, i.e. cleaning up/reformatting the attributes and providing one or two very simple examples of operation? @raynol-dsouza @samwaseda, have either of you or a student you work with got sufficient familiarity to make this task straightforward?

raynol-dsouza:
Add an example of how to submit notebooks using ScriptJob

### DIFF
--- a/pyiron_base/job/script.py
+++ b/pyiron_base/job/script.py
@@ -74,8 +74,8 @@ class ScriptJob(GenericJob):
 
                 # Optional: get an input value from 'submit_example_job.ipynb', the notebook which submits
                 #   'example.ipynb'
-                external_input = pr.get_external_input()
-                job.server.cores = external_input['number_of_cores']
+                number_of_cores = pr.data.number_of_cores
+                job.server.cores = number_of_cores
 
                 job.run()  # run the job
 
@@ -93,13 +93,14 @@ class ScriptJob(GenericJob):
                                                          # directory as 'submit_example_job.ipynb'
 
                 # Optional: to submit the notebook to a queueing system
-                number_of_cores = 4  # number of cores to be used
+                number_of_cores = 1  # number of cores to be used
                 scriptjob.server.cores = number_of_cores
                 scriptjob.server.queue = 'cmfe'  # specify the queue to which the ScriptJob job is to be submitted
                 scriptjob.server.run_time = 120  # specify the runtime limit for the ScriptJob job in seconds
 
                 # Optional: save an input, such that it can be accessed by 'example.ipynb'
-                scriptjob.input['number_of_cores'] = number_of_cores
+                pr.data.number_of_cores = number_of_cores
+                pr.data.write()
 
                 # run the ScriptJob job
                 scriptjob.run()
--- a/pyiron_base/job/script.py
+++ b/pyiron_base/job/script.py
@@ -40,10 +40,15 @@ class ScriptJob(GenericJob):
                     pr = Project('example')
                     job = pr.create_job(job_type=pr.job_type.Lammps, job_name='job')  # we name the job 'job'
                     job.structure = pr.create_ase_bulk('Fe')  # specify structure
-                    job.server.cores = 4  # Optional: use 4 cores to run this job
+
+                    # Optional: get an input value from 'submit_example_job.ipynb', the notebook which submits
+                    #   'example.ipynb'
+                    external_input = pr.get_external_input()
+                    job.server.cores = external_input['number_of_cores']
+
                     job.run()  # run the job
 
-                    # save a custom output
+                    # save a custom output, that can be used by the notebook 'submit_example_job.ipynb'
                     job['user/my_custom_output'] = 16
                     ```
 
@@ -58,9 +63,13 @@ class ScriptJob(GenericJob):
                                                              # directory as 'submit_example_job.ipynb'
 
                     # Optional: to submit the notebook to a queueing system
+                    number_of_cores = 4  # number of cores to be used
+                    scriptjob.server.cores = number_of_cores
                     scriptjob.server.queue = 'cmfe'  # specify the queue to which the ScriptJob job is to be submitted
-                    scriptjob.server.cores = 4  # specify the number of cores to be used
                     scriptjob.server.run_time = 120  # specify the runtime limit for the ScriptJob job in seconds
+
+                    # Optional: save an input, such that it can be accessed by 'example.ipynb'
+                    scriptjob.input['number_of_cores'] = number_of_cores
 
                     # run the ScriptJob job
                     scriptjob.run()
--- a/pyiron_base/job/script.py
+++ b/pyiron_base/job/script.py
@@ -31,6 +31,39 @@ class ScriptJob(GenericJob):
         project (ProjectHDFio): ProjectHDFio instance which points to the HDF5 file the job is stored in
         job_name (str): name of the job, which has to be unique within the project
 
+    Example:
+        To submit a notebook as a ScriptJob job to the puiron job management system,
+
+            1. Create the notebook to be submitted, for ex. 'example.ipynb', and save it -- Can contain any code like:
+                ```
+                from pyiron import Project
+                pr = Project('.')
+                job = pr.create_job(pr.job_type.Lammps, 'job')
+                job.structure = pr.create_ase_bulk('Fe')
+                job.run()
+                print(job.status)
+                print(job.output.energy_pot[-1])
+                ```
+
+            2. Create the notebook which submits the notebook created in 1, for ex. 'sumbit_example_job.ipynb',
+                which can have the following code:
+                ```
+                from pyiron import Project
+                pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
+                scriptjob = pr.create_job(pr.job_type.ScriptJob, 'scriptjob')  # create a ScriptJob named 'scriptjob'
+                scriptjob.script_path = 'example.ipynb'  # specify the PATH to the notebook you want to submit.
+                                                         # In this example case, 'example.ipynb' is in the same
+                                                         # directory as 'sumbit_example_job.ipynb'
+
+                # to submit the notebook to a queueing system
+                scriptjob.server.queue = 'cmfe'  # specify the queue to which the ScriptJob job is to be submitted
+                scriptjob.server.cores = 4  # specify the number of cores to be used
+                scriptjob.server.run_time = 120  # specify the runtime limit for the ScriptJob job in seconds
+
+                # run the ScriptJob job
+                scriptjob.run()
+                ```
+
     Attributes:
 
         attribute: job_name
--- a/pyiron_base/job/script.py
+++ b/pyiron_base/job/script.py
@@ -46,6 +46,7 @@ class ScriptJob(GenericJob):
                 pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
                 scriptjob = pr.create.job.ScriptJob('scriptjob')  # create a ScriptJob named 'scriptjob'
                 scriptjob.script_path = 'example.ipynb'  # specify the PATH to the notebook you want to submit.
+                ```
 
         Step 3. Check the job table to get details about 'scriptjob' by using:
                 ```
@@ -114,7 +115,8 @@ class ScriptJob(GenericJob):
         Step 4. Reload and analyse the child 'job/s': If the status of a child 'job' is 'finished', it can be loaded
                 into the 'submit_example_job.ipynb' notebook using:
                 ```
-                job = pr.load('job')  # remember in 1., we wanted to run a job named 'job', which has now 'finished'
+                job = pr.load('job')  # remember in Step 1., we wanted to run a job named 'job', which has now
+                                      # 'finished'
                 ```
                 this loads 'job' into the 'submit_example_job.ipynb' notebook, which can be then used for analysis,
                 ```
--- a/pyiron_base/job/script.py
+++ b/pyiron_base/job/script.py
@@ -31,67 +31,96 @@ class ScriptJob(GenericJob):
         project (ProjectHDFio): ProjectHDFio instance which points to the HDF5 file the job is stored in
         job_name (str): name of the job, which has to be unique within the project
 
-    Example:
-        To submit a notebook as a ScriptJob job to the pyiron job management system,
+    Simple example:
+        Step 1. Create the notebook to be submitted, for ex. 'example.ipynb', and save it -- Can contain any code like:
+                ```
+                import json
+                with open('script_output.json','w') as f:
+                    json.dump({'x': [0,1]}, f)  # dump some data into a JSON file
+                ```
 
-            1. Create the notebook to be submitted, for ex. 'example.ipynb', and save it -- Can contain any code like:
-                    ```
-                    from pyiron import Project
-                    pr = Project('example')
-                    job = pr.create_job(job_type=pr.job_type.Lammps, job_name='job')  # we name the job 'job'
-                    job.structure = pr.create_ase_bulk('Fe')  # specify structure
+        Step 2. Create the submitter notebook, for ex. 'submit_example_job.ipynb', which submits 'example.ipynb' to the
+                pyiron job management system, which can have the following code:
+                ```
+                from pyiron_base import Project
+                pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
+                scriptjob = pr.create.job.ScriptJob('scriptjob')  # create a ScriptJob named 'scriptjob'
+                scriptjob.script_path = 'example.ipynb'  # specify the PATH to the notebook you want to submit.
 
-                    # Optional: get an input value from 'submit_example_job.ipynb', the notebook which submits
-                    #   'example.ipynb'
-                    external_input = pr.get_external_input()
-                    job.server.cores = external_input['number_of_cores']
+        Step 3. Check the job table to get details about 'scriptjob' by using:
+                ```
+                pr.job_table()
+                ```
 
-                    job.run()  # run the job
+        Step 4. If the status of 'scriptjob' is 'finished', load the data from the JSON file into the
+                'submit_example_job.ipynb' notebook by using:
+                ```
+                import json
+                with open(scriptjob.working_directory + '/script_output.json') as f:
+                    data = json.load(f)  # load the data from the JSON file
+                ```
 
-                    # save a custom output, that can be used by the notebook 'submit_example_job.ipynb'
-                    job['user/my_custom_output'] = 16
-                    ```
+    More sophisticated example:
+        The script in ScriptJob can also be more complex, e.g. running its own pyiron calculations.
+        Here we show how it is leveraged to run a multi-core atomistic calculation.
 
-            2. Create the notebook which submits the notebook created in 1., for ex. 'submit_example_job.ipynb',
-                which can have the following code:
-                    ```
-                    from pyiron import Project
-                    pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
-                    scriptjob = pr.create_job(pr.job_type.ScriptJob, 'scriptjob')  # create a ScriptJob named 'scriptjob'
-                    scriptjob.script_path = 'example.ipynb'  # specify the PATH to the notebook you want to submit.
-                                                             # In this example case, 'example.ipynb' is in the same
-                                                             # directory as 'submit_example_job.ipynb'
+        Step 1. 'example.ipynb' can contain pyiron_atomistics code like:
+                ```
+                from pyiron_atomistics import Project
+                pr = Project('example')
+                job = pr.create.job.Lammps('job')  # we name the job 'job'
+                job.structure = pr.create.structure.ase_bulk('Fe')  # specify structure
 
-                    # Optional: to submit the notebook to a queueing system
-                    number_of_cores = 4  # number of cores to be used
-                    scriptjob.server.cores = number_of_cores
-                    scriptjob.server.queue = 'cmfe'  # specify the queue to which the ScriptJob job is to be submitted
-                    scriptjob.server.run_time = 120  # specify the runtime limit for the ScriptJob job in seconds
+                # Optional: get an input value from 'submit_example_job.ipynb', the notebook which submits
+                #   'example.ipynb'
+                external_input = pr.get_external_input()
+                job.server.cores = external_input['number_of_cores']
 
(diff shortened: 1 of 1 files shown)


pyiron_base/job/script.py
```diff
@@ -31,6 +31,39 @@ class ScriptJob(GenericJob):
         project (ProjectHDFio): ProjectHDFio instance which points to the HDF5 file the job is stored in
         job_name (str): name of the job, which has to be unique within the project
 
+    Example:
+        To submit a notebook as a ScriptJob job to the pyiron job management system,
+
+            1. Create the notebook to be submitted, for ex. 'example.ipynb', and save it -- Can contain any code like:
+                ```
+                from pyiron import Project
+                pr = Project('.')
```
samwaseda: Since we don't really recommend to use `Project('.')` in general, it's probably better either to give a name or use `default_working_directory=True`.

pyiron_base/job/script.py
```diff
@@ -31,6 +31,39 @@ class ScriptJob(GenericJob):
...
+                from pyiron import Project
+                pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
+                scriptjob = pr.create_job(pr.job_type.ScriptJob, 'scriptjob')  # create a ScriptJob named 'scriptjob'
+                scriptjob.script_path = 'example.ipynb'  # specify the PATH to the notebook you want to submit.
+                                                         # In this example case, 'example.ipynb' is in the same
+                                                         # directory as 'submit_example_job.ipynb'
+
+                # to submit the notebook to a queueing system
+                scriptjob.server.queue = 'cmfe'  # specify the queue to which the ScriptJob job is to be submitted
+                scriptjob.server.cores = 4  # specify the number of cores to be used
```
samwaseda: The number of cores is not consistent with what's used in the notebook.

pyiron_base/job/script.py
```diff
@@ -31,6 +31,39 @@ class ScriptJob(GenericJob):
...
+                scriptjob.server.run_time = 120  # specify the runtime limit for the ScriptJob job in seconds
+
+                # run the ScriptJob job
+                scriptjob.run()
+                ```
```
samwaseda, line 107: Generally I find it also helpful to know how to retrieve or analyse output data, but I wouldn't insist on it because it might be trivial if the user looks at `job_table()`.

samwaseda (commented):
Nice! I really like it!

pyiron_base/job/script.py
raynol-dsouza: I agree, it does look a little odd. I will change this.

pyiron_base/job/script.py
raynol-dsouza: Will change the example notebook to take 4 cores as input.

pyiron_base/job/script.py
raynol-dsouza: I will add a bit more documentation on how to check if the notebook is submitted/running/finished.

pmrv: Agree, please add how to retrieve the results from the notebooks.  If I remember an issue from last year, it's not quite intuitive.

raynol-dsouza: I usually do it by reloading the job executed within the submitted notebook. Is there another way of doing this?

samwaseda: Maybe you can mention in the notebook to submit (i.e. for the LAMMPS job) that the user can store data in `job['user']`, by something like:

```
job['user/output_value'] = some_output
```

And if `Project(default_working_directory=True)` is used, the job can be loaded from the submitter notebook via the job name just like any other job in the same notebook. If a different project name is used, I don't know how it works (that's something I never do).

raynol-dsouza: It so happens that ScriptJob does not care in which project 'job' is saved, as long as it has been executed in the submitted notebook (here, 'example.ipynb'). Which I think is really nice! :)

samwaseda:
Maybe it would be good to mention that we have also `scriptjob.input['something']`, which in this example fits well if we have something like:

```
number_of_cores = 4
scriptjob.server.cores = number_of_cores
scriptjob.input['number_of_cores'] = number_of_cores
```

and you can get the value in the notebook via:

```
external_input = pr.get_external_input()
job.server.cores = external_input['number_of_cores']
```

niklassiemer:
I really like the documentation! However, right now we reflect pyiron-only usage and Lammps is actually not part of `pyiron_base`. I just think we would run a lammps job without wrapping it in a scrip job, wouldn't we? It is very nice we actually can do it so easy (I didn't know about `get_external_input()` for example). 
Especially for someone just wanting to run any script he/she has with this functionality of `pyiron_base` it could be very confusing to see the interplay.

raynol-dsouza:
@niklassiemer The 'example' notebook is just a simple 'pyiron job' example. It can be any complicated notebook, which can exchange data with the submit notebook, as Sam has mentioned in his comments (I was also unaware of this fancy exchange functionality before today!). I do agree with you that the documentation is very 'pyiron job' specific, but I think it includes most of what one would want to do with the ScriptJob functionality.

niklassiemer:
I understand, however, I fear someone 'external' wanting to use this for his easy script could be overwhelmed. I do not know if it is reasonable, however, we could give two examples? One very easy one computing, just a polynomial or the like (which of course could not be loaded!) and the fancy pyiron one? 
I just think for the non-pyiron case I would like something like 
```
# ScriptJob
from pyiron import Project
import json
pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
scriptjob = pr.create_job(pr.job_type.ScriptJob, 'scriptjob')  # create a ScriptJob named 'scriptjob'
scriptjob.script_path = 'example.ipynb'  # specify the PATH to the notebook you want to submit.
pr.job_table()
# After 'finished':
with open('script_output.json') as f:
  parsed_output_from_script = json.load(f)
scriptjob['output/user'] = parsed_output_from_script 
```
```
# example.ipynb
import numpy as np
import json
x=np.linspace(0, 10, 10**4) 
y= 4*x**2 + 3*x + 1
with open('script_output.json') as f:
  json.dump({'x': x.to_list(), 'y': y.to_list()})
```

raynol-dsouza:
I would very much like having two examples, since @niklassiemer, your example doesn't involve any 'jobs', but at the same time shows that simpler notebooks without 'jobs' can also be executed using ScriptJob. But sadly, the docstring will only get longer. :(

However, I am open to having just one or both examples. If we have a consensus for one or the other, I can modify the docs accordingly.

niklassiemer:
If we use example a) (just a script) and b) (integrated pyiron) it would not grow a lot, however, it will be a pretty long Docstring in each case... I do not know which is the best procedure.

samwaseda:
I don't think it's a problem to have a very long doc string on the class level and actually I support @niklassiemer's claim that there should be an example which runs without plugins. The only one thing from my side is his example could be maybe simplified to the minimum to reduce as many lines as possible.

As always, I would leave the decision to the community so don't feel obliged to do anything with my comments.

niklassiemer:
We could of course reduce the number of lines to minimum, e.g.
```
# example.ipynb
with open('script_output.txt','w') as f:
    f.writelines([str(x) + '  ' + str(4*x**2 + 2*x + 3)+' \n' for x in [0.01*i for i in range(0,1000)]])
```
which gets a bit nasty to read... 
```
# ScriptJob
from pyiron import Project
pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
scriptjob = pr.create_job(pr.job_type.ScriptJob, 'scriptjob')  # create a ScriptJob named 'scriptjob'
scriptjob.script_path = 'example.ipynb'  # specify the PATH to the notebook you want to submit.
pr.job_table()
# After 'finished':
with open('output','r') as f:
    xy_pairs = [[float(line.split()[0]), float(line.split()[1])] for line in f.readlines()]
scriptjob['output/user'] = xy_pairs
```

samwaseda:
I meant something like this:

```
# example.ipynb
import json
with open('script_output.json') as f:
    json.dump({'x': [0,1]})
```

pyiron_base/job/script.py
```diff
@@ -31,6 +31,68 @@ class ScriptJob(GenericJob):
...
         job_name (str): name of the job, which has to be unique within the project
 
+    Example:
+        To submit a notebook as a ScriptJob job to the pyiron job management system,
+
+            1. Create the notebook to be submitted, for ex. 'example.ipynb', and save it -- Can contain any code like:
+                    ```
+                    from pyiron import Project
+                    pr = Project('example')
+                    job = pr.create_job(job_type=pr.job_type.Lammps, job_name='job')  # we name the job 'job'
```
liamhuber: Nit: uses old creation syntax; please use new `pr.create.job.Lamms('job')`.

pyiron_base/job/script.py
```diff
@@ -31,6 +31,68 @@ class ScriptJob(GenericJob):
         project (ProjectHDFio): ProjectHDFio instance which points to the HDF5 file the job is stored in
         job_name (str): name of the job, which has to be unique within the project
 
+    Example:
+        To submit a notebook as a ScriptJob job to the pyiron job management system,
+
+            1. Create the notebook to be submitted, for ex. 'example.ipynb', and save it -- Can contain any code like:
+                    ```
+                    from pyiron import Project
```
liamhuber: Shouldn't we be importing fmor `pyiron.atomistics`?

pyiron_base/job/script.py
```diff
@@ -31,6 +31,68 @@ class ScriptJob(GenericJob):
...
+                    pr = Project('example')
+                    job = pr.create_job(job_type=pr.job_type.Lammps, job_name='job')  # we name the job 'job'
+                    job.structure = pr.create_ase_bulk('Fe')  # specify structure
+
+                    # Optional: get an input value from 'submit_example_job.ipynb', the notebook which submits
```
liamhuber: After reading the discussion, I think I come down on the side that the external input method is making this docstring too complex. I'd rather drop these three lines. (Keep the core specification, but just make it a number)

pyiron_base/job/script.py
```diff
@@ -31,6 +31,68 @@ class ScriptJob(GenericJob):
...
+                    job.run()  # run the job
+
+                    # save a custom output, that can be used by the notebook 'submit_example_job.ipynb'
+                    job['user/my_custom_output'] = 16
+                    ```
+
+            2. Create the notebook which submits the notebook created in 1., for ex. 'submit_example_job.ipynb',
+                which can have the following code:
+                    ```
+                    from pyiron import Project
```
liamhuber: Here it should indeed be `from pyiron`, but `from pyiron.base` might be even more explicit? Not sure if we have the syntax in `pyiron/pyiron` set up that way yet though ;)

liamhuber: still, could be `from pyiron_base` safely

pyiron_base/job/script.py
```diff
@@ -31,6 +31,68 @@ class ScriptJob(GenericJob):
...
+                    pr = Project('scriptjob_example')  # save the ScriptJob in the 'scriptjob_example' project
+                    scriptjob = pr.create_job(pr.job_type.ScriptJob, 'scriptjob')  # create a ScriptJob named 'scriptjob'
```
liamhuber: Nit: use new syntax

pyiron_base/job/script.py
```diff
@@ -31,6 +31,68 @@ class ScriptJob(GenericJob):
...
 
+    Example:
+        To submit a notebook as a ScriptJob job to the pyiron job management system,
+
+            1. Create the notebook to be submitted, for ex. 'example.ipynb', and save it -- Can contain any code like:
+                    ```
+                    from pyiron import Project
+                    pr = Project('example')
+                    job = pr.create_job(job_type=pr.job_type.Lammps, job_name='job')  # we name the job 'job'
+                    job.structure = pr.create_ase_bulk('Fe')  # specify structure
```
liamhuber: Nit: use new syntax

liamhuber (changes requested):
Very nice! I'm super pleased with the depth and breadth of care this PR is getting :)

Like Sam, I am not concerned about having a long class-level docstring. I would rather use the very simple json-writing example in step 1, and simplify step 2 accordingly (e.g. no need to specify cores as the default of `cores=1` is fine). Then, take your more complex example and include it below the first example with the heading "More sophisticated example:\nThe script in ScriptJob can also be more complex, e.g. running its own pyiron calculations. Here we show how it is leveraged to run a multi-core atomistic calculation." or similar.

pyiron_base/job/script.py
niklassiemer: Absolutely!

pyiron_base/job/script.py
samwaseda: The problem with not having these lines here is that there's no way for the user to know that such a functionality exists (or at least I don't know how to look it up without leaving the notebook). And I'm pretty sure that many people want to pass input arguments from the submitter to the job notebook. So I'd say making them google it is bigger evil than having three (three?) lines here.

raynol-dsouza: I would also side with @samwaseda on this. As a user, I wouldn't know if this functionality existed unless specified in some example notebook. But since we have a 'docstring version' of the notebook here in the docstring of ScriptJob itself, I don't see a better way of advertising `pr.get_external_input()` within the context of submitting notebooks.

pmrv: This would still be in a sphinx-parsed docstring, so just make a reference like this
```
See :method:`~.Project.get_external_input()` for further reading how to pass input to a notebook.
```

pyiron_base/job/script.py
raynol-dsouza: Is it not `pyiron_atomistics`?

niklassiemer: Right now, all three variants work (`from pyiron`, `from pyiron_atomistics`, and `from pyiron.atomistics` all result in the same `Project`). In the future we will deprecate `from_pyiron`. The question is now if we use the original pyiron package as main module and the atomistic as sub-module (`pyiron.atomistics`) or if we directly use the `pyiron_atomistics` repository (both will work in the future - `pyiron_atomistics` will be accessible via `pyiron.atomistics`.
Since I would like to promote the overarching `pyiron` framework with its several sub-modules (`atomistics`, `experimental`, `continuum`), I would go for `pyiron.atomistics`

pyiron_base/job/script.py
liamhuber: I like this even less now.



> The problem with not having these lines here is that there's no way for the user to know that such a functionality exists (or at least I don't know how to look it up without leaving the notebook). 

If this is really the case, then there is a fundamental problem with `Project.get_external_output` and it should be changed. 

What is the actual intended use of this method? Is it *only* for `ScriptJob` or is there another application I'm missing? It's a public method on our base-level `Project` object; if it's only application is for `ScriptJob` then we really need to figure out a way to relocate it, or make a special `ScriptProject` or something.

I'm extremely uncomfortable with the idea that the only way to get meaningfully informed about a `pyiron_base.Project` method is by looking at the docstring of `ScriptJob`. That smells to high heaven.

raynol-dsouza:
I have a new concern stemming from the simple example that @niklassiemer and @samwaseda suggested. If my 'example' notebook looks like this:
```
# example.ipynb

import json
with open('script_output.json','w') as f:
    json.dump({'x': [0,1]}, f)
```

and the 'submit' notebook looks like this:
```
# submit.ipynb

from pyiron.base import Project
pr = Project('scriptjob_example') 
scriptjob = pr.create.job.ScriptJob('scriptjob')
scriptjob.script_path = 'scriptjob_example.ipynb'
scriptjob.run()
```

then to access the json file written, I need to do this:
```
# submit.ipynb continued...

import json
with open('scriptjob_example/scriptjob_hdf5/scriptjob/script_output.json') as f:
    data = json.load(f)
```

This line `'scriptjob_example/scriptjob_hdf5/scriptjob/script_output.json'` bugs me a lot. Of course I can load the project string `'scriptjob_example'` using `pr.name`,  and the job string `scriptjob` as `scriptjob.jobname`, but I would be inclined NOT to have to do that. Is this a limitation of `Scriptjob`, that it has to save all of the output files inside of `'scriptjob_hdf5/scriptjob/'`?

For loading pyiron jobs, it doesn't matter, since the child jobs are already linked to the master project. It only applies when loading output files that are not 'jobs'.

niklassiemer:
> This line `'scriptjob/scriptjob_hdf5/scriptjob/script_output.json'` bugs me a lot. Of course I can load the string `'scriptjob'` using `pr.name`, but I would be inclined NOT to have to do that. Is this a limitation of `Scriptjob`, that it has to save all of the output files inside of `'scriptjob_hdf5/scriptjob/'`?
> 

We have the option to just use
`with open(scriptjob.working_directory + '/script_output.json') as f:`

raynol-dsouza:
Ah yes. Very good. This helps a lot!

pyiron_base/job/script.py
samwaseda: Currently, it occurs only in `ScriptJob` that we need a project level external input.

Actually I don't really understand what troubles you so much.

pyiron_base/job/script.py
raynol-dsouza: `pyiron.atomistics` doesn't seem to have `Project` at all, but `pyiron_atomistics` does... I'm a bit confused.

pyiron_base/job/script.py
pmrv: Agree with @liamhuber, since it's a ScriptJob specific feature it should live in there as well, but I'm not sure where to put it.  The problem is that the notebook running in the ScriptJob does not have a reference to itself, but only one to the Project.

pyiron_base/job/script.py
niklassiemer: You are right.... I raised an issue in `pyiron` since this should become the nomenclature to use. However, right now we have a name clash between the `pyiron_atomistics.atomistics` module and the `pyiron_atomistics` of which the first is bound to `pyiron.atomistics` for backwards compatability...

pyiron_base/job/script.py
samwaseda: It's *currently* a `ScriptJob` specific feature. Since it's conceptually not necessary elsewhere, I don't see a problem with putting it only in this example here.

pyiron_base/job/script.py
raynol-dsouza: Cool! I will follow that issue and update the docs as soon as it has been fixed.

pyiron_base/job/script.py
liamhuber: `pyiron_base.Project` is sacrosanct because it's the most fundamental object in the user hierarchy. Thus, decorating it with a public method that only applies to one specific job type is undesirable. However, I share @pmrv's concern that I don't know where else to put it right now.

Let's leave the documentation as you have it, and discuss alternative tricks in the pyiron meeting.

samwaseda: As @liamhuber suggested, I'm perfectly fine with talking about it at the next pyiron meeting. I just want to say I think `get_external_input` does what I think it should, since I can totally imagine that we're gonna have project level external input in the future. So even if it's not used by `ScriptJob`, I'd like it to be there as it is now.

liamhuber (approved):

liamhuber:
I just want to bump the stalebot off its course here. 

I'm content with this as it stands, but we never wound up talking about `get_external_input` at the pyiron meeting. Let's do that this Friday and then you can incorporate any conclusions before merging, @raynol-dsouza

pyiron_base/job/script.py
```diff
@@ -31,6 +31,99 @@ class ScriptJob(GenericJob):
...
+        Step 1. 'example.ipynb' can contain pyiron_atomistics code like:
+                ```
+                from pyiron_atomistics import Project
+                pr = Project('example')
+                job = pr.create.job.Lammps('job')  # we name the job 'job'
+                job.structure = pr.create.structure.ase_bulk('Fe')  # specify structure
+
+                # Optional: get an input value from 'submit_example_job.ipynb', the notebook which submits
+                #   'example.ipynb'
+                external_input = pr.get_external_input()
```
liamhuber: can just use `pr.data.number_of_cores` as of #192

pyiron_base/job/script.py
```diff
@@ -31,6 +31,99 @@ class ScriptJob(GenericJob):
...
+                                                         # directory as 'submit_example_job.ipynb'
+
+                # Optional: to submit the notebook to a queueing system
+                number_of_cores = 4  # number of cores to be used
+                scriptjob.server.cores = number_of_cores
+                scriptjob.server.queue = 'cmfe'  # specify the queue to which the ScriptJob job is to be submitted
+                scriptjob.server.run_time = 120  # specify the runtime limit for the ScriptJob job in seconds
+
+                # Optional: save an input, such that it can be accessed by 'example.ipynb'
+                scriptjob.input['number_of_cores'] = number_of_cores
```
liamhuber: Also here can save data at the project level for use in script jobs with `pr.data.number_of_cores = number_of_cores; pr.data.write()` as of #192

liamhuber (changes requested):

pyiron_base/job/script.py
raynol-dsouza: Done.

pyiron_base/job/script.py
raynol-dsouza: and done.

raynol-dsouza: @liamhuber if this looks good, shall I merge this pull request?

liamhuber (approved):

pyiron_base/job/script.py
liamhuber: @raynol-dsouza I guess wait until the end of the week, by then Jan should have had a chance to review #192 and we can merge both.

raynol-dsouza: @liamhuber I accidentally merged it before you posted this comment. O_O'

I will revert the pull request if further changes to #192 affect this PR.

PR #176, opened 2021-02-03
Fix unpack 

jan-janssen:

PR #177, opened 2021-02-03
Fix loading jobs from imported projects

pmrv:
Previously loading a job from an imported caused an error due to some wrong paths (see [here](https://github.com/pyiron/pyiron_base/pull/177/commits/38e3f0c4ace3a7d939ef3b4527c2f3741e082d8b)).  There was also an independent bug in `JobPath` when constructing the paths to the HDF5 file that was triggered by this. 

@muh-hassani I had to do some extra boilerplate in the new test cases because the other cases re-use `imp_pr`.  Without re-packing the original project it fails, can you take a look?   Generally please structure tests so that they may be run independently from each other, it greatly simplifies adding new tests.

### DIFF
--- a/pyiron_base/job/path.py
+++ b/pyiron_base/job/path.py
@@ -473,7 +473,7 @@ class JobPath(JobPathBase):
             job_path = db_entry["projectpath"]
         else:
             job_path = ''
-        job_path += db_entry["project"] + hdf5_file + db_entry["subjob"]
+        job_path = os.path.join(job_path, db_entry["project"], hdf5_file) + db_entry["subjob"]
         super(JobPath, self).__init__(job_path=job_path)
 
         if "hamilton" in db_entry.keys():
--- a/pyiron_base/archiving/import_archive.py
+++ b/pyiron_base/archiving/import_archive.py
@@ -37,16 +37,15 @@ def extract_archive(archive_directory):
     tar.close()
 
 
-def update_project_paths(project_instance, archive_directory, df):
+def update_project_paths(receiving_project, archive_directory, df):
     """Update project paths in given dataframe to point to new project."""
-    pr_import = project_instance.open(os.curdir)
     archive_name = getdir(path=archive_directory)
     df["project"] = [os.path.join(
-        pr_import.project_path, os.path.relpath(p, archive_name))
+        receiving_project.project_path, os.path.relpath(p, archive_name))
         for p in df["project"].values
     ]
-    df['projectpath'] = len(df) * [pr_import.root_path]
-    return pr_import, df
+    df['projectpath'] = len(df) * [receiving_project.root_path]
+    return df
 
 def import_jobs(
     project_instance, directory_to_import_to, archive_directory,
@@ -84,7 +83,7 @@ def import_jobs(
     if compressed:
         rmtree(archive_directory)
 
-    pr_import, df = update_project_paths(project_instance, archive_directory, df)
+    df = update_project_paths(project_instance, archive_directory, df)
 
     # Add jobs to database
     job_id_lst = []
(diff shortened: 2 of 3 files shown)


pyiron_base/job/path.py
```diff
@@ -473,7 +473,7 @@ def __init__(self, db, job_id=None, db_entry=None, user=None):
             job_path = db_entry["projectpath"]
         else:
             job_path = ''
-        job_path += db_entry["project"] + hdf5_file + db_entry["subjob"]
+        job_path = os.path.join(job_path, db_entry["project"], hdf5_file) + db_entry["subjob"]
```
jan-janssen, line 476: As far as I understand this is the line that cause the bug of not being able to load the jobs, correct? If this is the case can you tell me what is the input and what is the output? In case there is a missing "/" in the database entry, then our search does not work, so this might be the wrong place to fix it.

jan-janssen:
Some general feedback: I currently struggle with the increasing number of pull requests, so it would be nice to separate cosmetics and bugfixes into two separate pull requests, that makes it easier to review.

max-hassani:
@pmrv, I still get the same error using this branch!

max-hassani:
> @muh-hassani I had to do some extra boilerplate in the new test cases because the other cases re-use `imp_pr`. Without re-packing the original project it fails, can you take a look? Generally please structure tests so that they may be run independently from each other, it greatly simplifies adding new tests.

Thanks for this tip, I will keep it in mind for the future unittests.

pmrv:
> @pmrv, I still get the same error using this branch!

Which? @jan-janssen fix is not applied yet I this branch, will rebase later.

pyiron_base/job/path.py
pmrv: The problem in this line is that the `db_entry["project"]` is not terminated with a '/'.  I also tried to ensure it is added [here](https://github.com/pyiron/pyiron_base/blob/38e3f0c4ace3a7d939ef3b4527c2f3741e082d8b/pyiron_base/archiving/import_archive.py#L97) before the job is added to the database, but it still got swallowed somewhere.  I did not investigate the details though.

I would rather keep this change, because it makes the relevant code more robust instead of trying to ensure we always provide correct input.

pmrv:
There's two independent bugs, that stopped jobs from loading

1. Around [this line](https://github.com/pyiron/pyiron_base/blob/38e3f0c4ace3a7d939ef3b4527c2f3741e082d8b/pyiron_base/archiving/import_archive.py#L82) the project paths where taken relative to the wrong directory, which caused the loading project to look at the wrong paths.  I pasted an example database entry in 38e3f0c4ace3a7d939ef3b4527c2f3741e082d8b.
2. After this the init in `JobPath` did not concatenate the project path and the hdf5 file name correctly, because the slash was missing.

So there's not cosmetics in this, both changes were needed to the job to load.  But I agree I didn't do a good job explaining the issue in the commit messages and will take more care.

pmrv:
Force push to fix the merge conflict and rebase on top of #176.

pyiron_base/job/path.py
jan-janssen: The issue is that in the database we have the option to search in the current project and in all sub projects and as far as I remember this is realised by having the same name e.g. `/project/name/` in both the project and the sub project.

jan-janssen: Here is the fix just for the job reload https://github.com/pyiron/pyiron_base/pull/178/

pmrv: It seems to me the old code would still break in that case, so if it is an option than it would have to take a different code path.  Since `os.path.join` only ensures there's '/' between the arguments and nothing more, I don't see how it could interfere with this.  Anyways #178 seems to work, so it is ok for now.

jan-janssen: I agree the code breaks and I am open to merge this change, but we should be careful to have all `project` entries in the database end with `/` otherwise the database search does not work - at least that is what I remember. There was an issue with using `/` in database queries.

pmrv:
Fixed in #176, #178.

PR #178, opened 2021-02-04
pack fix - part 2

jan-janssen:
With this fix the reload works again


pmrv (approved):
Will add tests from #177 in a separate PR.

PR #179, opened 2021-02-05
Allow to override resource folder for binaries

pmrv:
This helps when job classes that need a specific executable are not located in their own top level directory like in [here](https://github.com/pyiron/pyiron_contrib/pull/65).

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1472,15 +1472,21 @@ class GenericJob(JobCore):
                                 "'delete_existing_job=True in create_job'".format(self.job_name))
             self.from_hdf()
 
-    def _executable_activate(self, enforce=False):
+    def _executable_activate(self, enforce=False, module=None):
         """
-        Internal helper function to koad the executable object, if it was not loaded already.
+        Internal helper function to load the executable object, if it was not loaded already.
+
+        Args:
+            enforce (bool, optional): set executable instance even if defined already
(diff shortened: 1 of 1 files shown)


jan-janssen (changes requested):
The function is called inside generic job, so I guess this change only makes sense if we set another variable on the job level, which is then transferred to this function.

pmrv:
I envisioned that classes that want to override the default executable just call this function again with `enforce=True` after the `GenericJob.__init__`, as the linked PR does in `AtomicRex.__init__`.  I'd hesitate to add a magic instance variable to modify this behavior, especially since you'd have to set it before you call `super().__init__()`.

jan-janssen:
But the function is also executed when the version is changed: https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L187

pmrv:
> But the function is also executed when the version is changed: https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L187

Ah, I didn't see that before.  This issue came up again though when moving the mtp tools.  Should we instead just use the temporary fixes from the atomicrex/mtp branches?  That is, instead of using the name of the second most top level module, we'll use the name of the module just above the job class?  In code

```python
module=self.__module__.split(".")[-2],
```

instead of

```python
module=self.__module__.split(".")[1],
```

For the existing codes this should work out the same.

PR #183, opened 2021-02-08
Add docstrings for interactive functions

pmrv:


pyiron_base/job/interactive.py
```diff
@@ -268,17 +268,17 @@ def interactive_flush(self, path="interactive", include_last_step=False):
 
     def interactive_open(self):
         """
+        Set the run mode to interactive.
 
-        Returns:
-
+        Must be called before :method:`.run()` is called.
```
samwaseda, line 275: Maybe you can also say that this is the same as `job.server.run_mode.interactive = True` (I don't insist)

pyiron_base/job/interactive.py
```diff
@@ -268,17 +268,17 @@ def interactive_flush(self, path="interactive", include_last_step=False):
...
         """
         self.server.run_mode.interactive = True
 
     def interactive_close(self):
         """
+        Stop interactive execution and sync interactive output cache.
 
-        Returns:
-
+        Sets the job status to finished, :method:`.run()` cannot be called after this.
```
samwaseda: I guess it should be `finished` and not just finished, since it's a pyiron term

pmrv:
I would also ask a for a module level description on how to use an interactive job, but more importantly how to write an interactive job class from someone with more interactive experience.

samwaseda:
> I would also ask a for a module level description on how to use an interactive job, but more importantly how to write an interactive job class from someone with more interactive experience.

I don't know exactly how it works now, but if I remember well (correct me if I'm wrong @jan-janssen ), the static and interactive modules are not really separated, but the distinction is made via `job.server.status` within the modules. So it might be difficult to get an interactive-class-specific doc string in the right place.

samwaseda (approved):
Brilliant!

pmrv:
> > I would also ask a for a module level description on how to use an interactive job, but more importantly how to write an interactive job class from someone with more interactive experience.
> 
> I don't know exactly how it works now, but if I remember well (correct me if I'm wrong @jan-janssen ), the static and interactive modules are not really separated, but the distinction is made via `job.server.status` within the modules. So it might be difficult to get an interactive-class-specific doc string in the right place.

I'm also open to adding a page outside of the API docs, but I feel that the interactive workflow of `open`, a mixture of `calc_*` and `run()` and then `close` is not that intuitive.

jan-janssen:
I do not think I am the only one who is aware how the interactive mode works, but let me give a small explanation: 

- The user sets `job.server.interactive = True`.
- Then when `job.run()` is called the first structure is evaluated and afterwards the process of the computing engine stays active. 
- So the user can set a new structure `job.structure = structure_step_2` and then call `job.run()` again to evaluate the next structure. 
- The above step can be repeated with many structures. Finally to finish the job we use `job.interactive_close()`, which afterwards sets the job status from `running` to `finished`. 

Besides this high level interface, there is a second interface which can be faster, if you know already the difference between the structures you submit to the interactive job. For example if only the positions change during an MD run you can use: `job.interactive_positions_setter = structure_step_2.positions`. In this case pyiron does not check the atomistic species, cell or magnetic moments, but instead directly communicates the new positions to the simulation engine, which improves the performance. Here is an example how to implement a Langevin Thermostat using this second method: https://github.com/jan-janssen/pyiron-md-example/blob/master/langevin.ipynb

With this I hand it over to @raynol-dsouza and @liamhuber who extended the interactive interface with protocols.

liamhuber:
> With this I hand it over to @raynol-dsouza and @liamhuber who extended the interactive interface with protocols.

Unfortunately I don't know how much more we can add. We don't really extend the base interactive classes at all. On top of Jan's response I can add that there are about a million different methods attached to the atomistic interactive jobs (`interactive_positions_getter`, `...forces_getter`, `...energy...`, `...pressures...`, etc) and one can (must?) use these. Also the attributes `interactive_flush_frequency` and `interactive_write_frequency` can make your job run inefficiently or simply with the wrong physics if you don't set their values correctly. But honestly I'd need to go back into the issues pages to look at what some of the answers to/reasons behind these problems were, or whether I'm even misremembering the problem... I can't say I've ever had a deep or clear understanding of how the interactive jobs do what they do.

You can see how we interface with them in protocols [here](https://github.com/pyiron/pyiron_contrib/blob/04b51ab03806c7b0893ca4ca97d83defdb9db849/pyiron_contrib/protocol/primitive/one_state.py#L157), but even things like the `job.interactive_initialize_interface()` calls were added by a sort of trial and error procedure, if I recall.

samwaseda:
I guess what @pmrv is trying to do right now is to make the docstring more enlightening, so probably it's not so much about what it does behind the scenes.

Probably the most pragmatic solution right now is to put it in the class level docstring anyway, regardless of whether it's static or interactive, maybe by putting a few examples.

pyiron_base/job/interactive.py
```diff
@@ -268,17 +268,19 @@ def interactive_flush(self, path="interactive", include_last_step=False):
 
     def interactive_open(self):
         """
+        Set the run mode to interactive.
 
-        Returns:
+        This is the same as setting :attribute:`.run_mode.interactive`.
```
jan-janssen:
```suggestion
        This is the same as setting :attribute:`.server.run_mode.interactive`.
```

jan-janssen (approved):

pmrv:
For now I'll merge this and leave an extensive example for the future.

PR #187, opened 2021-02-10
Enhance file data handling

niklassiemer:
The handling of different file types is enhanced via a new `load_file` function.
In addition, a new `FileData` class is introduced to provide a generic interface in preparation of a project browser (#156).

### DIFF
--- a/pyiron_base/generic/filedata.py
+++ b/pyiron_base/generic/filedata.py
@@ -1,0 +1,139 @@
+"""
+Generic File Object.
+"""
+
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+import json
+import os
+import pandas
+
+from pyiron_base import ImportAlarm, FileHDFio
+
+__author__ = "Niklas Siemer"
+__copyright__ = (
+    "Copyright 2020, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "0.1"
+__maintainer__ = "Niklas Siemer"
+__email__ = "siemer@mpie.de"
+__status__ = "development"
+__date__ = "Feb 02, 2021"
+
+
+_has_imported = {}
+import_alarm = ImportAlarm()
+_not_imported = ''
+try:
+    from PIL import Image
+    _has_imported['PIL'] = True
+except ImportError:
+    _has_imported['PIL'] = False
+try:
+    from IPython.core.display import display
+    _has_imported['IPython'] = True
+except ImportError:
+    _has_imported['IPython'] = False
+if all(_has_imported.values()):
+    pass
+else:
+    for k, j in _has_imported.items():
+        if j and len(_not_imported) > 0:
+            _not_imported += ', '
+        if j:
+            _not_imported += k
+    import_alarm = ImportAlarm(
+        "Reduced functionality, since " + _not_imported + " could not be imported."
+    )
+
+
+def load_file(filename):
+    """
+        Load the file and return an appropriate object containing the data.
+ 
+        Args:
+            filename (str): path to the file to be displayed.
+    """
+    def _load_txt(file):
+        with open(file) as f:
+            return f.readlines()
+
+    def _load_json(file):
+        with open(file) as f:
+            return json.load(f)
+
+    def _load_csv(file):
+        return pandas.read_csv(file)
+ 
+    def _load_img(file):
+        return Image.open(file)
+ 
+    def _load_default(file):
+        try:
+            return _load_txt(file)
+        except UnicodeDecodeError:
+            return file
+
+    _, filetype = os.path.splitext(filename)
+    if filetype.lower() in ['.h5', '.hdf']:
+        return FileHDFio(file_name=filename)
+    if filetype.lower() in ['.json']:
+        return _load_json(filename)
+    elif filetype.lower() in ['.txt']:
+        return _load_txt(filename)
+    elif filetype.lower() in ['.csv']:
+        return _load_csv(filename)
+    elif _has_imported['PIL'] and filetype.lower() in Image.registered_extensions():
+        return _load_img(filename)
+    else:
+        return _load_default(filename)
+
+
+class FileData:
+
+    """ FileData stores an instance of a data file, e.g. a single Image from a measurement """
+    def __init__(self, source=None, data=None, filename=None, metadata=None, filetype=None):
+        """
+            FileData class to store data and associated metadata.
+
+            Args:
+                source (str/None): path to the data file
+                data (object/None): object containing data
+                filename (str/None): filename associated with the data object, Not used if source is given!
+                metadata (dict/InputList): Dictionary of metadata associated with the data
+                filetype (str): File extension associated with the type data,
+                                If provided this overwrites the assumption based on the extension of the filename.
+        """
+        if (source is None) and (data is None):
+            raise ValueError("No data given")
+        self._data = data
+        if source is not None:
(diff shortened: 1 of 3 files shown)


niklassiemer:
I'll just give up to satisfy codacy....
```
Multi-line docstring summary should start at the first line (D212)
"""
```
```
Multi-line docstring summary should start at the second line (D213)
"""FileData class to store data and associated metadata.
```

niklassiemer:
@jan-janssen This is now part one of the PR #156 handling data files within pyiron. However, now the generic `FileData` class is not used so far.

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,130 @@
...
+__status__ = "development"
+__date__ = "Feb 02, 2021"
+
+
+_has_imported = {}
+import_alarm = ImportAlarm()
+_not_imported = ''
+try:
+    from PIL import Image
+    _has_imported['PIL'] = True
```
pmrv, line 28: Would it make sense to augment `ImportAlarm` to do something similar?  I guess you have to do this work-around because you can't use the decorator?

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,130 @@
...
+
+            Args:
+                source (str/None): path to the data file
+                data (object/None): object containing data
+                filename (str/None): filename associated with the data object, Not used if source is given!
+                metadata (dict/InputList): Dictionary of metadata associated with the data
+                filetype (str): File extension associated with the type data,
+                                If provided this overwrites the assumption based on the extension of the filename.
+        """
+        if (source is None) and (data is None):
```
pmrv: Does it makes sense to collapse `source` and `filename` into one argument?  So when `data` is given it is interpreted as the filename and if not as the source?

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,130 @@
...
+        if source is not None:
+            self.filename = os.path.split(source)[1]
+            self.source = source
+        elif filename is None:
+            raise ValueError("No filename given")
+        else:
+            self.filename = filename
+        if (filetype is None) and (self.filename is not None):
+            filetype = os.path.splitext(self.filename)[1]
+            if len(filetype[1:]) == 0:
```
pmrv: This checks whether `filetype == ''`? If so I'd prefer the explicit check.

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,130 @@
...
+        else:
+            self.filetype = filetype
+        if metadata is None:
+            self.metadata = {}
+        else:
+            self.metadata = metadata
+        self._hasdata = True if self._data is not None else False
+
+    @property
+    def data(self):
```
pmrv, line 135: Maybe we can cache this with 
```python
@property
@lru_cache
def data(self):
    ...
```
but if you think this is premature, I won't insist.

pyiron_base/generic/filedata.py
```diff
@@ -0,0 +1,130 @@
...
+        if j:
+            _not_imported += k
+    import_alarm = ImportAlarm(
+        "Reduced functionality, since " + _not_imported + " could not be imported."
+    )
+
+
+def load_file(filename):
+    """
+        Load the file and return an appropriate object containing the data.
```
pmrv, line 40: Should list supported file types and the returned types.

tests/generic/test_filedata.py
```diff
@@ -0,0 +1,120 @@
...
+        hdf = load_file(self.current_dir+'/test_data.h5')
+        self.assertIsInstance(hdf, FileHDFio)
+        self.assertEqual(hdf['content/key'], 'value')
+
+    def test_load_file_default(self):
+        """Test default load for text file and h5 file without extension."""
+        filename = self.current_dir + '/test_data'
+        default = load_file(filename)
+        # For some reason on linux handles this different than windows:
+        self.assertTrue(filename == default or default[0] == '‰HDF\n')
```
pmrv: I don't understand what that second test is checking for.

pmrv (changes requested):

tests/generic/test_filedata.py
niklassiemer: Here, I actually want to have a failure of a normal `readlines` and thus `load_file` to return the filename. However, on windoof the read of the hdf file returns some cryptic byte like strings and the first one happens to be this... I really would like the windows `readlines` to throw an error on this...

pyiron_base/generic/filedata.py
niklassiemer: This would be possible, especially, since I am using the `filename` from the `source` even if `filename` is given.
I will call ist `file` then.

pyiron_base/generic/filedata.py
niklassiemer: Actually, the warning is only given once, if I am not mistaken. Thus, I should really _use_ the decorator for `load_file`. The reason for the lengthy loop over `_not_imported` was due to different packages I tried to use and this might easily taken out. The `_has_imported` dictionary exists for the same reason. I might just use 
`_has_imported_pil = True` here.
I need to know which (optional) packages are installed to allow for a working solution with reduced functionality without additional try/except later on.

pyiron_base/generic/filedata.py
pmrv: I didn't see this on first pass, but I don't like that `_load_default` returns the filename on error, it should raise an exception, maybe `ValueError` or `OSError`.  I checked with the [docs](https://docs.python.org/3/library/functions.html?highlight=open#open) and it seems the platform dependent behavior comes from the `encoding` argument to `open()`, maybe we can pass a fixed value to force the same behavior.

niklassiemer: Well in the end this would be an `IOError` if we want to raise an error. Fot the `encoding` argument, is this silently fixed differently for different platforms or is one platform just better at guessing a nice encoding? If the encoding is guessed by the `open()` I am not sure if we should fix it...

pyiron_base/generic/filedata.py
niklassiemer: It does not hurt, does it?

pyiron_base/generic/filedata.py
niklassiemer: Now I use a fixed encoding (utf8) and raise an error on failing to load the file (`IOError`).

pmrv: Ah yes, I hadn't considered that windows might not use the same UTF encoding.  Can we maybe just also support '.h5' files and return a `ProjectHDFio` in this case?

niklassiemer: I just give the encoding ('utf8') in the open now, so, therefore, it should fail to simply `readlines()` for non human readable files. Which is what we want, don't we? At least also windoof fails now to read the crappy bytes into a list :)

pmrv: Sounds good to me then.

pyiron_base/generic/filedata.py
pmrv: I'm not opposed to the solution for now, but if it comes up again, we should probably see if we can move that logic to `ImportAlert`.

pmrv (approved):
Looks good to me.

niklassiemer:
Such that we add a variable to hold the packages which have not been imported? Something like `_import_alarm_imported = {'package_name' : True}` which would live on the function decorated?

pmrv:
> Such that we add a variable to hold the packages which have not been imported? Something like `_import_alarm_imported = {'package_name' : True}` which would live on the function decorated?

I think we could have a method on the alam object that can check which libraries have been imported during the with statement it is used in, but I haven't looked into anything.

niklassiemer:
In my opinion this is ready to be merged.

PR #192, opened 2021-02-19
Project data

liamhuber:
Projects now have an input list attached to themselves for storing project-level data. The motivation was to come up with a more transparent implementation of the current `Notebook` class and associated `Project.get_external_input` method. The original code is only used for `ScriptJob`, and I extended the test suite there to make sure this new functionality does indeed work for it, but I could imagine it being useful in other circumstances too.

@pmrv there are two outstanding issues I'm really having trouble with, and I left failing tests in for them both:
- `InputList` recursion isn't working well and complex data types get loaded as empty containers.
- I want projects to try to auto-load data on instantiation but I get caught in some horrible error loop I don't understand. I left a bit of commented out code in `Project.__init__` that I really *wish* worked.

@raynol-dsouza note that this will impact the syntax for #171.

I'm also open to some critical feedback on any bad side-effects this might have that I'm missing.

### DIFF
--- a/pyiron_base/project/data.py
+++ b/pyiron_base/project/data.py
@@ -3,8 +3,7 @@
 # Distributed under the terms of "New BSD License", see the LICENSE file.
 
 """
-A modified data container for storing associating data with a project. Here is more worthless text trying to see what
-it takes to get codacy to shut up.
+A modified data container for storing associating data with a project.
 
 Spec:
     The `Project` class should have data associated with it which can be stored to-file in our supported format(s)
@@ -39,8 +38,7 @@ class ProjectData(InputList):
 
     def __init__(self, *args, project=None, **kwargs):
         """
-        A data storage container which can store itself to/retrieve itself from file at the project level. This is some
-        extra text to try to get Codacy to shut up.
+        A data storage container which can store itself to/retrieve itself from file at the project level.
 
         Args:
             project (pyiron_base.Project): The project instance the storage is attached to.
--- a/pyiron_base/project/data.py
+++ b/pyiron_base/project/data.py
@@ -48,16 +48,12 @@ class ProjectData(InputList):
         self._project = project
 
     def read(self):
-        """
-        Read existing data from project-level storage.
-        """
+        """Read existing data from project-level storage."""
         hdf = ProjectHDFio(self._project, file_name="project_data")
         if self.table_name not in hdf.list_groups():
             raise KeyError(f"Table name {self.table_name} was not found -- Project data is empty.")
         self.from_hdf(hdf=hdf)
 
     def write(self):
-        """
-        Write data to project-level storage.
-        """
+        """Write data to project-level storage."""
         self.to_hdf(ProjectHDFio(self._project, file_name="project_data"))
--- a/pyiron_base/project/data.py
+++ b/pyiron_base/project/data.py
@@ -7,14 +7,13 @@ A modified data container for storing associating data with a project. Here is m
 it takes to get codacy to shut up.
 
 Spec:
+    The `Project` class should have data associated with it which can be stored to-file in our supported format(s)
+    (right now just hdf5). This should meet the following requirements:
 
-The `Project` class should have data associated with it which can be stored to-file in our supported format(s) (right
-now just hdf5). This should meet the following requirements:
-
-- Data storage is immediately accessible, i.e. appearing in the project tab-completion module
-- Data stored there should be readable and writeable with a single parameter-free call
-- When instantiated, new projects should automatically read any available data (not *quite* satisfied right now!)
-- The `Project` tab completion menu should not get cluttered (i.e. nest everything under `Project.foo`)
+    - Data storage is immediately accessible, i.e. appearing in the project tab-completion module
+    - Data stored there should be readable and writeable with a single parameter-free call
+    - When instantiated, new projects should automatically read any available data (not *quite* satisfied right now!)
+    - The `Project` tab completion menu should not get cluttered (i.e. nest everything under `Project.foo`)
 """
 
 from pyiron_base.generic.inputlist import InputList
--- a/pyiron_base/project/data.py
+++ b/pyiron_base/project/data.py
@@ -4,6 +4,16 @@
 
 """
 A modified data container for storing associating data with a project.
+
+Spec:
+
+The `Project` class should have data associated with it which can be stored to-file in our supported format(s) (right
+now just hdf5). This should meet the following requirements:
+
+- Data storage is immediately accessible, i.e. appearing in the project tab-completion module
+- Data stored there should be readable and writeable with a single parameter-free call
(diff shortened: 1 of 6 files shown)


niklassiemer:
I tied an InputLists to the project level in the RDM branch in contrib to store Metadata. RDM/project if I am not mistaken. I'll look it up later.

liamhuber:
Cool I'll peek at that later too and we can synchronize the approach taking the best of both 😄

pmrv:
The initialization problem comes again from `ProjectData` changing the signature of `__init__`.  What's the `project` back reference used for?  Wouldn't it be easier to just add an `InputList` attribute project to the `Project` class?

liamhuber:
> The initialization problem comes again from `ProjectData` changing the signature of `__init__`. 

Ok, thanks for taking a look @pmrv! Isn't this a pretty big weakness? I feel that `InputList` is (or is becoming) a cornerstone of `pyiron_base`, so an inability to change its signature when subclassing scares me. Did I just change its signature in the wrong way?

> What's the `project` back reference used for? Wouldn't it be easier to just add an `InputList` attribute project to the `Project` class?

The `project` back reference allows me to specify the `InputList`'s HDF5 location. This is then used in super-short `pr.data.read()` and `pr.data.write()` method which invoke `from/to_hdf` with the right `ProjectHDFio`. 

I think it should be possible to instead add this path as a regular `InputList` attribute which is assigned by the parent project immediately after instantiation, but then the `InputList` will write its own HDF5 location to the HDF file. This isn't problematic just...ugly.

I added a very brief spec commit, but I'll paste it here:

>Spec:
>
>The `Project` class should have data associated with it which can be stored to-file in our supported format(s) (right now just hdf5). This should meet the following requirements:
> - Data storage is immediately accessible, i.e. appearing in the project tab-completion module
> - Data stored there should be readable and writeable with a single parameter-free call
> - When instantiated, new projects should automatically read any available data
> - The `Project` tab completion menu should not get cluttered (i.e. nest everything under `Project.foo`)

Then the usage should look like this:

```python
pr = Project('test')
pr.data.foo = 'foo'  # Acts like InputList
pr.data.write()  # Easy writing

pr2 = Project('test')
print(pr.2data.foo)  # Automatic loading
> 'foo'
pr2.data.bar = 42
pr2.data.write()

pr.data.read()  # Easy reading (in the edge case it doesn't happen automatically with instantiation)
print(pr.data.bar)
> 42
# Only new item in Project tab completion is 'data'
```

pmrv:
If the `InputList` lives directly on the project then it will be read/written directly in its `from_hdf`/`to_hdf` so no separate `read`/`write` should be needed, if I'm not missing anything.

liamhuber:
> If the `InputList` lives directly on the project then it will be read/written directly in its `from_hdf`/`to_hdf` so no separate `read`/`write` should be needed, if I'm not missing anything.

I'm confused. `Project` doesn't have `to/from_hdf` methods, and these methods have no default `hdf` arguments on `InputList`. How is the `Project.InputList :: pr.data` supposed to know where and when to save/load? I don't see how being an attribute of the project is already sufficient to pass this information.

pmrv:
> > If the `InputList` lives directly on the project then it will be read/written directly in its `from_hdf`/`to_hdf` so no separate `read`/`write` should be needed, if I'm not missing anything.
> 
> I'm confused. `Project` doesn't have `to/from_hdf` methods, and these methods have no default `hdf` arguments on `InputList`. How is the `Project.InputList :: pr.data` supposed to know where and when to save/load? I don't see how being an attribute of the project is already sufficient to pass this information.

No, that's correct, I just mentally misplaced some stuff. :/  I'll think about it some more.

liamhuber:
> No, that's correct, I just mentally misplaced some stuff. :/ I'll think about it some more.

Ok, no worries and thanks. Better that you slipped up quickly thinking about it for a few seconds than that I slipped up after spending more than an hour thinking about it and writing a whole new class haha 😆

pyiron_base/project/data.py
```diff
@@ -0,0 +1,51 @@
...
+        instance = super().__new__(cls, *args, **kwargs)
+        object.__setattr__(instance, "project", None)
+        return instance
+
+    def __init__(self, project, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.project = project
+
+    def read(self):
+        hdf = ProjectHDFio(self.project, file_name="project_data")
```
niklassiemer: why not `self.project.create_hdf(self.project.path, 'project_data)`  which opens the `hdf` at `h5_path='/project_data'`?

pmrv:
> > No, that's correct, I just mentally misplaced some stuff. :/ I'll think about it some more.
> 
> Ok, no worries and thanks. Better that you slipped up quickly thinking about it for a few seconds than that I slipped up after spending more than an hour thinking about it and writing a whole new class haha laughing

I think the appropriate fix should then be to add then project as an keyword argument to `__init__`, then it should interfere with the updating of the list when read from HDF5.

pmrv:
> > The initialization problem comes again from `ProjectData` changing the signature of `__init__`.
> 
> Ok, thanks for taking a look @pmrv! Isn't this a pretty big weakness? I feel that `InputList` is (or is becoming) a cornerstone of `pyiron_base`, so an inability to change its signature when subclassing scares me. Did I just change its signature in the wrong way?

As I said above, adding a new keyword argument should work.  If it does, I'll update the docs of `InputList` accordingly.

pyiron_base/project/data.py
liamhuber: I have one small objection to this, and that is `Project.create_hdf` re-instantiates another `Project` inside itself, which feels wasteful when that would just be a duplicate of the project calling the method.

I'd still be open to this change, but unfortunately it raises the error caught by the if-clause that immediately follows `if self.table_name not in hdf.list_groups():`.  So does

```python
hdf = ProjectHDFio(self.project, file_name="project_data", h5_path="/project_data")
```

So somehow providing the `h5_path` breaks things.

liamhuber:
> As I said above, adding a new keyword argument should work. If it does, I'll update the docs of `InputList` accordingly.

Beautiful. Yes, switching `project` from an arg to a kwarg resolves the recursive loading problem 😄

pyiron_base/project/data.py
niklassiemer: Ok, actually, I just realize I don't like my idea then. Since the `Project` is hard coded to be a `base` project and `self.__class__` is not available in a `@staticmethod`.

liamhuber:
Ok, I had to do a bit of a workaround to get the data of new projects to automatically load (on accessing the property it first checks if the private attribute has length zero, and if so tries a load). If anyone has advice for how to get around the infinite recursion issue and do this properly in `__init__` that would be super, but such elegance might need to wait until other parts of the code get changed. For now I'm content with this and feel it's ready to merge.

niklassiemer (approved):

PR #193, opened 2021-02-25
Set job class name from type(self), so sub classes don't have to

pmrv:
When creating new job classes, I keep forgetting to set `self.__name__` correctly.  This updates `GenericJob` to do it automatically.

EDIT: This could still make problems when deriving from sub-classes of `GenericJob` (i.e. masters and atomistic jobs), because they might overwrite it again, but I haven't checked it yet.


pmrv:
I've removed the name also from affected jobs in pyiron_atomistics [here](https://github.com/pyiron/pyiron_atomistics/pull/86).

jan-janssen (approved):
Looks good to me

PR #194, opened 2021-02-25
Re-export FlexibleMaster

pmrv:
For some reason all the other master are re-exported at the pyiron_base top-level, but this one isn't.


jan-janssen (approved):
Looks good to me - maybe we should also think about documenting this somewhere basically which classes a developer should reuse from pyiron base and which classes suite which usecase.

pmrv:
> Looks good to me - maybe we should also think about documenting this somewhere basically which classes a developer should reuse from pyiron base and which classes suite which usecase.

Agree, at least for the master job I have some notes by now, that I'll turn into a sphinx page once everything has settled a bit.

PR #195, opened 2021-02-25
Move initialization of ref_job to GenericMaster._init_child_job

pmrv:
This way also FlexibleMaster can use it and we save some lines in the other masters.


jan-janssen (approved):
Looks good to me

PR #209, opened 2021-02-27
Remove dicttoxml

### LINKED ISSUE #208: Remove dicttoxml
It seems like I missed the addition of `dicttoxml`:
* the package is no longer updated for 5 years https://github.com/quandyfactory/dicttoxml 
* the conda-forge package is not working - so the update of our conda forge package fails https://github.com/conda-forge/pyiron_base-feedstock/pull/52 

Given these constraints I would suggest to remove it @pmrv and @muh-hassani

jan-janssen:
Here is my current progress in debugging https://github.com/jan-janssen/xml-debug - xml is a mess - it seems like you can not use int as keys so what xmltodict does is adding an n in front of every int. While this produces a correct xml file I do not think this is a established convention.

### DIFF
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -4,7 +4,6 @@ dependencies:
   - coveralls
   - coverage
   - codacy-coverage
-  - dicttoxml =1.7.4
   - dill =0.3.3
   - future =0.18.2
   - gitpython =3.1.13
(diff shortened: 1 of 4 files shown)


pmrv:
It seems the error is because `unparse` doesn't write out a well-formatted xml file, but I can't look deeply into it right now.

jan-janssen:
@pmrv and @muh-hassani Can you take a look at this? I fixed the `dicttoxml` package but I would still prefer to remove it from our dependencies.

jan-janssen:
Close in favor of https://github.com/pyiron/pyiron_base/pull/215

PR #210, opened 2021-02-27
Fix executable path for nested job classes

pmrv:
As discussed in #179.

I still think midterm we should add a way for a job class to specify a sub folder in the resources folder (maybe kwarg to `GenericJob.__init__`), but for now this should work.


jan-janssen (approved):
Looks good to me

jan-janssen:
@pmrv From my perspective this is ready to be merged.

PR #214, opened 2021-03-02
Replace eval with explicit indices

pmrv:
This should fix [this error](https://github.com/pyiron/pyiron_contrib/pull/65).  Running the tests locally I'm getting some problems with `copy_to` that I don't understand yet.

PR #217, opened 2021-03-02
Replace eval with explicit indices

jan-janssen:
https://github.com/pyiron/pyiron_base/pull/214


jan-janssen:
@pmrv This should be independent of the `InputList` part, so I would suggest to merge it now and then directly trigger a new release to be on the save side.

PR #223, opened 2021-03-04
Allow recursive appending

samwaseda:
I have to admit I know little about `InputList`, but I thought it would nice to have recursive appending, for example:

```
some_input = InputList()
some_input['entry/with/new/value'] = 0
```

-> this feature is included with this PR.

### DIFF
--- a/pyiron_base/generic/inputlist.py
+++ b/pyiron_base/generic/inputlist.py
@@ -207,7 +207,7 @@ class InputList(MutableMapping):
 
         if isinstance(key, tuple):
             if key[0] not in self.keys():
-                self[key[0]] = InputList()
+                self[key[0]] = type(self)()
             self[key[0]][key[1:]] = val
         elif isinstance(key, int):
             if key < len(self):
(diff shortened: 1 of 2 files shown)


samwaseda:
I'll write tests as soon as the idea is approved.

liamhuber:
> I'll write tests as soon as the idea is approved.

Seems like a nice idea to me. Let's see first if the existing tests pass.

I would suggest also to promote/use/also test the syntax `some_input.entry.with.new.value = 0`. Let's avoid strings where possible 😸

pyiron_base/generic/inputlist.py
```diff
@@ -206,6 +206,8 @@ def __setitem__(self, key, val):
         key = _normalize(key)
 
         if isinstance(key, tuple):
+            if key[0] not in self.keys():
+                self[key[0]] = InputList()
```
pmrv:
```suggestion
                self[key[0]] = type(self)()
```
This plays nicer with sub-classes.

pmrv:
I don't see an obvious case where this would break anything, so it's fine with me.  @liamhuber suggestions should work directly with this change.

liamhuber (approved):

PR #224, opened 2021-03-04
Add additional argument string to the executable

### LINKED ISSUE #190: How to pass command line arguments
This one is closely related to [this issue](https://github.com/pyiron/pyiron_atomistics/issues/75). Now I'd like to pass command line arguments to the executable. Currently, the LAMMPS executable contains:

```
#!/bin/bash
module load pyiron/dev
mpiexec -n $1 lmp_mpi -in control.inp
```

But I'd like to extend it to something like

```
#!/bin/bash
module load pyiron/dev
mpiexec -n $1 lmp_mpi -in control.inp ${@:2}
```

and from LammpsBase I'd love to pass custom arguments. Is there a possibility to do something like this?

niklassiemer:
Closes #190

I introduce a `additional_arguments` string at the `Executable` class which is added to the system call of the executable. I store the additional command line arguments in `hdf["RUNARGS"]`.
@samwaseda with these changes it should be possible to extend the shell scripts as you proposed in #190.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -717,10 +717,7 @@ class GenericJob(JobCore):
         try:
             if self.server.cores == 1 or not self.executable.mpi:
                 out = subprocess.check_output(
-                    [
-                        str(self.executable),
-                        self.executable.additional_arguments,
-                    ],
+                    [str(self.executable)] + self.executable.additional_arguments.split(),
                     cwd=self.project_hdf5.working_directory,
                     shell=True,
                     stderr=subprocess.STDOUT,
@@ -732,8 +729,7 @@ class GenericJob(JobCore):
                         self.executable.executable_path,
                         str(self.server.cores),
                         str(self.server.threads),
-                        self.executable.additional_arguments,
-                    ],
+                    ] + self.executable.additional_arguments.split(),
                     cwd=self.project_hdf5.working_directory,
                     shell=False,
                     stderr=subprocess.STDOUT,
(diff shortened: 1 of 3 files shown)


niklassiemer:
Well, on windows this does not work, yet. I will look into it tomorrow.

pyiron_base/job/generic.py
```diff
@@ -1514,6 +1516,10 @@ def _type_from_hdf(self):
                 self.executable.version = self._hdf5["VERSION"]
             except ValueError:
                 self.executable.executable_path = self._hdf5["VERSION"]
+            try:
+                self.executable.additional_arguments = self._hdf5["RUNARGS"]
```
jan-janssen, line 1524: rather than `try except` I would prefer `if "RUNARGS" in self._hdf5.list_nodes():`

niklassiemer:
Although all tests pass, I do not yet trust this implementation. I am also confused the Linux tests did not fail on the previous commit...

niklassiemer:
@jan-janssen How do I get a echo executable to work for linux and windows? Or do I actually have to have two of them?...

pmrv:
> @jan-janssen How do I get a echo executable to work for linux and windows? Or do I actually have to have two of them?...

Try 
```python
job.executable = "echo $1 $2 $@"
```
for linux and a comparable string for windows, you can decide the path based on `sys.platform`.  I've never used `subprocess` on windows though, does it run stuff in cmd.exe? :S

niklassiemer:
I actually tried to write the stuff to a file and call that (works on Linux) however I will try this solution :)

niklassiemer:
Great, I got it working locally on Windows and Linux with a file based solution. However, CI does not seem to like that.

niklassiemer:
The most important test I don't get running on CI so far. Any ideas how to verify that the first two arguments are the number of cores and threads for mpi executables?

PR #226, opened 2021-03-10
Add ProjectBrowser feature

niklassiemer:
This adds a GUI element to any pyiron Project, a browser to access the data system of the Project via it's list_* functions.
2nd part of #156
(1st part #187 is merged)

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -119,6 +119,7 @@ class Project(ProjectPath):
         self._inspect_mode = False
         self._store = None
         self._creator = Creator(project=self)
+        self._project_browser = None
 
         if not s.database_is_disabled:
             s.open_connection()
@@ -173,6 +174,16 @@ class Project(ProjectPath):
     @property
     def create(self):
         return self._creator
+
+    @property
+    def browser(self):
+        """Provides a browser to inspect the data system of the project."""
+        if self._project_browser is None:
+            from pyiron_base.project.project_browser import ProjectBrowser
+            self._project_browser = ProjectBrowser(project=self,
+                                                   show_files=False,
+                                                   Vbox=None)
+        return self._project_browser
 
     @property
     def data(self):
--- a/pyiron_base/project/project_browser.py
+++ b/pyiron_base/project/project_browser.py
@@ -303,11 +303,13 @@ class ProjectBrowser:
         file_chosen_color = '#FFBBBB'
         file_color = '#DDDDDD'
 
-        dirs, files = self.dirs, self.files
-        if self._node_as_dirs:
-            dirs += self.nodes
-        else:
-            files += self.nodes
+        # item layout definition
+        item_layout = widgets.Layout(width='80%',
+                                     height='30px',
+                                     min_height='24px',
+                                     display='flex',
+                                     align_items="center",
+                                     justify_content='flex-start')
 
         def on_click_group(b):
             if self._busy_check():
@@ -323,32 +325,30 @@ class ProjectBrowser:
             self._update_filebox(filebox)
             self._busy_check(False)
 
-        buttons = []
-        item_layout = widgets.Layout(width='80%',
-                                     height='30px',
-                                     min_height='24px',
-                                     display='flex',
-                                     align_items="center",
-                                     justify_content='flex-start')
-
-        for f in dirs:
-            button = widgets.Button(description=f,
+        def gen_dir_button(dirname):
+            button = widgets.Button(description=dirname,
                                     icon="fa-folder",
                                     layout=item_layout)
             button.style.button_color = dir_color
             button.on_click(on_click_group)
-            buttons.append(button)
+            return button
 
-        for f in files:
-            button = widgets.Button(description=f,
+        def gen_file_button(filename):
+            button = widgets.Button(description=filename,
                                     icon="fa-file-o",
                                     layout=item_layout)
-            if os.path.join(self.path, f) in self._clickedFiles:
+            if os.path.join(self.path, filename) in self._clickedFiles:
                 button.style.button_color = file_chosen_color
             else:
                 button.style.button_color = file_color
             button.on_click(on_click_file)
-            buttons.append(button)
+            return button
+
+        dirs = self.dirs + self.nodes if self._node_as_dirs else self.dirs
+        files = self.files if self._node_as_dirs else self.files + self.nodes
+
+        buttons = [gen_dir_button(name) for name in dirs]
+        buttons += [gen_file_button(name) for name in files]
 
         filebox.children = tuple(buttons)
 
--- a/pyiron_base/project/project_browser.py
+++ b/pyiron_base/project/project_browser.py
@@ -252,7 +252,7 @@ class ProjectBrowser:
         self.output.clear_output(True)
         try:
             data = self.project[filename]
-        except(KeyError):
+        except(KeyError, IOError):
             data = None
         with self.output:
             if data is not None and str(type(data)).split('.')[0] == "<class 'PIL":
--- a/pyiron_base/project/project_browser.py
+++ b/pyiron_base/project/project_browser.py
@@ -64,7 +64,9 @@ class ProjectBrowser:
         self.pathbox = widgets.HBox(layout=widgets.Layout(width='100%', justify_content='flex-start'))
         self.optionbox = widgets.HBox()
         self.filebox = widgets.VBox(layout=widgets.Layout(width='50%', height='100%', justify_content='flex-start'))
-        self.path_string_box = widgets.Text(description="(rel) Path", width='min-content')
+        self.path_string_box = widgets.Text(description="(rel) Path",
+                                            layout=widgets.Layout(width='min-content')
+                                            )
         self.refresh()
 
     @property
@@ -202,7 +204,7 @@ class ProjectBrowser:
             set_path_button.disabled = True
         childs = [set_path_button, self.path_string_box]
 
-        button = widgets.Button(description="Reset selection", width='min-content')
+        button = widgets.Button(description="Reset selection", layout=widgets.Layout(width='min-content'))
         button.on_click(click_option_button)
         childs.append(button)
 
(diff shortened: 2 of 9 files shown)


pyiron_base/project/generic.py
```diff
@@ -174,6 +175,16 @@ def name(self):
     def create(self):
         return self._creator
 
+    @property
+    def browser(self):
+        """Provides a browser to inspect the data system of the project."""
+        if self._project_browser is None:
+            from pyiron_base.project.project_browser import ProjectBrowser
```
liamhuber, line 182: Nit: move import to top of file

liamhuber: I like the on-demand generation though. Nice speed optimization for a feature that might not be used on every project instance.

liamhuber: Ahaaaa, is this down here to avoid circular import? Doesn't it complain about circular imports anyway? If moving it down here is protecting from the circular import somehow I am still concerned this is a yellow flag to rethink the architecture so such a workaround is not needed.

pyiron_base/project/project_browser.py
```diff
@@ -0,0 +1,337 @@
+import os
```
liamhuber, line 5: Add encoding and copyright to head of file:
```python
# coding: utf-8
# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
# Distributed under the terms of "New BSD License", see the LICENSE file.
```

pyiron_base/project/project_browser.py
```diff
@@ -0,0 +1,337 @@
...
+__version__ = "0.1"
+__maintainer__ = "Niklas Siemer"
+__email__ = "siemer@mpie.de"
+__status__ = "development"
+__date__ = "Feb 02, 2021"
+
+
+class ProjectBrowser:
+    """
+        Project Browser Widget
```
liamhuber, line 29: Nit: docstring over-indented

pyiron_base/project/project_browser.py
```diff
@@ -0,0 +1,337 @@
...
+        """
+            ProjectBrowser to browse the project file system.
+
+            Args:
+                project: Any pyiron project to browse.
+                Vbox(:class:`ipython.widget.VBox`/None): Widget used to display the browser (Constructed if None).
+                fix_path (bool): If True the path in the file system cannot be changed.
+                show_files(bool): If True files (from project.list_files()) are displayed.
+        """
+        self.project = project.copy()
```
liamhuber: I'm not yet super familiar with usage, so feel free to disregard this -- should `project` be write-protected as a private variable + `@property`?

pyiron_base/project/project_browser.py
```diff
@@ -0,0 +1,337 @@
...
+        if isinstance(path, str):
+            rel_path = os.path.relpath(path, self.path)
+            if rel_path == '.':
+                return
+            self._update_project_worker(rel_path)
+        else:
+            self.project = path
+        self.refresh()
+
+    def _update_pathbox(self, box):
```
liamhuber, line 292: soft recommendation: refactor out a couple chunks into well-named methods to improve readability. This is the first method long/complex enough that I have trouble following it.

pyiron_base/project/project_browser.py
```diff
@@ -0,0 +1,337 @@
...
+            self._clickedFiles = [filepath]
+
+    def _update_filebox(self, filebox):
+        # color definitions
+        dir_color = '#9999FF'
+        node_color = '#9999EE'
+        file_chosen_color = '#FFBBBB'
+        file_color = '#DDDDDD'
+
+        if self._node_as_dirs:
```
liamhuber: This is a little hard to read. Alternatives:

Basically same length but more transparent:
```python
dirs, files = self.dirs, self.files
if self._node_as_dirs:
    dirs += self.nodes
else:
    files += self.nodes
```

Shorter! But maybe harder to read?
```python
dirs = self.dirs + self.nodes if self._nodes_as_dirs else self.dirs
files = self.files if self._nodes_as_dirs else self.files + self.nodes
```

pyiron_base/project/project_browser.py
```diff
@@ -0,0 +1,337 @@
...
+
+        buttons = []
+        item_layout = widgets.Layout(width='80%',
+                                     height='30px',
+                                     min_height='24px',
+                                     display='flex',
+                                     align_items="center",
+                                     justify_content='flex-start')
+
+        for f in dirs:
```
liamhuber: soft recommendation: I think the entire loops could use reduction into well-named methods.

liamhuber (commented):
Very nice! I have a question about the import resolution, but otherwise just some nits and then some recommendations that you can take or leave.

pyiron_base/project/generic.py
niklassiemer: It does not complain on circular imports or the like. I just tried to not overload the `Project` import and I just don't see a reason why a `Project` should require `ipywidgets` and `IPython` to run. Of course I could also wrap this in an `@import_alarm` and tell the user to install these to have the browser functionality.

pyiron_base/project/project_browser.py
niklassiemer: Thanks, forgot that.

pyiron_base/project/generic.py
liamhuber: > Of course I could also wrap this in an @import_alarm and tell the user to install these to have the browser functionality.

I like this solution best

pyiron_base/project/project_browser.py
niklassiemer: Then I actually do not need to copy, do I?

pyiron_base/project/generic.py
niklassiemer: Oh, I just re-tested and it is a circular import... And I just need to import to check if it `isinstance(BaseProject)`.

pyiron_base/project/project_browser.py
niklassiemer: The first one does not work correctly `if not self._node_as_dirs`, it is duplicating the file list on each click. The second one works and will be used.

pyiron_base/project/project_browser.py
liamhuber: Nope, should be safe

pyiron_base/project/generic.py
pmrv: Try if it works with [relative imports](https://docs.python.org/3/reference/import.html#relativeimports) I seem to vaguely recall that this sometimes help with circular imports.

niklassiemer: I tried this, however, it does not seem to help for the circular import. What actually helps is to 
`import pyiron_base.project.generic` in the `project_browser.py`. I think each time one uses 
`from some.module import SomeClass` the `some.module` is init to get `SomeClass`. Just providing the `some.module` as namespace does not need an init at this point which breaks the circular import. 
In the present case, however, I would rather stick to the old implementation since otherwise we would need to have a `@import_warning` on the `Project` level which I do not like at all.

liamhuber:
@niklassiemer what are the options for testing? I recognize that with GUI work we may really be dealing with a different beast, but it would be nice if we can whittle down the hit from its current -1.6%.

niklassiemer:
It is certainly possible to test some of the routines, I will look into how much is possible.

niklassiemer:
Well sure, this feature requires ipywidgets which is not in our pyiron requirements. Now, we have two options of we want to test: adding ipywidgets to our requirements or to extend the environment while testing.

liamhuber:
> Well sure, this feature requires ipywidgets which is not in our pyiron requirements. Now, we have two options of we want to test: adding ipywidgets to our requirements or to extend the environment while testing.

Right, fair enough. I guess for now we should stick to having `ipywidgets` in [`extras_require`](https://github.com/pyiron/pyiron_contrib/blob/e3407b1b8863cbbd1e540ad41e601cc4906fffa0/setup.py#L40) instead of the main dependencies list. Then it would be nice to have the tests but somehow wrap them in a `try/except` so they (at most) throw  a warning if the necessary package isn't there. I guess this won't help the coverage badge, but at least we could test on our local machines! Maybe there is an even better solution?

niklassiemer:
Well, we could add an extra requirements environment and install these for the unit tests as well. E.g. for the notebooks we have a bigger environment.

pyiron_base/project/generic.py
niklassiemer: Now, I am using `static_isinstance` and do not need to import any `Project` in the `ProjectBrowser`.

liamhuber:
'static_is instance' seems dangerous for renaming refactorization, is there a way around it? Also feel free to ignore this live nitpicking feedback as I'm just hassling you from my phone and am missing the reason you switched over to start with 😝

niklassiemer:
> 'static_is instance' seems dangerous for renaming refactorization, is there a way around it? Also feel free to ignore this live nitpicking feedback as I'm just hassling you from my phone and am missing the reason you switched over to start with 😝

The only reason to import `Project` from `pyiron_base` is to check `isinstance`, thus with our `static_isinstance` util I don't need to import the Project any more 😄

liamhuber:
> > 'static_is instance' seems dangerous for renaming refactorization, is there a way around it? Also feel free to ignore this live nitpicking feedback as I'm just hassling you from my phone and am missing the reason you switched over to start with 😝
> 
> The only reason to import `Project` from `pyiron_base` is to check `isinstance`, thus with our `static_isinstance` util I don't need to import the Project any more 😄

Ah, ok. I guess this makes the test run faster because the project import is slow? If this is the only reason I'd actually be more comfortable going with the slow import than the string comparison.

With strings I always get very worried that the reference will get broken when someone does a refactor:rename or refactor:move, because either the IDE misses it or the human ignores some or all of the string matches.

niklassiemer:
Fine with me, I will revert this since the only reason was to exclude possible circular imports 😆. I actually have not checked timings for the tests. However, they are in the order of a few seconds for all of them.
What costs a bit is to install the additional requirements - about 30 sec per test.

liamhuber (approved):

niklassiemer:
As discussed I close this here and move it to a new pyiron_gui repo

PR #227, opened 2021-03-10
Fix Sphinx Warning

### LINKED ISSUE #228: Sphinx calculation gives visible depreciation warning
Basic Sphinx job: 
```
job = pr_ev.create.job.Sphinx(job_name)
job.structure = pr_ev.create_ase_bulk("Cu", a=a)
job.set_exchange_correlation_functional("PBE")
job.plane_wave_cutoff = 350
job.set_kpoints(mesh=[4, 4, 4])
job.set_occupancy_smearing("fermi", 0.1)
job.calc_minimize()
job.run()
```

Warning: 
```
/opt/conda/lib/python3.8/site-packages/pyiron_base/generic/hdfio.py:718: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  np.array([np.array(v) for v in value]),
```

jan-janssen:


pmrv:
Doesn't that stop multidimensional arrays from to being stored efficiently as one numerical numpy array?

jan-janssen:
We just change the type of the array as suggested by the warning - from my perspective the HDF5 content remains the same.

samwaseda:
I share the same concern with @pmrv. We need a better understanding of the data handling, especially regarding the following points:

- Disk space efficiency
- Speed
- Correct handling of data type when `object` is not required

jan-janssen:
https://github.com/h5io/h5io/blob/master/h5io/_h5io.py#L188

samwaseda:
> https://github.com/h5io/h5io/blob/master/h5io/_h5io.py#L188

What's the message here?

jan-janssen:
> > https://github.com/h5io/h5io/blob/master/h5io/_h5io.py#L188
> 
> What's the message here?

I wrote the code so it works.

samwaseda:
> > > https://github.com/h5io/h5io/blob/master/h5io/_h5io.py#L188
> > 
> > 
> > What's the message here?
> 
> I wrote the code so it works.

Hm I don't really see the connection between disk space efficiency and the lines given on the page in the link...

jan-janssen:
All this pull request does is remove the warning we currently have - there is no risk in this change basically because this behaviour is the same we had before, it is basically a bit more explicit. Previously the same type was chosen implicitly.

samwaseda (approved):
> All this pull request does is remove the warning we currently have - there is no risk in this change basically because this behaviour is the same we had before, it is basically a bit more explicit. Previously the same type was chosen implicitly.

So I'd then in principle trust you, but I just want to know what the line you shared actually does, because the function names are not quite self-explanatory. First of all, it's the writing part, right? It looks to me like "if it's not `object` then convert it to `ndarray` - otherwise create indices and store data in a somewhat clever way". Is this correct? If so, now with every single array converted to `object` with the change of this PR, every array will fall into `else`, right? Doesn't make it less efficient?

And lastly, what happens when the data is read?

jan-janssen:
When we have an array with unequal shapes, then pyiron handles this as list of arrays in contrast h5io expects an array, so when we communicate to h5io we create an array. Now in the latest version 1.20 numpy complains that the elements of the array have unequal shape so it warns about converting it to objects, unless we explicitly require it to be an object. But in this special case we only expect list of arrays with unequal shapes here, so that is expected.

samwaseda:
Ok I see I had missed [this line](https://github.com/pyiron/pyiron_base/blob/7650a8b07fdb8bfd3eda7786dd450a711ae09458/pyiron_base/generic/hdfio.py#L714). You can merge it now. I'm gonna clean the lines a little bit after this PR because I see quite a few repetitions in the code.

Issue #234, opened 2021-03-11
Recursive InputList subclasses get type changed to top parent


liamhuber:
When I subclass off of `InputList` and put this subclass into another `InputList`-like object, everything works fine until I load it from HDF...then they all get their type converted to whatever the parent type is. Am I missing something stupid like assigning a `__name__` attribute or something?

MWE:
```python
from pyiron_base import Project, InputList
pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)
hdf = pr.create_hdf(path=pr.path, job_name='storage')

# Basic example
class Sub(InputList):
    pass

data = InputList()
data.sub = Sub()
print(data)  # fine
> InputList({'sub': Sub([])})
data.to_hdf(hdf=hdf)
print(data)  # still fine
> InputList({'sub': Sub([])})
data.from_hdf(hdf=hdf)
print(data)  # bork
> InputList({'sub': InputList([])})

# Advanced: doesn't just convert to InputList, but to top-most class
class Sub1(InputList):
    pass

class Sub2(InputList):
    pass

data2 = Sub(table_name='another')
data2.sub1 = Sub1()
data2.sub1.sub2 = Sub2()
print(data2)  # fine
> Sub({'sub1': Sub1({'sub2': Sub2([])})})
data2.to_hdf(hdf=hdf)
print(data2)  # still fine
> Sub({'sub1': Sub1({'sub2': Sub2([])})})
data2.from_hdf(hdf=hdf)  
print(data2)  # bork
> Sub({'sub1': Sub({'sub2': Sub([])})})
```

My actual use case looks something like this:

```python
JobInput(InputList):
    def __init__(self, init=None, table_name='input'):
        super().__init__(init=init, table_name=table_name)
        self.sub_input = SubInput()

SubInput(InputList):
    pass
```

So when it converts the child to the parent type it winds up getting copies of itself inside itself! Somehow it doesn't get into an infinite loop, but it still gets really ugly.

liamhuber:
Nevermind, I checked out #216 and it works fine there -- it's just a matter of time.

E.g. this works beautifully:

```python
from pyiron_atomistics import Project
from pyiron_base import DataContainer
pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)
pr.remove(enable=True)

pr = Project('scratch')
hdf = pr.create_hdf(path=pr.path, job_name='storage')

class Sub(DataContainer):
    pass

data = DataContainer()
data.sub = Sub()
[...]

PR #216, opened 2021-03-02
Changes to InputList

jan-janssen:

PR #235, opened 2021-03-12
Use FileHDFio instead of ProjectHDFio to remove workaround

### LINKED ISSUE #229: load project_data on __init__
Concerning the recursive problem in #192 I just realized that the only reason to pass the `Project` itself to the `ProjectData` is the knowledge of the project path in `ProjectData`. All the advanced functionality of `ProjectHDFio` compared to a plain `FileHDFio` like database connection and knowledge of how to create a working directory are not used (and right now, they actually can't be used since the `ProjectHDFio` is not assigned). 
Therefore, we could either store `self._project_path` instead of  the whole `self._project` or just use the `self._project.path` to create a `FileHDFio` to read/write from.

niklassiemer:
closes #229


liamhuber:
I like it, but it is somehow breaking the test for storing complex data (in this case a nested list). I played around with it for a few minutes, but don't understand what's going wrong. I'm going to play around with #234 for a bit this morning, so hopefully I'll pick up some insight into `InputList` that helps to resolve this. Automatic reading is absolutely the behaviour we want!

liamhuber:
@niklassiemer I didn't explicitly find a solution, but I think that maybe we should wait until #216 is merged as that has a significant impact on the storage landscape.

jan-janssen (approved):
Looks good to me

jan-janssen:
@niklassiemer and @liamhuber as https://github.com/pyiron/pyiron_base/pull/216 is merged I guess this one can now be merged as well.

liamhuber (approved):
Nice! I will merge 😄

PR #237, opened 2021-03-16
Add workflow to run atomistic tests on push to master

pmrv:
Just a test for now, if this works I'll add it for contrib too.

### DIFF
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -30,8 +30,9 @@ jobs:
         conda env update --name test --file .ci_support/environment-notebooks.yml
         mkdir -p tests/compat
         git clone https://github.com/pyiron/pyiron_atomistics tests/compat/atomistics
-        pip install --no-deps tests/compat/atomistics
-        conda env update --name test --file tests/compat/atomistics/.ci_support/environment.yml
+        conda install -c conda-forge pyiron_atomistics
+        # pip install --no-deps tests/compat/atomistics
+        # conda env update --name test --file tests/compat/atomistics/.ci_support/environment.yml
     - name: Test
       shell: bash -l {0}
       run: |
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -36,5 +36,6 @@ jobs:
     - name: Test
       shell: bash -l {0}
       run: |
+        conda activate test
         cd tests/compat/atomistics
         python -m unittest discover tests/
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -6,6 +6,8 @@ name: Compatibility with pyiron_atomistics
 on:
   push:
     branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


pmrv:
@jan-janssen The tests are complaining about the periodic table not being there.  I suppose this is related to `mendeleev`, does this need extra setup?  I think I have it setup just like the [atomistic unit tests](https://github.com/pyiron/pyiron_atomistics/blob/master/.github/workflows/unittests.yml), so I'm not sure how to fix this.

EDIT: this is fixed by instaling pyiron_atomistics via conda, but now it complains that the atomistic job classes are undefined, which surprises me more.

.github/workflows/atomistics-compat.yml
```diff
@@ -0,0 +1,40 @@
...
+      shell: bash -l {0}
+      run: |
+        python .ci_support/pyironconfig.py
+        pip install --no-deps .
+        conda env update --name test --file .ci_support/environment-notebooks.yml
+        mkdir -p tests/compat
+        git clone https://github.com/pyiron/pyiron_atomistics tests/compat/atomistics
+        conda install -c conda-forge pyiron_atomistics
+        # pip install --no-deps tests/compat/atomistics
+        conda env update --name test --file tests/compat/atomistics/.ci_support/environment.yml
```
jan-janssen:
```suggestion
        cd  tests/compat/atomistics
        conda env update --name test --file .ci_support/environment.yml
        python .ci_support/pyironconfig.py
        pip install --no-deps .
```
Following https://github.com/pyiron/pyiron_atomistics/blob/master/.github/workflows/unittests.yml

.github/workflows/atomistics-compat.yml
```diff
@@ -0,0 +1,41 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: conda-incubator/setup-miniconda@v2
+      with:
+        auto-update-conda: true
+        python-version: 3.8
```
jan-janssen:
```suggestion
        python-version: 3.9
```

jan-janssen (approved):
It took a couple of iterations but from my perspective this is ready to be merged now.

pmrv:
Thanks for fixing this! I'll just change it to run on merge only.

jan-janssen:
> Thanks for fixing this! I'll just change it to run on merge only.

I am still not fully convinced that this is the best solution - for example the pull request with the `inputlist` it would be nice to test this functionality there right? Does this mean the developer has to manually enable the test then?

pmrv:
> > Thanks for fixing this! I'll just change it to run on merge only.
> 
> I am still not fully convinced that this is the best solution - for example the pull request with the `inputlist` it would be nice to test this functionality there right? Does this mean the developer has to manually enable the test then?

We discussed this originally, because we anticipated that the integration tests take a lot of time and it would be sufficient to fix this before a new version release.  Currently this takes ~22min, which is not exactly fast, but may be acceptable.  I am open to change it back.

jan-janssen:
We could add an integration label which the reviewers can use to trigger additional test for a given pull request which most likely affects other parts of the code as well: 
```
if: ${{ github.event.label.name == 'label_name' }}
```
https://stackoverflow.com/a/62331521/6996746

pmrv:
> We could add an integration label which the reviewers can use to trigger additional test for a given pull request which most likely affects other parts of the code as well:
> 
> ```
> if: ${{ github.event.label.name == 'label_name' }}
> ```
> 
> https://stackoverflow.com/a/62331521/6996746

Seems to work in #245 will add here, after it is merged.

PR #238, opened 2021-03-16
Clarify job name restrictions

pmrv:
Job names have to be valid python identifiers, came up in [https://github.com/pyiron/pyiron/issues/1320].


jan-janssen (approved):
Looks good to me

PR #245, opened 2021-03-18
Add workflow to to run contrib tests on push to master

pmrv:
Same as #237, but for pyiron_contrib.


pmrv:
It seems pyiron_base and pyiron_contrib currently have incompatible requirements.  Since it works for atomistics though, it this is what we need for #216 I suggest to merge this now and then discuss the contrib situtation tomorrow.

PR #251, opened 2021-03-18
Fix HDF5 interface 

jan-janssen:
Sometimes small changes have a big impact


jan-janssen:
The second part was even worse: 
```
w = False
v = False,
if v: 
    print("v")
if w: 
    print("w")
```
prints `v` ... because `False,` is `True`

PR #256, opened 2021-03-20
Replace PyironObject with meaningful prototype

liamhuber:
Right now `PyironObject` is just a sort of pseudo-abstract base class. It suggests a handful of attributes and methods to implement and then raises a `NotImplementedError` for each of them `job.core.JobCore`, `generic.parameters.GenericParameters`, and `server.generic.Server` in `pyiron_base`, and `pyiron_atomistics.dft.waves.bandstructure.Bandstructure`. *Only* `JobCore` actually implements all the recommended methods, the rest fall woefully short.

Removing the class entirely had no effect on any of the tests (indeed can't since all it does is make suggestions). I'd like to get rid of it entirely to free up the very useful name `PyironObject` so we can develop a beefier version of the same thing cleanly.

EDIT: following Jan's suggestion, removing the old pseudo-documentation class is now done simultaneously with reintroducing a new class of the same name. The new version is not yet hooked up to anything (an ABC with no children) but preserves the `PyironObject` space in the codebase and serves as a guiding prototype.


liamhuber:
Crap, added the integration tag too soon. Atomistic will fail, but once I merge atomistics#134 over there then I'll add it again. At that point I expect it to go smoothly.

jan-janssen:
I am confused about this pull request, I thought the idea is that all objects should be pyiron objects, no this goes in the opposite direction.

liamhuber:
Yes, that's where I'd like to end up. Right now the namespace `PyironObject` is taken up by a class without any influence. I want to clean it up so we can make an influential version.

jan-janssen:
But would not it be easier to extend the class?

liamhuber:
I don't think so. We definitely want to remove dependence on it from the other classes so that we actually know who is *actually* using the real `PyironObject` and who is not. As for extension, the things we do want are just empty function names, and there is also things we'd need to remove anyhow.

At the end of the day, in the current version of pyiron this is literally dead code. IMO we should just prune the dead code.

liamhuber:
The PR itself obviously doesn't require much review though, so this discussion is why I added you as reviewer.

I am not 100% dead set against keeping the dead class around, but I do think we should have a good reason if we're going to do that.

jan-janssen:
The main intention of the pyiron object class is a place holder for all objects which should ideally be independent self contained objects. At the current stage the pyiron objects are mainly used to assist developers - that is why the import errors were added. This allowed us to draw the correct UML diagram. But if you want to start working on this that is fine for me, still I would be happy to discuss these kind of changes in a pyiron meeting before. As you already started with https://github.com/pyiron/pyiron_atomistics/pull/134 I would hand it over to you and then you can present it in the next pyiron meeting.

jan-janssen:
As this contains the changes to `InputList` please wait for https://github.com/pyiron/pyiron_base/pull/257 to be merged first.

jan-janssen:
To explain this a bit more and in particular as the InputList migration crashed again: 
* I do not want to prevent anybody from implementing new ideas in pyiron. 
* Still I have the feeling that in the current stage we try a lot of things but only a few of them are going to be executed to the end.
* As a consequence I am open to any pull request which improves pyiron, but I am reluctant to merge pull requests which delete existing code with the promise to implement a better solution later.

liamhuber:
> As this contains the changes to InputList please wait for #257 to be merged first.

This commit is completely unrelated to `InputList`. I guess those changes got inserted as a result of all the reversions with #256. 

I rebased and force pushed so this is cleanly just the actual changes I'm proposing. I'll let the tests run, but then I'll remove draft.

liamhuber:
> To explain this a bit more and in particular as the InputList migration crashed again:
> 
> * I do not want to prevent anybody from implementing new ideas in pyiron.
> * Still I have the feeling that in the current stage we try a lot of things but only a few of them are going to be executed to the end.
> * As a consequence I am open to any pull request which improves pyiron, but I am reluctant to merge pull requests which delete existing code with the promise to implement a better solution later.

I am 100% on board with these principles. My assertion is two-fold: (1) this PR improves pyiron, and (2) this PR doesn't remove any code.

I realize that the second claim sounds counter-intuitive since I delete a class here, but I think there is an extremely strong case that this is documentation masquerading as code. The evidence for this is that removing the class and re-basing all its children of `object` had *zero* impact on our functionality.

> The main intention of the pyiron object class is a place holder for all objects which should ideally be independent self contained objects. At the current stage the pyiron objects are mainly used to assist developers - that is why the import errors were added. This allowed us to draw the correct UML diagram.

This is absolutely a valid argument, but I don't find it a very convincing argument. I actually feel that this code-as-documentation makes things *more* complicated for developers because of the inconsistent application of the principle. Many objects that should be `PyironObject` children are not. Of those that are, only `JobCore` actually implements everything `PyironObject` wants. Thus, even if we were regularly advertising the UML diagram to new developers it would be misleading.

The first point is actually much easier to make: right now only `JobCore` actually implements all of `PyironObject`'s fields. That means that it is cluttering up the tab completion lists of `GenericParameters`, `Server`, and `Bandstructure` with garbage that just complain's that it's not implemented -- and never will be! Why would the `Server` need a `master_id`? This is obviously minor, but it's also obviously undesirable.

I feel strongly that the `PyironObject` deleted here should be removed from the inheritance tree (commit 2fc...). IMO all the good things about `PyironObject` are re-declared and re-documented in `JobCore` so I would also like to directly delete the `template` file (commit 7e78...), but I definitely feel less strongly on this point and can be convinced to leave it around as a stub.

Definitely we can hold off any merge until after discussing on Friday.

liamhuber:
@jan-janssen should be good for review at your convenience now. The new code isn't hooked up to anything, but I think it gives a solid foundation. 

After the meeting today my understanding is that we will develop a new class hierarchy on top of this instead of trying to keep compatibility to the old classes -- mostly due to backwards compatibility in the storage format.

pyiron_base/generic/object.py
```diff
@@ -0,0 +1,93 @@
...
+__version__ = "0.0"
+__maintainer__ = "Liam Huber"
+__email__ = "huber@mpie.de"
+__status__ = "development"
+__date__ = "Mar 23, 2021"
+
+s = Settings()
+
+
+class HasStorage(ABC):
```
pmrv, line 27: To help "super-users" creating new classes, I want a class that saves its attributes automatically.  This came up with @Leimeroth in pyiron_contrib for example.  This would work nicely with `HasStorage` since we could just route `__getattr__`/`__setattr__` to `self.storage`.  I could add this here or in a separate class, thoughts?

jan-janssen (approved):
Looks good to me

pyiron_base/generic/object.py
liamhuber: @pmrv , my trite answer is that we already have this class -- it's `DataContainer` and it's beautiful.

More seriously, I guess you mean such a class that *also* has the database functionality, e.g. to be logged there and referenced by other pyiron objects later.

I definitely like the direction you're going with the setters. I don't want this to be the standard behaviour, because there are many times we *don't* want an attribute to be serialized, e.g. consider `job.project` where we really don't want to be serializing the project repeatedly with every job. And we see already in `DataContainer` that once serialization is the default getting around it is a pain. However, I see it could be useful in a bunch of cases.

I think we'll probably need to add more bells and whistles than just the (ge)setters at the end of the day, but it's a good start. What about something like this:

```python
class PyironData(PyironObject):
  def __getattr__(self, key):
    return self.storage[key]

  def __setattr__(self, key, value):
    return self.storage[key] = value
```

It's just not totally clear to me how this sort of magic method overwrite will play out in the init of the super classes where we do stuff like `self._storage = DataContainer(table_name='storage')`.

Another option is to brute force it by putting wrappers around *each* field we care about:

```python
class MyData(PyironObject):
  def __init__(self, *args, **kwargs):
    super().__init__()
    self.storage._field1 = 'foo'
    self.storage._field2 = 42

  @property 
  def field1(self):
    return self.storage._field1

  @field1.setter 
  def field1(self, val):
    self.storage._field1 = val

  @property 
  def field2(self):
    return self.storage._field2

  @field2.setter 
  def field2(self, val):
    self.storage._field2 = val
```

If we have to go the second route, it would be nice to have some sort of wrapper to procedurally add these getters and setters so we can write something very roughly like

```python
@expose_storage('field1', 'field2')
class MyData(PyironObject):
  def __init__(self, *args, **kwargs):
    super().__init__()
    self.storage._field1 = 'foo'
    self.storage._field2 = 42
```

Anyhow, as Jan already approved this version of the PR I'll merge it later tonight and we can continue these data ideas in a new PR 😉

PR #265, opened 2021-03-23
Add note about subclassing DataContainer

pmrv:
I added some of the points discussed [here](https://github.com/orgs/pyiron/teams/pyiron/discussions/51/comments/3) to reduce friction in the future.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -153,9 +153,10 @@ class DataContainer(MutableMapping):
     >>> list(pl.keys())
     [0, 1, 2, 3]
 
-    .. attention::
+    .. attention:: Subclasses beware!
 
-        Subclasses beware!  DataContainer require some careful treatment when creating subclasses.
+        DataContainer require some careful treatment when creating subclasses.
+
         1. Since DataContainers are expected to recursively instantiate themselves subclasses need to keep their
         `__init__ compatible to the base class.  That means being able to be instantiated without arguments, if
         arguments are given the first one (or `init`) has to accept a Mapping or Iterable.  Additional arguments may be
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -168,7 +168,7 @@ class DataContainer(MutableMapping):
         3. Subclasses should always be thought of as general data structures, if you want to subclass to have access to
         the HDF5 functionality or the way the DataContainer is shown in jupyter notebooks, but only have a fixed number
         of attributes it is better to create a new class that has an DataContainer as an attribute and dispatch to the
-        :method:`.DataContainer.from_hdf`, :method:`.DataContainer.to_hdf` and :method:`.DataContainer._repr_json_`
+        :method:`DataContainer.from_hdf`, :method:`DataContainer.to_hdf` and :method:`DataContainer._repr_json_`
         methods.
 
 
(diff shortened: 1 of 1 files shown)


jan-janssen (approved):
Loks good to me

pyiron_base/generic/datacontainer.py
```diff
@@ -152,6 +152,18 @@ class DataContainer(MutableMapping):
     [3, 2, 1, 0]
     >>> list(pl.keys())
     [0, 1, 2, 3]
+
+    .. attention:: Subclasses beware!
```
liamhuber, line 156: Can you double check this is the right docstring format? I thought we used `Attention:` rather than `.. attention::`

pyiron_base/generic/datacontainer.py
```diff
@@ -152,6 +152,18 @@ class DataContainer(MutableMapping):
...
+
+        DataContainer require some careful treatment when creating subclasses.
+
+        1. Since DataContainers are expected to recursively instantiate themselves subclasses need to keep their
+        `__init__ compatible to the base class.  That means being able to be instantiated without arguments, if
+        arguments are given the first one (or `init`) has to accept a Mapping or Iterable.  Additional arguments may be
+        added, but must be after `init` and must have a default.
+        2. Creating new instance attributes that don't live in the container itself, you need to use
```
liamhuber: This one is especially confusing

liamhuber (changes requested):
Great idea!

I think it would benefit from at least one concrete example.

pyiron_base/generic/datacontainer.py
pmrv: This is reST specific syntax to get rendered as a little pop up box, see the [pre-render](https://pyiron-base.readthedocs.io/en/docs/apidoc/pyiron_base.generic.datacontainer.html).  There's actually a small issue with the list inside not getting properly rendered.

pyiron_base/generic/datacontainer.py
pmrv: ralph

I'll re-word it a bit. :/

pmrv:
> Great idea!
> 
> I think it would benefit from at least one concrete example.

Agree, will add a small example subclass.

pyiron_base/generic/datacontainer.py
liamhuber: 🤣 I just meant the concept, not that the phrasing is so bad. I think this case desperately needs to be shown not just told.

pyiron_base/generic/datacontainer.py
```diff
@@ -161,9 +161,62 @@ class DataContainer(MutableMapping):
         `__init__ compatible to the base class.  That means being able to be instantiated without arguments, if
         arguments are given the first one (or `init`) has to accept a Mapping or Iterable.  Additional arguments may be
         added, but must be after `init` and must have a default.
-        2. Creating new instance attributes that don't live in the container itself, you need to use
+
+        2. Creating new instance attributes that don't live in the container itself is possible, but you need to use
         `object.__setattr__` the first time you define that attribute.  Afterwards using normal assignment syntax works.
 
+        3. Subclasses should always be thought off as general data structures, if you want to subclass to have access to
```
liamhuber: off -> of

liamhuber (approved):
Super

pyiron_base/generic/datacontainer.py
pmrv:
```suggestion
        3. Subclasses should always be thought of as general data structures, if you want to subclass to have access to
```

PR #266, opened 2021-03-23
Fix ScriptJob

jan-janssen:

PR #267, opened 2021-03-23
Fix #264 and add tests.

### LINKED ISSUE #264: ScriptJob broken with data container
Same issue as https://github.com/pyiron/pyiron_base/issues/44

```
from pyiron import Project
pr = Project('JOB')
job = pr.create_job('ScriptJob', 'job')
job.input['value'] = 300
job.script_path = 'test.ipynb'
job.run()
```
in `test.ipynb`: 

```
from pyiron import Notebook as nb
coeff_tot = nb.get_custom_dict()
print(coeff_tot['value'])
```

The error message I get: 

```
<ipython-input-1-34e7bd0a0faf> in <module>
      1 from pyiron import Notebook as nb
----> 2 coeff_tot = nb.get_custom_dict()
      3 print(coeff_tot['value'])

~/PycharmProjects/pyiron_base/pyiron_base/job/external.py in get_custom_dict()
     38         if Path(hdf_file).exists():
     39             hdf = FileHDFio(hdf_file)
---> 40             custom_dict = hdf[folder + '/input/custom_dict/data']
     41             custom_dict["project_dir"] = str(project_folder)
     42             return custom_dict

~/PycharmProjects/pyiron_base/pyiron_base/generic/hdfio.py in __getitem__(self, item)
    833                     hdf_object = self.copy()
    834                     hdf_object.h5_path = "/".join(item_abs_lst[:-1])
--> 835                     return hdf_object[item_abs_lst[-1]]
    836 
    837     def _read(self, item):

~/PycharmProjects/pyiron_base/pyiron_base/generic/hdfio.py in __getitem__(self, item)
    805                         obj = hdf_item.copy()
    806                         return obj
--> 807                 raise ValueError("Unknown item: {}".format(item))
    808             else:
    809                 if item_lst[0] == "":  # item starting with '/', thus we have an absoute HDF5 path

ValueError: Unknown item: data
```

pmrv:
Seems you were a bit quicker, but from my test #47 still works fine and has the tests already.

### DIFF
--- a/pyiron_base/job/external.py
+++ b/pyiron_base/job/external.py
@@ -36,7 +36,7 @@ class Notebook(object):
         hdf_file = project_folder / folder
         hdf_file = str(hdf_file) + ".h5"
         if Path(hdf_file).exists():
-            hdf = FileHDFio(hdf_file)._create_project_from_hdf5()
+            hdf = FileHDFio(hdf_file).create_project_from_hdf5()
             custom_dict = hdf[folder + '/input/custom_dict'].to_object()
             custom_dict["project_dir"] = str(project_folder)
             return custom_dict
--- a/pyiron_base/job/external.py
+++ b/pyiron_base/job/external.py
@@ -37,7 +37,7 @@ class Notebook(object):
         hdf_file = str(hdf_file) + ".h5"
         if Path(hdf_file).exists():
             hdf = FileHDFio(hdf_file).create_project_from_hdf5()
-            custom_dict = hdf[str(folder) + '/input/custom_dict'].to_object()
+            custom_dict = hdf[folder + '/input/custom_dict'].to_object()
             custom_dict["project_dir"] = str(project_folder)
             return custom_dict
         elif Path("input.json").exists():
--- a/pyiron_base/job/external.py
+++ b/pyiron_base/job/external.py
@@ -37,7 +37,7 @@ class Notebook(object):
         hdf_file = str(hdf_file) + ".h5"
         if Path(hdf_file).exists():
             hdf = FileHDFio(hdf_file).create_project_from_hdf5()
-            custom_dict = hdf[folder + '/input/custom_dict'].to_object()
+            custom_dict = hdf[str(folder) + '/input/custom_dict'].to_object()
             custom_dict["project_dir"] = str(project_folder)
             return custom_dict
         elif Path("input.json").exists():
--- a/pyiron_base/job/external.py
+++ b/pyiron_base/job/external.py
@@ -37,6 +37,7 @@ class Notebook(object):
         hdf_file = str(hdf_file) + ".h5"
         if Path(hdf_file).exists():
             hdf = FileHDFio(hdf_file).create_project_from_hdf5()
+            print("Debug_notebook:", folder)
             custom_dict = hdf[folder + '/input/custom_dict'].to_object()
             custom_dict["project_dir"] = str(project_folder)
             return custom_dict
--- a/pyiron_base/job/external.py
+++ b/pyiron_base/job/external.py
@@ -8,7 +8,7 @@ Load input parameters for jupyter notebooks from external HDF5 or JSON file
 import json
 from pathlib2 import Path
 import warnings
-from pyiron_base.generic.hdfio import FileHDFio
+from pyiron_base.generic.datacontainer import DataContainer
 
 __author__ = "Osamu Waseda"
 __copyright__ = (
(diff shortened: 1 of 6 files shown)


pyiron_base/job/external.py
```diff
@@ -36,8 +36,8 @@ def get_custom_dict():
         hdf_file = project_folder / folder
         hdf_file = str(hdf_file) + ".h5"
         if Path(hdf_file).exists():
-            hdf = FileHDFio(hdf_file)
-            custom_dict = hdf[folder + '/input/custom_dict/data']
+            hdf = FileHDFio(hdf_file)._create_project_from_hdf5()
```
jan-janssen: I guess using an internal function is not recommended. Maybe we should make it a public function.

pmrv: The same method is also used by `GenericJob`, so making it public is probably ok.

pmrv: See #268.

jan-janssen (approved):
The code looks good to me - so once the tests pass feel free to merge it. 

Ps: in addition we should run the integration tests once as maybe somebody - e.g. me - was using the private function in `pyiron_atomistics`.

jan-janssen:
@pmrv I am a bit pushing on this pull request as we have users on the cluster who need this rather quickly.

pmrv:
> The code looks good to me - so once the tests pass feel free to merge it.
> 
> Ps: in addition we should run the integration tests once as maybe somebody - e.g. me - was using the private function in `pyiron_atomistics`.

I grepped through atomistics and contrib but there doesn't seem to be any calls of `_create_project_from_hdf5` in either.

pmrv:
@niklassiemer I guess neither me nor @jan-janssen have a windows installation ready to go, can you have a look what's going on with the windows paths?

niklassiemer:
Yes, I will have a look this evening - probably starting around 8pm.

pmrv:
> Yes, I will have a look this evening - probably starting around 8pm.

Thanks a lot!

jan-janssen:
Finally I used the solution proposed in https://github.com/pyiron/pyiron_base/pull/266

Looking at the previous implementation in detail, we create a FileHDFio object, from it we create a Project, from the project we open the HDF5 to get a ProjectHDF object and then we call to_object() to get a DataContainer. In the new implementation I go from a FileHDFio object directly to DataContainer.

tests/job/test_external.py
```diff
@@ -0,0 +1,38 @@
...
+
+from pyiron_base import Project
+
+job_py_source = """
+from pyiron_base import Notebook as nb
+coeff_tot = nb.get_custom_dict()
+print(coeff_tot['value'])
+"""
+
+class TestScriptJob(unittest.TestCase):
```
liamhuber, line 12: While we're touching this it would be very nice to convert it over to subclass of `pyiron_base._tests.TestWithProject` and remove a bunch of the setup and teardown code.

jan-janssen: As we need the changes in the pull request rather urgently I would move that to a separate pull request

tests/job/test_scriptjob.py
```diff
@@ -0,0 +1,30 @@
+import os
+import unittest
+
+from pyiron_base import Project
+
+class TestScriptJob(unittest.TestCase):
```
liamhuber, line 6: Similarly here. 

I don't object if anyone wants to go through and implement the new parentage everywhere it applies, but IMO it's sufficient if we just update things as we come across them.

tests/job/test_external.py
liamhuber: It's five minutes. How urgent is urgent? Let me put my kid to bed and I'll do it (~20:15).

jan-janssen: I proposed the fix 4 hours ago https://github.com/pyiron/pyiron_base/pull/266  and we still need to create the conda package plus wait for the release to be available on the cluster. I guess the update of the test can also be done in a separate pull request and it is not required on the cluster.

pmrv:
> Finally I used the solution proposed in #266
> 
> Looking at the previous implementation in detail, we create a FileHDFio object, from it we create a Project, from the project we open the HDF5 to get a ProjectHDF object and then we call to_object() to get a DataContainer. In the new implementation I go from a FileHDFio object directly to DataContainer.

Agree that our Project/HDF layer has too many indirections.

jan-janssen:
@muh-hassani Can you update `pyiron_base` to 0.2.8 to fix the ScriptJob issue?

PR #271, opened 2021-03-23
Use test class filepath instead of parent class filepath

liamhuber:
`TestWithProject` was using its own filepath for creating projects, which lead to headaches as discussed [here]() when tests were run in parallel. Now it uses the filepath of the child class, so each test module gets its own unique project name.

@pmrv and @jan-janssen I still don't know enough about how parallelized the remote testing environment is -- is it sufficient if each test module gets its own project, or does each test class within that project need one? e.g. right now the tests for `pyiron_base/generic/object` make three test classes inside a single module.

### DIFF
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -5,10 +5,11 @@
 """Classes to help developers avoid code duplication when writing tests for pyiron."""
 
 import unittest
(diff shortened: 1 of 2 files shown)


jan-janssen (approved):
Looks good to me

PR #272, opened 2021-03-24
New tables

liamhuber:
As discussed in the last tech dev meeting, we will need new pyiron database tables going forward.

Git is the one true way to iteratively collaborate on written material, so I am just immediately including this in the codebase. I also extracted out [the existing table](https://github.com/pyiron/pyiron_base/blob/e59f4682ba6228f725c0a8d10c8b2ed93551890e/pyiron_base/database/generic.py#L106) the exact same way, so we can clearly see how to use these. Once we have something we're happy with we can also just inline the tables again if we don't want this extra module kicking around.

Let's play around with it! I really streamlined down the existing table so the columns should be compatible even with data objects (as opposed to our jobs which actually run) and the needs of experimentalists, although @niklassiemer you should give feedback on that front. I also included a simple pair-relationship table as suggested by @pmrv.

### DIFF
--- a/pyiron_base/database/tables.py
+++ b/pyiron_base/database/tables.py
@@ -1,0 +1,79 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+"""
+Classes defining the shape of pyiron's database tables.
+"""
+
+from sqlalchemy import (
+    Column,
+    DateTime,
+    Integer,
+    String,
+    Table,
+)
+from abc import ABC
+
+__author__ = "Liam Huber, Murat Han Celik"
+__copyright__ = (
+    "Copyright 2020, Max-Planck-Institut für Eisenforschung GmbH"
+    " - Computational Materials Design (CM) Department"
+)
+__version__ = "0.0"
+__maintainer__ = "Liam Huber"
+__email__ = "huber@mpie.de"
+__status__ = "development"
+__date__ = "Mar 24, 2021"
+
+
+class PyironTable(ABC):
+    def __init__(self, table_name, metadata):
+        self._table = None
+
+    @property
+    def table(self):
+        return self._table
+
+
+class ObjectTable(PyironTable):
+    """
+    A table for storing pyiron objects. Only very primitive metadata is stored and more complex queries should use
+    a `PyironTable`.
+    """
+    def __init__(self, table_name, metadata):
+        super().__init__(table_name, metadata)
+        self._table = Table(
+            table_name,
+            metadata,
+            Column("id", Integer, primary_key=True, autoincrement=True),  # Globally unique
+            Column("location", String(511)),  # How expensive would it be to make this even bigger???
+            Column("name", String(50)),  # Unique within location
+            # Note: location+name provides exact location of serialized data
+            Column("pyiron", String(20)),  # e.g. pyiron_base, pyiron_contrib
+            Column("version", String(20)),  # Which version of pyiron used
+            Column("class", String(255)),  # Where to locate the class within pyiron
+            # e.g.
+            # pyiron=pyiron_base, class=ScriptJob; pyiron=pyiron_atomistics, class=not.a.thing.in.init.ClassName; etc.
+            # pyiron+version+class provides all necessary information to re-instantiate the right object
+            # (pyiron+version+class)+(location+name) gives all necessary information to fully recreate object instance
+            Column("created", DateTime),  # When the object was first saved
+            Column("status", String(20)),  # For data-like objects always simple like "created", for job-like is complex
+            Column("username", String(20)),  # Who's doing this (for access filtering)
+            extend_existing=True,
+        )
+
+
+class RelationTable(PyironTable):
+    """
+    A table for storing relational links between pyiron objects. Only the existence of the link is stored, the nature
+    of the link must be found by examining the pyiron object itself.
+    """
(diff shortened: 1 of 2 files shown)


liamhuber:
I would also love to get this as classes instead of functions, because I think `self.simulation_table = SimulationTable(table_name, self.metadata)` is much prettier than `self.simulation_table = simulation_table(table_name, self.metadata)`. But I had a hell of a time with inheriting from `Table` and couldn't get anything that worked.

liamhuber:
@jan-janssen @muh-hassani @Ahmed-Aslam I don't want to assign reviewers right now as this is just for brainstorming, but I want to ping you explicitly so you can give your thoughts when you have time 😄

pyiron_base/database/tables.py
```diff
@@ -0,0 +1,91 @@
...
+def relation_table(table_name, metadata, extend_existing=True):
+    """
+    A table for storing relational links between pyiron objects. Only the existence of the link is stored, the nature
+    of the link must be found by examining the pyiron object itself.
+    """
+    return Table(
+        table_name,
+        metadata,
+        Column("id1", Integer),
+        Column("id2", Integer),
```
niklassiemer, line 89: How about to include a `relation` column to specify the relation. This might be optional, however, I assume it would be helpful to know that we have e.g. a parent-child relation.
```suggestion
        Column("id2", Integer),
        Column("relation", String(20)),
```

niklassiemer:
For the meta data of different objects itself (i.e. what we 'have to include' in CoScInE for the SFB) we probably would need another table e.g. to store an ID-string and some other of the metadata we collect to allow for SQL queries instead of scrolling through all data. As mentioned by JNmpi, this could be generated by a `PyironTable` as intermediate solution.

pyiron_base/database/tables.py
liamhuber: Yeah, in general I like this. I.e. the table can be translated to "`id1` has `relation` `id2`". For something like an MD job having a structure this is perfect and (I think? basically) cost-free. For something like parent-child though we get storage inefficiencies as we have `1, parent, 2` and `2, child, 1`. I seem to recall Joerg and @pmrv having some concern about this. From my side your suggestion is good though.

liamhuber:
> For the meta data of different objects itself (i.e. what we 'have to include' in CoScInE for the SFB) we probably would need another table e.g. to store an ID-string and some other of the metadata we collect to allow for SQL queries instead of scrolling through all data. As mentioned by JNmpi, this could be generated by a `PyironTable` as intermediate solution.

I had understood `PyironTable` as the target solution rather than intermediary, and pyiron-table-queries rather than SQL queries? In this context I see all the CoScInE 'have to include' fields being everything we need to populate the corresponding `PyironObject`.

niklassiemer:
We also discussed to include a data base table with often queried fields from the 'global' `PyironTable`. However, this would be an additional step to speed up these queries in terms of SQL queries instead of following all connections 'by hand'. However, I might be wrong 😅

pyiron_base/database/tables.py
```diff
@@ -77,3 +78,29 @@ def __init__(self, table_name, metadata):
...
+            table_name,
+            metadata,
+            Column("id", Integer, primary_key=True, autoincrement=True),
+            Column("parentid", Integer),
+            Column("masterid", Integer),
+            Column("projectpath", String(50)),
+            Column("project", String(255)),
+            Column("job", String(50)),
+            Column("subjob", String(255)),
+            Column("chemicalformula", String(30)),
```
max-hassani: I don't know how exactly we can change this, but `chemicalformula`, `hamilton`, and `hamversion` are specific to atomistics jobs. One way can be to have separate tables for different pyiron modules (continuum, atomistics, etc), which I believe is not desirable. On the other hand we can keep the atomistic columns and add continuum/experimental related columns as well, then we would have a sparse table.

liamhuber: Yes, they are definitely atomistic-specific. And in the case of the hamilton stuff they're not even well named any more. I would propose to stop using this table entirely and start using only the sparse table. I refactored the existing table out here for pedagogical reasons, but it also highlights how little this code belongs in `pyiron_base`!

I hadn't thought of including sub tables for each repo, but I'm open to it!

jan-janssen:
Maybe it is easier to move these new developments to pyiron_contrib or pyiron_experimental first before integrating them in pyiron_base this allows us to change the new interface more frequently until it is converged to a general solution.

liamhuber:
> Maybe it is easier to move these new developments to pyiron_contrib or pyiron_experimental first before integrating them in pyiron_base this allows us to change the new interface more frequently until it is converged to a general solution.

I would rather not. From the suggest of pyiron_experimental I first want to clarify that I imagined the new tables being the universal tables -- that simulations would also use this reduced `object_table` set. @muh-hassani had a nice idea for domain-specific sub-tables, but I see that as an extension to the leaner main table not an alternative.

pyiron_contrib is not a crazy suggestion, but I think it is not a perfect fit. For instance, a naive interpretation would lead to circular imports. E.g. if we really moved `simulation_table` over there and started using it. From my perspective these are base-centric objects and the development belongs in base.

With that said, I don't have any expectation of merging this into master any day soon. If we're serious about building new pyiron_basic infrastructure, I don't see a more sensible place to do that than on a branch of pyiron_basic. With that said, I'm not a fan of a bunch of stub code or parallel solutions floating around on `master`, so I don't want to actually *merge* any of this until we're all happy with it and ready to do a major version bump. I'm open to other workflows, but I would want a clear explanation of how we'd handle the technicalities of doing this base development on pyiron_contrib before moving anything there.

PR #274, opened 2021-03-25
Update _tests.py

niklassiemer:
Since `tearDownClass` is a `classmethod` we have to redefine `cls.file_location` and `cls.project`, don't we?


liamhuber:
It seems to not be needed. I don't have a python answer for you, but empirically (a) if I add `print(cls.file_location)` to the existing teardown I get the right path, and (b) my projects are getting properly removed as-is.

pmrv:
Since `setupClass` is called before `tearDownClass` `file_name` etc. are already defined as class attributes, so it should be fine.

PR #286, opened 2021-04-12
Fix iterating for an empty project while passing `status`

pmrv:
As per the commit messages, previously calling `Project().iter_jobs(status=...)` caused a weird error because we indexed an empty pandas table.

### DIFF
--- a/tests/project/test_project.py
+++ b/tests/project/test_project.py
@@ -39,6 +39,21 @@ class TestProjectData(unittest.TestCase):
         self.assertEqual(1, len(project2.data))
         self.assertEqual(self.project.data.foo, project2.data.foo)
 
+    def test_iterjobs(self):
+
+        try:
+            for j in self.project.iter_jobs():
+                pass
+        except:
+            self.fail("Iterating over empty project should not raise exception.")
+
+
+        try:
+            for j in self.project.iter_jobs(status="finished"):
+                pass
+        except:
+            self.fail("Iterating over empty project with set status flag should not raise exception.")
+
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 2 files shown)


tests/project/test_project.py
```diff
@@ -39,6 +39,21 @@ def test_data(self):
         self.assertEqual(1, len(project2.data))
         self.assertEqual(self.project.data.foo, project2.data.foo)
 
+    def test_iterjobs(self):
+
+        try:
+            for j in self.project.iter_jobs():
+                pass
+        except:
+            self.fail("Iterating over empty project should not raise exception.")
```
niklassiemer, line 48: Since a test fails if an exception is raised and this way we do not actually see the exception any more I would prefer
```suggestion
        except Exception as e:
            raise AssertionError("Iterating over empty project should not raise exception.") from e
```

tests/project/test_project.py
```diff
@@ -39,6 +39,21 @@ def test_data(self):
...
+
+
+        try:
+            for j in self.project.iter_jobs(status="finished"):
+                pass
+        except:
+            self.fail("Iterating over empty project with set status flag should not raise exception.")
```
niklassiemer, line 55: Same here
```suggestion
        except Exception as e:
            raise AssertionError("Iterating over empty project should not raise exception.") from e
```

niklassiemer (commented):
Looks good! I would just retain the initial error message in the test output.

tests/project/test_project.py
pmrv: When I check the test output at 92d87d4 the key error also pops up in the output.  In this case I'd prefer calling the `fail` method, because that makes it clear as a test failure not an exception.  I know raising `AssertionError` is handled by `unittest` as a test failure, so I won't insist, it's a just a style choice.

niklassiemer: I agree, its just about style - and as long as the `self.fail` also shows/includes the raised Exception from the code I am happy with either solution.

niklassiemer:
I triggered the integration tests and would merge once everything is passing.
After the merge, I would like to trigger a new release.

PR #295, opened 2021-04-23
workaround the limitation for the length of the chemical formula in the database

max-hassani:
This fixes the issue with the limitation of the database with chemical formulas longer than 30 characters. The new code tests the length and truncates the formula to the last element.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -372,6 +372,33 @@ class DatabaseAccess(object):
             output_list += [dict(zip(col.keys(), tmp_values))]
         return output_list
 
+    def _truncate_chem_formula(self, chem_formula):
+        """
+        this functions truncate the chemical formula to the last element,
+        up to the limit of the database.
+        arg:
+        chem_form(str): the chemical formula
+        """
+        acceptable_len=30
+        if len(chem_formula) >= acceptable_len:
+            for i in range(acceptable_len):
+                if not chem_formula[acceptable_len-i].isdigit():
+                    return chem_formula[:acceptable_len-i+1]
+            return chem_formula[:acceptable_len]
+        else:
+            return chem_formula
+
+    def _check_chem_formula_length(self, par_dict):
+        """
+        performs a check whether the length of chemical formula exceeds the defined limit
+        args:
+        par_dict(dict): dictionary of the parameter
+        """
+        for key, value in par_dict:
+            if key == 'chemicalformula':
+               par_dict[key] = self._truncate_chem_formula(value)
+        return par_dict
+
     # Item functions
     def add_item_dict(self, par_dict):
         """
@@ -400,6 +427,7 @@ class DatabaseAccess(object):
         """
         if not self._viewer_mode:
             try:
+                par_dict = self._check_chem_formula_length(par_dict)
                 par_dict = dict(
                     (key.lower(), value) for key, value in par_dict.items()
                 )  # make keys lowercase
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -39,6 +39,35 @@ __status__ = "production"
 __date__ = "Sep 1, 2017"
 
 
+def _truncate_chem_formula(chem_formula):
+    """
+    this functions truncate the chemical formula to the last element,
+    up to the limit of the database.
+    arg:
+    chem_form(str): the chemical formula
+    """
+    acceptable_len=30
+    if len(chem_formula) >= acceptable_len:
+        for i in range(acceptable_len):
+            if not chem_formula[acceptable_len-i].isdigit():
+                return chem_formula[:acceptable_len-i+1]
+        return chem_formula[:acceptable_len]
+    else:
+        return chem_formula
+
+
+def _check_chem_formula_length( par_dict):
+    """
+    performs a check whether the length of chemical formula exceeds the defined limit
+    args:
+    par_dict(dict): dictionary of the parameter
+    """
+    for key, value in par_dict.items():
+        if key == 'chemicalformula' and not value is None:
+            par_dict[key] = _truncate_chem_formula(value)
+    return par_dict
+
+
 class AutorestoredConnection:
     def __init__(self, engine):
         self.engine = engine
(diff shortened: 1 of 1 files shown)


pyiron_base/database/generic.py
```diff
@@ -372,6 +372,33 @@ def get_items_sql(self, where_condition=None, sql_statement=None):
...
         return output_list
 
+    def _truncate_chem_formula(self, chem_formula):
+        """
+        this functions truncate the chemical formula to the last element,
+        up to the limit of the database.
+        arg:
+        chem_form(str): the chemical formula
+        """
+        acceptable_len=30
```
pmrv: Can we pull that out and use the value from the database column definition or is that defined somewhere else?  Leave it here if it takes more than 5 minutes.

pyiron_base/database/generic.py
```diff
@@ -372,6 +372,33 @@ def get_items_sql(self, where_condition=None, sql_statement=None):
...
+        if len(chem_formula) >= acceptable_len:
+            for i in range(acceptable_len):
+                if not chem_formula[acceptable_len-i].isdigit():
+                    return chem_formula[:acceptable_len-i+1]
```
pmrv: This doesn't quite work, e.g. 

```python
_truncate_chem_formula("Fe1000B3000X333333G33333444444000000") == 'Fe1000B3000X333333G'
```
but should be (from my understanding) 
```python
'Fe1000B3000X333333'
```

A hack-ish version that works, but needs `dropwhile` from `itertools`
```suggestion
            if chem_formula[acceptable_len].isalpha() and chem_formula[acceptable_len-1].isdigit():
                return chem_formula[:acceptable_len] # acceptable length happens to be just between two elements
            else:
                return "".join(dropwhile(str.isalpha, dropwhile(str.isdigit, chem_formula[acceptable_len::-1])))[::-1]
```

(check for typos :S)

pmrv (changes requested):

pyiron_base/database/generic.py
liamhuber: Yeah, it's [here](https://github.com/pyiron/pyiron_base/blob/e59f4682ba6228f725c0a8d10c8b2ed93551890e/pyiron_base/database/generic.py#L116), but it's hard-coded inline with the table instantiation. It could be refactored out as a parameter and referenced though...

liamhuber:
@sudarsan-surendralal pushed a patch for this [over here](https://github.com/pyiron/pyiron_atomistics/pull/172/files). What I like about that solution is that it lives right in `pyiron_atomistics`, which is the chunk of the code that *should* be dealing with chemical formulas.

I'll give this a quick review anyhow and let the two of you (@muh-hassani and @sudarsan-surendralal) sort out which solution you want to use and whether to merge thoughts from both.

liamhuber (commented):
I don't have anything to add beyond what Marvin already noted, except this: even truncating the formula in a sensible way is still returning *the wrong formula* and doing it silently! As I said over on Sudarsan's PR, if the formula can't be represented accurately I'm much more comfortable returning a value of `OVERFLOW_ERROR` or similar so it is immediately clear that the data the user is looking for is not (reliably) there.

jan-janssen:
I do not think we should make cluster dependent bug fixes, as the issue is already fixed in https://github.com/pyiron/pyiron_base/pull/290 . If you want to modify the version on our cluster that is fine for me but I am against including this workaround in the master branch.

liamhuber:
> I do not think we should make cluster dependent bug fixes, as the issue is already fixed in #290 . If you want to modify the version on our cluster that is fine for me but I am against including this workaround in the master branch.

@jan-janssen how is this cluster-dependent? The chemical formula table is hard-coded to 30 characters [right here in `pyiron_base`](https://github.com/pyiron/pyiron_base/blob/e59f4682ba6228f725c0a8d10c8b2ed93551890e/pyiron_base/database/generic.py#L116)

jan-janssen:
> > I do not think we should make cluster dependent bug fixes, as the issue is already fixed in #290 . If you want to modify the version on our cluster that is fine for me but I am against including this workaround in the master branch.
> 
> @jan-janssen how is this cluster-dependent? The chemical formula table is hard-coded to 30 characters [right here in `pyiron_base`](https://github.com/pyiron/pyiron_base/blob/e59f4682ba6228f725c0a8d10c8b2ed93551890e/pyiron_base/database/generic.py#L116)

That is fixed in https://github.com/pyiron/pyiron_base/pull/290

liamhuber:
> > > I do not think we should make cluster dependent bug fixes, as the issue is already fixed in #290 . If you want to modify the version on our cluster that is fine for me but I am against including this workaround in the master branch.
> > 
> > 
> > @jan-janssen how is this cluster-dependent? The chemical formula table is hard-coded to 30 characters [right here in `pyiron_base`](https://github.com/pyiron/pyiron_base/blob/e59f4682ba6228f725c0a8d10c8b2ed93551890e/pyiron_base/database/generic.py#L116)
> 
> That is fixed in #290

1) that involves updating the database itself. Due to the interuptions that would cause, we discussed in the meeting today to find an alternative, simple approach.

2) Increasing from 30 to 50 does not solve the root problem: when the value is larger than an arbitrary cutoff you hit a hard wall. For non-critical information like the chemical formula, I'm strongly in favour of allowing violating jobs to proceed and just log the data as an "ERROR".

pyiron_base/database/generic.py
max-hassani: I have added a variable to the class, `_chem_formula_lim_length`, which takes the value of 30 and defines the length of the chamical formula in the db.

pyiron_base/database/generic.py
max-hassani: Now I got Liam's suggestion in the meeting, and I changed the formula to `OVERFLOW_ERROR`.

max-hassani:
> 
> 
> @sudarsan-surendralal pushed a patch for this [over here](https://github.com/pyiron/pyiron_atomistics/pull/172/files). What I like about that solution is that it lives right in `pyiron_atomistics`, which is the chunk of the code that _should_ be dealing with chemical formulas.
> 
> I'll give this a quick review anyhow and let the two of you (@muh-hassani and @sudarsan-surendralal) sort out which solution you want to use and whether to merge thoughts from both.

I also used the `OVERFLOW_ERROR` then the user knows that something is wrong with the job table. Otherwise, the chemical formula would be misleading.

pyiron_base/database/generic.py
```diff
@@ -142,7 +114,7 @@ def __init__(self, connection_string, table_name):
             Column("project", String(255)),
             Column("job", String(50)),
             Column("subjob", String(255)),
-            Column("chemicalformula", String(30)),
+            Column("chemicalformula", String(self._chem_formula_lim_lenght)),
```
liamhuber: typo: length

liamhuber (approved):
👍

pyiron_base/database/generic.py
```diff
@@ -372,6 +373,19 @@ def get_items_sql(self, where_condition=None, sql_statement=None):
...
         return output_list
 
+    def _check_chem_formula_length(self, par_dict):
+        """
+        performs a check whether the length of chemical formula exceeds the defined limit
+        args:
+        par_dict(dict): dictionary of the parameter
+        limit(int): the limit for the length of checmical formular
+        """
+        for key, value in par_dict.items():
```
jan-janssen, line 383: This is very inefficient, it would be better to just check if the chemicalformular key exists.

PR #296, opened 2021-04-23
Accelerate database check

jan-janssen:


max-hassani (approved):

liamhuber (approved):
👍 
I like this PR. The next one in the chain should extend `key_limited` to a dictionary with (key, length) and we can put safety valves on some other columns too. I guess then some care will need to be taken if the key is critical, like maybe the job name or the path? That is all for another day, but I think this code sets us up nicely for that anyhow.

PR #303, opened 2021-05-01
weekly dependency updates

max-hassani:
During the last pyiron meeting, it was suggested that it would be better to schedule the dependency updates weekly rather than daily. If this pull request is approved, I will do the same for our other repositories.


jan-janssen (approved):
That is fine with me

PR #305, opened 2021-05-05
Force PIL to look up registered extension on import

niklassiemer:
This prevents import order dependent behavior as recognized in https://github.com/pyiron/pyiron_gui/issues/6.

### DIFF
--- a/pyiron_base/generic/filedata.py
+++ b/pyiron_base/generic/filedata.py
@@ -5,8 +5,9 @@
 
 import json
 import os
-import pandas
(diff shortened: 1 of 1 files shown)


pmrv:
It seems that `registered_extension` forces some internal initialization of PIL, [here](https://pillow.readthedocs.io/en/stable/_modules/PIL/Image.html#registered_extensions).  I can't find another way to trigger (or re-trigger) it, so I guess it's ok for now.

pmrv (approved):

PR #306, opened 2021-05-05
Add ipynb file support

niklassiemer:


niklassiemer:
I'll probably drop the nice `self._repr_html_` feature again, since I do not like a class definition in try/except.

pyiron_base/generic/filedata.py
```diff
@@ -26,28 +27,53 @@
...
+else:
     import_alarm = ImportAlarm(
-        "Reduced functionality, since " + 'PIL' + " could not be imported."
+        "Reduced functionality, since " +
+        str([package for package in _has_imported.keys() if not _has_imported[package]]) +
+        " could not be imported."
     )
 
 
+try:
```
pmrv: Why not 
```suggestion
if _has_imported['nbformat']:
```

pyiron_base/generic/filedata.py
```diff
@@ -83,17 +109,21 @@ def _load_default(file):
         filetype = '.' + filetype
 
     if filetype.lower() in ['.h5', '.hdf']:
-        return FileHDFio(file_name=filename)
+        if project is None:
```
pmrv, line 116: Does it make sense to keep this branch, when `load_file` is always called with a valid `project`?

pmrv (approved):
Is there a case where `load_file` is not called from `Project`?  If not I'd make `project` a normal argument and always return `ProjectHDFio`.  It still feels a bit weird to have to pass a project just for one special file type, but for it's good.

pyiron_base/generic/filedata.py
niklassiemer: Good point! I am happy to force everyone who wants to use this function directly to provide an actual `Project`.

niklassiemer: Well, I just recognized the `FileData` class directly below calls this without a `Project` 😅. And I actually like to have an `ProjectHDFio` if possible, since otherwise a `__getitem__` of a path super to the root from this object would provide a `pyiron_base.Project`. On the other hand, I doubt that this function will be called on `hdf` files frequently since all 'our' hdf files are handled natively and excluded from `list_nodes()`...

PR #307, opened 2021-05-06
Allow to recursivly create new project paths with absolute paths

### LINKED ISSUE #304: Project directory creation relative vs absolute path inconsistency
When creating a project using an absolute path directories are only created if all but the last directory already exist, while directories are created recursively when using a relative path.
Example:
path /home/pyiron/projects/one exists
jupyter lab running in /home/pyiron/projects
pr = Project("/home/pyiron/projects/one/two") will create directory two
pr = Project("/home/pyiron/projects/one/two/three") will not create the directories two and three
pr = Project("one/two/three") will create directories two and three.

Priority probably very low, but can be annoying when trying to run jobs in different directories results in a file not found error.

pmrv:
Fixes #304.

### DIFF
--- a/pyiron_base/project/path.py
+++ b/pyiron_base/project/path.py
@@ -368,9 +368,7 @@ class ProjectPath(GenericPath):
                 )
                 self._create_path(path_local, path)
                 path = posixpath.join(path_local, path)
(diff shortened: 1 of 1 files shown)


Leimeroth:
Works and tests passed, so I will merge it.

PR #308, opened 2021-05-06
Cache access to ProjectHDFio

pmrv:
Sketch of cached HDF5 access.  In case of loading a bunch of structures from an atomistic job with `get_structure()` as [here](https://github.com/pyiron/pyiron_atomistics/pull/189), this affords a 10x speed up.

### DIFF
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -4,6 +4,7 @@ dependencies:
   - coveralls
   - coverage
   - codacy-coverage
+  - cachetools =4.2.2
   - dill =0.3.3
   - future =0.18.2
   - gitpython =3.1.14
--- a/pyiron_base/__init__.py
+++ b/pyiron_base/__init__.py
@@ -8,7 +8,7 @@ from pyiron_base.generic.hdfio import FileHDFio, ProjectHDFio
 from pyiron_base.generic.datacontainer import DataContainer
 from pyiron_base.generic.inputlist import InputList
 from pyiron_base.generic.parameters import GenericParameters
-from pyiron_base.generic.util import deprecate, deprecate_soon, ImportAlarm
+from pyiron_base.generic.util import deprecate, deprecate_soon, ImportAlarm, cached
 from pyiron_base.job.executable import Executable
 from pyiron_base.job.external import Notebook
 from pyiron_base.job.generic import GenericJob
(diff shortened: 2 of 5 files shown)


pmrv:
I replaced the standard library cache with an LFU cache from [cachetools](https://github.com/tkem/cachetools) with a default size limit of 8GiB.

.ci_support/environment.yml
```diff
@@ -4,6 +4,7 @@ dependencies:
   - coveralls
   - coverage
   - codacy-coverage
+  - cachetools =4.2.2
```
liamhuber, line 7: Seeing the 100% coverage badge on this library
please

liamhuber (commented):
Very cool. A few questions:

(1) Is the 8GB shared across a single universal cache, or is it 8GB/cached function? The latter seems excessive.

(2) What is the behaviour when the cache is smaller than a single instance of the returned value?

(3) Can ~~we~~EDIT: you(😉) get the demo implemented as a simple unit test? Also testing (2) and verifying (at least implicitly through timing stuff) that the cache clearing is working as expected?

pmrv:
> Very cool. A few questions:
> 
> (1) Is the 8GB shared across a single universal cache, or is it 8GB/cached function? The latter seems excessive.

Per function, I can change it.  I just want to make sure that even large structures lists etc. get cached.  But yeah, it's on the larger side.

> (2) What is the behaviour when the cache is smaller than a single instance of the returned value?

It just won't end up in the cache.

> (3) Can ~we~EDIT: you(wink) get the demo implemented as a simple unit test? Also testing (2) and verifying (at least implicitly through timing stuff) that the cache clearing is working as expected?

Sure, will get that in.

jan-janssen:
I am not exactly sure if this works for our setup. If I have a job running in non modal mode and I have a second process accessing the the job, how does the cache recognize the file changed? I guess that can be tested with a simple function which reads from a file, a separate process is writing to, to see if the cache is recognizing this.

pmrv:
If a second job updates/adds an HDF entry in the file that does not reside in the cache, we're good.

If a second job updates an HDF entry in the file, while it still resided in the cache, that would indeed result in incoherence.  I guess this would be a problem for interactive jobs, if another job is processing them while they're still running.

I could invalidate the cache if the 'last-modified` file stamp changes between function calls, but that's maybe a bit round about.  Alternatively we optimize the method to be faster.  Something definitely has to change, because a factor 10 slowdown because our HDF layer is weird is not acceptable.

PR #309, opened 2021-05-10
Change lookup order in JobCore.__getitem__

pmrv:
This checks the job HDF5 first and only then looks in HDF5 files of
child jobs.  This avoids database lookups (for the child jobs) in the
common case of accessing a jobs HDF5 storage.

As discussed [here](https://github.com/pyiron/pyiron_atomistics/pull/189) this speeds up HDF5 access quite significantly.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -857,18 +857,28 @@ class JobCore:
 
     def __getitem__(self, item):
         """
-        Get/read data from the HDF5 file or access log files.
+        Get/read data from the HDF5 file, child jobs or access log files.
 
         If the job is :method:`~.decompress`ed, item can also be a file name to
         access the raw output file of that name of the job.  See available file
         with :method:`~.list_files()`.
 
+        `item` is first looked up in this jobs HDF5 file, then in the HDF5 files of any child jobs and finally it is
+        matched against any files in the job directory as described above.
+
         Args:
             item (str, slice): path to the data or key of the data object
 
         Returns:
-            dict, list, float, int: data or data object
+            dict, list, float, int, None: data or data object; if nothing is found None is returned
         """
+        try:
+            hdf5_item = self._hdf5[item]
+        except ValueError:
+            hdf5_item = None
+        if hdf5_item is not None:
+            return hdf5_item
+
         name_lst = item.split("/")
         item_obj = name_lst[0]
         if item_obj in self._list_ext_childs():
@@ -879,13 +889,6 @@ class JobCore:
                 return child
             else:
                 return child["/".join(name_lst[1:])]
-
-        try:
-            hdf5_item = self._hdf5[item]
-        except ValueError:
-            hdf5_item = None
-        if hdf5_item is not None:
-            return hdf5_item
 
         if name_lst[0] in self.list_files():
             file_name = posixpath.join(self.working_directory, "{}".format(item_obj))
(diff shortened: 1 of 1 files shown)


pyiron_base/job/core.py
```diff
@@ -857,18 +857,28 @@ def __del__(self):
...
+        matched against any files in the job directory as described above.
+
         Args:
             item (str, slice): path to the data or key of the data object
 
         Returns:
-            dict, list, float, int: data or data object
+            dict, list, float, int, None: data or data object; if nothing is found None is returned
         """
+        try:
```
liamhuber, line 875: Simplicity nit: It looks to me like these six lines are equivalent to the following four:

```python
try:
    return self._hdf5[item]
except ValueError:  # Item not found
    pass
```

I have mixed feelings about tucking return values away inside clauses and try/exceptions, but I definitely don't veto it and it's consistent with the rest of the method.

ahaaaa, wait. The current implementation stops the case that the item *does* exist but is literally `None`. However, isn't this a mistake? `None` != "not data". If we try to load something and it's there and `None`, isn't this the value we *want* to return? Isn't digging into children then odd behaviour?

liamhuber (commented):
No objection to your changes, but I'm a little confused why the method works the way it does to begin with.

pyiron_base/job/core.py
pmrv, line 875: I agree, I don't like returning `None` for non-existent keys, but I also don't want to refactor this, since presumably a lot of other code depends on this.  So it is what it is.  I can put your suggestion in though, since it is cleaner.

pmrv: I agree, I don't like returning None for non-existent keys, but I also don't want to refactor this, since presumably a lot of other code depends on this. So it is what it is. I can put your suggestion in though, since it is cleaner.

liamhuber: No, that's not quite my concern. I'm really not confident I understand the functionality here, so let me try to explain in full detail and let me know if I'm just being an idiot.

Suppose I had a job, `'myJob'`, that had an HDF field `'foo'` *and* a child job with the name `'foo'`. Then my understanding is that our project has  something like `'myJob_hdf5/foo'` *and* `'myJob_hdf5/foo_hdf5'`

case 1) `'myJob_hdf5/foo'` has value `None`
- Orignal code:  Children get searched first and `job['foo']` returns the child job object
- Your code: We find the `None` value in `'myJob_hdf5/foo'` but explicitly ignore it with the if clause, go on to find and return the child job.
- My suggestion: `None` is found and returned.

case 2) `'myJob_hdf5/foo'` has value `'anything_but_none'`
- Original code: Children get searched first and `job['foo']` returns the child job object
- Your code: Finds and returns `'anything_but_none'`
- My suggestion: Finds and returns `'anything_but_none'`

If nodes and groups can be guaranteed not to have the same name, then this situation never comes up, and original=your=my, and I like mine best for its brevity.

Otherwise, if we're going to let this silly situation pass silently at all, then I *like* my version best -- return node values first and if it's `None` then dammit that's a valid value. But this is really a behaviour change and we shouldn't push it without wider review.

pmrv: Ok, I understand now.  Your version is nice I think, but it changes behavior a bit more, so discussing on Friday is fine.

liamhuber: Next Friday 😉

jan-janssen (approved):
Looks good to me

PR #311, opened 2021-05-11
Document gotcha in analysis_project

pmrv:
Just a docstring touch up

### DIFF
--- a/pyiron_base/table/datamining.py
+++ b/pyiron_base/table/datamining.py
@@ -571,6 +571,11 @@ class TableJob(GenericJob):
 
     @property
     def analysis_project(self):
+        """
+        :class:`.Project`: which pyiron project should be searched for jobs
+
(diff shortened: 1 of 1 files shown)


niklassiemer:
I am not really used to work with pyiron tables so far. However, the setter uses `system_function_lst=self._system_function_lst` which would retain the analysis functions, wouldn't it?

pmrv:
Honestly this is from February and I just found the commit... I'll have a look at it again tomorrow, but I thought this tripped me up before.

pmrv:
I've checked just now that this is still an issue.  I agree that the code appears to catch this, but I don't have time right now to debug this.  I would merge this and add an "unsolved" tag, so this can be fixed later, if no one complains loudly.

pmrv:
[Nevermind.](https://github.com/pyiron/pyiron_base/pull/198)

PR #312, opened 2021-05-11
Use dill to save filter functions

pmrv:
I haven't tested this yet, so I'll leave it as a draft until I get around to it or someone urgently needs it.

### DIFF
--- a/pyiron_base/table/datamining.py
+++ b/pyiron_base/table/datamining.py
@@ -651,7 +651,7 @@ class TableJob(GenericJob):
             group_name:
         """
         super(TableJob, self).from_hdf(hdf=hdf, group_name=group_name)
-        hdf_version = hdf.get("HDF_VERSION", "0.1.0")
+        hdf_version = self.project_hdf5.get("HDF_VERSION", "0.1.0")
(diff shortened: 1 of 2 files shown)


pmrv:
@niklassiemer I'm not sure how familiar you are with the table internals, but only me and @jan-janssen ever touch this file, so I thought I add you. ;)

niklassiemer:
I never touched/had a look at them at all 😂 However, I will try my best to review this 😊

niklassiemer (commented):
So you now always use dill to store the 'whole' function to get the data? I assume the function object will actually not be a lot larger than the string to generate that - from that perspective I think this is a good idea. How about the transferability? Is a `dill`ed object stable against moving to another machine/ version update? 
Other than that, I am not really sure what your update does. However, I did not see any big issues either :)

pmrv:
I haven't actually looked at the size difference.  The main point is that storing functions as their "string" simply does not work generally and especially not for lambdas.  Dill at least promises that the format is stable.  I'm not sure that if they update the format it is in a backwards-compatible manner, but you can ask (via their API) for an older format to be read as well.  So we could think about specifying which version of the dill format we want, but we don't right now for the other functions, so I left it.

PR #314, opened 2021-05-11
Always add HDF5 file extension to path

pmrv:
#235 introduced a bug where previously project data is not read in again, because the filename was missing the correct extension.  This was due to the inconsistent interface between `ProjectHDFio` and `FileHDFio` so I fixed that.

I'm not sure why the tests didn't catch that, but we should update them before merging this.

EDIT: Actually there seem to be more problems. `FileHDFio` doesn't have the `to_object` method so actually reading data from `ProjectData` also fails, but I don't have time to investigate right now.

I pushed a hotfix [here](https://github.com/pyiron/pyiron_base/tree/project_data_hotfix).  I don't think reverting is necessary, but should work on a fix here, but I needed a working version right now.

### DIFF
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -111,8 +111,7 @@ class FileHDFio(object):
     """
 
     def __init__(self, file_name, h5_path="/", mode="a"):
-        if not file_name.endswith(".h5"):
-            file_name += ".h5"
+        file_name += ".h5" if not file_name.endswith(".h5") else ""
         if not os.path.isabs(file_name):
             raise ValueError("file_name must be given as absolute path name")
         self._file_name = None
@@ -1047,8 +1046,7 @@ class ProjectHDFio(FileHDFio):
     """
 
     def __init__(self, project, file_name, h5_path=None, mode=None):
-        if not file_name.endswith(".h5"):
-            file_name += ".h5"
+        file_name += ".h5" if not file_name.endswith(".h5") else ""
         self._file_name = file_name.replace("\\", "/")
         if h5_path is None:
             h5_path = "/"
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -111,6 +111,8 @@ class FileHDFio(object):
     """
 
     def __init__(self, file_name, h5_path="/", mode="a"):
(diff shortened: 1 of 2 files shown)


liamhuber:
I guess the broken test is really explicitly in conflict with this PR; the test tries to write to both `file` and `file.h5`, but now `file` is automatically getting converted to `file.h5` so when the test later tries to remove `file` it isn't there.

I don't understand the rationale of why the test wants to do both these things. @niklassiemer I checked the blame and you touched it last 😉 I'll let you decide whether it's the test or code that needs to change.

niklassiemer (commented):
I like it, also to switch to `endswith` is way saver. I made a PR in here to fix the test which used the old non-h5 extension behavior.

pyiron_base/generic/hdfio.py
```diff
@@ -111,6 +111,8 @@ class FileHDFio(object):
     """
 
     def __init__(self, file_name, h5_path="/", mode="a"):
+        if not file_name.endswith(".h5"):
```
liamhuber: How do you feel about this? I'm not actually 100% sure it's any clearer, but it has the (small) advantage of being 50% shorter in line-space 😉 

```python
file_name += ".h5" if not file_name.endswith(".h5") else ""
```

liamhuber (approved):
lgtm

pyiron_base/generic/hdfio.py
pmrv: No strong feelings, I can change it.

pmrv: Purely aesthetically I like the current version more, but I don't mind changing it.

pmrv:
> EDIT: Actually there seem to be more problems. `FileHDFio` doesn't have the `to_object` method so actually reading data from `ProjectData` also fails, but I don't have time to investigate right now.

This issue persists, but I'm not sure how to fix it. @niklassiemer why did we change to `FileHDFio`?

niklassiemer:
Due to the recursive project instantiation project > project data > ProjectHDFio > Project which prohibits to load the project data on project init. However, I was not aware of the missing to_object. Therefore, we probably need to revert the old PR to get this functionality back.

pmrv:
Even with #323 around I'd say we keep this, since it's still an API defect in my book.

PR #315, opened 2021-05-11
Copy file.h5 to file instead of creating file via FileHDFio

niklassiemer:


pmrv:
I don't completely understand the test case, but if this fixes it, I'm good.

pmrv (approved):

PR #320, opened 2021-05-17
Add hook after copy_to on Jobs

pmrv:
Looking for ways to procrastinate I see [this](https://github.com/pyiron/pyiron_atomistics/pull/196), so I added a hook that runs after `copy_to` is called on jobs to handle any special needs of sub classes without them having to worry about all the arguments `copy_to` provides.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -762,10 +762,9 @@ class JobCore:
 
         # Remove output if it should not be copied
         if input_only:
-            if "output" in new_job_core.project_hdf5.list_groups():
-                del new_job_core.project_hdf5[
-                    posixpath.join(new_job_core.project_hdf5.h5_path, "output")
-                ]
+            for group in new_job_core.project_hdf5.list_groups():
+                if "output" in group:
+                    del new_job_core.project_hdf5[posixpath.join(new_job_core.project_hdf5.h5_path, group)]
             new_job_core.status.initialized = True
         return new_job_core
 
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -28,6 +28,7 @@ from pyiron_base.job.util import \
     _job_remove_folder
 from tables import NoSuchNodeError
 import shutil
+import warnings
 
 __author__ = "Jan Janssen"
 __copyright__ = (
@@ -751,6 +752,7 @@ class JobCore:
         """
         # Update flags
         if input_only and new_database_entry:
+            warnings.warn("input_only conflicts new_database_entry; setting new_database_entry=False")
             new_database_entry = False
 
         new_job_core, _, _, reloaded = self._internal_copy_to(
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -753,13 +753,13 @@ class JobCore:
         if input_only and new_database_entry:
             new_database_entry = False
 
-        new_job_core, _, _, reload = self._internal_copy_to(
+        new_job_core, _, _, reloaded = self._internal_copy_to(
             project=project,
             new_job_name=new_job_name,
             new_database_entry=new_database_entry,
             copy_files=copy_files
         )
-        if reload_flag:
+        if reloaded:
             return new_job_core
 
         # Remove output if it should not be copied
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -739,18 +739,34 @@ class JobCore:
 
         Args:
             project (JobCore/ProjectHDFio/Project): project to copy the job to
-            new_database_entry (bool): [True/False] to create a new database entry - default True
+            input_only (bool): [True/False] Whether to copy only the input. (Default is False.)
+            new_database_entry (bool): [True/False] Whether to create a new database entry. If input_only is True then
+                new_database_entry is False. (Default is True.)
             copy_files (bool): [True/False] copy the files inside the working directory - default True
 
         Returns:
             JobCore: JobCore object pointing to the new location.
         """
+        # Update flags
+        if input_only and new_database_entry:
+            new_database_entry = False
+
         new_job_core, _, _, reload = self._internal_copy_to(
             project=project,
             new_job_name=new_job_name,
             new_database_entry=new_database_entry,
             copy_files=copy_files
         )
+        if reload_flag:
+            return new_job_core
+
+        # Remove output if it should not be copied
+        if input_only:
+            if "output" in new_job_core.project_hdf5.list_groups():
+                del new_job_core.project_hdf5[
+                    posixpath.join(new_job_core.project_hdf5.h5_path, "output")
+                ]
+            new_job_core.status.initialized = True
         new_job_core._after_copy_to(reload=reload)
         return new_job_core
 
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -739,6 +739,8 @@ class JobCore:
 
         Args:
             project (JobCore/ProjectHDFio/Project): project to copy the job to
+            new_job_name (str): The new name to assign the duplicate job. Required if the project is `None` or the same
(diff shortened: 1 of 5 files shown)


pyiron_base/job/core.py
```diff
@@ -739,18 +739,34 @@ def copy_to(self, project, new_job_name=None, new_database_entry=True, copy_file
...
             new_job_name=new_job_name,
             new_database_entry=new_database_entry,
             copy_files=copy_files
         )
+        if reload_flag:
+            return new_job_core
+
+        # Remove output if it should not be copied
+        if input_only:
+            if "output" in new_job_core.project_hdf5.list_groups():
```
liamhuber: *sigh* we really, really, really need to get v1.0 out and fix our janky storage arrangement.

So since we never updated all the code to stop using `GenericInput` and `GenericOutput`, when I make new classes that use `DataContainer` for IO, I have to give them paths *other* than `"input"` and `"output"`. 

I guess here we can make a simple fix like

```python
for group in new_job_core.project_hdf5.list_groups():
    if "output" in groups:
        blah blah
```

That will cover us for now, but I am really longing for a more consistent data management scheme!

liamhuber:
I really like this idea. Will give a more detailed review when you're satisfied with it, but I'll poke my nose in and add single comments until then.

raynol-dsouza (approved):

pyiron_base/job/core.py
pmrv: I don't quite understand this.  This is catch all groups in HDF that have "output" somewhere in the name?

liamhuber: Yeah, that's my concern; that there are "output_foo" fields that should be excluded along with "output".

pmrv: Ah, that is new functionality then, but I can add it, sure.

pyiron_base/job/core.py
```diff
@@ -739,18 +739,33 @@ def copy_to(self, project, new_job_name=None, new_database_entry=True, copy_file
 
         Args:
             project (JobCore/ProjectHDFio/Project): project to copy the job to
-            new_database_entry (bool): [True/False] to create a new database entry - default True
+            input_only (bool): [True/False] Whether to copy only the input. (Default is False.)
+            new_database_entry (bool): [True/False] Whether to create a new database entry. If input_only is True then
```
liamhuber, line 746: I think I would rather raise an error if these are conflicting. I'm open to push-back on that though.

pyiron_base/job/core.py
```diff
@@ -739,18 +739,33 @@ def copy_to(self, project, new_job_name=None, new_database_entry=True, copy_file
...
+        if input_only and new_database_entry:
+            new_database_entry = False
+
+        new_job_core, _, _, reload = self._internal_copy_to(
             project=project,
             new_job_name=new_job_name,
             new_database_entry=new_database_entry,
             copy_files=copy_files
         )
+        if reload_flag:
```
liamhuber: typo? Should be `reload`?

pyiron_base/job/core.py
```diff
@@ -739,18 +739,33 @@ def copy_to(self, project, new_job_name=None, new_database_entry=True, copy_file
...
+            return new_job_core
+
+        # Remove output if it should not be copied
+        if input_only:
+            for group in new_job_core.project_hdf5.list_groups():
+                if "output" in group:
+                    del new_job_core.project_hdf5[posixpath.join(new_job_core.project_hdf5.h5_path, group)]
+            new_job_core.status.initialized = True
```
liamhuber: @raynol-dsouza will this annoy us? @pmrv isn't the job still `created` and not `initialized`? I am not confident because I still don't grok the status flow, I just want to throw a yellow flag.

raynol-dsouza: @liamhuber I think it shouldn't cause any issues, since the existing version also set the status to `initialized`.

pyiron_base/job/core.py
pmrv: Since this is just shuffling existing behavior I'd prefer a warning.  As far as grep can tell `input_only=True` only in protocols (protocol.primitive.one_state:254) though, so maybe we do a hard switch in a separate PR and fix it in contrib before the next version bump.

pyiron_base/job/core.py
pmrv: This shouldn't change functionality, as this is the same as before it just lives on `JobCore` now instead of `GenericJob`.  I don't get the status hierarchy as well, but it's been on my backlog for some archeology.

pmrv: I looked it up and `initialized` comes before `created`, so I suppose we should be good.

liamhuber: > I think it shouldn't cause any issues, since the existing version also set the status to initialized.

And I thought the existing behaviour was a pain in the butt for us 😉 

Anyhow, maintaining the existing behaviour is totally fine for now.

> but it's been on my backlog for some archeology.

I like this phrasing 😄

pyiron_base/job/core.py
liamhuber: Sure thing, a warning is completely sufficient for now. Actually probably much safer than suddenly making it an error, as per second point:

Yeah, I wasn't aware of heavy usage in the code base, but since it's public I guess we also need to think about it being used in notebooks we know nothing about. (And thus for a transition period a warning is important!)

pyiron_base/job/core.py
raynol-dsouza: > And I thought the existing behaviour was a pain in the butt for us 😉

@liamhuber The thing that was problematic for us, was that the reference job itself needed to be initialized, not it's copy. Creating a job and saving it without running, sets the status to `created`, and not `initialized`. The funny part is that while making a copy of the reference job, the existing code sets the copy's status to `initialized`, while the reference job still remains `created`. 😄 I do not know if this needs fixing.

pyiron_base/job/core.py
```diff
@@ -733,24 +734,42 @@ def _internal_copy_to(self, project=None, new_job_name=None, new_database_entry=
...
         )
+        if reloaded:
+            return new_job_core
+
+        # Remove output if it should not be copied
+        if input_only:
+            for group in new_job_core.project_hdf5.list_groups():
+                if "output" in group:
+                    del new_job_core.project_hdf5[posixpath.join(new_job_core.project_hdf5.h5_path, group)]
+            new_job_core.status.initialized = True
```
jan-janssen: The job status of a job core object is just a string, the job status object is only introduced for the generic jobs.

jan-janssen: The job status of a job core object is just a string, the job status object is only introduced for the generic jobs.

pyiron_base/job/generic.py
```diff
@@ -524,30 +526,27 @@ def copy_to(self, project=None, new_job_name=None, input_only=False, new_databas
...
-        if input_only:
-            if "output" in new_job_core.project_hdf5.list_groups():
-                del new_job_core.project_hdf5[
-                    posixpath.join(new_job_core.project_hdf5.h5_path, "output")
-                ]
-            new_job_core.status.initialized = True
+        new_job_core._copy_to(new_database_entry=new_database_entry, reloaded=reloaded)
         return new_job_core
 
+    def _copy_to(self, new_database_entry, reloaded):
```
jan-janssen: Can you rename this function to be a bit more specific? Class specific copy to or anything similar would be fine.

pmrv: It's called `_specific_copy_to` now.

liamhuber: Honestly I think I liked `_after_copy_to` most so far since it was even more descriptive. What about combining these two ideas for maximum descriptivity: `_after_generic_copy_to`?

liamhuber (approved):

pmrv:
Took a few goes at it (because I'm a nit), but most of the changes should be covered now.  I'll merge when I'm home.

PR #322, opened 2021-05-18
Add workflow to update the environment.yml in PRs from dependabot

niklassiemer:
This workflow uses the title of the dependabot PR `Bump package_name from some_version to updated_version` to replace the `some_version` of `package_name` by `updated_version` in our `environment.yml` within the dependabot PR. 
To get this working we need other credentials than the usual `secrets.GITHUB_TOKEN` since if this one is used in a workflow no subsequent workflow run (our CI) is triggered (see [here](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#using-the-github_token-in-a-workflow)). 
Therefore, @muh-hassani do we have another token which might be used? Otherwise we would need another token (I called it `WORFKLOW_TOKEN` in the action for now) with write access to the repository.

### DIFF
--- a/.github/workflows/UpdateDependabotPR.yml
+++ b/.github/workflows/UpdateDependabotPR.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.ref }} # Check out the head of the actual branch, not the PR
           fetch-depth: 0 # otherwise, you will fail to push refs to dest repo
-          token: ${{ secrets.WORFKLOW_TOKEN }}
+          token: ${{ secrets.DEPENDABOT_WORKFLOW_TOKEN }}
       - name: UpdateEnvironmentFile
         shell: bash -l {0}
         run: |
@@ -29,5 +29,5 @@ jobs:
       - name: UpdateDependabotPR push
         uses: ad-m/github-push-action@master
         with:
-          github_token: ${{ secrets.WORFKLOW_TOKEN }}
+          github_token: ${{ secrets.DEPENDABOT_WORKFLOW_TOKEN }}
           branch: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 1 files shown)


niklassiemer:
I tested this at my fork of `pyiron_base` in this dependabot [PR](https://github.com/niklassiemer/pyiron_base/pull/4) using the `main` branch as default target (I did not want to destroy my master). I made a Personal Access Token from my user account and stored it in the repository secrets (in this case `MY_WORKFLOW_TOKEN`) and this way it is triggering the CI run for the added commit.

max-hassani:
@niklassiemer 
I added a token as secret in the repository, called: `DEPENDABOT_WORKFLOW_TOKEN`.
I thought a meaningful name can help us understand later on what is the purpose of the token.
Shall we test it only for pyiron_base and then add similar workflow to the other repo, or you prefer to do it for all repos now?

niklassiemer:
Thanks a lot! I updated the Token in the PR

max-hassani (approved):
Looks good to me! thanks @niklassiemer for this nice addition to our dependency updates.

pmrv:
I'm a bit confused what this does.  Is that because of dependabot now being part of github?  If so shouldn't we kick out the config too then?  This also only update the environment.yml and not the setup.py, right?

niklassiemer:
We after all rely on the dependabot (So we would not change any settings for this). However, right now we have to change our `environment.yml` always 'by hand', i.e. we have to make the same change to it as dependabot did for the `setup.py`. This workflow _only_ acts on PR opened by dependabot and adds a commit to such PR that changes the `environmen.yml` automatically (at least if the package name is the same). It also should not affect the behavior of dependabot on its PR (rebasing by dependabot etc. works after all).

pmrv:
Ah, I missed that.  I just checked the changed files for one of the dependabot PRs and didn't see that the changed environment.yml came from @muh-hassani My bad, lgtm in that case.

pmrv (approved):

niklassiemer:
Ok, then let's test this here first and if everything works as expected we push it to the other repos as well 😃

PR #323, opened 2021-05-19
Project data hotfix

pmrv:
Following [this](https://github.com/pyiron/pyiron_base/pull/314) discussion reverts the change to `FileHDFio`, delays the instantiation of `ProjectData` to break the recursion on init and expands to the test to catch this.


niklassiemer (approved):
lgtm

liamhuber (approved):
In the face of the recursion problems, this is a totally acceptable compromise 👍

PR #324, opened 2021-05-19
Remove obsolete ProjectStore

pmrv:
I saw that we actually have something like `ProjectData` already previously under `Project._store`.  However there's only write access via `Project.__setitem__`, no read access and I haven't found any references to it in other code, so I've taken the liberty to delete it.


niklassiemer:
Seems to be a good idea to remove it. We could replace (extend) `Project.__setitem__` (`Project.__getitem__`) to write (read) to (from) the `Project.data`.

pmrv:
I'd prefer have the `[...]` syntax be exclusively for project navigation and keep `project.data` as it is.  It's not too much typing and we'll run into all sort of issues like what happens when there's a sub project and a data item with the same name.  @liamhuber and I had a similar discussion [here](https://github.com/pyiron/pyiron_base/pull/309).

liamhuber (approved):
❤️ a PR that's just +0/-207. Not being sarcastic, it is literally a beautiful thing.

liamhuber:
> I'd prefer have the `[...]` syntax be exclusively for project navigation and keep `project.data` as it is. 

Yeah, agreed. In this case I feel the safety that comes with being explicit about it is preferable to the convenience of the shortcut.

PR #325, opened 2021-05-22
Move create pipeline

### LINKED ISSUE #326: pipeline cannot be created
`GenericJob` has the method `create_pipeline` which internally binds to `self.project.create_pipeline`. `Project` has no such method.

This method exists over in `pyiron_atomistics.Project`, which in turn binds to the `pipe` method of `pyiron_atomistics.atomistics.parallel.pipe`. Both of these make explicit reference to structures or atomistics classes.

Proposed simple fix: move `create_pipeline` off of `pyiron_base.GenericJob` and onto `pyiron_atomistics.AtomisticGenericJob`. Implemented [here](https://github.com/pyiron/pyiron_atomistics/pull/206) and [here](https://github.com/pyiron/pyiron_base/pull/325).

I can't make a MWE example because of the [issue](https://github.com/pyiron/pyiron_base/issues/327) that came up with instantiating a `GenericJob`

liamhuber:
It references `self.project.create_pipeline`, but `pyiron_base.Project` has no such method.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1033,18 +1033,6 @@ class GenericJob(JobCore):
         job._init_child_job(self)
         return job
 
-    def create_pipeline(self, step_lst, delete_existing_job=False):
-        """
-        Create a job pipeline
-
-        Args:
-            step_lst (list): List of functions which create calculations
-
-        Returns:
-            FlexibleMaster:
-        """
-        return self.project.create_pipeline(job=self, step_lst=step_lst, delete_existing_job=delete_existing_job)
-
     def update_master(self):
         """
         After a job is finished it checks whether it is linked to any metajob - meaning the master ID is pointing to


pmrv (approved):
lgtm

I sort of see, that pipelines are something we'd want in base as well, but as long as we don't have it on `ProjectCore` it makes sense to not have it.  Looking over `pipe` it seems the implementation is not too atomistic specific, but I don't see an urgent need to generalize it.  Moving this back once `ProjectCore` can create pipelines is easy enough.

liamhuber:
Yeah, having them on core makes more sense to me too, it was just more of a PR than I wanted to write on a Saturday morning 😝 I also agree that it's probably low priority, so we can clean house now and renovate later

niklassiemer:
What is the status here? Anything holding back this merge?

liamhuber:
Not from my perspective. It should be merged contemporaneously with the appearance of the same code over in atomistics though

niklassiemer:
I would suggest a changed workflow for such moves in general: First update the target repo, i.e. overwrite the method of the source repo there. Afterwards delete the source repo method. This way, at any time, there will be the method available and the stability is increased. 
Therefore, I would like to merge the PR in `pyiron_atomistics` before the next release and shift this PR to be merged after the next release.

PR #329, opened 2021-05-28
Set master ID for child jobs

jan-janssen:


Leimeroth (approved):
Fixes the problem when the master jobs work correctly. 
I am not sure when a child job fails and causes the master job to abort. I will update the issue accordingly.

pmrv (approved):
Is there a case where we also need to update the parent id?

PR #331, opened 2021-05-28
Add build docs workflow like in pyiron_atomistics

niklassiemer:


niklassiemer:
@pmrv We have a similar issue as in [here](https://github.com/pyiron/pyiron_atomistics/pull/208).

niklassiemer:
The docs run also fails for a pervious version of the code base where readthedocs was building... see dummy PR #332

pmrv:
Might be related to [this](https://github.com/spatialaudio/nbsphinx/issues/563). Let's check our jinja+nbsphinx versions on Monday.

niklassiemer:
Restricting to jinja2 <3 seems to fix the issue! Actually, why do we have different `environment` files for building the docs in our repos and for deploying on `readthedocs`? I would suggest to straighten this up and to include everything docs related into the docs folder, i.e. using `docs/environment.yml` in the build docs workflow?

pmrv:
Maybe it's better to fix `nbsphinx >= 0.8.5`, since we don't directly depend on jinja, see [here](https://github.com/spatialaudio/nbsphinx/releases/tag/0.8.5).

pmrv (approved):

niklassiemer:
I just try to fix `nbsphinx` only, like you suggested. I like the `>=` way more than the `<` 😄

PR #332, opened 2021-05-28
Test revert for docs

niklassiemer:
I just wanted to run the Docs test at a point in our history, where the readthedocs was working


niklassiemer:
Ok, obviously I made some error on reverting (probably missed to remove some new test files). However, the Docs run actually terminates with the same error... 
Therefore, building the docs fails due to an update in its environment?

PR #333, opened 2021-05-29
Add Interface for `list_nodes`/`list_groups`

pmrv:
A lot of pyiron objects exist with methods called "list_nodes" and "list_groups".  We generally use them to walk certain hierarchies in a type agnostic way.  I've had this PR kicking around for a while that makes all of those objects derive from a single ABC and "defines" in way what we expect from these methods.

There's not a lot of utility in this alone, but I want this Interface as part of the "storage back end" interface, so I thought I'd break it out and get your comments early.

I'll break some of the docstring out into a spec file next week, too.


pyiron_base/generic/datacontainer.py
```diff
@@ -10,6 +10,7 @@
 import warnings
 import numpy as np
 from .fileio import read, write
+from .has_groups import HasGroups
```
liamhuber: Is there any danger to the ease of future refactoring from using local imports instead of the full `pyiron_base.generic.has_groups` import?

pyiron_base/generic/has_groups.py
```diff
@@ -0,0 +1,93 @@
+"""
```
liamhuber: Just floating the idea: instead of shoving more stuff into `generic`, what about making a new submodule `mixins`/`interfaces`/`mixin_interfaces`/`interface_mixins` and sticking this there?

pyiron_base/generic/has_groups.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def list_groups(self):
+        """
+        Return a list of names of all nested groups.
+
+        Indexing the object with names returned from here, must return instances of :class:`.HasGroups`.
+
+        Returns:
+            list of str: group names
+        """
```
niklassiemer: Do I understand correctly, that the intent is to have this docstring as the only user facing docstring for `list_groups` functions? Since all derived classes just implement the private `_list_groups` method docstrings would not be visible. 
In this case I would prefer a more user friendly docstring and shift this docstring into a developer comment?

pyiron_base/generic/has_groups.py
```diff
@@ -0,0 +1,93 @@
...
+from abc import ABC, abstractmethod
+
+class HasGroups(ABC):
+    """
+    Abstract mixin to capture hierarchical structure of pyiron objects.
+
+    Necessary overrides are :method:`.__getitem__()`, :method:`._list_groups()` and :method:`._list_nodes()`.  Sub
+    classes may override :method:`._list_all()`, but the default implementation will call the other two methods.
+
+    A hierarchical object has a number of children and associates them with string names.  :method:`.__getitem__()`
```
liamhuber: Comment: in the long run I feel like it will be nice to synchronize our master control with this functionality. E.g. `MasterJob` can loop over its `list_groups` and check their type for `Job` and those are it's children, none of this separate `childs` attribute. Maybe there are problems with this that I don't see right now, but I dream of a pyiron with a more unified set of interfaces and control structures.

niklassiemer (commented):
Thanks a lot for taking a shot at unifying the `list_*` interface!

liamhuber (approved):
I like it. A couple suggestions and a comment, but if you don't like them it can be merged as-is IMO.

pyiron_base/generic/datacontainer.py
pmrv: No, full imports are fine.  It's just habit, I've changed it. ;)

pyiron_base/generic/has_groups.py
pmrv: Good idea, I was thinking about removing mentions of children from `GenericJob`, but hadn't come up with anything.  This sounds like a way to go.

pyiron_base/generic/has_groups.py
pmrv: Good point, I'll move this to `_list_groups()` for devs to read.  I'm still not very sure about the user docs, since sub classes might have different ideas about what "groups" and "nodes" are them.

pyiron_base/generic/has_groups.py
pmrv: We should then also think about moving `HasStorage`, etc. there at some point, but that's probably best left to a time when they are actually used.

liamhuber: Absolutely. We also need to think a little bit about those mixins which tamper with `__init__` (and therefore need some special treatment so we don't run afoul of MRO issues and missing an important init) and ones like this that are safer. Can definitely do the move together later.

PR #336, opened 2021-05-31
Add generic database interface -- IsDatabase

pmrv:
Inspired by Friday, I've started worked to create a uniform interface for `DatabaseAccess` and `FileTable` called `IsDatabase`.

For now I've only touched `view_mode` and `job_table`, but the general recipe for this PR is to go through the generic interface

https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/jobtable.py

and then move each function there to an abstract method in `IsDatabase` and split the implementation to methods in `DatabaseAccess` and `FileTable`.  Since the latter is newer, this mostly ends up moving stuff to methods in the former.

I'll be moving functions from time to time when I feel like it, but it's mostly copying code between places, so feel free to pick up some methods yourself, too.  In any case test coverage is light on `DatabaseAccess`, so tread lightly.  I would suggest that once we have a stable interface here, we add tests to cover at least the sqlite code path.

### DIFF
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -147,12 +147,14 @@ class FileTable(IsDatabase, metaclass=Singleton):
         return self._job_table.columns.values
 
     def _get_job_table(
-        self, 
-        project=None, 
-        recursive=True, 
-        columns=None, 
-        sort_by="id", 
-        max_colwidth=200, 
+        self,
+        sql_query,
+        user,
+        project=None,
+        recursive=True,
+        columns=None,
+        sort_by="id",
+        max_colwidth=200,
         job_name_contains=''
     ):
         self.update()
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -164,8 +164,7 @@ class FileTable(IsDatabase, metaclass=Singleton):
             else:
                 df = self._job_table[self._job_table.project == project]
         else:
-            df = self._job_table
-        return df[columns]
+            return self._job_table
 
     def get_jobs(self, project=None, recursive=True, columns=None):
         if project is None:
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -150,7 +150,7 @@ class FileTable(IsDatabase, metaclass=Singleton):
         self,
         sql_query,
         user,
-        project=None,
+        project_path=None,
         recursive=True,
         columns=None,
         sort_by="id",
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -5,13 +5,14 @@ import datetime
 import h5io
 from pyfileindex import PyFileIndex
 from pyiron_base.generic.util import Singleton
+from pyiron_base.database.generic import IsDatabase
 
 
 def filter_function(file_name):
     return '.h5' in file_name
 
 
-class FileTable(metaclass=Singleton):
+class FileTable(IsDatabase, metaclass=Singleton):
     def __init__(self, project):
         self._fileindex = None
         self._job_table = None
@@ -21,9 +22,8 @@ class FileTable(metaclass=Singleton):
                          'username']
         self.force_reset()
 
-    @property
-    def viewer_mode(self):
-        return None
+    def _get_view_mode(self):
+        return False
 
     def force_reset(self):
         self._fileindex = PyFileIndex(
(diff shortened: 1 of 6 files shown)


pmrv:
There's some inconsistency between `DatabaseAccess` and `FileTable` when it comes to pre-filtering the jobs according to a custom sql query and user names that I've noted [here](https://github.com/pyiron/pyiron_base/pull/336/commits/9ddd11b01003428e86360982eefbf599e23040ed).  The problem with `sql_query` is that it is fairly deeply embedded into `Project` and therefore not so easy to move to `DatabaseAccess`.

pyiron_base/database/filetable.py
```diff
@@ -21,9 +22,8 @@ def __init__(self, project):
                          'username']
         self.force_reset()
 
-    @property
-    def viewer_mode(self):
-        return None
+    def _get_view_mode(self):
```
liamhuber, line 25: Getters with no arguments can easily just be properties. Is there a good reason to include this method and not just make `view_mode` `@property @absractmethod`? (There might well be! I just don't see it atm.)

pyiron_base/database/generic.py
```diff
@@ -38,6 +41,121 @@
...
+    """
+
+    @abstractmethod
+    def _get_view_mode(self):
+        pass
+
+    @property
+    def view_mode(self):
+        """
+        Get viewer_mode - if viewer_moded is enable pyiron has read only access to the database.
```
liamhuber: Docstring still uses deprecated method name

pyiron_base/database/filetable.py
pmrv: Imo properties are nice for UI.  Since `IsDatabase` has a property defined this is taken care.  It's possible to make it `@property @abstractmethod ...` and then `@property ...` in the implementation, but it seems to me just one level of indirection and more typing.  It's cosmetic only, so I'm easily overruled.

pyiron_base/database/generic.py
pmrv: I fixed this one, but there's a similar inconsistency on `Project` with `switch_to_viewer_mode`.  I left that for now as it's externally visible.

pmrv:
Btw. I just found out you can unsubscribe from failed workflow notifications [here](https://docs.github.com/en/github/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#github-actions-notification-options). ;)

pyiron_base/database/filetable.py
liamhuber: Hmm, I think we've talked past each other although I don't know who's misunderstanding.

I'm trying to propose one fewer level of indirection by setting `IsDatabase.view_mode` as an `@property@abstractmethod` and removing `@abstractmethod def _get_view_mode` entirely. In this case I guess it's true that devs need to remember `@property` when overriding, is that your point? I guess then it's not a cosmetic difference so much as a cost/benefit of an extra line in the overrides vs one less method (and one less layer of indirection). I lean towards more typing and less layers of indirection, but only weakly.

pmrv: No, I think I got some things mixed up, as you said the only consequence should be more typing, which is ok with me.  I'll change accordingly.

pmrv: Looking at the code again, one advantage of having a proper property and abstract getter/setters is that you only need to document the property once.

liamhuber: > Looking at the code again, one advantage of having a proper property and abstract getter/setters is that you only need to document the property once.

I'm afraid it's unclear to me whether this is advocating for or against removing `def _get_view_mode` in favour of `@property@abstractmethod def view_mode`. IMO the documentation doesn't come into play as the attribute should be documented at the class level. I don't specifically object to property-level docstrings, but since non-property attributes need to get documented at the class level anyhow, I just figure it's unnecessary work.

Honestly though, I'm still torn between whether it's better to demand the extra typing or to introduce an extra layer of abstraction...

pmrv:
I won't have time to finish in one go and I don't want it to get outdated, so I propose I'll merge this at the end of the week with some additional tests and convert this into an issue to keep track of the progress.

pmrv:
I finally got around to updating this, if you don't oppose @liamhuber I'd merge this and iteratively work on this.

liamhuber:
@pmrv sounds good. Please skim over #437 and #441. These were problematic so I just shut them down, but it will give you a rough idea of the direction I'm thinking in. I'll also iterate on this topic over the next few weeks.

PR #340, opened 2021-06-08
Database Hotfix

pmrv:
@raynol-dsouza alerted me yesterday that pyiron was generally slow on the cluster, saving jobs etc could take up to a minute.  It turns out that the problem was with the database.  Our `DatabaseAccess` has an attribute `_keep_connection` which was set to `False` in the case of postgres, which in turns means every database look up would open a connection, do its thing, then closing it again.  One database connection takes ~1s, but e.g. saving a job requires ~20 database accesses so it stacks.

For now I've simply set `_keep_connection` to `True` in every case, but I'm not sure if that has any unintended side effects.  I'm not really sure why this wasn't an issue before.  I've briefly scanned the sqlalchemy changelog, but didn't find anything interesting.

On a side note there's a problem with logging in the database module right now. I would like log (re-)connections to the database, so it's quicker to debug in the future, but I also cannot import the settings module to access the global logger, since that would create a circular import (settings needs database access).  I have left simple print calls now, but this should be changed before merging.

### DIFF
--- a/pyiron_base/settings/generic.py
+++ b/pyiron_base/settings/generic.py
@@ -432,12 +432,12 @@ class Settings(metaclass=Singleton):
                 self._configuration["sql_view_user_key"] = parser.get(
                     section, "VIEWERPASSWD"
                 )
-            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", default=60)
+            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", fallback=60)
         elif self._configuration["sql_type"] == "SQLalchemy":
             self._configuration["sql_connection_string"] = parser.get(
                 section, "CONNECTION"
             )
-            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", default=60)
+            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", fallback=60)
         else:  # finally we assume an SQLite connection
             if parser.has_option(section, "FILE"):
                 self._configuration["sql_file"] = parser.get(section, "FILE").replace(
--- a/pyiron_base/settings/generic.py
+++ b/pyiron_base/settings/generic.py
@@ -60,6 +60,7 @@ class Settings(metaclass=Singleton):
             "user": "pyiron",
             "resource_paths": [],
             "project_paths": [],
+            "connection_timeout": 60,
             "sql_connection_string": None,
             "sql_table_name": "jobs_pyiron",
             "sql_view_connection_string": None,
@@ -200,6 +201,7 @@ class Settings(metaclass=Singleton):
             self._database = DatabaseAccess(
                 self._configuration["sql_connection_string"],
                 self._configuration["sql_table_name"],
+                timeout=self._configuration["connection_timeout"]
             )
 
     def switch_to_local_database(self, file_name="pyiron.db", cwd=None):
@@ -416,10 +418,12 @@ class Settings(metaclass=Singleton):
                 self._configuration["sql_view_user_key"] = parser.get(
                     section, "VIEWERPASSWD"
                 )
+            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", default=60)
         elif self._configuration["sql_type"] == "SQLalchemy":
             self._configuration["sql_connection_string"] = parser.get(
                 section, "CONNECTION"
             )
+            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", default=60)
         else:  # finally we assume an SQLite connection
             if parser.has_option(section, "FILE"):
                 self._configuration["sql_file"] = parser.get(section, "FILE").replace(
--- a/pyiron_base/settings/generic.py
+++ b/pyiron_base/settings/generic.py
@@ -192,6 +192,20 @@ class Settings(metaclass=Singleton):
         """
         return self._configuration["resource_paths"]
 
+    @property
+    def connection_timeout(self):
+        """
+        Get the connection timeout in seconds.  Zero means close the database after every connection.
+
+        Returns:
+            int: timeout in seconds
+        """
+        return self._configuration["connection_timeout"]
+
+    @connection_timeout.setter
+    def connection_timeout(self, val):
+        self._configuration["connection_timeout"] = val
+
     def open_connection(self):
         """
         Internal function to open the connection to the database. Only after this function is called the database is
--- a/pyiron_base/settings/generic.py
+++ b/pyiron_base/settings/generic.py
@@ -432,12 +432,12 @@ class Settings(metaclass=Singleton):
                 self._configuration["sql_view_user_key"] = parser.get(
                     section, "VIEWERPASSWD"
                 )
-            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", fallback=60)
+            self._configuration["connection_timeout"] = parser.get(section, "CONNECTION_TIMEOUT", fallback=60)
         elif self._configuration["sql_type"] == "SQLalchemy":
             self._configuration["sql_connection_string"] = parser.get(
                 section, "CONNECTION"
             )
-            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", fallback=60)
+            self._configuration["connection_timeout"] = parser.get(section, "CONNECTION_TIMEOUT", fallback=60)
         else:  # finally we assume an SQLite connection
             if parser.has_option(section, "FILE"):
                 self._configuration["sql_file"] = parser.get(section, "FILE").replace(
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -152,6 +152,9 @@ options can be added to the :code:`~/.pyiron`:
 
 * :code:`JOB_TABLE` the name of the database table. pyiron is commonly using one table per user. 
 
+* :code:`CONNECTION_TIMEOUT` the time in seconds before an idle connection to the database server is closed, set to 0 to
+  close after every transaction.  Default is 60 seconds.
+
 A typical :code:`.pyiron` configuration with a `PostgreSQL <https://www.postgresql.org>`_ database might look like this: 
 
 .. code-block:: bash
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -116,12 +116,13 @@ class ConnectionWatchDog(Thread):
 
 
 class AutorestoredConnection:
-    def __init__(self, engine):
+    def __init__(self, engine, timeout=60):
         self.engine = engine
         self._conn = None
         self._lock = Lock()
         self._watchdog = None
         self._logger = pyiron_base.settings.logger.get_logger()
+        self._timeout = timeout
 
     def execute(self, *args, **kwargs):
         while True:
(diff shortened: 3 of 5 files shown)


jan-janssen:
I guess the risk with the change is that a VASP calculation which runs for 3 days keeps the connection open while it does not communicate anything with the database. The breaking of the connection was implemented initially to reduce time outs during high throughout DFT studies.

pmrv:
I see the problem of leaking long running connections, but I also don't want to slow everyone else down because of it.  We could think about closing the connection after a certain timeout?  Googling I found [this](https://wiki.postgresql.org/wiki/PgBouncer) or we start a little [timer](https://docs.python.org/3/library/threading.html?highlight=timer#threading.Timer) that does it from a different thread.

pmrv:
Not sure if I understand the comment about high throughput studies.   Did their connections get slowed down by idle connections from other jobs?

jan-janssen:
As far as I remember the issue on the high throughput side was the number of connections. At some point the database decided to close the connection and the python objects still believed the connection was active and that resulted in issues.

jan-janssen:
@muh-hassani Can you take a look at the load of the database server right now?

max-hassani:
> 
> 
> @muh-hassani Can you take a look at the load of the database server right now?

@jan-janssen I have looked at it. Unfortunately, I don't have any previous reference to compare it with. Therefore, I took some snapshots and attached here.

pmrv:
Do I read this correctly that there's only ~30 connections to the DB open on average?  That seems plausible (one per CMler), but also not any load that should cause problem, no?

raynol-dsouza:
I get the following Runtime Error while trying to run Protocol jobs now:

```
File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/primitive/one_state.py", line 225, in command
    self._job_reload()

  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/primitive/one_state.py", line 267, in _job_reload
    self._job = pr.load(self._job_name)

  File "/u/rads/pyiron_repos/pyiron_base/pyiron_base/project/generic.py", line 762, in load
    job_id=job_id, convert_to_object=convert_to_object

  File "/u/rads/pyiron_repos/pyiron_atomistics/pyiron_atomistics/project.py", line 210, in load_from_jobpath
    job_id=job_id, db_entry=db_entry, convert_to_object=convert_to_object

  File "/u/rads/pyiron_repos/pyiron_base/pyiron_base/project/generic.py", line 781, in load_from_jobpath
    job = jobpath(db=self.db, job_id=job_id, user=self.user)

  File "/u/rads/pyiron_repos/pyiron_base/pyiron_base/job/path.py", line 468, in __init__
    db_entry = db.get_item_by_id(job_id)

  File "/u/rads/pyiron_repos/pyiron_base/pyiron_base/database/generic.py", line 585, in get_item_by_id
    return self.__get_items("id", int(item_id))[-1]

  File "/u/rads/pyiron_repos/pyiron_base/pyiron_base/database/generic.py", line 485, in __get_items
    row = result.fetchall()

  File "/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/sqlalchemy/engine/result.py", line 983, in fetchall
    return self._allrows()

  File "/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/sqlalchemy/engine/result.py", line 402, in _allrows
    made_rows = [make_row(row) for row in rows]

  File "/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/sqlalchemy/engine/result.py", line 402, in <listcomp>
    made_rows = [make_row(row) for row in rows]

RuntimeError: number of values in row (1) differ from number of column processors (16)
```

max-hassani:
@pmrv, as this has happened since the latest update to `sqlalchemy=1.4.17`, shall I revert the update to avoid the issue at the moment till we figure out a solution?

pmrv:
> @pmrv, as this has happened since the latest update to `sqlalchemy=1.4.17`, shall I revert the update to avoid the issue at the moment till we figure out a solution?

We can try.  As I said looking at the code, I'm not sure how this wasn't a problem previously, but it's better than nothing.

max-hassani:
The problem with downgrading sqlalchemy is that we have sqlalchemy>1.4.17 as the dependency in conda-forge feedstock!

pmrv:
> The problem with downgrading sqlalchemy is that we have sqlalchemy>1.4.17 as the dependency in conda-forge feedstock!

Have you tried any combination of `conda install --force` or `conda install --no-deps`?  For now we might get away with something like
```
conda install --force -no-deps sqlalchemy=1.4.15
```
but I don't know if this also forces downgrades.

max-hassani:
@pmrv thanks, it worked. Can you please check whether this has been effective in resolving the issue?

pmrv:
> @pmrv thanks, it worked. Can you please check whether this has been effective in resolving the issue?

Doesn't seem to change anything actually.  With the master branch I still get ~30s to create and save a job + a profile that shows all time is in the database connect, where with this branch it only takes ~2s.

niklassiemer:
Hmm, could you try with our previous release? Since it worked before, didn't it? I just don't understand what actually changed?? Or did we just miss the performance drop before and this is caused by an older update?

max-hassani:
@niklassiemer I remember 0.2.14 had some issue, I guess if I need to downgrade is to pyiron_base=0.2.13. Am I right?

pmrv:
> @niklassiemer I remember 0.2.14 had some issue, I guess if I need to downgrade is to pyiron_base=0.2.13. Am I right?

I will run git bisect with the timing test case as soon as I have time for it.

FWIW, I've been testing this with this snippet previously

```python

from pyiron import Project

pr = Project('lmp')

%%time 

import cProfile
p = cProfile.Profile()
p.enable()

j = pr.create.job.Lammps("test_svae", delete_existing_job=True)
j.structure = pr.create.structure.bulk("Fe", cubic=True).repeat(5)
j.calc_md(temperature=500, n_ionic_steps=100_000)
j.potential = '1997--Ackland-G-J--Fe--LAMMPS--ipr1'
j.save()

p.disable()

p.dump_stats("create_job_save.pyprof")
```

pmrv:
Way older commits don't seem to have the performance problems, but I haven't yet isolated the offending commit. But from this it seems sqlalchemy was a red herring and it's related to something we did recently.

max-hassani:
> 
> 
> > @niklassiemer I remember 0.2.14 had some issue, I guess if I need to downgrade is to pyiron_base=0.2.13. Am I right?
> 
> I will run git bisect with the timing test case as soon as I have time for it.
> 
> FWIW, I've been testing this with this snippet previously
> 
> ```python
> from pyiron import Project
> 
> pr = Project('lmp')
> 
> %%time 
> 
> import cProfile
> p = cProfile.Profile()
> p.enable()
> 
> j = pr.create.job.Lammps("test_svae", delete_existing_job=True)
> j.structure = pr.create.structure.bulk("Fe", cubic=True).repeat(5)
> j.calc_md(temperature=500, n_ionic_steps=100_000)
> j.potential = '1997--Ackland-G-J--Fe--LAMMPS--ipr1'
> j.save()
> 
> p.disable()
> 
> p.dump_stats("create_job_save.pyprof")
> ```

I tried your code snippet with pyiron_base=0.2.13, and still the wall time~40s :(

niklassiemer:
@pmrv what is 'way older' in this context, i.e. what is the newest release of base which is working (you found so far)?

pmrv:
> @pmrv what is 'way older' in this context, i.e. what is the newest release of base which is working (you found so far)?

Might have been a fluke.  Running git bisect it finds commit where the runtime of the snippet is substantially smaller, but the commits I've found so far depend on the bisect range and appear to be completely unrelated to the database.  I've also tested all pyiron_base tags back to version 0.2.1 and they all run slow.  Maybe it is related to the database setup.

niklassiemer:
> Might have been a fluke. Running git bisect it finds commit where the runtime of the snippet is substantially smaller, but the commits I've found so far depend on the bisect range and appear to be completely unrelated to the database. I've also tested all pyiron_base tags back to version 0.2.1 and they all run slow. Maybe it is related to the database setup.

Ok, did we change anything in the database setup? Otherwise, we would need to look into sqlalchemy again - maybe checking different versions on one user account? I am kind of lost what is going wrong here...!

max-hassani:
By the recent increase of the available memory to the database, my test shows a considerable improvement in running @pmrv's code snippet. I got the wall time of ~850ms. I think the issue is solved and we can close the ticket and this pull request.

pmrv:
> By the recent increase of the available memory to the database, my test shows a considerable improvement in running @pmrv's code snippet. I got the wall time of ~850ms. I think the issue is solved and we can close the ticket and this pull request.

Thanks for checking this and also opening the ticket!

I'd want to keep this open for a bit longer though.  I still feel we are leaving some performance on the table by having a new connection for every access.  I'm open for other solutions though, I just want to keep this on the agenda for a bit.

pmrv:
I've added a small watchdog that closes the connection after 60s.  In my small benchmark of saving a lammps job, this decreases the time spent talking to the database by factor 10 and the total time by factor 3.

We discussed in the database meeting that there should be a way to control `_keep_connection` from the `JobWrapper`.  I'll add that once I've tested this a bit on the cluster.

pmrv:
I've been test driving this for a while now without problems, but I'd sleep better if someone else (ping: 
@samwaseda @raynol-dsouza @sudarsan-surendralal @niklassiemer) could also try it during their workflows.  The only thing that should change is that from time to time a string "Reconnecting to DB; connection closed." might pop up.  This is normal, it just means the connection died/was closed and is automatically reopened.  I'd like that to be in the log, but haven't figured it out yet, because there is a circular import problem as mentioned above.

niklassiemer:
I fear I am not using the db connection that rigorously since I am actually not performing calculations. Nonetheless, I will use this branch if I am not developing `pyiron_base` to check for issues.
For the logger connection I had one idea, see PR #407 (Not yet tested...)

niklassiemer:
I tried with PR #407 into here, however, I somehow never triggered any of the `pyiron_log - INFO` so far...

pmrv:
Btw. you can check whether you currently have a running connection to the database with

```
lsof -i @pyiron-db-cmmc.esc.rzg.mpg.de
```

if this gives output, yes, if not, then no.

pmrv:
> I tried with PR #407 into here, however, I somehow never triggered any of the `pyiron_log - INFO` so far...

Same, which is weird, but it also doesn't give any errors. :|

niklassiemer:
Maybe try with the warning level? I assumed we want to have it at the 'INFO' level - not at the 'WARN' level.

pmrv:
I've added a config flag to turn the 'timeout' off and use it when running jobs from a cluster node (i.e. via job_wrapper).  Log messages do appear now for me, but maybe it's even a bit too chatty, as you'll get a message whenever the timer ran out and you initiate a new database connection.  Then again it's on the `INFO` level, so it's easily muted.

I haven't noticed any hick-ups so far, but let me know if you do, otherwise I'll merge this after the next pyiron meeting.

pyiron_base/settings/generic.py
```diff
@@ -416,10 +432,12 @@ def _config_parse_file(self, config_file):
                 self._configuration["sql_view_user_key"] = parser.get(
                     section, "VIEWERPASSWD"
                 )
+            self._configuration["connection_timeout"] = parser.get(section, "connection_timeout", fallback=60)
```
pmrv: String name should be documented.

pmrv:
Have a quick logger with database statistics from pyiron-base.

PR #298, opened 2021-04-27
Fix swmr warnings

jan-janssen:
```
UserWarning: swmr=True only affects read ('r') mode. For swmr write mode, set f.swmr_mode = True after opening the file.
  with h5py.File(self.file_name, mode="a", libver="latest", swmr=True) as h:
```

PR #346, opened 2021-06-14
Remove HDFStoreIO

pmrv:
For historic reasons we used a subclass from pandas to get groups/nodes information from an HDF file.  I've replaced these parts using h5py directly.  This makes the code a bit easier and also seems to speed up removing jobs (where list_groups() could take a long time).

### DIFF
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -25,6 +25,13 @@ __email__ = "janssen@mpie.de"
 __status__ = "production"
 __date__ = "Sep 1, 2017"
 
+def open_hdf5(filename, mode="r", swmr=False):
+    if swmr and mode != "r":
+        store = h5py.File(filename, mode=mode, libver="latest")
+        store.swmr = True
+        return store
+    else:
+        return h5py.File(filename, mode=mode, libver="latest", swmr=swmr)
 
 class FileHDFio(object):
     """
@@ -176,7 +183,7 @@ class FileHDFio(object):
             bool: [True/False]
         """
         if self.file_exists:
-            with h5py.File(self.file_name) as h:
+            with open_hdf5(self.file_name) as h:
                 return len(h.keys()) == 0
         else:
             return True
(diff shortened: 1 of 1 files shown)


pyiron_base/generic/hdfio.py
```diff
@@ -26,58 +25,13 @@
...
-        Returns:
-            HDFStoreIO: self - in contrast to the original implementation in pandas.
-        """
-        super(HDFStoreIO, self).open(**kwargs)
-        return self
-
+def open_hdf5(filename, mode="r", swmr=False):
+    if swmr and mode != "r":
+        store = h5py.File(filename, mode=mode, libver="latest")
+        store.swmr = True
```
niklassiemer, line 31: what is the difference between 
`h5py.File(..., swmr=True)` and `store = h5py.File(...); store.swmr = True`?
Since if this is the same, all the logic is not needed?

pyiron_base/generic/hdfio.py
```diff
@@ -893,16 +843,6 @@ def _get_h5_path(self, name):
         """
         return posixpath.join(self.h5_path, name)
 
-    def _get_store_node(self, store):
-        """
-        Index store with the current h5_path set, but avoid to index by '/', because that raises an IndexError.
```
niklassiemer, line 898: This is not an issue when directly using the `h5py.File` interface, right?

niklassiemer (commented):
I like the streamlining with the `open_hdf5` call and the  `HDFStoreIO` is indeed only used to access items in the hdf5 file itself. It looks good to me as far as I can tell (did not test this)

niklassiemer:
And we currently have a failing test in atomistics
```
FAIL: test_import (vasp.test_vasp_import.TestVaspImport)
  File "/home/runner/work/pyiron_base/pyiron_atomistics/tests/vasp/test_vasp_import.py", line 51, in test_import
    self.assertEqual(len(w), 4)
AssertionError: 3 != 4
```
EDIT:
This is a warnings catcher which now has one warning less. However, the tests do not tell why (and which) warnings are expected in this test.

pyiron_base/generic/hdfio.py
pmrv: When opening the file writing/amending h5py ignores the `swmr` flag as in #298, this has somehow something to do with what you are allowed to do with a file [swmr mode](https://docs.h5py.org/en/stable/swmr.html).  It's a bit academic, because we never use `swmr=True`, so I could throw it out.  It's only in, because I started working on this before #298 and only rebased after.

pyiron_base/generic/hdfio.py
pmrv: No, `File`s are also `Group`s, so you can index them directly.

pmrv:
> And we currently have a failing test in atomistics
> 
> ```
> ======================================================================
> FAIL: test_import (vasp.test_vasp_import.TestVaspImport)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/runner/work/pyiron_base/pyiron_atomistics/tests/vasp/test_vasp_import.py", line 51, in test_import
>     self.assertEqual(len(w), 4)
> AssertionError: 3 != 4
> ```
> 
> EDIT:
> This is a warnings catcher which now has one warning less. However, the tests do not tell why (and which) warnings are expected in this test.

I suppose this has something to do with an swmr warning disappearing.  Thanks for checking, I hadn't picked up that it's only a warnings issue.

pmrv:
I've confirmed the extra warning comes from pytables (about a np.object type error).  I'd suggest we ignore this here and fix the error on the next pyiron_base version in atomistics.

niklassiemer (approved):
Agreed, then let's merge this shortly before a new base release to allow for an easy identification of further integration tests on other PRs and fix the atomistics test in the update of the base version number.

pmrv:
> Agreed, then let's merge this shortly before a new base release to allow for an easy identification of further integration tests on other PRs and fix the atomistics test in the update of the base version number.

Yup, can wait until Tuesday.

PR #354, opened 2021-06-24
Update generic.py

samwaseda:
`if None` to `if is None`

### DIFF
--- a/pyiron_base/master/generic.py
+++ b/pyiron_base/master/generic.py
@@ -147,7 +147,7 @@ class GenericMaster(GenericJob):
         Returns:
             list: list of child job ids
         """
-        if self._child_id_func:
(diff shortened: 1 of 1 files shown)


niklassiemer (approved):

pyiron_base/master/generic.py
```diff
@@ -265,7 +265,7 @@ def _after_generic_copy_to(self, original, new_database_entry, reloaded):
         if reloaded:
             return
 
-        if self.job_id and new_database_entry and original._job_id:
+        if self.job_id is not None and new_database_entry and original._job_id is not None:
```
jan-janssen, line 268:
```suggestion
        if self.job_id is not None and new_database_entry is not None and original._job_id is not None:
```

PR #356, opened 2021-06-25
Remove identical functions

### LINKED ISSUE #355: Why are these functions defined in both `GenericJob` and `GenericMaster`?
https://github.com/pyiron/pyiron_base/blob/30aeb75325c6722242fbf42f867f8eed32c7db05/pyiron_base/master/generic.py#L387

There are quite a few functions e.g. `run_if_interactive`, `collect_output` etc. which are defined both in `GenericJob` and `GenericMaster`, even though `GenericMaster` is anyway derived from `GenericJob`. Some of them are `pass` in `GenericMaster` and `NotImplementedError` in `GenericJob`, others are exactly identical (`NotImplementedError` in both of them). Are there particular reasons that they are repeated?

samwaseda:
Some functions are defined in `GenericMaster` and `GenericJob` which return both `NotImplementedError` and some other functions return `None` in `GenericMaster` and `NotImplementedError` in `GenericJob`. I removed all of them in `GenericMaster` and see what happens.

Closes #355

### DIFF
--- a/pyiron_base/master/generic.py
+++ b/pyiron_base/master/generic.py
@@ -511,15 +511,6 @@ class GenericMaster(GenericJob):
         """
         pass
 
-    def run_if_refresh(self):
-        """
-        Internal helper function the run if refresh function is called when the job status is 'refresh'. If the job was
-        suspended previously, the job is going to be started again, to be continued.
-        """
-        raise NotImplementedError(
-            "Refresh is not supported for this job type for job  " + str(self.job_id)
-        )
-
     def _run_if_busy(self):
         """
         Run if busy is not implemented for MetaJobs
--- a/pyiron_base/master/generic.py
+++ b/pyiron_base/master/generic.py
@@ -277,24 +277,6 @@ class GenericMaster(GenericJob):
                 if new_database_entry and child.master_id:
                     new_child.master_id = self.job_id
 
-    def update_master(self):
-        """
-        After a job is finished it checks whether it is linked to any metajob - meaning the master ID is pointing to
-        this jobs job ID. If this is the case and the master job is in status suspended - the child wakes up the master
-        job, sets the status to refresh and execute run on the master job. During the execution the master job is set to
-        status refresh. If another child calls update_master, while the master is in refresh the status of the master is
-        set to busy and if the master is in status busy at the end of the update_master process another update is
-        triggered.
-        """
-        master_id = self.master_id
-        project = self.project
-        self._logger.info("update master: {} {} {}".format(master_id, self.get_job_id(), self.server.run_mode))
-        if master_id is not None:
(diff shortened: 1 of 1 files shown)


pyiron_base/master/generic.py
```diff
@@ -277,24 +277,6 @@ def _after_generic_copy_to(self, original, new_database_entry, reloaded):
...
-        """
-        master_id = self.master_id
-        project = self.project
-        self._logger.info("update master: {} {} {}".format(master_id, self.get_job_id(), self.server.run_mode))
-        if master_id is not None:
-            self._reload_update_master(
-                project=project,
-                master_id=master_id
-            )
-
```
jan-janssen, line 297: We need this

pyiron_base/master/generic.py
```diff
@@ -357,57 +339,6 @@ def get_child_cores(self):
...
-        """
-        interactive flush is not implemtned for MetaJobs
-        """
-        pass
-
-    def run_if_interactive_non_modal(self):
-        """
-        Run if interactive non modal is not implemented for MetaJobs
-        """
-        pass
```
jan-janssen, line 409: We need this

pyiron_base/master/generic.py
```diff
@@ -580,21 +511,6 @@ def _executable_activate_mpi(self):
...
-        raise NotImplementedError(
-            "Refresh is not supported for this job type for job  " + str(self.job_id)
-        )
-
-    def _run_if_busy(self):
-        """
-        Run if busy is not implemented for MetaJobs
-        """
-        pass
-
```
jan-janssen, line 597: We need this

jan-janssen:
When proposing these kind of changes - please always run the integration tests, by adding the corresponding label

niklassiemer:
Right now, I have the feeling we should write some tests that fail on this PR 😆 Removing the duplicated `NotImplementedError`s should be fine. However, the change from `NotImplementedError` to `pass` should make a difference.

liamhuber:
> Right now, I have the feeling we should write some tests that fail on this PR 😆

💯 this is the key issue. I've got to do some science this morning, but I'll come back to this PR later in the week to either review what's been done or do something myself depending on how the week goes for the rest of you.

pyiron_base/master/generic.py
samwaseda: Why?

pyiron_base/master/generic.py
samwaseda: Why?

pmrv (commented):
Not sure on the details, but I agree with the general idea of keeping overrides that silence `NotImplementedError`s with `pass`, but throw out those methods that just also raise them.

PR #361, opened 2021-06-28
Postgres performance

max-hassani:
The new class is intended to monitor important parameters of pyiron central database. Here, it is assumed that the database is a Postgresql database. The final data is presented as pandas dataframe.

the use case should be :
```
from pyiron_base import DatabaseStatistics
db_stat = DatabaseStatistics()
df = db_stat.performance()
```
The references for the queries can be found:
[1] https://russ.garrett.co.uk/2015/10/02/postgres-monitoring-cheatsheet/
[2] https://pgdash.io/blog/essential-postgres-monitoring-part2.html
[3] https://wiki.postgresql.org/wiki/Index_Maintenance


max-hassani:
Up to an extent, I converted the sql queries to sqlalechemy expressions, but some of them are really long and complicated to convert. But if the reviewers see it essential, I will try to convert all the queries to sqlalchemy expressions.

max-hassani:
Of course a unittest will be added. I have to first figure out how to create a postgres database via a github workflow.

pyiron_base/database/performance.py
```diff
@@ -0,0 +1,170 @@
+import pandas as pd
```
jan-janssen, line 6: Can you add the copyright statement on top?

pyiron_base/database/performance.py
```diff
@@ -0,0 +1,170 @@
...
+        s = Settings()
+        self._connection_string = s._configuration['sql_connection_string']
+        self._job_table = s._configuration['sql_view_table_name']
+        if "postgresql" not in self._connection_string:
+            raise RuntimeError(
+                """
+                The detabase statistics is only available for a Postgresql database
+                """
+            )
+        self._table = s._configuration['sql_table_name']
```
jan-janssen: This variable is not used - if I see it correctly.

pyiron_base/database/performance.py
```diff
@@ -0,0 +1,170 @@
...
+)
+__version__ = "1.0"
+__maintainer__ = "Muhammad Hassani"
+__email__ = "hassani@mpie.de"
+
+
+class DatabaseStatistics:
+    def __init__(self):
+        s = Settings()
+        self._connection_string = s._configuration['sql_connection_string']
```
jan-janssen: This can be a local variable, as it is only used during the initialisation.

pyiron_base/database/performance.py
```diff
@@ -0,0 +1,170 @@
...
+        self._stat_view = Table('pg_stat_activity', self._metadata, autoload_with=self._engine)
+        self._locks_view = Table('pg_locks', self._metadata, autoload_with=self._engine)
+
+    def _num_conn(self, conn):
+        """
+        return the number of connections
+        """
+        stmt = select(func.count()).select_from(self._stat_view)
+        result = conn.execute(stmt)
+        self._performance_dict['total num. connection'] = result.fetchone()[0]
```
jan-janssen: I find this code hard to read. From my perspective it would be easier to return the entry rather than writing to the dictionary. This also simplifies debugging later.

jan-janssen:
@muh-hassani I added a few comments but apart from those I guess it would be great to merge this functionality soon.

pyiron_base/database/performance.py
max-hassani: Done!

pyiron_base/database/performance.py
max-hassani: Done!

pyiron_base/database/performance.py
max-hassani: Do you mean the whole code or this function in particular? There are multiple functions which modifies the `self._performance_dict`. Some of them, like [here](https://github.com/pyiron/pyiron_base/blob/b1c02b722ce05a4dde3225e5d7ef73f93799d9dd/pyiron_base/database/performance.py#L60) and [here](https://github.com/pyiron/pyiron_base/blob/b1c02b722ce05a4dde3225e5d7ef73f93799d9dd/pyiron_base/database/performance.py#L133) would be more complex to be returned an entry. But definitely can be done.

jan-janssen: Yes, I would like each function to return a sub dictionary and then use `dict.update()` to merge them in the end. https://python-reference.readthedocs.io/en/latest/docs/dict/update.html This way we can debug the individual functions, look at the output they return. Following the zen of python: Explicit is better than implicit. https://www.python.org/dev/peps/pep-0020/

max-hassani: Thanks for the links. Now each method returns a dictionary which is used at the end to update `self._performance_dict`.

pyiron_base/database/performance.py
max-hassani: I removed it.

jan-janssen:
I like the class still I can not stop wondering if it would not be easier for the user to have just one function? 
```
from pyiron_base import get_database_statistics
get_database_statistics()
```

This function could be defined as: 
```
def get_database_statistics():
    return DatabaseStatistics().performance()
```

max-hassani:
> 
> 
> I like the class still I can not stop wondering if it would not be easier for the user to have just one function?
> 
> ```
> from pyiron_base import get_database_statistics
> get_database_statistics()
> ```
> 
> This function could be defined as:
> 
> ```
> def get_database_statistics():
>     return DatabaseStatistics().performance()
> ```

I agree, it makes the usage much easier. I have done it, the way you suggested. 
I also made two methods as separate functions, as they did not need to be part of the class. You can find them [here](https://github.com/pyiron/pyiron_base/blob/151f1da34b6cb75ba5131d5956fe73147f9f54f8/pyiron_base/database/performance.py#L26) and [here](https://github.com/pyiron/pyiron_base/blob/151f1da34b6cb75ba5131d5956fe73147f9f54f8/pyiron_base/database/performance.py#L45)

max-hassani:
@jan-janssen, if you approve the current status of this PR, I can merge it.

jan-janssen (approved):
> @jan-janssen, if you approve the current status of this PR, I can merge it.

Looks good to me.

PR #363, opened 2021-07-02
Use correct project in copy_to

### LINKED ISSUE #362: copy to bug
If I did not miss something, the `file_project` is not defined here:
https://github.com/pyiron/pyiron_base/blob/5e69910af320fd06cf48d95558edb4ed76cc1d55/pyiron_base/master/generic.py#L272

pmrv:
Should fix #362 

If I see this correctly `file_project` originally referred to the project from `_get_project_for_copy`, which in turn should be the project the new job is located in turn and since the `_after_generic_copy_to` is called on the new job this would be identical to `self.project`.


niklassiemer:
At least it fixed the copy problem I had. I will merge it now. However, we really need to write more tests for such cases!

Issue #364, opened 2021-07-03
Better node labels when storing a `DataContainer` in HDF5 files


sudarsan-surendralal:
Suppose I define a class inherited from a data container:

```
class A(DataContainer):

    def __init__(self, a, b):
        super().__init__()
        self.a =a
        self.b =b

obj = A(3, 4)
obj.to_hdf(hdf5, group_name="A")

```
Currently, `a` and `b` are stored as `hdf5["A/a__index_0"]` and `hdf5["A/b__index_1"]`.  I understand that the reason the attributes are stored in such a way is to preserve the order when re-creating the object using `from_hdf()`. But it would be great if we can use the node labels consistent with the attribute names (`hdf5["A/a"]` and `hdf5["A/b"]`) especially when accessing data using the inspect mode.

pmrv:
I see the problem, but no clear solution right now.  Technically there's attributes on HDF groups that we could use to store this, but iirc @jan-janssen and I had timed this once and it was a bit slow.  (It'd be possible look into it again, though)

A thing that we had discussed before (and that is on my back burner), is to lazily read `DataContainer`s from HDF5 as you tab-complete through it.  If that is implemented, we could always load `DataContainer`s (even in inspect mode) without the speed penalty and then show the correct names, but it's a bit low on my priority right now.

liamhuber:
I actually am totally comfortable with the HDF tags being hideous; I would really like to get to the point that users are never touching these anyhow -- strings are the devil. This means getting some sort of `inspect` routine or attribute running that reproduces the full HDF string path from the `.`-style path. In between now and then I'd rather live with ugly paths. -- Aha, I read the second half of Marvin's comment and he's already on this 😉

jan-janssen:
I have the feeling we have two concepts colliding here:

- In the current pyiron version the HDF5 path try to emulate the object hierarchy. So `job.output.energy_tot` equals `job["output/generic/energy_tot"]. This enables the inspect mode to accelerate the analysis. Still there are limitations to this approach, as nested objects like `job.structure.positions` is not stored in `job["input/generic/structure/positions"]`.
- The second approach and I believe that is what @pmrv is currently working on is to initialise the objects in a lazy way. So the content is only loaded when it is accessed. While I like this concept, I am not sure how much we have to implement and how difficult it is to implement this for new developers.

pmrv:
[...]

PR #367, opened 2021-07-07
Allow lazy loading of DataContainers

pmrv:
With this data containers (and potentially other data types) can be loaded "on demand" from HDF5 instead of all at once.  In my simple example below this saves ~90% time on load.

The way it works is that in `from_hdf` values of the data container are not loading right away but just the hdf object and the path are saved in a "stub" object, that is then later transparently loaded upon item access.

Tests and design spec follow when we discussed this.

This should help with #364 and inspect mode in general.  My vision would be that expensive to load objects should be using data container for all storage and can then be `load()`ed with `lazy=True` instead of `inspect()`ed.

Here's an example that can be run with `ipython -i`

```python
# coding: utf-8
import pyiron_base.generic.datacontainer as D
import numpy as np

d = D.DataContainer()
pr = Project('asdf')
h = pr.create_hdf(pr.path, 'test')

K = 'abcdefghijklmnop'
for c in K:
    dd = d.get(c, create=True)
    for c in K:
        dd[c] = np.random.rand(1024)

d.to_hdf(h, 'lazy')
# ~2s
get_ipython().run_line_magic('time', "d_normal = h['lazy'].to_object()")
# ~150ms
get_ipython().run_line_magic('time', "d_lazy = h['lazy'].to_object(lazy=True)")

assert (d_normal.a.a == d_lazy.a.a).all()
```

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -250,8 +250,9 @@ class DataContainer(MutableMapping, HasGroups):
 
         return instance
 
-    def __init__(self, init=None, table_name=None):
+    def __init__(self, init=None, table_name=None, lazy=False):
         self.table_name = table_name
+        self._lazy = lazy
         if init is not None:
             self.update(init, wrap=True)
 
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -14,6 +14,7 @@ import numpy as np
 
 from pyiron_base.interfaces.has_groups import HasGroups
 from .fileio import read, write
+from .hdfstub import SimpleStub, ObjectStub
 
 __author__ = "Marvin Poul"
 __copyright__ = (
@@ -245,6 +246,7 @@ class DataContainer(MutableMapping, HasGroups):
         object.__setattr__(instance, "_indices", {})
         object.__setattr__(instance, "table_name", None)
         object.__setattr__(instance, "_read_only", False)
+        object.__setattr__(instance, "_lazy", False)
 
         return instance
 
@@ -272,13 +274,21 @@ class DataContainer(MutableMapping, HasGroups):
 
         elif isinstance(key, int):
             try:
-                return self._store[key]
+                v = self._store[key]
+                if not isinstance(v, HDF5Stub):
+                    return v
+                else:
+                    return v.realize()
             except IndexError:
                 raise IndexError("list index out of range") from None
 
         elif isinstance(key, str):
             try:
-                return self._store[self._indices[key]]
+                v = self._store[self._indices[key]]
+                if not isinstance(v, HDF5Stub):
+                    return v
+                else:
+                    return v.realize()
             except KeyError:
                 raise KeyError(repr(key)) from None
 
@@ -740,9 +750,9 @@ class DataContainer(MutableMapping, HasGroups):
             for n in hdf.list_nodes():
                 if n in _internal_hdf_nodes:
                     continue
-                items.append( (*normalize_key(n), hdf[n]))
+                items.append( (*normalize_key(n), hdf[n] if not self._lazy else SimpleStub(hdf, n)) )
             for g in hdf.list_groups():
-                items.append( (*normalize_key(g), hdf[g].to_object()))
+                items.append( (*normalize_key(g), hdf[g].to_object() if not self._lazy else ObjectStub(hdf, g)) )
             for _, k, v in sorted(items, key=lambda x: x[0]):
                 self[k] = v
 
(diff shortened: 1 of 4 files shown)


pyiron_base/generic/datacontainer.py
```diff
@@ -10,6 +10,7 @@
 import warnings
 import numpy as np
 from .fileio import read, write
+from .hdfstub import HDFStub
```
jan-janssen: Personally I prefer global imports, or is there a specific reason to use relative imports here?

jan-janssen:
I still have a bit of trouble to understand the full implications of this change. For example for a structure object stored in the HDF5 file, when are the positions loaded at `job.structure` or only when I select the positions `job.structure.positions` ? So I guess we should discuss this in more detail tomorrow.

pyiron_base/generic/datacontainer.py
pmrv: Jup, will fix.

pmrv:
Currently they would be loaded when `job.structure` is accessed, assuming we'd have the following setup

```python
class Job:
  ...
  def from_hdf(self, hdf, group_name):
    self._structure = HDFStub(hdf, group_name + '/structure')
  ...
  @property
  def structure(self):
    if isinstance(self._structure, HDFStub):
      self._structure = self._structure.load() # all HDF access for this structure occurs here
    return self._structure
```

In the current setup this change only affects how `DataContainer` is read from HDF, but `HDFStub` would be easily used to delay loading of other types as well.

However `HDFStub` already allows to customize this, with the `register` method.  If the `Atoms` class supported to read itself lazily from HDF (which would have to be implemented separately), then you could do something like

```python
HDFStub.register('Atoms', lambda hdf, group: Atoms.lazy_from_hdf(hdf, group))
```

>  So I guess we should discuss this in more detail tomorrow.

Yeah, will prepare a small demo.

pmrv:
Only thing left from my side is whether I should enable `lazy` by default or not.

pmrv:
I rebased to fix the merge conflicts and will merge will lazy loading enabled by default, if no one objects.

tests/generic/test_datacontainer.py
```diff
@@ -459,6 +460,54 @@ def test_subclass_preservation(self):
...
+        self.assertTrue(all(isinstance(v, HDFStub) for v in ll0._store),
+                        "Recursive datacontainers not lazily loaded!")
+
+        self.assertEqual(ll[0].foo, self.pl[0].foo,
+                         "Lazily loaded list not equal to orignal list!")
+
+        self.assertTrue(not isinstance(ll._store[0], HDFStub),
+                        "Loaded value not stored back into container!")
+
+    def test_stub(self):
```
liamhuber: I wanted to peek at codacy before approving and it actually caught something meaningful 😝 This is a redefinition of the test in 463!

tests/generic/test_datacontainer.py
```diff
@@ -459,6 +460,54 @@ def test_subclass_preservation(self):
...
 
+    def test_stub(self):
+        """Lazily loaded containers should contain only stubs and only force them when directly accessed."""
+
+        self.pl.to_hdf(self.hdf, "lazy")
+        ll = self.hdf["lazy"].to_object(lazy=True)
+        self.assertTrue(all(isinstance(v, HDFStub) for v in ll._store),
+                        "Not all values loaded as stubs!")
+
+        ll_repr = repr(ll)
```
liamhuber: Codacy also complains this is unused. If I understand, you're just making sure the repr doesn't throw an error, yeah? In that case I guess there's no need to assign it to a variable. Similar down in line 475.

pyiron_base/generic/hdfstub.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Args:
+            hdf (:class:`.ProjectHDFio`): hdf object to load from
+            group_name (str): node or group name to load from the hdf object
+        """
+        self._hdf = hdf
+        self._group_name = group_name
+
+    @classmethod
+    def register(cls, type, load):
```
liamhuber, line 71: Finally, codacy complains that this overrides a built-in type...do you know what it's talking about? In pycharm I don't get any complaint, and I was expecting like when you try to use "id" or "dict" as a variable.

liamhuber (changes requested):
Codacy actually said something meaningful! Otherwise lgtm.

pyiron_base/generic/hdfstub.py
pmrv: The argument `type` is a builtin.  It is the base class of all classes and also a function to create new classes (iirc).  Since I'm not using this inside this short method, I think it's ok, but I can also come up with a different name.

tests/generic/test_datacontainer.py
pmrv: It's to make sure that calling `repr` doesn't force the stub inside a lazily loaded container, I'll remove it, but I fear codacy will still complain.

PR #370, opened 2021-07-12
Update dependabot pr

niklassiemer:


.ci_support/update_environment.py
```diff
@@ -0,0 +1,37 @@
...
+with open('.ci_support/environment.yml', 'r') as f:
+    environment = f.readlines()
+
+with open('.ci_support/environment.yml', 'w') as f:
+    for line in environment:
+        f.write(re.sub(
+            r'(' + package_name + '.*)' + from_version,
+            r'\g<1>' + to_version,
+            line
+        ))
```
jan-janssen: Is there a specific reason not to use a YAML parser? Here is an example project how to work with `environment.yml` files https://github.com/amitbeka/conda-merge

niklassiemer: No, I just directly translated my former `sed` call into python. I will have a look at the parser and use that instead.

niklassiemer: Since the outcome of `yaml.save_load(file_stream)` is a dictionary like 
```
{'channels': ['conda-forge'],
 'dependencies': [
  'pyiron_base =0.2.17',
  'pyiron_atomistics =0.2.17',
  'numpy =1.21.0' 
]}
```
One after all needs to use `re` to change the line. However, at least it is only applied to the dependencies so it is a bit safer this way 😄

.ci_support/update_environment.py
```diff
@@ -0,0 +1,45 @@
...
+    updated_dependencies.append(re.sub(
+         r'(' + package_name + '.*)' + from_version,
+         r'\g<1>' + to_version,
+         dep
+    ))
+
+environment['dependencies'] = updated_dependencies
+
+with open('.ci_support/environment.yml', 'w') as f:
+    yaml.safe_dump(environment, f)
```
jan-janssen: Can you split this in some more functions? At least the updating of the environment should be a separate function from my perspective. So the next person who updates the code knows what to change.

niklassiemer:
I would merge this soon if nobody objects.

jan-janssen:
Maybe I am a bit picky, but would not it be possible to manage the `pyiron_` dependencies also with the `JSON` file. Currently it feels like a very specific solution as the replacement of `-` with `_` only happens for certain packages and so on.

jan-janssen:
In addition if I understand the tests from @pmrv  in https://github.com/pmrv/submodule_user then it should be possible to move both the `.ci_support` as well as the `.github` folder to a separate repository. Maybe even merge both in the `.github` folder by adding some pyiron subfolder `.github/pyiron`. Anyway that is something we can also do in the next pull request.

niklassiemer:
Well, yes it would be possible and easy to include the `pyiron-module` to `pyiron_module` conversion in the `json` file. I just had the feeling we would stick to our naming pattern and adding new pyiron_modules would not need a modification. However, explicit is better than implicit 👍 
To the submodule of `.github`. This is currently _not working_ as intended. We could use a submodule and copy everything into the right places and commit with a workflow. However, it is not possible to just use `.github` as a submodule since this is not checked out. This is also true for links into another sub-module from the `.github` (since this is also not checked out with the main repo).
Therefore, our idea was to use the [template repo](https://github.com/pyiron/pyiron_module_template) and to make each merge to `master` there a patch to open a PR on all other repos.

.ci_support/update_environment.py
```diff
@@ -0,0 +1,68 @@
...
+        with open(environment_file, 'r') as f:
+            self.environment = yaml.safe_load(f)
+
+        self.package_name = self._convert_package_name(package_name)
+
+    def _convert_package_name(self, name):
+        try:
+            result = self._name_conversion_dict[name]
+        except KeyError:
+            result = name
```
jan-janssen: You can check if `name` is in `self._name_conversion_dict.keys()` then you do not need the `try` and `except` part.

.ci_support/update_environment.py
```diff
@@ -0,0 +1,68 @@
+import json
```
jan-janssen, line 5: Can you add the copyright statement at the beginning of the file?

jan-janssen (approved):
Looks good to me

PR #371, opened 2021-07-13
Add _repr_html_ to display DataContainer in notebook mode

niklassiemer:
I just had the issue of an awful representation of the `DataContainer` in some Notebook again and tried this. Does it break the lab view?


pmrv (approved):
Doesn't seem to change anything in lab and in notebooks it's ok now, I think, thank!

PR #372, opened 2021-07-13
Remove dead code

pmrv:


pmrv:
I grepped all our repos for this variable, but it's nowhere used, btw.

PR #373, opened 2021-07-13
Properly cleanup copyto tests

pmrv:
Previously some hdf5 files were left that caused the tests to fail if run multiple times.


tests/job/test_copyto.py
```diff
@@ -21,6 +21,7 @@ def tearDownClass(cls):
         sub_project.remove(enable=True)
         sub_project = project.open("sub_project")
         sub_project.remove(enable=True)
+        project.remove(enable=True)
```
niklassiemer: Could be based on `TestWithProject` instead, could it not?

PR #376, opened 2021-07-16
Add a apply method to map functions over stored arrays to FlattenedStorage

pmrv:
Quick sketch following our discussion today about handling numpy functions and ragged arrays.

Once `FlattenedStorage` has convenience methods that return ragged and filled arrays, I think this could be easily used in e.g. the `Neighbors` classs.

Tests follow if we think this is useful.


pyiron_base/generic/flattenedstorage.py
```diff
@@ -364,6 +364,102 @@ def has_array(self, name):
         return {"shape": a.shape[1:], "dtype": a.dtype, "per": per}
 
 
+    def apply(self, name, function, inplace=True, per="element"):
```
niklassiemer: I do not like `inplace` since this sounds like you would not return anything but change the array itself. I.e. I understand inplace as 
```
source = self...
result = function(source) 
...
self.... = result
``` 
instead of 
```
source = self...
result = function(source) 
...
return result
```

pyiron_base/generic/flattenedstorage.py
```diff
@@ -364,6 +364,102 @@ def has_array(self, name):
...
+                result = np.array([ function(source[self._get_per_element_slice(i)]) for i in range(len(self)) ])
+            else:
+                source = self._per_chunk_arrays[name][:self.num_chunks]
+                result = np.array([ function(source[i]) for i in range(len(self)) ])
+            result.resize((self.num_chunks_alloc,) + result.shape[1:])
+            store = self._per_chunk_arrays
+        else:
+            raise ValueError(f"per must \"element\" or \"chunk\", not {per}")
+
+        if inplace:
```
niklassiemer: If this save in this case? Ok, we probably do not want to support `0` or `''` as name. However, wouldn't a `1` cause problems?
```suggestion
        if inplace is True or isinstance(inplace, str) and len(inplace)>0:
```
I just realize that it is indeed modified inplace if `name` and `inplace` match. Do we want this?
This would also have very nasty side effects for `'length'` e.g. as result for a norm function!

niklassiemer (commented):
I did not test if it does what it is supposed to do. Especially, the part to extract the flattened array per chunk or per element, I did not fully follow. However, as far as I can tell it looks like a valuable method. One thing I would like is to not allow for 'inplace' modification of existing arrays.

pyiron_base/generic/flattenedstorage.py
pmrv: Agree, this should check if the target name already exists in the storage.

pmrv: `0` or `1` would make problems indeed, but the array names are explicitly documented as strings.  I will add code to enforce this.

pyiron_base/generic/flattenedstorage.py
pmrv: I felt always returning the result is most convenient, but am open to other namings.

PR #379, opened 2021-07-23
Extend `load_file` to handle file objects.

niklassiemer:


pyiron_base/generic/filedata.py
```diff
@@ -83,55 +84,74 @@ def load_file(filename, filetype=None, project=None):
...
     else:
-        return _load_default(filename)
+        return _load_default(fp, is_file=filename_is_str)
+
+
+class FileDataTemplate(ABC):
+    @property
+    def data(self):
+        """Return the associated data."""
+        raise NotImplementedError
```
pmrv:
```suggestion
    @property
    @abstractmethod
    def data(self):
        """Return the associated data."""
        pass  
```
Deriving from `ABC` only checks methods that are marked with `abstractmethod`.

pyiron_base/generic/filedata.py
```diff
@@ -83,55 +84,78 @@ def load_file(filename, filetype=None, project=None):
...
     """
-    def _load_txt(file):
-        with open(file, encoding='utf8') as f:
-            return f.readlines()
+    def _load_txt(file, is_file):
+        if is_file:
+            with open(file, encoding='utf8') as f:
+                return f.readlines()
+        else:
+            return file.readlines()
```
pmrv: I'm not sure the extra flag is needed, maybe something like

```suggestion
        if instance(file, io.IOBase):
            return file.readlines()
        else:
            with open(file, encoding='utf8') as f:
                return _load_txt(f)
```
is better.

pyiron_base/generic/filedata.py
```diff
@@ -83,55 +84,78 @@ def load_file(filename, filetype=None, project=None):
...
+        return _load_ipynb(fp)
     elif _has_imported['PIL'] and filetype in Image.registered_extensions():
-        return _load_img(filename)
+        return _load_img(fp)
     else:
-        return _load_default(filename)
+        return _load_default(fp, is_file=filename_is_str)
+
+
+class FileDataTemplate(ABC):
```
pmrv, line 148: I didn't see it before, but where else is the `FileDataTemplate` supposed to be used?

pyiron_base/generic/filedata.py
niklassiemer: Probably we should then go the whole way and use this check instead of generating the `filename_is_string` at all.

pyiron_base/generic/filedata.py
niklassiemer: I hope to have it as a Base class for storing File related data. I am using this also for the `S3FileData` [here](https://github.com/pyiron/pyiron_contrib/blob/5e4106f17e4ad0f5f2899df383e9d8b16753f806/pyiron_contrib/generic/s3io.py#L24).

PR #384, opened 2021-07-26
Enable to remove a job with loading the full job object.

niklassiemer:
For external storage types, we would need to remove the job using the fully loaded job object with the information how to reach the external storage. Therefore, I added a flag at the hdf level which is used to trigger a full load of the job.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -6,12 +6,15 @@ The JobCore the most fundamental pyiron job class.
 """
 
 import copy
+import math
 import os
 import posixpath
-import math
-from pyiron_base.settings.generic import Settings
+import shutil
+import warnings
+
+from tables import NoSuchNodeError
+
 from pyiron_base.interfaces.has_groups import HasGroups
-from pyiron_base.generic.util import static_isinstance
 from pyiron_base.job.util import \
     _get_project_for_copy, \
     _copy_database_entry, \
@@ -27,9 +30,7 @@ from pyiron_base.job.util import \
     _job_delete_files, \
     _job_delete_hdf, \
     _job_remove_folder
-from tables import NoSuchNodeError
-import shutil
-import warnings
+from pyiron_base.settings.generic import Settings
 
 __author__ = "Jan Janssen"
 __copyright__ = (
@@ -429,6 +430,23 @@ class JobCore(HasGroups):
         internal function to remove command that removes also child jobs.
         Do never use this command, since it will destroy the integrity of your project.
         """
+        # Check if the job requires to be removed from the full object (This is the case for external Storage)
+        print(f"Enter base/job/core/JobCore.remove_child with self={type(self)} with path={self.path}")
+        try:
+            print("try")
+            requires_full_object = self.get('REQ_OBJ_RM')
+            print("Found!")
+        except ValueError as e:
+            print(e)
+            requires_full_object = False
+
+        print("Check if requires full object:")
+        if requires_full_object:
+            print("require full obj!")
+            job = self.to_object()
+            job.remove_child()
+
+        print('Go on with unchanged implementation')
         # Delete job from HPC-computing-queue if it is still running.
         job_status = str(self.status)
         if (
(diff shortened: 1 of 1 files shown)


pyiron_base/job/core.py
```diff
@@ -429,6 +430,16 @@ def remove_child(self):
         internal function to remove command that removes also child jobs.
         Do never use this command, since it will destroy the integrity of your project.
         """
+        # Check if the job requires to be removed from the full object (This is the case for external Storage)
+        try:
+            requires_full_object = self.get('REQ_OBJ_RM')
+        except ValueError as e:
+            requires_full_object = False
```
jan-janssen: I am not a fan of `try` and `except`, you could just check if the key exists in the HDF5 file

niklassiemer: True, will change this 👍

pmrv:
```suggestion
        requires_full_object = self.get('REQ_OBJ_RM', default=False)
```
should work.

I'd like more descriptive name for the flag, though.

pyiron_base/job/core.py
```diff
@@ -429,6 +430,16 @@ def remove_child(self):
...
+
+        if requires_full_object:
+            job = self.to_object()
+            job.remove_child()
+
```
pmrv: Is the normal case still necessary after calling `remove_child()` or could it try to delete certain things twice?

pmrv (approved):
lgtm, but there should be some mention that this is a work-around for the RDM job, so we don't forget to throw it out later again.

pyiron_base/job/core.py
niklassiemer: It should not be needed, since I actually call `super()` as well. However, all remove functions here don't mind if things are not there.

niklassiemer:
Thanks a lot. I really like the get with default version! Any suggestions for a nice flag name?
`REQUIRES_FULL_OBJECT_FOR_REMOVAL` is quite lengthy...

niklassiemer:
> Thanks a lot. I really like the get with default version! Any suggestions for a nice flag name?
> `REQUIRES_FULL_OBJECT_FOR_REMOVAL` is quite lengthy... 

Well the `get` seems not to work as expected... Maybe we should implement that behavior? 😅

pmrv:
> > Thanks a lot. I really like the get with default version! Any suggestions for a nice flag name?
> > `REQUIRES_FULL_OBJECT_FOR_REMOVAL` is quite lengthy...
> 
> Well the `get` seems not to work as expected... Maybe we should implement that behavior? sweat_smile

Ah, didn't read carefully enough yesterday. The `get()` with default behavior exists on `ProjectHDFio`, since you want to look up the flag there, I think you can just change `self.get` to `self.project_hdf5.get`.

pyiron_base/job/core.py
pmrv: I just have some trouble following the control flow here now.  As far as I can tell now

1. `Project.remove_jobs()` inspects jobs and calls `JobPath.remove_child()`.
2. This calls the functions modified here, which loads the job and calls `RDMStorage.remove_child()` to do the actual deletion of the data in S3.
3. But presumably that also calls `super().remove_child()`, which propagates back to `JobPath.remove_child()`, so we go back to 2.

Am I missing where the recursion is broken?  Also because of that `super()` call I'd prefer a 
```python
if requires_full_object:
  ...
else:
  ...
```

niklassiemer: I am also happy with the if/else. However, I after all have to break the recursion 'by hand'. Since, as long as the `REQUIRE_FULL_OBJ_FOR_RM` Flag is `True` in the hdf, we would loop through the same code over and over again. Therefore I use
```
def remove_child(self):
   remove_s3_storage_access()
   self['REQUIRE_FULL_OBJ_FOR_RM']=False
   super().remove_child()
```
to break this.

Reading my own answer, I think it would be easier to have a `_before_generic_remove_child` method on the job which is called instead of `job.remove_child()`. This breaks the recursion and is much cleaner.

```suggestion
        if requires_full_object:
            job = self.to_object()
            job._before_generic_remove_child()
            
```

pmrv: Sounds good!

niklassiemer:
Great, thanks a lot for the reviews! I will merge this once the CI (with integration tests) passes 👍

PR #385, opened 2021-07-27
Add project browser from pyiron_gui

niklassiemer:


pyiron_base/project/generic.py
```diff
@@ -182,6 +183,17 @@ def data(self):
                 pass
         return self._data
 
+    @property
+    def browser(self):
+        try:
+            from pyiron_gui import ProjectBrowser
```
jan-janssen, line 189: I am not so happy with `pyiron_base` importing `pyiron_gui` that feels like spaghetti code to me. Can not the issue be solved by creating a `Project` class in `pyiron_gui` ?

niklassiemer: That's what currently is done in the [`activate_gui` function](https://github.com/pyiron/pyiron_gui/blob/3df43717c29e0de4c0d212a004bee9870a5cdc62/pyiron_gui/project/project.py#L20) in the `pyiron_gui` repo - even for all possible `Projects` (base, atomisitics, ...). However, it was requested in the ADIS meeting yesterday, that everything should be possible with a plain `from pyiron(_module) import Project`, i.e. without further imports.

pmrv: It's also a bit weird, because you'd have to do `from pyiron_gui import ProjectAtomisticGui`, `from pyiron_gui import ProjectContribGui`, etc.

niklassiemer:
@jan-janssen can you live with some 'spaghettization' of the code in this way?

jan-janssen:
Can not we use the `ImportAlarm` from @liamhuber here? Like https://github.com/pyiron/pyiron_atomistics/blob/master/pyiron_atomistics/gpaw/gpaw.py#L24

pyiron_base/project/generic.py
liamhuber: IMO it's better to break "everything comes from 'from pyiron_x import foo'" than it is to make spaghetti of the code. I just stopped by to correct the attribution of 'ImportAlarm" from myself to Marvin, so I may be missing something. If it's really user pattern vs code structure, I strongly prefer to maintain the atypical use pattern in order to preserve a more sensible codebase. I'm confident we can get a solution that does both in the future.

niklassiemer:
The idea of the `ImportAlarm` is to alarm a user of a potentially missing module further downstream, i.e. fail (in this case warn) as early as possible. However, we do certainly not want to warn on import of `Project`, especially, if this feature is just a nice display and no added functionality. Therefore, raising a nice error message on trying to access the `browser` would be the best option we have.

niklassiemer:
I convert this PR to draft until we discussed this with everyone in one of the upcoming pyiron meetings, probably the one on 16.08. with Liam back on board :)

freyso:
Have you considered doing a fixup of the existing project class(es) by the pyiron_gui module? If this is only about adding functionality (add new functions, no rewrite), it is safe to define
```python
def great_new_feature(self, ...):
   ...
```
and then do in the pyiron_gui `__init__`
```
setattr (pyiron.Project, 'new_function_name', great_new_feature)
```
at any point after the initial `import pyiron`.

There will be no spaghetti code, and it is pretty obvious in the extension module how it extends the pyiron functionality. It will be somewhat harder, of course, to figure out for developers where a certain functionality after multiple extension imports is implemented, although this is nothing pyiron is particularly good at anyway.

From the users' perspective, it's transparent: if you want to have the "browse ()" feature in your projects, import the pyiron_gui module in addition. Not everyone will need it.

pmrv:
> Have you considered doing a fixup of the existing project class(es) by the pyiron_gui module? If this is only about adding functionality (add new functions, no rewrite), it is safe to define
> 
> ```python
> def great_new_feature(self, ...):
>    ...
> ```
> 
> and then do in the pyiron_gui `__init__`
> 
> ```
> setattr (pyiron.Project, 'new_function_name', great_new_feature)
> ```
> 
> at any point after the initial `import pyiron`.
> 
> There will be no spaghetti code, and it is pretty obvious in the extension module how it extends the pyiron functionality. It will be somewhat harder, of course, to figure out for developers where a certain functionality after multiple extension imports is implemented, although this is nothing pyiron is particularly good at anyway.
> 
> From the users' perspective, it's transparent: if you want to have the "browse ()" feature in your projects, import the pyiron_gui module in addition. Not everyone will need it.

Imo monkey-patching this way is the least dev-friendly way of doing it, because it means I have to manually trace what methods exist at run time while debugging.  It might also lead to all kinds of weirdness with import order as we used to have with monkey-patching `Atoms` from ASE.

Personally I find @niklassiemer solution quite pragmatic, especially since it concerns a small localized change only, i.e. it's unlikely we'll start adding dozens of different types of browsers to all our projects. 

If we really, really don't want this I'd rather add a function to `pyiron_gui`, like
```python
def load_gui(project):
  return ProjectBrowser(project, ...)
```
or even
```python
def add_gui(project):
  class GuiProject(project.__class__, ProjectBrowser):
    ...
  return GuiProject(project.path)
```
similar to what @niklassiemer did elsewhere.  For using you'd then have to import explicitly from `pyiron_gui`
```python
from pyiron_gui import load_gui
load_gui(pr)
```
which we usually don't want, aber naja...


There's one way I can see going in @freyso's direction, but then really committing to it.  All pyiron related subproject seem to want their own project features at some point, atomistics adds the structure creators, continuum will want 3D models at some point, gui wants a GUI, ontology?  Embracing this we could add well-defined hooks where subprojects could register new functionality with `pyiron_base.Project` without sub-classing at all.  Small sketch for the creators:

in pyiron_base
```python
class Creator:
  _registry = []
  @classmethod
  def register(cls, name, creator):
    cls._registry[name] = creator
  def __getattr__(self, name):
    return self._registry[name]
  def __dir__(self):
    return self._registry.keys()

class Project:
  create = Creator()
  @classmethod
  def register_creator(cls, name, creator):
    cls.create._register(name, creator)
```
and then in atomistics
```python
import pyiron_base
from ... import AtomisticFactory # not sure what the correct input is right now

pyiron_base.Project.register_creator('structure', AtomisticFactory())
```
and then in gui
```python
import pyiron_base
from ... import ProjectBrowserCreator

pyiron_base.Project.register_creator('gui', ProjectBrowserCreator())
```
etc.

`register_creator` is a convenience method and could be dropped for `Project.creator.register`.

This would have all the advantages @freyso mentions, but none of the developer drawbacks, because the `Creator` knows which pyiron subprojects have registered things with it and this is therefore transparent to the developer stepping through the code while debugging.

(It's debatable whether the gui should live on the creator, but I hope it's clear enough how this could be generalized to other things living on `Project`.)

freyso:
As for the developers convenience: I found the pyiron flow of control extraordinarily difficult to follow. I now mostly use the inspect module. inspect.getsource/getsourcefile/getsourceline gives me quite detailed information of how and where a particular existing function or class is implemented.

In the end, it's a question of code philosophy. Patching up existing classes in selected cases (=extra functionality is rarely used, and garantueed not to interfere with the main (or other pluggable) functionality of the class) adds a good deal of flexibility and speeds up development (no boilerplate to integrate). But it cannot serve as a general-purpose coding pattern unless you want to drive people crazy.
 
As for the spaghetti question: From my experience with the SPHInX code, it is essential to have some conceptual layering of the code base, i.e., to know which part of the code relies on what, and how generic functionality is plugged into the basis from derived classes. Putting any dependency into something called "_base" should ring the alarm bells (and apparently does for you, too). I spent quite a lot of time cleaning such things up in SPHInX, and it was a nightmare. If you get that conceptually right, however, it becomes much easier to decide where a certain new functionality belongs.

Anyway, I had my say on this and won't interfere any longer.

pmrv:
Few sketches from today's discussion

## Have register method

```python
class Project:
  _representation = {}
  @classmethod
  def register_repr(cls, name, func):
    # func takes a project instance and must return a IPython.display.Display
    cls._representation[name] = func
  def _ipython_display_(self): # or whatever it is named
    if 'gui' in self._representation:
      return display(self._representation['gui'](self))
    else:
      ... # return the text repr
```
then in `pyiron_gui.__init__.py`
```python
from ... import ProjectBrowser

Project.register_repr('gui', ProjectBrowser)
```

If `pyiron` imports `pyiron_gui` automatically, then as a user you never have to touch it.  If it exists then the GUI will be there, if not then not.

## From other project
https://github.com/dask/dask-cloudprovider/blob/main/dask_cloudprovider/utils/logs.py 

```python
class PyironObject:
  ...
  def _widget(self):
    ...
  
  def _ipython_display(self):
    display(self._widget())

class Projec(PyironObject):
  ...
  def _widget(self):
    import pyiron_gui
    return pyiron_gui.ProjectBrowser(self)
```

Still the spaghetti imports, but pretty simple from a usage perspective and also invisible if not run in a notebook.

## Assorted Comments

- we didn't want to register widgets to `pyiron_atomistics`, `pyiron_contrib` objects, because for this to work `pyiron_gui` would have to depend on all pyiron repositories.
- probably don't want to have very complicated guis by default to keep things responsive
- instead of GUIs by default, power users might want to have simpler text/html based output, this should be configured?
  - i.e. in .pyiron: `STANDARD_VIEW = HTML`
  - and objects offer a range of represenations: `pr.representation.html/gui/text`?
- mid-term merge pyiron_gui classes directly into pyiron sub-repos again and develop them inline
- for now borrow from @freyso and monkey-patch `_ipython_display_` methods on all objects pyiron_gui supports now (on import) so that it's easy for us devs to play around with it
- cool GUI idea: `pr.create` should return drop-down menu as could `Atoms.analyze`.

Issue #232, opened 2021-03-11
20 character limit for job type causing DB error


raynol-dsouza:
All the protocol jobs started throwing a DB related error during job submission (on the head node) which looks like this:

```
During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-6-6cf202eba93f> in <module>
     39     # fts_job.format_whitelist()
     40 
---> 41     fts_job.run()

~/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/generic.py in run(self, delete_existing_job, repair, debug, run_mode, continue_run)
    650             self.status.created = True
    651         super(CompoundVertex, self).run(delete_existing_job=delete_existing_job, repair=repair, debug=debug,
--> 652                                         run_mode=run_mode)
    653 
    654     def collect_output(self):

~/pyiron_repos/pyiron_base/pyiron_base/generic/util.py in decorated(*args, **kwargs)
    211                         stacklevel=2
    212                     )
--> 213             return function(*args, **kwargs)
    214         return decorated
    215 

~/pyiron_repos/pyiron_base/pyiron_base/job/generic.py in run(self, delete_existing_job, repair, debug, run_mode, run_again)
    667                 self._run_if_repair()
    668             elif status == "initialized":
--> 669                 self._run_if_new(debug=debug)
    670             elif status == "created":
    671                 self._run_if_created()

~/pyiron_repos/pyiron_base/pyiron_base/job/generic.py in _run_if_new(self, debug)
   1327             print("job exists already and therefore was not created!")
   1328         else:
-> 1329             self.save()
   1330             self.run()
   1331 

~/pyiron_repos/pyiron_base/pyiron_base/job/generic.py in save(self)
   1174         """
   1175         self.to_hdf()
-> 1176         job_id = self.project.db.add_item_dict(self.db_entry())
   1177         self._job_id = job_id
   1178         self.refresh_job_status()

~/pyiron_repos/pyiron_base/pyiron_base/database/generic.py in add_item_dict(self, par_dict)
    411                 return result
    412             except Exception as except_msg:
--> 413                 raise ValueError("Error occurred: " + str(except_msg))
    414         else:
    415             raise PermissionError("Not avilable in viewer mode.")

ValueError: Error occurred: (psycopg2.errors.StringDataRightTruncation) value too long for type character varying(20)

[...]

PR #386, opened 2021-07-28
Increase chemical formula limit to 50 characters

max-hassani:
I have increase the character limit of chemical formula to 50 in our database, therefore I changed the limit in the code as well.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -102,7 +102,7 @@ class DatabaseAccess(object):
         except Exception as except_msg:
             raise ValueError("Connection to database failed: " + str(except_msg))
 
-        self._chem_formula_lim_length=30
+        self._chem_formula_lim_length = 50
         self.__reload_db()
         self.simulation_table = Table(
             str(table_name),


pmrv:
I've double-checked the limit is adjusted in database, so looks good to me.  Since we have the staging db now, we probably also want to come back to https://github.com/pyiron/pyiron_base/issues/232 and bump `hamilton` also to an even 50.  The other columns that are shorter that 50 are: `status`, `username` and I'd think they can stay that way.

EDIT: Ah, I remember now, we can't update the columns without recreating the view, right?  Naja, next time the database restarts we should get #232 in, I had just forgotten about it until now.

pmrv (approved):

max-hassani:
> 
> 
> I've double-checked the limit is adjusted in database, so looks good to me. Since we have the staging db now, we probably also want to come back to #232 and bump `hamilton` also to an even 50. The other columns that are shorter that 50 are: `status`, `username` and I'd think they can stay that way.



> 
> 
> I've double-checked the limit is adjusted in database, so looks good to me. Since we have the staging db now, we probably also want to come back to #232 and bump `hamilton` also to an even 50. The other columns that are shorter that 50 are: `status`, `username` and I'd think they can stay that way.
> 
> EDIT: Ah, I remember now, we can't update the columns without recreating the view, right? Naja, next time the database restarts we should get #232 in, I had just forgotten about it until now.

Yesterday, I tested the change in the staging database. It did not require any restart of the database and was effective instantly. Just a few lines of sql commands sufficed. 
```sql
UPDATE pg_attribute SET atttypmod = 50+4  
WHERE attrelid IN ('jobs_cmmc_admin'::regclass, 'jobs_cmmc'::regclass)
AND attname = 'chemicalformula';
```
This updates the column in both the view and the table simultaneously, therefore there is no need to remove and recreate the view. 
Similarly, I will change the limit for Hamilton column as well. I will report the update in the corresponding issue. But this will require no further changes in the code, am I right? If not, I will merge this one.

pmrv:
> This updates the column in both the view and the table simultaneously, therefore there is no need to remove and recreate the view.

Great news! Is this noted in the cluster maintenance wiki yet? 

> Similarly, I will change the limit for Hamilton column as well. I will report the update in the corresponding issue. But this will require no further changes in the code, am I right? If not, I will merge this one.

For the `hamilton` column only the wrapper has to be updated [here](https://github.com/pyiron/pyiron_base/blob/064f1eb985fdaff49b1426266a7f1dc0aac0cd0c/pyiron_base/database/generic.py#L119) I think.  This one is ok to merge though.

PR #388, opened 2021-07-29
Run integration test on linux, windows, and mac os (only python 3.9)

niklassiemer:


niklassiemer:
@pmrv I expected the windows tests to fail (run forever). However, I did not expect the contrib tests to fail. Are some changes to the flattened storage not yet taken into account in contrib?

pmrv:
Yeah,  I think I borked the HDF of `FlattenedStorage`, will push PRs today.

pmrv:
See #390, that fixes the contrib failure in my local tests.

PR #389, opened 2021-07-30
Fix HDF of FlattenedStorage

pmrv:
Before both per element and per chunk arrays were stored in one HDF group, which meant that on reading I had to guess which are which.  In the case of storages that have only have chunks of length one both types of arrays have the same overall length, so the code got confused.  I fixed that now by storing them in different HDF groups.

@Leimeroth Since this changes HDF group names it might interact with the atomicrex classes.  Loading old versions should be compatible, but please check if there's other problems.

### DIFF
--- a/pyiron_base/generic/flattenedstorage.py
+++ b/pyiron_base/generic/flattenedstorage.py
@@ -487,8 +487,13 @@ class FlattenedStorage:
     def from_hdf(self, hdf, group_name="flat_storage"):
         with hdf.open(group_name) as hdf_s_lst:
             version = hdf_s_lst.get("HDF_VERSION", "0.0.0")
-            num_chunks = hdf_s_lst["num_chunks"] or hdf_s_lst["num_structures"]
-            num_elements = hdf_s_lst["num_elements"] or hdf_s_lst["num_atoms"]
+            try:
+                num_chunks = hdf_s_lst["num_chunks"]
+                num_elements = hdf_s_lst["num_elements"]
+            except KeyError:
+                num_chunks = hdf_s_lst["num_structures"]
+                num_elements = hdf_s_lst["num_atoms"]
+
             self._num_chunks_alloc = self.num_chunks = self.current_chunk_index = num_chunks
             self._num_elements_alloc = self.num_elements = self.current_element_index = num_elements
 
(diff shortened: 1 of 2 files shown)


pyiron_base/generic/flattenedstorage.py
```diff
@@ -469,41 +467,76 @@ def to_hdf(self, hdf, group_name="flat_storage"):
...
+            def read_array(name, hdf):
+                a = np.array(hdf[name])
+                if a.dtype.char == "S":
+                    # if saved as bytes, we wrote this as an encoded unicode string, so manually decode here
+                    # TODO: string arrays with shape != () not handled
+                    a = np.array([s.decode("utf8") for s in a],
+                                # itemsize of original a is four bytes per character, so divide by four to get
+                                # length of the orignal stored unicode string; np.dtype('U1').itemsize is just a
+                                # platform agnostic way of knowing how wide a unicode charater is for numpy
+                                dtype=f"U{a.dtype.itemsize//np.dtype('U1').itemsize}")
```
Leimeroth: I didn't test anything yet, but why do you define read_array and write_array inside other functions? Does

pmrv: I nested the functions, because I don't want them as methods, since they a) don't access the instance and b) would bloat the class API.  Alternatively I could make them private top-level functions, but then the definition is far away from the usage and it's more difficult to read.  Creating nested functions is [not](https://stackoverflow.com/questions/14055843/what-is-the-performance-overhead-of-nested-functions) for [free](https://stackoverflow.com/questions/7839632/is-there-an-overhead-when-nesting-functions-in-python/7839697#7839697), but it's also not expensive.  If it ever comes up in a profile, we can still pull it out.

niklassiemer: I am not against defining the function in a nested way, however, I would prefer to have function definitions at the beginning of the main function and not being nested again in the `with open... as f` context.

Leimeroth (approved):
I did some tests and as far as I can tell everything still works. From my side this can be merged.

jan-janssen:
Does this resolve the issue in https://github.com/pyiron/pyiron_contrib/pull/214

niklassiemer:
Does not seem to resolve it based on the integration tests

niklassiemer:
I would like to bring up https://github.com/pyiron/pyiron_contrib/issues/215 again. Is this related to changes here?

pmrv:
No, the problem should be addressed by #390, but I haven't gotten around to that yet.

niklassiemer:
Thanks a lot for the clarification.

PR #390, opened 2021-07-30
Strip first axis when creating array in add_chunk

pmrv:
Previously this was only checked when the array already exists.  Adapted
tests to cover this as well.

This is also related to the error in
https://github.com/pyiron/pyiron_contrib/pull/197

Should fix the current failure in https://github.com/pyiron/pyiron_base/pull/388

### DIFF
--- a/pyiron_base/generic/flattenedstorage.py
+++ b/pyiron_base/generic/flattenedstorage.py
@@ -431,8 +431,13 @@ class FlattenedStorage:
                 if len(a.shape) > 0 and a.shape[0] == n:
                     self.add_array(k, shape=a.shape[1:], dtype=a.dtype, per="element")
                 else:
-                    self.add_array(k, shape=a.shape, dtype=a.dtype, per="chunk")
-            # if the first axis was added by the caller to force to add a per chunk array, remove it again here
+                    shape = a.shape
+                    # if the first axis was added by the caller to force to add a per chunk array, remove it again here
+                    if len(shape) > 0 and a.shape[0] == 1:
+                        shape = shape[1:]
+                    self.add_array(k, shape=shape, dtype=a.dtype, per="chunk")
+            # same as above: if the first axis was added by the caller to force to add a per chunk array, remove it
+            # again here
             if k in self._per_chunk_arrays and len(a.shape) > 0 and a.shape[0] == 1:
                 a = a[0]
             self.set_array(k, self.current_chunk_index, a)
--- a/tests/generic/test_flattenedstorage.py
+++ b/tests/generic/test_flattenedstorage.py
@@ -146,6 +146,10 @@ class TestFlattenedStorage(unittest.TestCase):
             self.fail("add_chunk should not raise an exception when passed a value for an existing per-chunk array.")
         self.assertTrue(np.array_equal(val, cont.get_array("perchunk", 0)),
                         "add_chunk did not remove first axis on a per chunk array!")
+        # test the same, but now let the array be created by add_chunk, instead of doing it on our own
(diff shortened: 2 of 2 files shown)


niklassiemer:
Are your local tests with the master of contrib? Since this should be tested here by the integration tests.

pmrv:
Yeah, I'm pretty sure it works locally with contrib's master branch, I'll have a look at the workflow again. o.O

pmrv:
It's weird, I checked out the latest contrib master and this branch on the cluster and it works without issue, just on the CI there's somehow a problem..

pmrv:
@niklassiemer Is it possible that [this line](https://github.com/pyiron/pyiron_base/blob/71e59b540bfb6061c41de66a1810b6b8587aec95/.github/workflows/contrib-compat.yml#L25) overrides the local pyiron_base version with the latest published one? :S

niklassiemer:
@pmrv This is concerning the checkout if the repo which is calling this itself, thus this `environment.yml` is installing the latest dependencies of `pyiron_base` as fetched from this branch. In the Setup step we then clone the `pyiron_contrib` exclude the `pyiron_base` dependency of contrib and install `pyiron_contrib` before we start the contrib unit tests. Thus the workflow should be fine.

pmrv:
What I did now

```bash
git clone https://github.com/pyiron/pyiron_base
cd pyiron_base
mamba -n pb -f .ci_support/environment.yml
mamba activate pb
pip install --no-deps .
cd ..
git clone https://github.com/pyiron/pyiron_contrib
cd pyiron_contrib
grep -v "pyiron_base" .ci_support/environment.yml > environment.yml
mambda env update -f environment.yml
pip install --no-deps .
python -m unittest discover tests
```
on my local machine, i.e. exactly what the workflow is supposed to be doing and the tests pass fine. I'm not sure why the workflow breaks, but I'm sure that this PR passes the tests.

niklassiemer:
@pmrv I tried to set up a similar workflow and I also locally get the error! (I assume your code snipped above is from memory and you have also switched the branch after checkout.)

pmrv:
Agree, there were some typos in the script above.  I've tested this now

```bash
git clone https://github.com/pyiron/pyiron_base
cd pyiron_base
micromamba create -n pb -f .ci_support/environment.yml
eval "$(micromamba shell hook -s zsh)"
micromamba activate pb
git cd shapefix
pip install --no-deps .
cd ..
git clone https://github.com/pyiron/pyiron_contrib
cd pyiron_contrib

grep -v "pyiron_base" .ci_support/environment.yml > environment.yml

micromamba env update -f environment.yml
pip install --no-deps .
python -m unittest discover tests
```
and indeed it fails.  However, I have checked in a debugger that this runs the old code and not the code from this PR for the contrib tests.  Additionally, if I then do 
```bash
cd ../pyiron_base
pip install --no-deps .
cd ../pyiron_contrib
python -m unittest discover tests
```
they do pass.  Also checking the output from mamba gives "ClobberWarnings" for pyiron_base related files.  All of this suggests to me  that the workflow does overwrite the pyiron_base version.  My guess is that this comes from installing `pyiron_atomistics` which will pull `pyiron_base` in again.

niklassiemer:
Sounds reasonable! Maybe we have to do it the other way round... installing the full `pyiron_module` to test and afterwards `conda remove --force -y pyiron_base` and then checkout of the present branch?...

niklassiemer:
There we go, the contrib compatibility passes 🎉 Thanks @pmrv for looking into the workflow again!

jan-janssen (approved):
Looks good to me

PR #394, opened 2021-08-02
Correct unpacking

### LINKED ISSUE #393: Project.pack and unpack do not handle paths well
Currently, the  `unpack()` function on the `Project` does not handle paths well or even ignores those:

```
import os

pr = Project('test')
job = pr.create.job.Lammps('test')
job.structure = pr.create.structure.bulk("Fe")
job.run()

os.listdir()
>>> ['test']

pr_nested = pr.open('nested')
pr.pack(destination_path='target', csv_file_name='export.csv')

os.listdir()
>>> ['export.csv', 'target.tar.gz', 'test']

pr_imp = pr_nested.open('imported')
pr_imp.unpack(origin_path='target', csv_file_name="export.csv")
pr_imp
>>> {'groups': [], 'nodes': []}

os.listdir()
>>> ['export.csv', 'imported', 'target.tar.gz', 'test']
```
Thus, the unpack function did _not_ unpack to the destination project but just at the current working directory!
Similar, unpack does not work if it is given a path from where to unpack the packed project:
```
pr.pack(destination_path='exported/target2', csv_file_name='exported/export2.csv')
pr_imp.unpack(origin_path='exported/target2', csv_file_name='exported/export2.csv')
>>> FileNotFoundError: [Errno 2] No such file or directory: 'target2.tar.gz'
```

max-hassani:
With this PR, the issue #393 should be solved.
Basically:
- wherever the archived project is stored, the unpacking should work now
- and copying the files to the destination project is handled correctly

### DIFF
--- a/pyiron_base/archiving/import_archive.py
+++ b/pyiron_base/archiving/import_archive.py
@@ -41,6 +41,10 @@ def import_jobs(
     project_instance, archive_directory,
     df, compressed=True
 ):
+    if archive_directory[-7:] == ".tar.gz":
+        archive_directory = archive_directory[:-7]
+        if not compressed:
+            compressed = True
     # Copy HDF5 files
     # if the archive_directory is a path(string)/name of the compressed file
     if static_isinstance(
@@ -62,7 +66,7 @@ def import_jobs(
     # destination folder
     des = project_instance.path
     # source folder; archive folder
-    src = os.path.abspath(os.path.join(os.path.dirname(archive_directory), os.path.basename(archive_directory)))
+    src = os.path.abspath(archive_directory)
     copy_tree(src, des)
     if compressed:
         rmtree(src)
--- a/pyiron_base/archiving/import_archive.py
+++ b/pyiron_base/archiving/import_archive.py
@@ -60,8 +60,7 @@ def import_jobs(
         extract_archive(archive_directory)
     archive_name = getdir(path=archive_directory)
     # destination folder
-    des = os.path.abspath(project_instance.path)
-
+    des = project_instance.path
     # source folder; archive folder
     src = os.path.abspath(os.path.join(os.path.dirname(archive_directory), os.path.basename(archive_directory)))
     copy_tree(src, des)
--- a/pyiron_base/archiving/import_archive.py
+++ b/pyiron_base/archiving/import_archive.py
@@ -41,10 +41,6 @@ def import_jobs(
     project_instance, archive_directory,
     df, compressed=True
 ):
-    if archive_directory[-7:] == ".tar.gz":
-        archive_directory = archive_directory[:-7]
-        if not compressed:
-            compressed = True
     # Copy HDF5 files
     # if the archive_directory is a path(string)/name of the compressed file
     if static_isinstance(
@@ -54,7 +50,12 @@ def import_jobs(
         ]
     ):
         archive_directory = archive_directory.path
-    elif not isinstance(archive_directory, str):
+    elif isinstance(archive_directory, str):
+        if archive_directory[-7:] == ".tar.gz":
+            archive_directory = archive_directory[:-7]
+            if not compressed:
+                compressed = True
+    else:
         raise RuntimeError(
             """the given path for importing from,
             does not have the correct format paths
--- a/pyiron_base/archiving/import_archive.py
+++ b/pyiron_base/archiving/import_archive.py
@@ -62,8 +62,10 @@ def import_jobs(
             as string or pyiron Project objects are expected"""
         )
     if compressed:
-        extract_archive(archive_directory)
+        extract_archive(os.path.relpath(archive_directory, os.getcwd()))
+
     archive_name = getdir(path=archive_directory)
+
     # destination folder
     des = project_instance.path
     # source folder; archive folder
@@ -74,6 +76,7 @@ def import_jobs(
 
     # # Update Database
     pr_import = project_instance.open(os.curdir)
+
     df["project"] = [os.path.join(
         pr_import.project_path, os.path.relpath(p, archive_name)) + "/"
         for p in df["project"].values
--- a/pyiron_base/archiving/import_archive.py
+++ b/pyiron_base/archiving/import_archive.py
@@ -38,22 +38,19 @@ def extract_archive(archive_directory):
 
 
 def import_jobs(
-    project_instance, directory_to_import_to, archive_directory,
+    project_instance, archive_directory,
     df, compressed=True
 ):
     # Copy HDF5 files
     # if the archive_directory is a path(string)/name of the compressed file
-    if isinstance(archive_directory, str):
-        archive_directory = os.path.basename(archive_directory)
-    # if the archive_directory is a project
-    elif static_isinstance(
+    if static_isinstance(
         obj=archive_directory.__class__,
         obj_type=[
             "pyiron_base.project.generic.Project",
         ]
     ):
         archive_directory = archive_directory.path
-    else:
+    elif not isinstance(archive_directory, str):
         raise RuntimeError(
             """the given path for importing from,
             does not have the correct format paths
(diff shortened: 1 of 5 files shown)


max-hassani:
From the debug messages in the test logs:
```
 origin_path
> /home/runner/work/pyiron_base/pyiron_base/archive_folder_comp
 os.listdir(os.path.dirname(origin_path))
> ['archive_folder', 'CONTRIBUTING.rst', 'setup.py', 'CODE_OF_CONDUCT.md', 'imported', 'specs', '.gitattributes', 'MANIFEST.in', '.coveragerc', 'LICENSE', '.gitignore', 'export.csv', '.coveralls.yml', '.git', 'archive_folder_comp.tar.gz', '.readthedocs.yml', 'binder', 'versioneer.py', 'notebooks', '.ci_support', 'pyiron_base', 'docs', '.gitpod', '.github', 'pyiron.log', 'README.rst', 'tests', 'test', 'setup.cfg']
```
So in [this line](https://github.com/pyiron/pyiron_base/blob/39f5ead9e471f17aaf98db8dfe36f8579b3f5e3b/tests/archiving/test_import.py#L116), the call to the method should be rewritten as : 
```
  pr_imp.unpack(origin_path=os.path.join(origin_path, self.arch_dir_comp),
                csv_file_name=os.path.join(cwd, 'export.csv'))
```
Since you can see above that `archive_folder_comp.tar.gz` is placed in the `origin_path`.
I think this will resolve the failing unittest. @niklassiemer, do you agree?

niklassiemer:
No, in the dirname of origin path the file resides. Thus origin path should be the correct path? Of course this is not pointing to an actual file since the `.tar.gz` extension is missing.

max-hassani:
> 
> 
> No, in the dirname of origin path the file resides. Thus origin path should be the correct path? Of course this is not pointing to an actual file since the `.tar.gz` extension is missing.

But the method assumes that you give the path to the file and only the extension `tar.gz` would be missing. If the compressed file is located in `/path_1/compressed.tar.gz`, to call this method, `orgin_path` should be equal to `/path_1/compressed`

max-hassani:
@niklassiemer,  I will try to  rewrite `test_unpack_from_other_dir(self)`, since I think with the current function even at the end when comparing the content of the compressed directory and the imported one fails. But to make sure I understood the purpose, the idea is that we have the compressed file in a directory other than the project path or current working directory and import it to an empty, nested or non-nested project. Am I right? 
For example,
```python3
pr.import(origin_path = '../exported_proj/compressed_file', csv_file_name = '../exported_proj/csvfile.csv')
```

niklassiemer:
Yes, exactly. I would add the requirement `pr.path != os.getcwd()`
Indeed, I would also like to support absolute paths.

niklassiemer:
> > No, in the dirname of origin path the file resides. Thus origin path should be the correct path? Of course this is not pointing to an actual file since the `.tar.gz` extension is missing.
> 
> But the method assumes that you give the path to the file and only the extension `tar.gz` would be missing. If the compressed file is located in `/path_1/compressed.tar.gz`, to call this method, `orgin_path` should be equal to `/path_1/compressed`

Yes, exactly, and  
```
origin_path
> /home/runner/work/pyiron_base/pyiron_base/archive_folder_comp
``` 
matches this requirement. While I actually would like to allow for names including the `.tar.gz` and then automatically setting `compressed` to `True` and not adding the extension.

max-hassani:
@niklassiemer, the new test ([here](https://github.com/pyiron/pyiron_base/blob/b17c8f5bcaf06871071e258562d469cebb149542/tests/archiving/test_import.py#L90)) should do it. Please note that I used `compressed=False` in calling `pack/unpack` methods, since this way we do not need to extract the archived file in another step, to compare the content of the destination and the source.
I agree that not including `tar.gz` in the filename is not really reasonable.  I have to modify some parts to get it working. I will re-request your review when I am done.

tests/archiving/test_import.py
```diff
@@ -64,6 +64,48 @@ def test_import_compressed(self):
...
+        print(f"pack_path_comp: {pack_path_comp}\n"
+              "pack_path_csv: {pack_path_csv}")
+        self.pr.pack(destination_path=pack_path_comp, csv_file_name=pack_path_csv, compress=False)
+        print(f"content of pack_path: {os.listdir(pack_path)}")
+        pr = self.pr.open("nested")
+        pr_imp = pr.open("imported")
+        pr_imp.unpack(origin_path=pack_path_comp, csv_file_name=pack_path_csv, compress=False)
+        compare_obj = dircmp(pack_path_comp, pr_imp.path)
+        self.assertEqual(len(compare_obj.diff_files), 0)
+
```
niklassiemer: Since we do not use the compressed file here, I would rather reflect this in the names. This confused me on reading.

```suggestion
    def test_unpack_from_other_dir(self):
        cwd = os.getcwd()
        print("DEBUG")
        print(f"cwd: {cwd}")
        pack_path = os.path.join(cwd, 'exported')
        print(f"pack_path: {pack_path}")
        os.mkdir(path=pack_path)
        pack_path_arch = os.path.join(pack_path, self.arch_dir)
        pack_path_csv = os.path.join(pack_path, 'export.csv')
        print(f"pack_path_arch: {pack_path_arch}\n"
              "pack_path_csv: {pack_path_csv}")
        self.pr.pack(destination_path=pack_path_arch, csv_file_name=pack_path_csv, compress=False)
        print(f"content of pack_path: {os.listdir(pack_path)}")
        pr = self.pr.open("nested")
        pr_imp = pr.open("imported")
        pr_imp.unpack(origin_path=pack_path_arch, csv_file_name=pack_path_csv, compress=False)
        compare_obj = dircmp(pack_path_arch, pr_imp.path)
        self.assertEqual(len(compare_obj.diff_files), 0)

```

pyiron_base/archiving/import_archive.py
```diff
@@ -62,19 +59,15 @@ def import_jobs(
...
-    if directory_to_import_to[-1] != '/':
-        directory_to_import_to = os.path.basename(directory_to_import_to)
-    else:
-        directory_to_import_to = os.path.basename(directory_to_import_to[:-1])
     # destination folder
-    des = os.path.abspath(os.path.join(os.curdir, directory_to_import_to))
+    des = project_instance.path
     # source folder; archive folder
-    src = os.path.abspath(os.path.join(os.curdir, archive_directory))
+    src = os.path.abspath(os.path.join(os.path.dirname(archive_directory), os.path.basename(archive_directory)))
```
niklassiemer: Isn't a `join` of `dirname` and `basename` identical to the input?
```suggestion
    src = os.path.abspath(archive_directory)
```

niklassiemer (approved):
Some minor things to ease reading the code but it seems to work now 😄

tests/archiving/test_import.py
max-hassani: I have updated the name of the method to reflect this. I simply added `_uncompress` to the end.

pyiron_base/archiving/import_archive.py
max-hassani: I corrected this!

max-hassani:
@niklassiemer, thanks for your comments. I also added a few lines to allow the `.tar.gz` extension in the given paths both in pack and unpack methods, and updating the `compress` flag. This minor change, allows us to keep old functionality.

max-hassani:
@niklassiemer I have updated both tests, and unpacking/packing methods. The extracting and archiving functions are very dependant to a relative path to the current working directory. The changes in the `import_archive.py` and `export_archive.py` are due to this dependence. For example, if you pass an absolute path (`/home/user/path1/archive.tar.gz`) to the archiving function, it will create `archive.tar.gz`, but when you extract it, you would get `/current_directory/home/user/path1/archive`.

niklassiemer:
@muh-hassani anything left to do here? I triggered the integration workflow, here. If it passes and you are finished here, merge it :)

niklassiemer:
The failure with `contrib` was expected (https://github.com/pyiron/pyiron_contrib/issues/215).

muh-hassani:
> 
> 
> The failure with `contrib` was expected ([pyiron/pyiron_contrib#215](https://github.com/pyiron/pyiron_contrib/issues/215)).

If it is expected, is this related to the pack/unpack function or it is irrelevant?
For pyiron_atomistics test, I see that badr package is missing.

max-hassani:
@niklassiemer, about the failure in the compatibility test with pyiron_atomisitics, `test_cached_speed` fails. As far as I checked this test has nothing to do with the packing/unpacking functionality!

niklassiemer:
That is the benchmark test which occasionally failed. Should now be solved with the extraction of the benchmark tests to a separate workflow.

max-hassani:
@niklassiemer, thanks for merging. I thought that I need to wait for the other tests to be fixed. But thanks anyway.

niklassiemer:
Well, since we fix a bug here and all the other issues are known and not introduced by us, I thought it would be nice to have it in the next release.

PR #396, opened 2021-08-04
Adding project.maintenance

max-hassani:
Introduces `Maintenance` class, which returns `get_database_statistics` as an attribute of a pyiron project. Of course, the maintenance class can be further extended to include other performance measures, like disk spaces, available memory, etc.

The use case is:
```python3
pr = Project()
pr.maintenance.database_statistics
```
this returns a dataframe including measure about postgres database statistics.

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -1519,7 +1519,7 @@ class Maintenance:
         self._local = None
 
     @property
-    def global(self):
+    def global_status(self):
         return self._global
 
 
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -18,7 +18,7 @@ from pyiron_base.project.path import ProjectPath
 from pyiron_base.database.filetable import FileTable
 from pyiron_base.settings.generic import Settings
 from pyiron_base.settings.publications import list_publications
-from pyiron_base import get_database_statistics
+from pyiron_base.database.performance import get_database_statistics
 from pyiron_base.database.jobtable import (
     get_db_columns,
     get_job_ids,
@@ -1502,7 +1502,15 @@ class Project(ProjectPath, HasGroups):
 
 
 class Maintenance:
+    """
+    The purpose of maintenance class is to provide
+    some measures of perfomance for pyiron
+    """
     def __init__(self):
+        """
+        initialize the flag self._check_postgres, to control whether pyiron is
+        set to communicate with a postgres database.
+        """
         s = Settings()
         connection_string = s._configuration['sql_connection_string']
         if "postgresql" not in connection_string:
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -1513,16 +1513,14 @@ class Maintenance:
     """
     def __init__(self):
         """
-        initialize the local and global attributes to None
+        initialize the local and global attributes
         """
-        self._global = None
+        self._global = GlobalMaintenance()
         self._local = None
 
     @property
-    def global_status(self):
-        if self._global is None:
-            self._global = GlobalMaintenance()
-            return self._global
+    def global(self):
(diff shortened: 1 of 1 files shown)


pyiron_base/project/generic.py
```diff
@@ -130,6 +131,8 @@ def __init__(self, path="", user=None, sql_query=None, default_working_directory
             self.db = FileTable(project=path)
         self.job_type = JobTypeChoice()
 
+        self.maintenance = Maintenance()
+
```
niklassiemer: I slightly changed my suggestion. This way, the `Maintenance` is only initiated if needed. This also triggers the non-postgres warning at the right place and not for everyone starting a new Project.
```suggestion
        self._maintenance = None


    @property
    def maintenance(self):
        if self._maintenance is None:
            self._maintenance = Maintenance()
        return self._maintenance
```

jan-janssen:
Is not it sufficient to import directly? like: 
```
from pyiron_base import get_database_statistics
get_database_statistics()
```
I do not see why the database statistics are related to a specific project.

max-hassani:
> 
> 
> Is not it sufficient to import directly? like:
> 
> ```
> from pyiron_base import get_database_statistics
> get_database_statistics()
> ```
> 
> I do not see why the database statistics are related to a specific project.

I agree! But Jörg required it yesterday in the technical meeting. I believe, he had the idea that the maintenance class can be extended to include more performance measures later on.

jan-janssen:
But then at least keep the same function name, rather than making it a property: 
```
pr = Project()
pr.maintenance.get_database_statistics()
```

niklassiemer:
> > Is not it sufficient to import directly? like:
> > ```
> > from pyiron_base import get_database_statistics
> > get_database_statistics()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I do not see why the database statistics are related to a specific project.
> 
> I agree! But Jörg required it yesterday in the technical meeting. I believe, he had the idea that the maintenance class can be extended to include more performance measures later on.

Yes, I raised the same argument that the database statistics is not related to a specific project. However, he really wants to be able to do everything with _one_ import statement 
```
from pyiron{_module} import Project
```
Therefore, also #385.

max-hassani:
> 
> 
> But then at least keep the same function name, rather than making it a property:
> 
> ```
> from pyiron_base import Project
> pr = Project()
> pr.maintenance.get_database_statistics()
> ```

Now, it is implemented as you suggested.

pyiron_base/project/generic.py
```diff
@@ -1510,8 +1508,24 @@ def unpack(self, origin_path, csv_file_name='export.csv', compress=True):
...
+        initialize the local and global attributes to None
+        """
+        self._global = None
+        self._local = None
+
+    @property
+    def global_status(self):
+        if self._global is None:
+            self._global = GlobalMaintenance()
+            return self._global
```
niklassiemer: I would just name the property `global` and I tend to init all Maintenance classes at once if someone is already accessing the maintenance: 
```suggestion
    def __init__(self):
        """
        initialize the local and global attributes
        """
        self._global = GlobalMaintenance()
        self._local = None

    @property
    def global(self):
        return self._global
```

max-hassani:
@niklassiemer, the reason that I chose `self.global_status` was because `global` is a keyword in python. Since the tests are failing too, I will revert it to `global_status`.

niklassiemer:
Sure, I never used the `global` and looking at this, I think I never will...

max-hassani:
@niklassiemer, if you agree, we can merge this. 
I will open another pull request to include a unittest and improve the coverage. Currently, I do not know exactly if a mocked up database connection is sufficient, which I doubt, or a running postgress database via github workflows is necessary. I have to learn more about it.

PR #397, opened 2021-08-09
A generic units class for pyiron

### LINKED ISSUE #400: Ontology for units
**Summary**

Ontology for handling unit conversions (between pyiron and code specific units) within pyiron 

**Detailed Description**

The conversion of inputs and outputs to appropriate units within pyiron is very critical and it would be nice to do this in a consistent way across different workflows.

**Further Information, Files, and Links**

A very basic (LAMMPS specific) implementation that may require some abstraction is implemented [here](https://github.com/pyiron/pyiron_atomistics/pull/217)

sudarsan-surendralal:
Leveraging the capabilities of [pint](https://pint.readthedocs.io/en/stable/index.html) to handle units for different simulation codes as well as modelling length and timescales.

### DIFF
--- a/pyiron_base/generic/units.py
+++ b/pyiron_base/generic/units.py
@@ -114,7 +114,8 @@ class PyironUnitRegistry:
             if quantity in self.unit_dict.keys():
                 self._quantity_dict[label] = quantity
             else:
-                raise KeyError("Quantity {} is not defined. Use `add_quantity` to register the unit of this label")
+                raise KeyError("Quantity {} is not defined. "
+                               "Use `add_quantity` to register the unit of this label".format(quantity))
 
     def __getitem__(self, item):
         """
--- a/pyiron_base/generic/units.py
+++ b/pyiron_base/generic/units.py
@@ -6,6 +6,8 @@ import functools
 import numpy as np
 import pint
 
+Q_ = pint.Quantity
+
 __author__ = "Sudarsan Surendralal"
 __copyright__ = (
     "Copyright 2021, Max-Planck-Institut für Eisenforschung GmbH - "
@@ -165,7 +167,7 @@ class UnitConverter:
     0.043364104241800934
 
     Alternatively, the unit converter can also be used as decorators for functions that return an array scaled into
-    appropriate units
+    appropriate units:
 
     >>> @unit_converter(quantity="energy", conversion="code_to_base")
     >>>    def return_ones():
@@ -173,6 +175,14 @@ class UnitConverter:
     >>> print(return_ones())
     [0.0433641 0.0433641 0.0433641 0.0433641 0.0433641]
 
+    The decorator can also be used to assign units for numpy arrays
+    (for more info see https://pint.readthedocs.io/en/0.10.1/numpy.html)
+
+    >>> @unit_converter(quantity="energy", conversion="base_units")
+    >>> def return_ones_ev():
+    >>>     return np.ones(5)
+    >>> print(return_ones_ev())
+    [1.0 1.0 1.0 1.0 1.0] electron_volt
     """
 
     def __init__(self, base_units, code_units):
@@ -257,8 +267,24 @@ class UnitConverter:
                 @functools.wraps(function)
                 def dec(*args, **kwargs):
                     return np.array(function(*args, **kwargs) * self.base_to_code_value(quantity),
-                                    dtype=self._base_units.get_dtype(quantity))
+                                    dtype=self._code_units.get_dtype(quantity))
                 return dec
             return _decorate_to_code
+        elif conversion == "base_units":
+            def _decorate_base_units(function):
+                @functools.wraps(function)
+                def dec(*args, **kwargs):
+                    return Q_(np.array(function(*args, **kwargs), dtype=self._base_units.get_dtype(quantity)),
+                              self._base_units[quantity])
+                return dec
+            return _decorate_base_units
+        elif conversion == "code_units":
+            def _decorate_code_units(function):
+                @functools.wraps(function)
+                def dec(*args, **kwargs):
+                    return Q_(np.array(function(*args, **kwargs), dtype=self._code_units.get_dtype(quantity)),
+                              self._code_units[quantity])
+                return dec
+            return _decorate_code_units
         else:
             raise ValueError("Conversion type {} not implemented!".format(conversion))
--- a/pyiron_base/generic/units.py
+++ b/pyiron_base/generic/units.py
@@ -104,7 +104,8 @@ class PyironUnitRegistry:
             quantity (str): Physical quantity associated with the labels
 
         Note: `quantity` should already be a key of unit_dict
-
+Raises:
+    ValueError: if quantity is not yet added with :method:`.add_quantity()`.
         """
         for label in labels:
             if quantity in self.unit_dict.keys():
--- a/pyiron_base/generic/units.py
+++ b/pyiron_base/generic/units.py
@@ -63,7 +63,7 @@ class PyironUnitRegistry:
     @property
     def dtype_dict(self):
         """
-        A dictionary of the different physical quantities and the corresponding datatype in which they are to be stored
+        A dictionary of the names of the different physical quantities to the corresponding datatype in which they are to be stored
 
         Returns:
             dict
--- a/pyiron_base/generic/units.py
+++ b/pyiron_base/generic/units.py
@@ -172,19 +172,19 @@ class UnitConverter:
     >>> base.add_quantity(quantity="energy", unit=pint_registry.eV)
     >>> code = PyironUnitRegistry()
     >>> code.add_quantity(quantity="energy",
-    >>>                         unit=pint_registry.kilocal / (pint_registry.mol * pint_registry.N_A))
+    ...                         unit=pint_registry.kilocal / (pint_registry.mol * pint_registry.N_A))
     >>> unit_converter = UnitConverter(base_registry=base, code_registry=code)
 
     The unit converter instance can then be used to obtain conversion factors between code and base units either as a
     `pint` quantity:
 
     >>> print(unit_converter.code_to_base_pint("energy"))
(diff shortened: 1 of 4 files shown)


pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+    def add_labels(self, labels, quantity):
+        """
+        Maps quantities with different labels to quantities already defined in the registry
+
+        Args:
+            labels (list/ndarray): List of labels
+            quantity (str): Physical quantity associated with the labels
+
+        Note: `quantity` should already be a key of unit_dict
+
```
pmrv, line 111:
```suggestion
Raises:
    ValueError: if quantity is not yet added with :method:`.add_quantity()`.
```

pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+        Returns:
+            pint.unit.Unit/pint.quantity.Quantity: The corresponding `pint` unit/quantity
+        """
+        if item in self._unit_dict.keys():
+            return self._unit_dict[item]
+        elif item in self._quantity_dict.keys():
+            return self._unit_dict[self._quantity_dict[item]]
+        else:
+            raise ValueError("Quantity/label '{}' not registered in this unit registry".format(item))
```
pmrv: Should probably be a `KeyError`.

pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+    def __getitem__(self, item):
+        """
+        Getter to return corresponding `pint` unit for a given quantity
+
+        Args:
+            item (str):
+
+        Returns:
+            pint.unit.Unit/pint.quantity.Quantity: The corresponding `pint` unit/quantity
+        """
```
pmrv, line 132:
```suggestion
Raises:
   ...
        """
```

pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+
+    >>> print(unit_converter.code_to_base_value("energy"))
+    0.043364104241800934
+
+    Alternatively, the unit converter can also be used as decorators for functions that return an array scaled into
+    appropriate units:
+
+    >>> @unit_converter(quantity="energy", conversion="code_to_base")
+    >>>    def return_ones():
+    >>>        return np.ones(5)
```
pmrv:
```suggestion
    ... def return_ones():
    ...    return np.ones(5)
```

pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+    >>> print(return_ones())
+    [0.0433641 0.0433641 0.0433641 0.0433641 0.0433641]
+
+    The decorator can also be used to assign units for numpy arrays
+    (for more info see https://pint.readthedocs.io/en/0.10.1/numpy.html)
+
+    >>> @unit_converter(quantity="energy", conversion="base_units")
+    >>> def return_ones_ev():
+    >>>     return np.ones(5)
```
pmrv:
```suggestion
    ... def return_ones_ev():
    ...     return np.ones(5)
```

pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+    >>> print(return_ones_ev())
+    [1.0 1.0 1.0 1.0 1.0] electron_volt
+
+    """
+
+    def __init__(self, base_units, code_units):
+        """
+
+        Args:
+            base_units (pyiron_base.generic.units.PyironUnitRegistry): Base unit registry
```
pmrv:
```suggestion
            base_units (:class:`pyiron_base.generic.units.PyironUnitRegistry`): Base unit registry
```
I think this is the right syntax to make links in the sphinx output.

pmrv: At least for references to our own classes we should use it, to make the docs more readable.

pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Args:
+            quantity (str): Name of quantity
+
+        Returns:
+            float: Conversion factor as a float
+        """
+        return self.base_to_code(quantity).magnitude
+
+    def __call__(self, conversion, quantity):
```
pmrv, line 285: It would be cool if instead of having to provide the right string to `conversion`, I could do something like

```python
@unitconverter.code_to_base(quantity='energy')
def foo(...):
    ...
```

That means making four decorators and having them as attributes on `UnitConverter`.  If that turns out to be too much work the current solution is perfectly fine.

tests/generic/test_units.py
```diff
@@ -0,0 +1,76 @@
...
+        base_units.add_quantity(quantity="energy", unit=pint_registry.eV)
+        self.assertRaises(ValueError, base_units.add_quantity, quantity="energy", unit="eV")
+        dim_less_q = pint_registry.Quantity
+        base_units.add_quantity(quantity="dimensionless_integer_quantity", unit=dim_less_q(1), data_type=int)
+        code_units = PyironUnitRegistry()
+        # Define unit kJ/mol
+        code_units.add_quantity(quantity="energy",
+                                unit=pint_registry.kilocal / (pint_registry.mol * pint_registry.N_A))
+        code_units.add_labels(labels=["energy_tot", "energy_pot"], quantity="energy")
+        # Raise Error for undefined quantity
```
pmrv, line 36: Comments like this could live inside `self.assertRaises(..., msg='Raise Error for undefined quantity')` to make the test output a little bit more readable.

tests/generic/test_units.py
```diff
@@ -0,0 +1,76 @@
...
+        self.assertEqual(1 * return_ones_ev().units, 1 * base_units["energy"])
+        self.assertTrue(np.allclose(return_ones_base(), np.ones(10) * 0.0433641))
+        self.assertTrue(np.allclose(return_ones_base() * return_ones_code(), np.ones(10) * 1))
+        self.assertRaises(ValueError, unit_converter, quantity="energy", conversion="gibberish")
+        # Define dimensionally incorrect units
+        code_units.add_quantity(quantity="energy",
+                                unit=pint_registry.N * pint_registry.metre ** 2)
+        unit_converter = UnitConverter(base_units=base_units, code_units=code_units)
+        # Check if dimensionality error raised
+        self.assertRaises(pint.DimensionalityError, unit_converter.code_to_base_value, "energy")
```
pmrv: Is it possible to raise the dimensionality error already in `add_quantity`?

pyiron_base/generic/units.py
pmrv: Generally there's a few places where the raised exception are not documented yet.

pyiron_base/generic/units.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Returns:
+            dict
+        """
+        return self._quantity_dict
+
+    @property
+    def dtype_dict(self):
+        """
+        A dictionary of the different physical quantities and the corresponding datatype in which they are to be stored
```
pmrv:
```suggestion
        A dictionary of the names of the different physical quantities to the corresponding datatype in which they are to be stored
```

pmrv (changes requested):
I left a few small remarks, but generally looks really nice!

I especially liked the detailed usage examples, do they pass under [doctest](https://docs.python.org/3/library/doctest.html?highlight=doctest#module-doctest)?  No problem if they don't, tough, but it's easy to check and a nice bonus.

pyiron_base/generic/units.py
pmrv: Continuation lines are marked with `...` instead of `>>>` because they indicate the lines as one prompt in a python shell.

tests/generic/test_units.py
sudarsan-surendralal: Not unless we know the dimensionality of the quantities beforehand which is a lot of additional work

pmrv: So here the error is raised, because `UnitConverter` receives two incompatible units for the same quantity name, e.g. `'energy'`, right?  I see that we don't have an a-priori mapping between our quantity names and their dimensionality, so we cannot check in `add_quantity`.  What about checking during the initialization of `UnitConverter`, like
```python
for quantity, unit in base_units.items():
  assert unit.dimensionality == code_units[quantity].dimensionality
```
I've just skimmed the pint docs, so I'm not sure that always makes sense.

pyiron_base/generic/units.py
pmrv: Thanks for adding this, love it!

pyiron_base/generic/units.py
```diff
@@ -103,15 +104,17 @@ def add_labels(self, labels, quantity):
...
-Raises:
-    ValueError: if quantity is not yet added with :method:`.add_quantity()`.
+
         """
         for label in labels:
             if quantity in self.unit_dict.keys():
                 self._quantity_dict[label] = quantity
             else:
-                raise ValueError("Quantity {} is not defined. Use `add_quantity` to register the unit of this label")
+                raise KeyError("Quantity {} is not defined. Use `add_quantity` to register the unit of this label")
```
niklassiemer: Here the `.format(label)` is missing

sudarsan-surendralal: Oops! Fixed now

tests/generic/test_units.py
sudarsan-surendralal: I've implemented new checks based on your ideas. I think for now they check that the units are compatible and not if they're correct.

pmrv: Looks good, thanks!

sudarsan-surendralal:
I've found a way to test the Docstrings properly as well. If this is also okay, I can merge this!

pmrv:
> I've found a way to test the Docstrings properly as well. If this is also okay, I can merge this!

Cool!  I had meant to just check it once, but this will make keeping the docstring up to date *much* easier, thanks!

pmrv (approved):

PR #401, opened 2021-08-12
Install pyiron_base last in contrib compat workflow

pmrv:
This avoids it being overwritten by the conda version when installing pyiron_atomistics from contrib deps.

### DIFF
--- a/.github/workflows/contrib-compat.yml
+++ b/.github/workflows/contrib-compat.yml
@@ -26,12 +26,12 @@ jobs:
     - name: Setup
       shell: bash -l {0}
       run: |
-        pip install --no-deps .
         cd ..
         git clone https://github.com/pyiron/pyiron_contrib
         cd pyiron_contrib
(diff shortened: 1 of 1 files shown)


niklassiemer (commented):
looks like a good idea 👍 I will trigger the `integration` tag to make it run once.

jan-janssen:
For me to understand this correctly - this should prevent issues like https://github.com/pyiron/pyiron_base/pull/390 to be recognised before a new `pyiron_base` version is released, right?

niklassiemer:
Yes, the idea of the whole `integration` tag is that we run the unittests of the other packages with the branch we are developing right now to avoid such issues. However, since `pyiron_contrib` installs itself `pyiron_base` via installing `pyiron_atomistics` it is not using the current development code at all, which is a bug for this compatibility workflow.

pmrv:
It runs, but the test still fails naturally, so I say it's good to go.

PR #407, opened 2021-08-20
use logging directly to get the logger

niklassiemer:


pmrv:
Could work, will test later today.

pmrv:
I've seen it in a log output now, so it should work.

PR #416, opened 2021-08-26
Add a Hdf mixin that takes care of type info + general boilerplate

pmrv:
It came up a few times that `DataContainer` was subclassed because of the HDF functionality, but wasn't otherwise very suitable, e.g. [here](https://github.com/pyiron/pyiron_contrib/pull/205) and [here](https://github.com/pyiron/pyiron_contrib/pull/228) and generally we have objects that write themselves to HDF but are not based on a general data structure like `DataContainer`.  I've had this sketch for a while now that should make it easier to implement this kind of classes.

As a demonstration I've derived `DataContainer` from this new mixin, but generally it should be easy to move old classes to this format (expect `GenericParameters` which handles HDF in a slightly different way).

Test + docs will follow after discussion whether we find this useful.

### DIFF
--- a/pyiron_base/interfaces/has_hdf.py
+++ b/pyiron_base/interfaces/has_hdf.py
@@ -42,12 +42,13 @@ class HasHDF(ABC):
         hdf["HDF_VERSION"] = self.__hdf_version__
 
     def from_hdf(self, hdf, group_name=None):
-        group_name = group_name or self._get_group_name()
+        group_name = group_name or self._get_hdf_group_name()
         with WithHDF(hdf, group_name) as hdf:
             version = hdf.get("HDF_VERSION", "0.1.0")
             self._from_hdf(hdf, version=version)
 
     def to_hdf(self, hdf, group_name=None):
+        group_name = group_name or self._get_hdf_group_name()
         with WithHDF(hdf, group_name) as hdf:
             if len(hdf.list_dirs()) > 0 and group_name is None:
                 raise ValueError("HDF group must be empty when group_name is not set.")
--- a/pyiron_base/interfaces/has_hdf.py
+++ b/pyiron_base/interfaces/has_hdf.py
@@ -35,7 +35,6 @@ class _WithHDF:
 
 class HasHDF(ABC):
 
-    __version__ = "0.1.0"
     __hdf_version__ = "0.1.0"
 
     @abstractmethod
--- a/pyiron_base/interfaces/has_hdf.py
+++ b/pyiron_base/interfaces/has_hdf.py
@@ -38,6 +38,7 @@ class HasHDF(ABC):
     def _type_to_hdf(self, hdf):
         hdf["NAME"] = self.__class__.__name__
         hdf["TYPE"] = str(type(self))
+        hdf["OBJECT"] = hdf["NAME"] # unused alias
         hdf["VERSION"] = self.__version__
         hdf["HDF_VERSION"] = self.__hdf_version__
 
--- a/pyiron_base/interfaces/has_hdf.py
+++ b/pyiron_base/interfaces/has_hdf.py
@@ -71,13 +71,13 @@ class HasHDF(ABC):
         hdf["HDF_VERSION"] = self.__hdf_version__
 
     def from_hdf(self, hdf, group_name=None):
-        group_name = group_name or self._get_hdf_group_name()
+        group_name = group_name if group_name is not None else self._get_hdf_group_name()
         with _WithHDF(hdf, group_name) as hdf:
             version = hdf.get("HDF_VERSION", "0.1.0")
             self._from_hdf(hdf, version=version)
 
     def to_hdf(self, hdf, group_name=None):
-        group_name = group_name or self._get_hdf_group_name()
+        group_name = group_name if group_name is not None else self._get_hdf_group_name()
         with _WithHDF(hdf, group_name) as hdf:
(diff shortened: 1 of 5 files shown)


niklassiemer:
I really like this idea a lot! It might make writing new self-hdf-able classes a lot easier! Maybe you should add the `@classmethod` `from_hdf_args(cls, hdf)` somehow, since this one allows to pass in needed `__init__` variables upon `to_object()`.

pyiron_base/interfaces/has_hdf.py
```diff
@@ -0,0 +1,63 @@
+from abc import ABC, abstractmethod
+
+class WithHDF:
```
liamhuber:
```suggestion
class _WithHDF:
```

Or?

pyiron_base/interfaces/has_hdf.py
```diff
@@ -0,0 +1,63 @@
...
+    @abstractmethod
+    def _from_hdf(self, hdf, version=None):
+        pass
+
+    @abstractmethod
+    def _to_hdf(self, hdf):
+        pass
+
+    @abstractmethod
+    def _get_hdf_group_name(self):
```
liamhuber:
```suggestion
    @property
    @abstractmethod
    def _hdf_group_name(self):
```

At least I think that's the right order for the decorators. 

Am I missing a side effect?

liamhuber (approved):
I love this. `WithHDF` took me a second read-through to understand, but I find the use of `__enter__` to handle the group name absolutely wonderful.

I posted a couple of minor suggestions to consider.

pyiron_base/interfaces/has_hdf.py
pmrv: It should work, I didn't do it before because it's a private method anyway, but I can change it now.

pyiron_base/interfaces/has_hdf.py
```diff
@@ -0,0 +1,63 @@
...
+    def _type_to_hdf(self, hdf):
+        hdf["NAME"] = self.__class__.__name__
+        hdf["TYPE"] = str(type(self))
+        hdf["OBJECT"] = hdf["NAME"] # unused alias
+        if hasattr(self, "__version__"):
+            hdf["VERSION"] = self.__version__
+        hdf["HDF_VERSION"] = self.__hdf_version__
+
+    def from_hdf(self, hdf, group_name=None):
+        group_name = group_name or self._get_hdf_group_name()
```
pmrv: Explicit check for `None`.

pyiron_base/interfaces/has_hdf.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def to_hdf(self, hdf, group_name=None):
+        group_name = group_name or self._get_hdf_group_name()
+        with _WithHDF(hdf, group_name) as hdf:
+            if len(hdf.list_dirs()) > 0 and group_name is None:
+                raise ValueError("HDF group must be empty when group_name is not set.")
+            self._to_hdf(hdf)
+            self._type_to_hdf(hdf)
+
+    def rewrite_hdf(self, hdf, group_name=None):
```
pmrv: Garbage collection for HDF?

pmrv:
Fixed the test, details in the commits.  The problem was that previously `DataContainer.from_hdf` didn't call `hdf.close()` if it descended into a group.  `HasHDF` does that now, but since `DataContainer` passes a reference to `HDFStub`, this in turn reached into the wrong location in the file (since where hdf points to changes on `hdf.close`).

pmrv:
Rebased to resolve merge conflict.

liamhuber:
Beautiful, tests are passing so I'm happy! If you have no further concerns, please merge it and let's get this thing live :)

Since `PyironObject`s are the trifecta of python, database, and file storage, we should consider if we want some sort of `class Storable(HasHDF, HasStorage)` as an intermediate that never needs to be located in the database.

EDIT: spelling of trifecta

pmrv:
> Beautiful, tests are passing so I'm happy! If you have no further concerns, please merge it and let's get this thing live :)
> 
> Since `PyironObject`s are the trifecta of python, database, and file storage, we should consider if we want some sort of `class Storable(HasHDF, HasStorage)` as an intermediate that never needs to be located in the database.
> 
> EDIT: spelling of trifecta

Since `HasStorage` just has the input/output container, I'd actually think, that it should just derive from `HasHDF`.

liamhuber:
> Since HasStorage just has the input/output container, I'd actually think, that it should just derive from HasHDF.

Yes, this can work. We just need to think a little bit whether (1) `HasStorage._to_hdf` calls `self._storage.to_hdf()`, or if (2) `HasStorage` overrides `to_hdf` with a `super` call and the storage saving.

(1) Pro: the ABC is satisfied and people can directly make new classes without ever worrying about storage (as long as everything they want serialized gets slapped onto the `storage` attribute. Con: If someone *does* want to save extra stuff they need to remember to call `super` in `_to_hdf`.

(2) Pro: `_to_hdf` maintains its status as "that thing I should do my own work in". Con: anyone inheriting from `HasStorage` is still stuck implementing `_to/_from_hdf` every single time, even if they don't need it.

I lean towards (1). Thoughts?

pmrv:
Yeah, (1) sounds cleaner.  Overloading the saving stuff should only be necessary for fundamental object and jobs anyway.

PR #417, opened 2021-08-26
Support Pandas DataFrame in the DataContainer

jan-janssen:
This is primarily used for Scriptjobs to submit the current state of a dataset for analysis.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -8,6 +8,7 @@ Data structure for versatile data handling.
 import copy
 import json
 import warnings
+import pandas
 from collections.abc import Sequence, Set, Mapping, MutableMapping
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


pmrv:
The underlying problem is that `DataFrame` defines its own `to_hdf` that is independent of ours?

pmrv:
I'm not quite happy with having to special case the `DataFrame` on both writing and reading.  I'll think about it today if there's not another way to do it.

pmrv:
I see that in `FileHDFio` we make a similar thing to allow saving dataframes

https://github.com/pyiron/pyiron_base/blob/20f31a7e30c03dc392cdf4d430d52acb90e6493d/pyiron_base/generic/hdfio.py#L646

The easiest solution is probably to check for `DataFrame` and pass them through to `FileHDFio` that seems to know how to handle them.  I've implemented this in #420.

jan-janssen:
Clou this in favor of https://github.com/pyiron/pyiron_base/pull/420

PR #419, opened 2021-08-27
Customized TestCase for pyiron

sudarsan-surendralal:
This PR includes the following changes

- Creating a generic `PyironTestCase` for all unittesting within pyiron which includes generic functions that can be used for all test modules
- Inheriting all test modules from this particular module
- Automatic testing of the docstrings of the `DataContainer` module

### DIFF
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -52,7 +52,7 @@ class PyironTestCase(unittest.TestCase, ABC):
         self.failIf(result.failed > 0, msg=output)
 
 
-class TestWithProject(PyironTestCase):
+class TestWithProject(PyironTestCase, ABC):
     """
     Tests that start and remove a project for their suite.
     """
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -72,7 +72,7 @@ class TestWithProject(PyironTestCase):
             pass
 
 
-class TestWithCleanProject(TestWithProject):
+class TestWithCleanProject(TestWithProject, ABC):
     """
     Tests that start and remove a project for their suite, and remove jobs from the project for each test.
     """
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -27,36 +27,7 @@ __status__ = "development"
 __date__ = "Mar 23, 2021"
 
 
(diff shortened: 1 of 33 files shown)


pyiron_base/_tests.py
```diff
@@ -47,35 +72,10 @@ def tearDownClass(cls):
             pass
 
 
-class TestWithCleanProject(TestWithProject, ABC):
+class TestWithCleanProject(TestWithProject):
```
```suggestion
class TestWithCleanProject(TestWithProject, ABC):
```

We still expect this to be a parent for individual testing suites.

pyiron_base/_tests.py
```diff
@@ -27,7 +27,32 @@
...
+
+        Output capturing adapted from https://stackoverflow.com/a/22434594/12332968
+        """
+        with StringIO() as buf, redirect_stdout(buf):
+            result = doctest.testmod(self.docstring_module)
+            output = buf.getvalue()
+        self.failIf(result.failed > 0, msg=output)
+
+
+class TestWithProject(PyironTestCase):
```
```suggestion
class TestWithProject(PyironTestCase, ABC):
```

We still expect this to be a parent for individual testing suites.

liamhuber (approved):
Please keep `ABC` for the all the parent test cases. Otherwise looks good!

sudarsan-surendralal:
> 
> 
> Please keep `ABC` for the all the parent test cases. Otherwise looks good!

But `PyironTestCase` is derived from `ABC`. So isn't that sufficient?

liamhuber:
> > Please keep `ABC` for the all the parent test cases. Otherwise looks good!
> 
> But `PyironTestCase` is derived from `ABC`. So isn't that sufficient?

Nope. I thought this was because inheriting from `ABC` makes you uninstantiable, but in fact that's only the case if you *also* have `abstractmethod`s.

What remains is that `ABC` indicates to the developer that this class is not intended to ever be instantiated, but only to act as a parent class for other classes. This is definitely the case for what we're doing here since these only serve as parents for actual test cases, but I will downgrade my complaint to a suggestion.

sudarsan-surendralal:
> What remains is that `ABC` indicates to the developer that this class is not intended to ever be instantiated, but only to act as a parent class for other classes

Thanks, I get it now. Applied your changes.

PR #420, opened 2021-08-29
Don't call pandas' to_hdf method

pmrv:
See #417 for discussion.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -11,6 +11,7 @@ import warnings
 from collections.abc import Sequence, Set, Mapping, MutableMapping
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


jan-janssen:
Just to understand this correctly, with this change it is possible to store pandas.Dataframes inside the DataContainer? If this works then I guess we could add a quick test and merge it.

pmrv:
Yes, it should work, since the HDF layer also does a similar check, but I haven't tested it.

jan-janssen:
I added a unit test - from my perspective this is ready to be merged.

jan-janssen:
The unit test for linux passed - the windows ones currently seem to have a general issue https://github.com/pyiron/pyiron_base/issues/432

PR #423, opened 2021-08-30
Add tqdm to Project.iter_jobs

pmrv:
Sometimes iteration over jobs in a project can be a bit slow-ish, so I often want to have some progress update, but I always forget about `tqdm`, so I thought I'd add it by default. 

Since is strictly QoL so I'm open to discussion to leave the default to `False` or not merge this PR at all, depending on y'alls opinion. ;)


jan-janssen (approved):
Looks good to me

PR #425, opened 2021-09-02
Fix downstream test suite

niklassiemer:


niklassiemer:
With these changes, the downstream test suites do not try to run the tests inside the conda installation of `pyiron_base`. See https://github.com/pyiron/pyiron_atomistics/pull/337.

pyiron_base/_tests.py
```diff
@@ -33,6 +33,12 @@ class PyironTestCase(unittest.TestCase, ABC):
     Tests that also include testing the docstrings in the specified modules
     """
 
+    @classmethod
+    def setUpClass(cls):
+        if cls is PyironTestCase:
```
```suggestion
        if np.any([cls is c for c in _TO_SKIP]):
```

with 

```python
_TO_SKIP = [PyironTestCase, TestWithProject, TestWithCleanProject]
```

So we only need the skipping logic in the parent-most class?

I'm not 100% sure I actually like this better, but I am a bit turned off by the code repetition right now.

Proof of concept:
```python
class Foo:
    @classmethod
    def do(cls):
        if np.any([cls is c for c in TO_SKIP]):
            print(f"Class is {cls.__name__}")
        else:
            print("Nope")

class Bar(Foo):
    pass

class Baz(Bar):
    pass

class Boa(Foo):
    pass

TO_SKIP = [Foo, Bar, Boa]

foo = Foo()
bar = Bar()
baz = Baz()
boa = Boa()
print(foo.do(), bar.do(), baz.do(), boa.do())
>>> Class is Foo
>>> Class is Bar
>>> Nope
>>> Class is Boa
```

liamhuber (approved):
Nice. I had an idea to reduce code repetition, but I'm not sure it's actually any better. Otherwise looks good and will let us move forward on the other PRs, which is important!

pyiron_base/_tests.py
niklassiemer: I think `any` is sufficient, but otherwise I like the general idea. I am just thinking if there is a nicer way to handle the global `TO_SKIP` list , e.g. to register such `BaseClass`es near their definition...

liamhuber: > I thing any is sufficient

Indeed probably, I'm just a creature of habit 😆 

> e.g. to register such BaseClasses near their definition...

Hmm, perhaps you mean something more sophisticated than I am thinking when you say "register", but we could always define an empty list `_SKIP_ABC_TESTS []` and immediately after each class definition give a `_SKIP_ABC_TESTS.append(PyironTestCase)`?

niklassiemer:
I am also 'running the tests' in pyiron/pyiron_atomistics#337.

pyiron_base/_tests.py
niklassiemer: Yes, I think this already would be clearer if we extend the feature - I just do not like the global variable to be around 😅 
However, for now I leave the list definition at the end of the file. I will merge this and release a new `pyiron_base` if all my tests pass such that we can update to the new base in atomistics.

PR #427, opened 2021-09-07
Creator registration

liamhuber:
Instead of having new pyiron submodules subclass `Project` and always worry about which child of `Project` we currently have, to simply on-the-fly enhance the existing `Project` by registering new methods with it. 

This was originally suggested by @freyso [here](https://github.com/pyiron/pyiron_base/pull/385) and has since been discussed at one of the meetings ([minutes](https://github.com/orgs/pyiron/teams/pyiron/discussions/79)) and is an official [issue](https://github.com/pyiron/pyiron_atomistics/issues/323) from @niklassiemer. I made a first crack at it and got a good response from @JNmpi and @muh-hassani in the tech dev meeting ([minutes](https://github.com/orgs/pyiron/teams/development/discussions/48)), so this is an expanded attack at that idea.

The
```python
pr = Project('scratch')
pr.base.job.ScriptJob('foo')

import pyiron_atomistics
job = pr.atomistics.job.Lammps('bar')  # Note we did not need to reinstantiate pr
job.structure = pr.atomistics.structure.bulk('Al')
```

A use case can be seen over in [pyiron_atomistics#347](https://github.com/pyiron/pyiron_atomistics/pull/347).

Pros:
- Drastically reduces use of strings in controlling the code behaviour
- Hopefully no more need to subclass of `Project` for every submodule (which right now is so we can re-instantiate the right class?)
- Submodule init files get a lot simpler
- Development in this direction can run fully in parallel to the existing routines

Cons:
- More typing maybe (i.e. with keyboard)? As it is now we lose the verb `create` which makes me sad, but I'm trying to minimize keystrokes a little
- Potentially harder to find things, e.g. "do I want `pr.atomistics.job.Foo` or `pr.continuum.job.Foo`?"
  - I am definitely still open to rearranging things, e.g. `pr.atomistics.job.Lammps` --> `pr.create.job.atomistics.Lammps`
- **Not yet tested for remote jobs.** As long as your remote python environment has imported (or imports on seeing the loading job type) all the same pyiron modules as you use in your notebook, it's fine...but I haven't tested this
  - I know we already have some solution for this, as I *already* make custom job classes that wouldn't work if I hadn't added bespoke modules to my `PYTHONPATH` in my `.bashrc`, I am just ignorant of *how* exactly we work this magic

TODO:
- ~~Attach job factories to these creators~~
- ~~Unit tests~~
- ~~Finish adding type-hinting~~
- Make sure it works for jobs run on a queue

Not TODO:
- Move stuff around to new modules (e.g. a new `pyiron_base.creators` module or similar)
  - Exact location can be adjusted if/when we're happy with everything in principle, as avoiding circular imports means some of it is a pain in the butt and I want to handle it separately.

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -48,8 +48,6 @@ from pyiron_base.server.queuestatus import (
 from pyiron_base.job.external import Notebook
 from pyiron_base.project.data import ProjectData
 from pyiron_base.archiving import import_archive, export_archive
-from abc import ABC
-from typing import Type
 
 __author__ = "Joerg Neugebauer, Jan Janssen"
 __copyright__ = (
@@ -63,17 +61,6 @@ __status__ = "production"
 __date__ = "Sep 1, 2017"
 
 s = Settings()
-
-
-class CreatorCore(ABC):
-    def __init__(self, project):
-        self._project = project
-
-
-class BaseCreator(CreatorCore):
-    def __init__(self, project):
-        super().__init__(project)
-        # TODO: Attach a job factory
 
 
 class Project(ProjectPath, HasGroups):
@@ -135,7 +122,6 @@ class Project(ProjectPath, HasGroups):
         self._inspect_mode = False
         self._data = None
         self._creator = Creator(project=self)
-        self._base = BaseCreator(project=self)
 
         if not s.database_is_disabled:
             s.open_connection()
@@ -1528,13 +1514,13 @@ class Project(ProjectPath, HasGroups):
         )
 
     @classmethod
-    def register_creator(cls, name: str, creator: Type[CreatorCore]):
+    def register_creator(cls, name: str, creator):
         """
         Add a new creator to the project class.
 
         Example)
-        >>> from pyiron_base import Project, CreatorCore
-        >>> class MyCreator(CreatorCore):
+        >>> from pyiron_base import Project, Toolkit
+        >>> class MyCreator(Toolkit):
         >>>     @property
         >>>     def foo(self):
         >>>         return 'foo'
@@ -1550,7 +1536,7 @@ class Project(ProjectPath, HasGroups):
 
         Args:
             name (str): The name for the newly registered property.
-            creator (CreatorCore): The creator to register.
+            creator (Toolkit): The creator to register.
         """
         setattr(cls, name, property(lambda self: creator(self)))
 
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -1539,6 +1539,10 @@ class Project(ProjectPath, HasGroups):
             name (str): The name for the newly registered property.
             tools (Toolkit): The tools to register.
         """
+        if hasattr(cls, name):
+            raise AttributeError(
+                f'{cls.__name__} already has an attribute {name}. Please use a new name for registration.'
+            )
         setattr(cls, name, property(lambda self: tools(self)))
 
 
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -47,8 +47,9 @@ from pyiron_base.server.queuestatus import (
 )
 from pyiron_base.job.external import Notebook
 from pyiron_base.project.data import ProjectData
-
 from pyiron_base.archiving import import_archive, export_archive
+from abc import ABC
+from typing import Type
 
 __author__ = "Joerg Neugebauer, Jan Janssen"
 __copyright__ = (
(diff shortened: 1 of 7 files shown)


pyiron_base/project/generic.py
```diff
@@ -1514,6 +1513,33 @@ def unpack(self, origin_path, csv_file_name='export.csv', compress=True):
             self, archive_directory=origin_path, df=df, compressed=compress
         )
 
+    @classmethod
+    def register_tools(cls, name: str, creator):
+        """
+        Add a new creator to the project class.
+
+        Example)
```
pmrv, line 1521:
```suggestion
        Example)
        
```
`>>>` have to start with empty line above, otherwise sphinx will mess up the formatting.

pyiron_base/project/generic.py
```diff
@@ -1514,6 +1513,33 @@ def unpack(self, origin_path, csv_file_name='export.csv', compress=True):
...
+        >>> from pyiron_base import Project, Toolkit
+        >>> class MyCreator(Toolkit):
+        >>>     @property
+        >>>     def foo(self):
+        >>>         return 'foo'
```
```suggestion
        ...     @property
        ...     def foo(self):
        ...         return 'foo'
```

pyiron_base/project/generic.py
```diff
@@ -1514,6 +1513,33 @@ def unpack(self, origin_path, csv_file_name='export.csv', compress=True):
             self, archive_directory=origin_path, df=df, compressed=compress
         )
 
+    @classmethod
+    def register_tools(cls, name: str, creator):
```
```suggestion
    def register_tools(cls, name: str, creator: Toolkit):
```

pyiron_base/project/generic.py
```diff
@@ -1514,6 +1513,33 @@ def unpack(self, origin_path, csv_file_name='export.csv', compress=True):
...
+
+        The intent is then that pyiron submodules (e.g. `pyiron_atomistics`) define a new creator and in their
+        `__init__.py` file only need to invoke `Project.register_creator('pyiron_submodule', SubmoduleCreator)`.
+        Then whenever `pyiron_submodule` gets imported, all its functionality is available on the project.
+
+        Args:
+            name (str): The name for the newly registered property.
+            creator (Toolkit): The creator to register.
+        """
+        setattr(cls, name, property(lambda self: creator(self)))
```
pmrv: Should probably do a `hasattr` check, so that we're not overwriting things without knowing.

pmrv (commented):
Looks great!  

I'd also be inclined to keep `create` around, however.  Maybe we can setup a `CombinedCreator` that can bundle multiple creators and by default makes all of their attributes available, but if naming conflicts exist also allow to explicitly specify the creator.   Usage could be

```python
project.create.register_subtools('atomistics', 'job', AtomisticJobFactory)
project.create.register_subtools('atomistics', 'structure', AtomisticStructureFactory)
project.create.job.Lammps == project.create.job.atomistics.Lammps
project.create.structure.bulk == project.create.structure.atomistics.bulk
```
Thinking about it, it might be a bit tricky to get it right implementing it though.

pyiron_base/project/generic.py
liamhuber: No can do. I use `Toolkit` in the example, but right now I can't actually import it in the `project/generic` module or we get circular imports.

Also, if I understand what we want is `creator: typing.Type[Toolkit]` to hint that any child of `Toolkit` will do.

I wasn't willing to fight with circular imports just to get type hinting working :wink:

liamhuber:
> Looks great!
> 
> I'd also be inclined to keep `create` around, however. Maybe we can setup a `CombinedCreator` that can bundle multiple creators and by default makes all of their attributes available, but if naming conflicts exist also allow to explicitly specify the creator. Usage could be
> 
> ```python
> project.create.register_subtools('atomistics', 'job', AtomisticJobFactory)
> project.create.register_subtools('atomistics', 'structure', AtomisticStructureFactory)
> project.create.job.Lammps == project.create.job.atomistics.Lammps
> project.create.structure.bulk == project.create.structure.atomistics.bulk
> ```
> 
> Thinking about it, it might be a bit tricky to get it right implementing it though.

Perfect, glad you like it! Thanks also for the corrections.

Yeah, I think we're going to need to have a wider conversation on Monday to think about the best way to handle the namespaces. I'm still leaning hard towards having `atomistics` (/`continuum`/etc) show up only once  (i.e. at the beginning). Your idea for combinations is neat.
  Regarding `create`, I definitely miss having the verb in the call chain. I also foresee that many of the submodule packages might wind up with methods/attributes that aren't explicitly a creation though...then keeping `create` runs back into my concern about namespace concentration, e.g. to avoid having `pr.create.atomstics.Job` and `pr.atomistics.method` both kicking around.
  We'll see though. I'm not married to any particular namespace organization yet, and your combo idea holds promise.

PR #428, opened 2021-09-07
Update template job to use ~~DataContainer~~ HasStorage

liamhuber:
@JNmpi here it is. I noticed also that this job type is not available to create, but is only accessible with `from pyiron_base import TemplateJob`. My gut feeling is that this is mostly a parent class and not actually instantiated, so it's OK that it doesn't show up on the creation list?

@jan-janssen sorry to pester you for review, but the only place this might have an impact is over in pyiron_experimental for [pystem](https://github.com/pyiron/pyiron_experimental/blob/master/pyiron_experimental/pystemjob.py), [temmeta](https://github.com/pyiron/pyiron_experimental/blob/master/pyiron_experimental/temmetajob.py), and [matchseries](https://github.com/pyiron/pyiron_experimental/blob/master/pyiron_experimental/matchseries.py). For the first two I can already see there's not going to be an impact, but matchseries is more complicated and it has no tests. Since you're the sole contributor there I'd appreciate if you'd quickly look over and/or set up a test.

### DIFF
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -47,7 +47,7 @@ class TemplateJob(GenericJob):
             hdf=hdf,
             group_name=group_name
         )
-        self._data.from_hdf(hdf=hdf, group_name=group_name)
+        self._data.from_hdf(hdf=self.project_hdf5, group_name=None)
 
 
 class PythonTemplateJob(TemplateJob):
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -40,7 +40,7 @@ class TemplateJob(GenericJob):
             hdf=hdf,
             group_name=group_name
         )
-        self._data.to_hdf(hdf=hdf, group_name=group_name)
+        self._data.to_hdf(hdf=self.project_hdf5, group_name=None)
 
     def from_hdf(self, hdf=None, group_name=None):
         super().from_hdf(
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -6,7 +6,7 @@ Template class to define jobs
 """
 
 from pyiron_base.job.generic import GenericJob
-from pyiron_base.generic.parameters import GenericParameters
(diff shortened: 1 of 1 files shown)


pyiron_base/job/template.py
```diff
@@ -23,23 +23,31 @@
...
+        return self._data.output
 
     def to_hdf(self, hdf=None, group_name=None):
         super().to_hdf(
             hdf=hdf,
             group_name=group_name
         )
-        with self.project_hdf5.open("input") as h5in:
-            self.input.to_hdf(h5in)
+        self._data.to_hdf(hdf=hdf, group_name=group_name)
```
pmrv:
```suggestion
        self._data.to_hdf(hdf=self.project_hdf5, group_name=None)
```

pyiron_base/job/template.py
```diff
@@ -23,23 +23,31 @@
...
 
     def from_hdf(self, hdf=None, group_name=None):
         super().from_hdf(
             hdf=hdf,
             group_name=group_name
         )
-        with self.project_hdf5.open("input") as h5in:
-            self.input.from_hdf(h5in)
+        self._data.from_hdf(hdf=hdf, group_name=group_name)
```
pmrv:
```suggestion
        self._data.from_hdf(hdf=self.project_hdf5, group_name=None)
```

pmrv (approved):
Suggested a gotcha in `to/from_hdf`.  `GenericJob.to_hdf/from_hdf` both set `self._hdf5` aka `self.project_hdf` to `hdf.open(group_name)` if `group_name` is defined.  It's therefore sufficient to point `self._data` to `self.project_hdf5` and leave `group_name` to the default `table_name`.  If that is not done, you'd end up writing both the job and its input in the same HDF5 group.

liamhuber:
@pmrv I think I can actually dodge the gotcha altogether by leveraging one of your other fine creations ;)

pmrv (approved):
lgtm now!

PR #429, opened 2021-09-09
Flattened storage store defaults

Leimeroth:
This makes sure that the fill value set when using add_arrays is also set when resizing arrays.


pmrv (approved):
lgtm!

pmrv:
Ah, please add a few unit tests `pyiron_base/generic/test_flattenedstorage.py`.  Should test at least
1. arrays without a fill are unchanged after resize (value-wise)
2. arrays with a fill have the correct value in the new portion of the array (and only there)
3. `_default_values` is correctly read from HDF.

Ah and also rename `_default_values` to `fill_values`.  The rest of the code refers to them as `fill`, so it should be consistent.

PR #431, opened 2021-09-14
Fix storage

liamhuber:
Had some trouble due to multiple inheritance.

### DIFF
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -36,8 +36,8 @@ class TemplateJob(GenericJob, HasStorage):
         return self._storage.output
 
     def to_hdf(self, hdf=None, group_name=None):
-        HasStorage.to_hdf(self, hdf=self.project_hdf5)
         GenericJob.to_hdf(self, hdf=hdf, group_name=group_name)
+        HasStorage.to_hdf(self, hdf=self.project_hdf5)
 
     def from_hdf(self, hdf=None, group_name=None):
         GenericJob.to_hdf(self, hdf=hdf, group_name=group_name)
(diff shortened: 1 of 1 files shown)


pyiron_base/job/template.py
```diff
@@ -35,6 +35,14 @@ def input(self):
     def output(self):
         return self._storage.output
 
+    def to_hdf(self, hdf=None, group_name=None):
+        HasStorage.to_hdf(self, hdf=self.project_hdf5)
+        GenericJob.to_hdf(self, hdf=hdf, group_name=group_name)
```
pmrv:
```suggestion
        GenericJob.to_hdf(self, hdf=hdf, group_name=group_name)
        HasStorage.to_hdf(self, hdf=self.project_hdf5)
```
I suspect they should be changed around, because `GenericJob.to_hdf` should be messing with `self.project_hdft`.

liamhuber: Yeah, that's how I had it originally then I played around seeing if I could get the job's HDF_VERSION to bump to 0.2.0 by invoking HasStorage first (obviously not, since HasStorage only deals with the storage hdf, it was a stupid idea) -- but it worked both ways in my example and I forgot to switch it back ;) GenericJob first is at the very least more aesthetically pleasing and may have some positive side effect I didn't see.

pmrv:
LGTM! I guess having to explicitly call both parents is a bit annoying, but that might go away in the future when any job has a `DataContainer` for itself.

liamhuber:
> LGTM! I guess having to explicitly call both parents is a bit annoying, but that might go away in the future when any job has a `DataContainer` for itself.

Yeah, we're stuck with the explicit calls as long as we take the 'composition' route, but I agree we should be able to hide it from users and most devs once we have some good forward-facing classes cooked up to inherit from.

PR #434, opened 2021-09-16
Smooth storage

liamhuber:
Rebase `HasStorage` to be a child of `HasHDF`. This way all children of `HaveStorage` handle the complete storage workflow: they can themselves get serialized and deserialized, and they have a spot for slapping any permanent data that should belong to them.

@pmrv it was all super easy, but I was a bit uncomfortable with this snippet:
```python
class HasStorage...

    def _get_hdf_group_name(self):
        return ""
```

Basically it means that `TYPE` etc will get stored at the base level of whatever is passed as `hdf` to the `HasStorage` child class, while `storage` is the sister to these and holds all the data _belonging_ to that object. This works beautifully, e.g. over in `TemplateJob` and in what I'm doing over in the `schroedinger` branch of `pyiron_continuum`, but do you foresee any side effects?

### DIFF
--- a/pyiron_base/generic/object.py
+++ b/pyiron_base/generic/object.py
@@ -44,8 +44,6 @@ class HasStorage(HasHDF, ABC):
     def _from_hdf(self, hdf: ProjectHDFio, version: str=None):
         self.storage.from_hdf(hdf=hdf)
 
-    def _get_hdf_group_name(self):
-        return ""
 
 
 class HasDatabase(ABC):
--- a/pyiron_base/generic/object.py
+++ b/pyiron_base/generic/object.py
@@ -8,6 +8,8 @@ The core class in pyiron, linking python to the database to file storage.
 
 from abc import ABC
 from pyiron_base import DataContainer
+from pyiron_base.interfaces.has_hdf import HasHDF
+from pyiron_base.generic.hdfio import ProjectHDFio
 from pyiron_base.settings.generic import Settings
 
 __author__ = "Liam Huber"
@@ -24,7 +26,7 @@ __date__ = "Mar 23, 2021"
 s = Settings()
 
 
-class HasStorage(ABC):
+class HasStorage(HasHDF, ABC):
     """
     A base class for objects that use HDF5 data serialization via the `DataContainer` class.
     """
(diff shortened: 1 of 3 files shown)


pmrv:
My first thought is a bit uncomfortable, but I have to meditate for a night about this.

pyiron_base/generic/object.py
```diff
@@ -33,32 +35,17 @@ def __init__(self, *args, **kwargs):
...
-            hdf = hdf.create_group(group_name)
-        self._storage.from_hdf(hdf=hdf)
+    def _to_hdf(self, hdf: ProjectHDFio):
+        self.storage.to_hdf(hdf=hdf)
+
+    def _from_hdf(self, hdf: ProjectHDFio, version: str=None):
+        self.storage.from_hdf(hdf=hdf)
+
+    def _get_hdf_group_name(self):
+        return ""
```
pmrv:
```suggestion
```
`HasHDF` checks for `None` explicitly and returning `""` has the same effect as `None`, so I'd just not provide it.

pmrv (approved):
Actually if `_get_hdf_group_name` returns `None` and no `group_name` is passed into `to_hdf` the code will works as documented by `HasHDF`, so I'm actually sort of ok with how it is.

I did a quick test with this branch and get the following structure in HDF.  Say I have a
```python
class MyObject(HasStorage):
  pass
```
then doing
```python
o = MyObject()
o.storage.foo = 42
o.to_hdf(hdf=...)
```
gives me
```
root
|-TYPE # type info of HasStorage-derived object
|-NAME
|-....
\-storage # nested group for datacontainer
```
whereas 
```python
o.to_hdf(hdf=..., group_name="my_store")
```
gives me
```
root
  \-my_store
    |-TYPE # type info of HasStorage-derived object
    |-NAME
    |-....
    \-storage # nested group for datacontainer
```
which I am ok with.

If we don't like this we could go back to `HasHDF` and only allow passing `group_name=None` if the objects `_get_hdf_group_name` returns something not-`None`.

pyiron_base/generic/object.py
liamhuber: Yes, but it's an `@abstractmethod` and Pycharm will whine if I don't define it here...?

pmrv: Oh, yes.  Then I'd say we change that to return `None` by default.

liamhuber:
> which I am ok with.

This was basically the line of reasoning I went through. I'm very happy to just have the `None` behaviour as the default. Once the tests are done I'll merge 😄

PR #435, opened 2021-09-17
Reference job file size increase fix (issue #411)

### LINKED ISSUE #411: Reference job file size increases while creating copies of it
Consider the following code:

```
import os

pr = Project('ref_job_size')

ref_job = pr.create.job.Lammps('ref_job')
ref_job.structure = pr.create.structure.bulk('Al', cubic=True).repeat(3)
ref_job.potential = '2008--Mendelev-M-I--Al--LAMMPS--ipr1'
ref_job.save()
```

The size of the .h5 file created is:

```
os.stat('ref_job_size/ref_job.h5').st_size / 1e6
>>> 0.267593  # in MB
```

If I now create copies of it, I expect the size of  `ref_job.h5` to remain the same. 

```
for i in range(10):
    job = ref_job.copy_to(
        project=pr,
        new_job_name='new_job_' + str(i),
        input_only=True,
        new_database_entry=False,
        delete_existing_job=True
        )
```

However, I see that the size of `ref_job.h5` has increased:

```
os.stat('ref_job_size/ref_job.h5').st_size / 1e6 
>>> 4.782905  # in MB
```

Is this unexpected behavior or an artifact of how the `copy_to()` function is intended to work?

raynol-dsouza:
There were a couple of things that were actually causing the file size explosion in issue #411:

1. Line 710 in `job/core.py` was not calling the `copy` method defined on line 658. I renamed the method to `copy_self`.
2. `to_hdf` was being called on the reference job every time a new copy of it was made. I added a hidden flag in the `JobCore` class called `_copy_exists` to record if an object of this class has previously been copied or not.

This PR should fix #411.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -655,7 +655,7 @@ class JobCore(HasGroups):
         h5_dict["groups"] += self._list_ext_childs()
         return h5_dict
 
-    def copy(self):
+    def copy_self(self):
         """
         Copy the JobCore object which links to the HDF5 file
 
@@ -707,7 +707,7 @@ class JobCore(HasGroups):
 
         # Create a new job by copying the current python object, move the content
         # of the HDF5 file and then attach the new HDF5 link to the new python object.
-        new_job_core = self.copy()
+        new_job_core = self.copy_self()
         new_job_core._name = new_job_name
         new_job_core._hdf5 = hdf5_project
         new_job_core._master_id = self._master_id
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -655,7 +655,7 @@ class JobCore(HasGroups):
         h5_dict["groups"] += self._list_ext_childs()
         return h5_dict
 
-    def copy(self):
+    def copy_self(self):
         """
         Copy the JobCore object which links to the HDF5 file
 
@@ -707,11 +707,7 @@ class JobCore(HasGroups):
 
         # Create a new job by copying the current python object, move the content
         # of the HDF5 file and then attach the new HDF5 link to the new python object.
-        if isinstance(project, JobCore):
-            new_job_core = self.copy()
-        else:
-            new_job_core = copy.copy(self)
-            new_job_core.reset_job_id()
+        new_job_core = self.copy_self()
         new_job_core._name = new_job_name
         new_job_core._hdf5 = hdf5_project
         new_job_core._master_id = self._master_id
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -655,7 +655,7 @@ class JobCore(HasGroups):
         h5_dict["groups"] += self._list_ext_childs()
         return h5_dict
 
-    def copy_self(self):
+    def copy(self):
         """
         Copy the JobCore object which links to the HDF5 file
 
@@ -707,7 +707,11 @@ class JobCore(HasGroups):
 
         # Create a new job by copying the current python object, move the content
         # of the HDF5 file and then attach the new HDF5 link to the new python object.
-        new_job_core = self.copy_self()
+        if isinstance(project, JobCore):
+            new_job_core = self.copy()
+        else:
+            new_job_core = copy.copy(self)
+            new_job_core.reset_job_id()
         new_job_core._name = new_job_name
         new_job_core._hdf5 = hdf5_project
         new_job_core._master_id = self._master_id
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -894,7 +894,7 @@ class JobCore(HasGroups):
         """
         try:
             return self._hdf5[item]
-        except ValueError:
+        except (ValueError, TypeError):
             pass
 
         name_lst = item.split("/")
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -894,7 +894,7 @@ class JobCore(HasGroups):
         """
         try:
             return self._hdf5[item]
-        except (ValueError, TypeError):
+        except ValueError:
             pass
 
         name_lst = item.split("/")
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -663,6 +663,7 @@ class JobCore(HasGroups):
             JobCore: New FileHDFio object pointing to the same HDF5 file
         """
         copied_self = copy.copy(self)
+        copied_self.reset_job_id()
         return copied_self
 
     def _internal_copy_to(self, project=None, new_job_name=None, new_database_entry=True,
@@ -893,7 +894,7 @@ class JobCore(HasGroups):
         """
         try:
             return self._hdf5[item]
-        except ValueError:
+        except (ValueError, TypeError):
             pass
 
         name_lst = item.split("/")
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -663,7 +663,6 @@ class JobCore(HasGroups):
             JobCore: New FileHDFio object pointing to the same HDF5 file
         """
         copied_self = copy.copy(self)
-        copied_self.reset_job_id()
         return copied_self
 
     def _internal_copy_to(self, project=None, new_job_name=None, new_database_entry=True,
--- a/pyiron_base/job/util.py
+++ b/pyiron_base/job/util.py
@@ -379,12 +379,13 @@ def _job_remove_folder(job):
         shutil.rmtree(working_directory)
 
 
-def _job_store_before_copy(job):
+def _job_store_before_copy(job, copy_exists=False):
     """
     Store job in HDF5 file for copying
 
     Args:
         job (GenericJob): job object to copy
+        copy_exists (bool): if a copy of the job has already been made
 
     Returns:
         bool: [True/False] if the HDF5 file of the job exists already
@@ -393,7 +394,8 @@ def _job_store_before_copy(job):
         delete_file_after_copy = True
     else:
         delete_file_after_copy = False
-    job.to_hdf()
+    if not copy_exists:
+        job.to_hdf()
     return delete_file_after_copy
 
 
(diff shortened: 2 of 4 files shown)


raynol-dsouza:
@jan-janssen

pyiron_base/job/core.py
```diff
@@ -655,7 +655,7 @@ def _list_all(self):
         h5_dict["groups"] += self._list_ext_childs()
         return h5_dict
 
-    def copy(self):
+    def copy_self(self):
```
pmrv, line 658:
```suggestion
    def copy(self):
```
Imo not necessary.

pyiron_base/job/core.py
```diff
@@ -894,7 +894,7 @@ def __getitem__(self, item):
         """
         try:
             return self._hdf5[item]
-        except ValueError:
+        except (ValueError, TypeError):
```
pmrv, line 897: What raises this new error?

pyiron_base/job/util.py
```diff
@@ -393,7 +394,8 @@ def _job_store_before_copy(job):
         delete_file_after_copy = True
     else:
         delete_file_after_copy = False
-    job.to_hdf()
+    if not copy_exists:
+        job.to_hdf()
```
pmrv, line 398: There's an issue, where someone might copy a job, modify it and then copy it again.  If I understand this correctly this would not be caught by this solution.

Don't have a better idea just now, though.

pyiron_base/master/generic.py
```diff
@@ -440,16 +441,28 @@ def __getitem__(self, item):
...
 
         Args:
             item (str): name of the job
 
         Returns:
 
         """
-        item_from_get_item = self.__getitem__(item=item)
-        if item_from_get_item is not None:
+        exception = False
```
pmrv, line 452: Is this related to this change or does it just pop up now on your machine?

pmrv (changes requested):

pyiron_base/job/core.py
liamhuber: I think it's to make sure the right copy gets called where there's multiple inheritance

pyiron_base/job/util.py
liamhuber: I also share this concern. Instead of the flag I'd like to try loading an instance of the job that's about to get saved and compare the pre-save, the go ahead and save if either the load failed or the two instances differ. However, I don't think we have such an equality comparison....

pyiron_base/job/core.py
raynol-dsouza: As Liam said. I could not isolate which copy it was calling in the first place, but it was definitely not calling the one defined in the class.

pyiron_base/job/util.py
raynol-dsouza: @pmrv In that case, you are absolutely right. But as @liamhuber said, unless there is another method that compares if any dataset in the job has been modified, I also can't think of a better way. I can foresee this part resulting in a larger issue.

pyiron_base/job/core.py
pmrv: Ah, that's big stinky on our part then.

pmrv:
I've tested the following modification now to your original snippet, which also works
```python
for i in range(10):
    job = ref_job.copy_to(
        project=pr,
        new_job_name='new_job_' + str(i),
        input_only=True,
        new_database_entry=False,
        delete_existing_job=True
        )
    !h5repack ref_job_size/ref_job.h5 ref_job_size/ref_job.h5.repack
    !mv ref_job_size/ref_job.h5.repack ref_job_size/ref_job.h5
    print(os.stat('ref_job_size/ref_job.h5').st_size / 1e6)
```
So that tells me it's really HDF5's copy-on-write semantics that kills us here.  We could then do something like
```python
os.system('h5repack ...')
shutil.move(...)
```
in the copy function. [(Apparently there's no API exposure of this.)](https://github.com/h5py/h5py/issues/1272)

Interestingly we also have `ProjectHDFio.rewrite_hdf5` that solves the same problem, see this snippet
```python
for i in range(10):
    job = ref_job.copy_to(
        project=pr,
        new_job_name='new_job_' + str(i),
        input_only=True,
        new_database_entry=False,
        delete_existing_job=True
        )
    ref_job.project_hdf5.rewrite_hdf5(ref_job.name)
    print(os.stat('ref_job_size/ref_job.h5').st_size / 1e6)
```
For some reason this gives me a 264kb file, while the first snippet gives me ~100kb. 

I think the following solution could work:
1. Call `rewrite_hdf5` in the copy function
2. Rip out the existing implementation of that and replace it with `subprocess.popen('h5rewrite ...')`.

raynol-dsouza:
@pmrv I read on stack that using `h5repack` can be dangerous for larger files as it reads every key+value in the existing `h5` file and writes it to a new `h5` file. This could add an unnecessary memory/time burden on larger jobs that one would want to create copies of...

Does `rewrite_hdf5` also work on the same principle as `h5repack`?

pmrv:
Yup, `rewrite_hdf5` does exactly the same, step through all groups in a file and copy them to a new file, then rename it.  It's correct that for large files and small deletions `h5repack` is no good, but since, in this case, we're always writing the whole job with `to_hdf` (and have to, since we don't know if it was modified), the size of the file should roughly double after a single copy, we're ok I think.  For reference jobs like we have here, it's not a relevant amount of I/O and even for gigabyte-sized jobs I supposed we'd rather incur ~1s of I/O rather than waste gigabytes.

raynol-dsouza:
I guess the overhead will need to be tested. It's a good fix and probably the best one we have so far.

pyiron_base/master/generic.py
raynol-dsouza: Yes, the following code from a `unittest` causes a `RecursionError`, if the now renamed `copy_self` method is called, without the exception part that I added:

```
child = pr.create_job( job_type=GenericJob,  job_name='child' )
master = pr.create_job(job_type=GenericMaster, job_name='master')
master.append(child)
master_copy = master.copy_to(project=pr, new_job_name='copy', new_database_entry=False)
```

pyiron_base/job/core.py
pmrv: I've traced it in a debugger and the issue is that `GenericJob.copy()` doesn't call `super()`.  We should fix that in separate issue.  Apparently this caused a recursion error in some master job before.

pyiron_base/master/generic.py
pmrv: I guess this is what [this](https://github.com/pyiron/pyiron_base/blob/232a144847e0ba1223eacaa6d76ac17e153c7b61/pyiron_base/job/generic.py#L458) is supposed to mean.  In another way it means someone owes us a beer.

pyiron_base/job/core.py
raynol-dsouza: Aha! And this is the same `RecursionError` that I try to get around by adding `exception = False` in `pyiron_base/master/generic.py` 

My solution then also requires me to add this exception `except (ValueError, TypeError):`. `__getitem__` throws a TypeError if the item is of NoneType, because I set `child_ids` to `None`.

pyiron_base/master/generic.py
raynol-dsouza: I'm gonna add the `code_smell` tag now... :P

pyiron_base/job/util.py
raynol-dsouza: I also believe this specific part can be a separate issue, since it has more dire implications than the (supposed) `RecursionError` fix that I have.

pyiron_base/job/core.py
liamhuber: 😬

pyiron_base/job/util.py
liamhuber: The only thing I worry about in fixing it later is that right now the code works slow but predictably; after these changes it will work fast but dangerously. (I realize I am very lazy with language here and (slow,fast)==(memory hog, memory efficient). In general I am much more comfortable with slow and predictable/safe.

pyiron_base/job/core.py
raynol-dsouza: @liamhuber someday, sometime in the future I will owe someone a beer. :P

In all seriousness, any suggestions to fix this without using what I consider my 'workaround'? I also do not like the fact that it may cause another Raynol in the future to spend time on this.

pyiron_base/job/util.py
raynol-dsouza: How about adding a tag in the `copy_to` method like `rewrite_hdf` which is by default `True`, and for use cases like mine, I can manually set it to `False`?

raynol-dsouza: This, plus @pmrv's suggestion on modifying and using `rewrite_hdf5` in `copy_to`.

jan-janssen (changes requested):
To me this pull request feels more like a workaround rather than a solution. With the renaming of `copy()` to `copy_self()` and the `Recursionerror` I have the feeling this pull request is going to cause more trouble than good in future. So I am against merging this.

raynol-dsouza:
@jan-janssen I do agree that this PR does indeed feel like a workaround. For the time being, I can use this branch for my simulations. 

What are your thoughts on @pmrv's suggession on using:

> I think the following solution could work:
>
>   Call rewrite_hdf5 in the copy function
>  Rip out the existing implementation of that and replace it with subprocess.popen('h5rewrite ...').

pyiron_base/job/core.py
liamhuber: You can replace `super().some_method(foo)` with `ParentClass.some_method(self, foo)`, provided you can find all the right places to do it.

I think the root of the evil is that our inheritance tree is....wiggly. And fixing that is a bigger job than can be attacked in this PR.

pyiron_base/job/util.py
liamhuber: Yes, if the dangerous behaviour has an energy barrier then I'm satisfied.

PR #436, opened 2021-09-17
Refactor tables

liamhuber:
Just a very simple refactoring moving the creation of the main database table out to its own module. To peek at motivation see the conversation in #272, although here there is absolutely no new code things are just moved.

### DIFF
--- a/pyiron_base/database/tables.py
+++ b/pyiron_base/database/tables.py
@@ -26,26 +26,27 @@ __status__ = "development"
 __date__ = "Sep, 2021"
 
 
-def simulation_table(table_name, metadata, extend_existing=True):
+class HistoricalTable(Table):
     """The historical table."""
-    return Table(
-        table_name,
-        metadata,
-        Column("id", Integer, primary_key=True, autoincrement=True),
-        Column("parentid", Integer),
-        Column("masterid", Integer),
-        Column("projectpath", String(50)),
-        Column("project", String(255)),
-        Column("job", String(50)),
-        Column("subjob", String(255)),
-        Column("chemicalformula", String(30)),
-        Column("status", String(20)),
-        Column("hamilton", String(20)),
-        Column("hamversion", String(50)),
-        Column("username", String(20)),
-        Column("computer", String(100)),
-        Column("timestart", DateTime),
-        Column("timestop", DateTime),
-        Column("totalcputime", Float),
-        extend_existing=extend_existing
-    )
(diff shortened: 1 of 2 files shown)


pyiron_base/database/tables.py
```diff
@@ -0,0 +1,52 @@
...
+            table_name,
+            metadata,
+            Column("id", Integer, primary_key=True, autoincrement=True),
+            Column("parentid", Integer),
+            Column("masterid", Integer),
+            Column("projectpath", String(50)),
+            Column("project", String(255)),
+            Column("job", String(50)),
+            Column("subjob", String(255)),
+            Column("chemicalformula", String(30)),
```
max-hassani: The limits for chemicalformula and hamilton have been increased to 50. In `generic.py`, it is set by `self._chem_formula_lim_length = 50`.

max-hassani:
@liamhuber I like the direction. But I see there are still several methods in the `DatabaseAccess` class, which deals directly with this table. I believe they belong to the `HistoricTable` class. Though I understand that if we move these methods to the new class (HistoricTable), we might need to make lots of changes in the job classes.

pyiron_base/database/tables.py
liamhuber: Good catch! I was just recklessly copying from the old or

max-hassani:
As a solution, I think we can keep for example `add_item()` method in `DatabaseAccess`, but it calls to `self.simulation_table.add_item()`. 
I am suggesting this so if we want to have multiple tables for experiment, continuum simulation, etc, they need their own add_item method (for example).

max-hassani:
But we can do this in another PR, when we have a solid idea of the additional tables.

pyiron_base/database/tables.py
max-hassani: I think it can be a good idea to have a limit for each column as variables. Then if necessary we can change them once somewhere, and it is reflected throughout the code.

liamhuber: Yeah, I 100% agree. I have an idea for a more ambitious refactoring that should accomplish this. Will try to get it up before the tech dev meeting

liamhuber:
@muh-hassani absolutely, I actually was thinking about this class while out for a run with the kids. From what you wrote I think we're on a super similar page.

I just pushed the chemical formula fix and will merge this shortly, since it really is just a very simple refactor and you and I are both happy with it. Then I'll crank out some code I've been thinking of and make a draft PR for you to look at the start of next week.

PR #437, opened 2021-09-17
Exploit table

liamhuber:
The idea is to increase encapsulation of our database tables by decoupling the details of the table from the more generic database access. 

I still leave in place all the old infrastructure, but now `GenericJob.save` exploits the new saving mechanism -- this still needs tests and then all the old stuff needs to be removed. I am happy to work on that, but can't do it now.

A couple issues I still don't like:
- `Settings` are passed from the job all the way to the table
- The job still talks to database access like `self.project.db.add(self, s, self.project.db.historical)`, but it would read much nicer to just have `self.project.db.historical.add(self, s)`, but I didn't figure out how to do this without passing too much responsibility from `DatabaseAccess` over to `HistoricalTable`...maybe it's just ok how it is...

Things I love:
- Column definition is handled in *one place*, not separately in `GenericJob.db_entry` *and* `DatabaseAccess`
- Encapsulation gives extensibility -- we have a template for adding more tables/breaking the historical one down into a table for objects and a table for jobs
- `DatabaseAccess` can be cleaner and not worry about crap like checking if the 'chemicalformula' is too long, which should clearly outside its responsibility. (Ok, this point is kind of just encapsulation again)

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -13,13 +13,8 @@ import time
 import os
 from datetime import datetime
 from sqlalchemy import (
-    Column,
     create_engine,
-    DateTime,
-    Float,
-    Integer,
     MetaData,
-    String,
     Table,
     text,
     and_,
@@ -30,6 +25,7 @@ from sqlalchemy.sql import select
 from sqlalchemy.exc import OperationalError, DatabaseError
 from threading import Thread, Lock
 from queue import SimpleQueue, Empty as QueueEmpty
+from pyiron_base.database.tables import simulation_table
 
 __author__ = "Murat Han Celik"
 __copyright__ = (
@@ -204,27 +200,7 @@ class DatabaseAccess(object):
 
         self._chem_formula_lim_length = 50
         self.__reload_db()
-        self.simulation_table = Table(
-            str(table_name),
-            self.metadata,
-            Column("id", Integer, primary_key=True, autoincrement=True),
-            Column("parentid", Integer),
-            Column("masterid", Integer),
-            Column("projectpath", String(50)),
-            Column("project", String(255)),
-            Column("job", String(50)),
-            Column("subjob", String(255)),
-            Column("chemicalformula", String(self._chem_formula_lim_length)),
-            Column("status", String(20)),
-            Column("hamilton", String(20)),
-            Column("hamversion", String(50)),
-            Column("username", String(20)),
-            Column("computer", String(100)),
-            Column("timestart", DateTime),
-            Column("timestop", DateTime),
-            Column("totalcputime", Float),
-            extend_existing=True,
-        )
+        self.simulation_table = simulation_table(str(table_name), self.metadata)
         self.metadata.create_all()
         self._viewer_mode = False
 
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -25,7 +25,8 @@ from sqlalchemy.sql import select
 from sqlalchemy.exc import OperationalError, DatabaseError
 from threading import Thread, Lock
 from queue import SimpleQueue, Empty as QueueEmpty
-from pyiron_base.database.tables import HistoricalTable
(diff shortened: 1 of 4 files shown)


liamhuber:
@muh-hassani take a look at this when you get a chance. There's still lots of work to be done, but I hope it heads very much in the direction you were imagining 😄

pyiron_base/database/generic.py
```diff
@@ -463,6 +472,20 @@ def _check_chem_formula_length(self, par_dict):
             par_dict[key_limited] = "OVERFLOW_ERROR"
         return par_dict
 
+    def add(self, obj, settings, target_manager: Type[TableManager]):
```
max-hassani, line 475: As far as I know, setting object does not change in job, therefore I think we can instantiate an instance of setting in the `__init__`of the TableManager class.

pyiron_base/database/generic.py
```diff
@@ -463,6 +472,20 @@ def _check_chem_formula_length(self, par_dict):
...
+                if not self._keep_connection:
+                    self.conn.close()
+                return result
+            except Exception as except_msg:
+                raise ValueError("Error occurred: " + str(except_msg))
+        else:
+            raise PermissionError("Not avilable in viewer mode.")
+
     # Item functions
     def add_item_dict(self, par_dict):
```
max-hassani, line 490: Doesn't the `add()` method replace `add_item_dict()`? I searched a bit through the repos in pyiron organizations, and found out that there might be couple of places, where we should replace `self.project.db.add_item_dict()` to `self.project.db.historical.add()`

max-hassani: Since we already have `_check_chem_formula_length()` in the `HistoricalTable` class, I think that this is absolute.

max-hassani (changes requested):
@liamhuber thanks for this nice encapsulation of the table-related methods inside the separate table classes. I made some comments, which might be useful. I would like to handle the character limits of different columns nicer, do you mind if I open a PR to this branch and discuss it further?

pyiron_base/database/generic.py
liamhuber: Agreed, but unfortunately I tried this and ran into circular import hell

pyiron_base/database/generic.py
liamhuber: Yes, I will go through later this week to kill all the dead code and update the tests. In the current commits I only add and use the new functionality without removing the old ways.

liamhuber:
> @liamhuber thanks for this nice encapsulation of the table-related methods inside the separate table classes. I made some comments, which might be useful. I would like to handle the character limits of different columns nicer, do you mind if I open a PR to this branch and discuss it further?

Not at all. For me what would be ideal would be to synchronize the columns currently defined in `_create_table` and `_object_to_entry` so that the column names appear only *once* in the code. With the changes so far at least they're now only appearing next to each other in the same file, but I think we can do even better.

From there I think it would be good if `_object_to_entry` exploited what is known about the column lengths to run a safety check on *all* columns so we can raise a clear error if *any* entry is too long. We have easy access to the limits for all the created columns with `self.table.columns.COLUMNAME.type.length` as I now exploit in `_check_chem_formula_length`.

I'm fairly confident this sort of clean bookkeeping is possible, but I don't have any particular ideas or plan for the implementation details yet. I'd be very happy if you take a crack at it and come up with something nice 😄 

On Wednesday and Thursday I would then focus on getting the old code out and tests adapted.

jan-janssen (changes requested):
I only briefly looked over this, but I have the feeling it breaks the compatibility of the filetable https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/filetable.py . Would not it be better to have one abstract class for the functions we need from our database and then implement them for the different interfaces? like we did it in https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/jobtable.py ? Rather than requiring an SQL based database which is sqlalchemy compatible?

PR #439, opened 2021-09-20
checking the character limits in the db

max-hassani:
Via the new `_check_char_limits` method, the length of the given strings for each database entries would be checked, though I am not using `self.table.columns.COLUMNAME.type.length`. Instead I use a dictionary which defines the limit and the string keys.

### DIFF
--- a/pyiron_base/database/tables.py
+++ b/pyiron_base/database/tables.py
@@ -32,6 +32,20 @@ __status__ = "development"
 __date__ = "Sep, 2021"
 
 
+class TooLongDBEntry(Exception):
+    """
+    Error: the value given as database exceeds the expected length
+    args:
+    key(str) : the column name which has the too long value
+    limit(int): the expected limit in the number of characters
+    val: the value given as the db entry
+    """
+    def __init__(self, key, limit, val):
+        message = f"Error: the value for the {key} in the database exceed the limit of {limit}" \
+                  f"the given value for the key: {val}"
+        super(TooLongDBEntry, self).__init__(message)
+
+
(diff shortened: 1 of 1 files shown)


pyiron_base/database/tables.py
```diff
@@ -104,18 +131,19 @@ def _object_to_entry(self, obj, settings) -> dict:  # : Type[GenericJob] circula
...
-        limit(int): the limit for the length of checmical formular
         """
-        key_limited = 'ChemicalFormula'
-        if key_limited in par_dict.keys() and \
-                par_dict[key_limited] is not None and \
-                len(par_dict[key_limited]) > self.table.columns.chemicalformula.type.length:
-            par_dict[key_limited] = "OVERFLOW_ERROR"
+        for key in par_dict.keys():
+            if par_dict[key] is not None and \
+                    len(par_dict[key]) > self._char_limit['key.lower()']:
```
liamhuber:
```suggestion
                    len(par_dict[key]) > self._char_limit[key.lower()]:
```

liamhuber (approved):
I like it.

One commit suggestion as it looks like theres a bug.

I also wonder if we can get a more descriptive name than `_sensitive_keys`. Once I looked at what this was doing it made tonnes of sense, but I didn't get it right away. I'm thinking `_keys_whose_length_limit_cannot_be_exceeded`, but this is reeeeaaally long...

max-hassani:
@liamhuber thanks for your feedback. Maybe a better name for this would be `restrictively_limited_size_keys`.

PR #440, opened 2021-09-20
fix an error

max-hassani:
@liamhuber this fixes the error which happens when len() is applied to a datetime.datetime object.


liamhuber:
I have to stop so it's still in a super rough state, but I drafted something over in #441 that I think will make this unnecessary.

PR #441, opened 2021-09-20
More table abstraction

liamhuber:
Towards the goal of only defining tables in **one place**.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -472,11 +472,35 @@ class DatabaseAccess(object):
             par_dict[key_limited] = "OVERFLOW_ERROR"
         return par_dict
 
-    def add(self, obj, settings, target_manager: Type[TableManager]):
+    def _add_item(self, insertion_fnc, *insertion_args):
+        """
+                Create a new database item
+
+                Args:
+                    par_dict (dict): Dictionary with the item values and column names as keys, like:
+                                      {'chemicalformula': 'BO',
+                                     'computer': 'localhost',
+                                     'hamilton': 'VAMPS',
+                                     'hamversion': '1.1',
+                                     'job': 'testing',
+                                     'subjob' : 'SubJob',
+                                     'parentid': 0L,
+                                     'myCol': 'Blubbablub',
+                                     'project': 'database.testing',
+                                     'projectpath': '/root/directory/tmp',
+                                     'status': 'KAAAA',
+                                     'timestart': datetime(2016, 5, 2, 11, 31, 4, 253377),
+                                     'timestop': datetime(2016, 5, 2, 11, 31, 4, 371165),
+                                     'totalcputime': 0.117788,
+                                     'username': 'Test'}
+
+                Returns:
+                    int: Database ID of the item created as an int, like: 3
+                """
         if not self._viewer_mode:
             try:
                 result = self.conn.execute(
-                    target_manager.add(obj, settings)
+                    insertion_fnc(*insertion_args)
                 ).inserted_primary_key[-1]
                 if not self._keep_connection:
                     self.conn.close()
@@ -485,6 +509,9 @@ class DatabaseAccess(object):
                 raise ValueError("Error occurred: " + str(except_msg))
         else:
             raise PermissionError("Not avilable in viewer mode.")
+
+    def add(self, obj, settings, target_manager: Type[TableManager]):
+        return self._add_item(target_manager.add, obj, settings)
 
     # Item functions
     def add_item_dict(self, par_dict):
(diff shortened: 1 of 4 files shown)


liamhuber:
@muh-hassani There are still uses of `add_item_dict` in `archiving.import_archive.import_jobs`, `job.generic.GenericJob.transfer_from_remote`, and `job.util._copy_database_entry` that give me cold sweats, but otherwise this branch manages to centralize column control+naming in just a *single location*. I'll keep working on it this week, but I'm really appreciating your feedback and modifications if you still have time to keep an eye on it.

jan-janssen (changes requested):
I only briefly looked over this, but I have the feeling it breaks the compatibility of the filetable https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/filetable.py . Would not it be better to have one abstract class for the functions we need from our database and then implement them for the different interfaces? like we did it in https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/jobtable.py ? Rather than requiring an SQL based database which is sqlalchemy compatible?

liamhuber:
> I only briefly looked over this, but I have the feeling it breaks the compatibility of the filetable https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/filetable.py .

@jan-janssen I'm afraid I'm not sure I know what you mean. First, this is meant to be purely a refactoring, so if it breaks *anything* then it's not doing its job. Right now all the tests are still passing -- including for filetable -- but our coverage is not amazing so before removing 'draft' I'll try to poke around to see if we're actually testing what needs to be tested.

`filtable` is a great example of what is really concerning me about our current implementations though -- the table column names are defined *three times* completely independently ([in `__init__`](https://github.com/pyiron/pyiron_base/blob/5fb4a7fb6baa162ea63d5906c4b0b7bb58d4cffa/pyiron_base/database/filetable.py#L19), [in `add_item_dict`](https://github.com/pyiron/pyiron_base/blob/5fb4a7fb6baa162ea63d5906c4b0b7bb58d4cffa/pyiron_base/database/filetable.py#L63), and [in `get_extract`](https://github.com/pyiron/pyiron_base/blob/5fb4a7fb6baa162ea63d5906c4b0b7bb58d4cffa/pyiron_base/database/filetable.py#L307)). This is a recipe for pain. I haven't gotten to filetable yet, but this is the sort of thing I'm trying to refactor away in other parts of the code.

> Would not it be better to have one abstract class for the functions we need from our database and then implement them for the different interfaces? like we did it in https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/jobtable.py ? Rather than requiring an SQL based database which is sqlalchemy compatible?

Again, I'm not entirely clear what you're driving at. If your concern is that the current highest level of abstraction for the table explicitly requires sqlalchemy types like `String` and `DateTime`, then I agree it should be possible to make an sqlalchemy-ambivalent layer on top of this storing the column names, types, and how to extract the relevant data from pyiron objects without ever making reference to their external classes. I don't see that as necessary at this moment, but I'm not opposed to it.

I admit I need to look more deeply into `job_table`, but atm I don't see how it's doing anything beyond/incompatible with what's done here. For instance, `get_db_columns` is anyhow under the hood just calling `database.get_table_headings`, right? So the current implementation looks to me like an abstraction layer *on top* of the sqlalchemy-compliant `DatabaseAccess` object, not the other way around.

liamhuber:
@muh-hassani as per Jan's comments, I think we will ultimately need to abstract to an even higher level. I.e. the true database table should exist and be able to be cast/projected down to/instantiated as either an SQL-compliant table, or a `FileTable`.

I think we can do this step-wise though, and here I'm only trying to accomplish abstraction of the SQL-compliant representation. Let me know what you think about this direction, incomplete though it may be.

 I don't think there is any concern with `JobTable`, since it is anyhow just reading from these representations and (a) tests are passing and (b) it really is just a refactor and I don't see how it would have broken anything.

max-hassani:
@liamhuber sorry for my delay. I like the stepwise approach, you suggested. But it might cause problem if we do not refactor simultaneously `FileTable` class according to the changes in `DatabaseAccess`. 
With regards to the successful tests, I have one concern. Since by the default installation of pyiron, the environment variable determines the database type (by default, I think it uses sqlite), so `Setting.database_is_disabled` is set to `False` and the database is not set for FileTable. Then by different tests of notebooks, or unittests, `FileTable` is not used at all. The only tests made for `FileTable` are checking its methods. We probably have to extend the unittests to include such case as well.
The part which takes care of setting database can be found:
 https://github.com/pyiron/pyiron_base/blob/18ec15fe898a931d108ce0de2dbbf895f10a07a6/pyiron_base/project/generic.py#L127
The problem might be that if `project.db= FileTable(project=path)`, then when `job.project.db.add()` is called, it would result in an error as the `FileTable` does not have an `add()` method.

liamhuber:
> The problem might be that if project.db= FileTable(project=path), then when job.project.db.add() is called, it would result in an error as the FileTable does not have an add() method.

@muh-hassani this is a super good observation, I guess this is exactly the broken compatibility with Filetable that Jan was talking about.

In principle this is probably still doable, but in practice I think it will be easier to take the lessons learned here and apply them to a new table that is much more abstract and then can be projected down to an SQL or file representation.

PR #445, opened 2021-09-21
Force overwriting into the same group

### LINKED ISSUE #443: ValueError when group_name is None
When trying to run the following job:

```
job1d_square = pr.create.job.TISE('tise1d_square', delete_existing_job=True)
job1d_square.input.potential = SquareWell(depth=10)
job1d_square.input.mesh = RectMesh(bounds=10, divisions=100)
job1d_square.input.n_states = 10
job1d_square.input.mass = 1
job1d_square.run()
```

I run into the following error:

```
~/pyiron/pyiron_continuum/pyiron_continuum/schroedinger/schroedinger.py in run_static(self)
    177         self.output.energy = eigenvalues
    178         self.output.psi = np.array([np.reshape(v, self.mesh.shape[1:]) for v in eigenvectors.T])
--> 179         self.to_hdf()
    180         self.status.finished = True
    181 

~/pyiron/pyiron_base/pyiron_base/job/template.py in to_hdf(self, hdf, group_name)
     37 
     38     def to_hdf(self, hdf=None, group_name=None):
---> 39         HasStorage.to_hdf(self, hdf=self.project_hdf5)
     40 
     41     def from_hdf(self, hdf=None, group_name=None):

~/pyiron/pyiron_base/pyiron_base/interfaces/has_hdf.py in to_hdf(self, hdf, group_name)
    175         with _WithHDF(hdf, group_name) as hdf:
    176             if len(hdf.list_dirs()) > 0 and group_name is None:
--> 177                 raise ValueError("HDF group must be empty when group_name is not set.")
    178             self._to_hdf(hdf)
    179             self._store_type_to_hdf(hdf)

ValueError: HDF group must be empty when group_name is not set.
```

This error message is raised since in the 2nd stack, group_name is None. It pops up from the change in line 176 in PR #434.

pmrv:
Fixes #443.  Passing `''` essentially silences the error and forces `HasHDF` to write into the group that `GenericJob.to_hdf` has already written into.  

As I said the underlying queezyness comes from the fact that `GenericJob.to_hdf` is not from `HasHDF` yet.

### DIFF
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -37,11 +37,11 @@ class TemplateJob(GenericJob, HasStorage):
 
     def to_hdf(self, hdf=None, group_name=None):
         GenericJob.to_hdf(self, hdf=hdf, group_name=group_name)
-        HasStorage.to_hdf(self, hdf=self.project_hdf5, group_name='storage')
+        HasStorage.to_hdf(self, hdf=self.project_hdf5, group_name='')
 
     def from_hdf(self, hdf=None, group_name=None):
         GenericJob.from_hdf(self, hdf=hdf, group_name=group_name)
-        HasStorage.from_hdf(self, hdf=self.project_hdf5, group_name='storage')
+        HasStorage.from_hdf(self, hdf=self.project_hdf5, group_name='')
 
 
 class PythonTemplateJob(TemplateJob):


liamhuber (approved):
'queezyness' is a good description. Your solution is definitely the elegance I was looking for though! At least now the actual final HDF structure comes out nice

PR #446, opened 2021-09-21
Use HasHDF in FlattenedStorage

pmrv:


pmrv:
This might break `StructureStorage` in atomistics next week, but it's an easy fix I'll do on the next version bump.

pyiron_base/generic/flattenedstorage.py
```diff
@@ -493,22 +484,20 @@ def write_array(name, array, hdf):
         self._resize_elements(self.num_elements)
         self._resize_chunks(self.num_chunks)
 
-        with hdf.open(group_name) as hdf_s_lst:
-            self._type_to_hdf(hdf_s_lst)
-            hdf_s_lst["num_elements"] =  self._num_elements_alloc
-            hdf_s_lst["num_chunks"] = self._num_chunks_alloc
+        hdf["num_elements"] =  self._num_elements_alloc
```
liamhuber, line 487: is there a reason `_num_elements_alloc` can't live on `self.storage` instead of `self`? If it lives in storage then it doesn't need to be explicitly handled here. Since it's something that is being stored, I would suggest to instead stick it in storage. You can still make a (set)getter for it if you want users to be able to access it directly from the object.

pmrv: This is only `HasHDF` not `HasStorage`, `FlattenedStorage` doesn't have a `DataContainer` associated and imo doesn't need one, since it's simple attributes (no nesting) anyway.

liamhuber: Ah, sorry, my bad! But if you made it with `HasStorage` then your `_to_hdf` could be much shorter 😉

PR #447, opened 2021-09-21
Use HasHDF in Server

pmrv:


pmrv:
`Server` is not actually `to_object()`-able, so I think it doesn't actually make sense to convert it.

PR #448, opened 2021-09-21
Update HasHDF docs when dual-inheriting

pmrv:
Inheriting from classes that have already an to_hdf/from_hdf and
HasHDF is a bit tricky.  Updated docstring to show how to do it
correctly.

See #444 and friends for where this popped up.

### DIFF
--- a/pyiron_base/interfaces/has_hdf.py
+++ b/pyiron_base/interfaces/has_hdf.py
@@ -109,7 +109,6 @@ class HasHDF(ABC):
     When using this class as a mixin that also derive from classes that have a
     legacy implementation here's a simple recipe
 
-    >>> from pyiron_base.jobs.generic import GenericJob
     >>> class MyOldClass:
     ...     def to_hdf(self, hdf, group_name):
     ...         ... # whatever you need to save
@@ -130,6 +129,7 @@ class HasHDF(ABC):
     If you're deriving from :class:`GenericJob` it will already take care of
     descending into group_name, so you can pass `""` as the group_name like so
 
+    >>> from pyiron_base import GenericJob
     >>> class MyHybridJob(GenericJob, HasHDF):
     ...     def to_hdf(self, hdf, group_name):
     ...         GenericJob(self, hdf=hdf, group_name=group_name)
--- a/pyiron_base/interfaces/has_hdf.py
+++ b/pyiron_base/interfaces/has_hdf.py
@@ -112,15 +112,16 @@ class HasHDF(ABC):
     >>> from pyiron_base.jobs.generic import GenericJob
     >>> class MyOldClass:
     ...     def to_hdf(self, hdf, group_name):
-    ...         pass # snip
+    ...         ... # whatever you need to save
     ...     def from_hdf(self, hdf, group_name):
-    ...         pass # snip
+    ...         ... # whatever you need to restore
     >>> class MyDerivedClass(MyOldClass, HasHDF):
     ...     def to_hdf(self, hdf, group_name):
     ...         MyOldClass.to_hdf(self, hdf=hdf, group_name=group_name)
     ...         HasHDF.to_hdf(self, hdf=hdf, group_name=group_name)
     ...     def from_hdf(self, hdf, group_name):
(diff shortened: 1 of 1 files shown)


pyiron_base/interfaces/has_hdf.py
```diff
@@ -106,6 +106,35 @@ class HasHDF(ABC):
     >>> hdf.remove_file()
     >>> pr.remove(enable=True)
 
+    When using this class as a mixin that also derive from classes that have a
+    legacy implementation here's a simple recipe
+
```
liamhuber, line 111:
```suggestion

    >>> from pyiron_base.jobs.generic import GenericJob
```

pyiron_base/interfaces/has_hdf.py
```diff
@@ -106,6 +106,35 @@ class HasHDF(ABC):
     >>> hdf.remove_file()
     >>> pr.remove(enable=True)
 
+    When using this class as a mixin that also derive from classes that have a
```
liamhuber:
```suggestion
    When using this class as a mixin that also derives from classes having a
```

pyiron_base/interfaces/has_hdf.py
liamhuber: I think this is what's causing the tests to fail

pyiron_base/interfaces/has_hdf.py
```diff
@@ -106,6 +106,35 @@ class HasHDF(ABC):
...
+    legacy implementation here's a simple recipe
+
+    >>> class MyOldClass:
+    ...     def to_hdf(self, hdf, group_name):
+    ...         pass # snip
```
liamhuber: I'm actually not familiar with `# snip`. Is this to imply that the old class might actually have code here?

pyiron_base/interfaces/has_hdf.py
```diff
@@ -106,6 +106,35 @@ class HasHDF(ABC):
...
+    ...     def from_hdf(self, hdf, group_name):
+    ...         pass # snip
+    >>> class MyDerivedClass(MyOldClass, HasHDF):
+    ...     def to_hdf(self, hdf, group_name):
+    ...         MyOldClass.to_hdf(self, hdf=hdf, group_name=group_name)
+    ...         HasHDF.to_hdf(self, hdf=hdf, group_name=group_name)
+    ...     def from_hdf(self, hdf, group_name):
+    ...         pass # snip
```
liamhuber: I'm confused, don't we need to `from_hdf` both parents here?

pyiron_base/interfaces/has_hdf.py
```diff
@@ -106,6 +106,35 @@ class HasHDF(ABC):
...
+
+    i.e. explicitly call both methods with the same group_name.  The call to
+    :meth:`.HasHDF.to_hdf` has to be last so that the type information is
```
liamhuber, line 126:
```suggestion
    :meth:`HasHDF.to_hdf` has to be last so that the type information is
```

I think? If that extra dot was intentional then ignore this suggestion and just explain to me what I'm missing ;)

liamhuber (changes requested):
Very good idea.

A few suggestions, one of which will hopefully resolve the failing tests.

pyiron_base/interfaces/has_hdf.py
pmrv: Ah, I was just lazy yesterday and thought "snip" is understandable as "more of the same".  I'll bulk it out later.

pyiron_base/interfaces/has_hdf.py
pmrv: Normally you have to give the full module path to an object so that sphinx does the referencing correctly.  With the dot you can skip part of it.

pyiron_base/interfaces/has_hdf.py
liamhuber: Maybe it is clear, it might just be my Boomer showing 😂

liamhuber (approved):

PR #449, opened 2021-09-22
Refactor settings

liamhuber:
Just some minor housekeeping for readability

### DIFF
--- a/pyiron_base/settings/generic.py
+++ b/pyiron_base/settings/generic.py
@@ -93,9 +93,11 @@ class Settings(metaclass=Singleton):
 
         # Take dictionary as primary source - overwrite everything
         self._read_external_config(config=config)
-        if "CONDA_PREFIX" in environment.keys() and os.path.exists(os.path.join(environment["CONDA_PREFIX"], "share", "pyiron")):
+        if "CONDA_PREFIX" in environment.keys() \
+                and os.path.exists(os.path.join(environment["CONDA_PREFIX"], "share", "pyiron")):
             self._configuration["resource_paths"].append(os.path.join(environment["CONDA_PREFIX"], "share", "pyiron"))
-        elif "CONDA_DIR" in environment.keys() and os.path.exists(os.path.join(environment["CONDA_DIR"], "share", "pyiron")):
+        elif "CONDA_DIR" in environment.keys() \
+                and os.path.exists(os.path.join(environment["CONDA_DIR"], "share", "pyiron")):
             self._configuration["resource_paths"].append(os.path.join(environment["CONDA_DIR"], "share", "pyiron"))
 
         self._configuration["project_paths"] = [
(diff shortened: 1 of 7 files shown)


liamhuber:
On my local machine the install test needs `os.remove(os.path.join(execution_path, "pyiron.log"))` in `TearDownClass` or I get a `pyiron.log` leftover in the test directory. But on the CI this gives 

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/pyiron_base/pyiron_base/tests/settings/pyiron.log'
```

For now I just wrapped the removal in a `try/except FileNotFoundError` because I want to keep my system clean, but I'd feel better understanding where `pyiron.log` winds up on the CI system if not at `execution_path`...

pyiron_base/settings/generic.py
```diff
@@ -226,15 +232,15 @@ def switch_to_local_database(self, file_name="pyiron.db", cwd=None):
             file_name (str): SQLite database file name
             cwd (str/None): directory where the SQLite database file is located in
         """
-        if not self._use_local_database:
+        if self._use_local_database:
+            print("Database is already in local mode or disabled!")
```
pmrv: Could be a logger call.

tests/settings/test_generic_file_config.py
```diff
@@ -51,7 +51,3 @@ def test_file_top_path(self):
 
     def test_file_login_user(self):
         self.assertEqual(self.file_config.login_user, "pyiron")
-
-
-if __name__ == "__main__":
-    unittest.main()
```
jan-janssen, line 57: Why is this part removed? I though we need it to run the tests in the continuous integration environment.

tests/settings/test_generic_test_config.py
```diff
@@ -66,7 +67,3 @@ def test_resource_paths(self):
 
     def test_login_user(self):
         self.assertEqual(self.test_config.login_user, "pyiron")
-
-
-if __name__ == "__main__":
-    unittest.main()
```
jan-janssen, line 72: same applies here

tests/settings/test_install.py
```diff
@@ -35,6 +39,3 @@ def test_install(self):
         self.assertIn("RESOURCE_PATHS", content[2])
         self.assertTrue(os.path.exists(os.path.join(self.execution_path, "project")))
         self.assertTrue(os.path.exists(os.path.join(self.execution_path, "resources")))
-
-if __name__ == "__main__":
-    unittest.main()
```
jan-janssen, line 40: same applies here

tests/settings/test_generic_file_config.py
liamhuber: That's a very good warning!!! It's not necessary on my local machine so I thought I was removing dead code. I'll just burn some CPU minutes and do a brute-force tests by throwing a failing test in one of these suites to see if the CI catches it.

liamhuber: Indeed, it's safe to remove the `__main__` boilerplate. I added this to one of the modules that had had the boilerplate stripped and the CI caught it no problem:

```python
    def test_should_fail(self):
        self.assertTrue(False)
```

I'm going to revert the change and resolve the other two instances, but it was a very good call to double check this behaviour!!!

PR #451, opened 2021-09-23
Provide version and git repo in maintenance

niklassiemer:


jan-janssen (approved):
Maybe it makes sense to move `Maintenance` and `GlobalMaintenance` to a separate module so it is easier to find for new developers, other than that this pull request looks fine to me.

PR #452, opened 2021-09-23
Cleanup

samwaseda:
This clears some duplicated functions and DocStrings. For me personally, this is the preparation for updating `InteractiveWrapper`, which will be a different PR.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1036,7 +1036,7 @@ class GenericJob(JobCore):
         job._init_child_job(self)
         return job
 
-    def update_master(self):
+    def update_master(self, force_update=False):
         """
         After a job is finished it checks whether it is linked to any metajob - meaning the master ID is pointing to
         this jobs job ID. If this is the case and the master job is in status suspended - the child wakes up the master
@@ -1048,16 +1048,16 @@ class GenericJob(JobCore):
         master_id = self.master_id
         project = self.project
         self._logger.info("update master: {} {} {}".format(master_id, self.get_job_id(), self.server.run_mode))
-        if (
-            master_id is not None
-            and not self.server.run_mode.thread
-            and not self.server.run_mode.modal
-            and not self.server.run_mode.interactive
-        ):
-            self._reload_update_master(
-                project=project,
-                master_id=master_id
-            )
+        if master_id is not None:
+            if force_update or (
+                not self.server.run_mode.thread
+                and not self.server.run_mode.modal
+                and not self.server.run_mode.interactive
+            ):
+                self._reload_update_master(
+                    project=project,
+                    master_id=master_id
+                )
 
     def job_file_name(self, file_name, cwd=None):
         """
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1052,12 +1052,14 @@ class GenericJob(JobCore):
         master_id = self.master_id
         project = self.project
         self._logger.info("update master: {} {} {}".format(master_id, self.get_job_id(), self.server.run_mode))
-        if master_id is not None:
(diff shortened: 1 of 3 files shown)


pyiron_base/master/generic.py
```diff
@@ -277,23 +277,10 @@ def _after_generic_copy_to(self, original, new_database_entry, reloaded):
...
-        self._logger.info("update master: {} {} {}".format(master_id, self.get_job_id(), self.server.run_mode))
-        if master_id is not None:
-            self._reload_update_master(
-                project=project,
-                master_id=master_id
-            )
+    def update_master(self, force_update=True):
+        super().update_master(force_update=force_update)
+
+    update_master.__doc__ = GenericJob.update_master.__doc__
```
liamhuber, line 283:
```suggestion
```

If it's a bare super call you can even just directly inherit the method.

Sorry for sticking my nose in, I was just jazzed to see a branch called 'cleanup' 😍

samwaseda: XD

yeah the problem is `force_update` which is `True` in `GenericMaster` and `False` in `GenericJob` (parent class).

Is it anyway something to apologise for if someone starts reviewing without being asked to do so?

liamhuber: > yeah the problem is force_update which is True in GenericMaster and False in GenericJob (parent class).

Ah, gotcha! Missed that subtlety. Shouldn't that mean the docstring can't be copy-pasted then?

> Is it anyway something to apologise for if someone starts reviewing without being asked to do so?

No, just being Canadian 😆 🇨🇦

samwaseda: > > yeah the problem is force_update which is True in GenericMaster and False in GenericJob (parent class).
> 
> Ah, gotcha! Missed that subtlety. Shouldn't that mean the docstring can't be copy-pasted then?

Ah good point. I'll change that.

> 
> > Is it anyway something to apologise for if someone starts reviewing without being asked to do so?
> 
> No, just being Canadian 😆 🇨🇦

ah good now I know in which country I'm gonna be socially dead XD

pyiron_base/job/generic.py
```diff
@@ -1036,28 +1037,31 @@ def create_job(self, job_type, job_name, delete_existing_job=False):
...
-            and not self.server.run_mode.thread
-            and not self.server.run_mode.modal
-            and not self.server.run_mode.interactive
-        ):
-            self._reload_update_master(
-                project=project,
-                master_id=master_id
-            )
+        if master_id is not None:
+            if force_update or (
```
jan-janssen: Why do you need two if statements?

samwaseda: Readability. I find separate if statements easier than one encapsulated one. But I can also put it in one if-statement if you prefer.

pyiron_base/job/generic.py
```diff
@@ -1036,28 +1037,31 @@ def create_job(self, job_type, job_name, delete_existing_job=False):
...
+                not self.server.run_mode.thread
+                and not self.server.run_mode.modal
+                and not self.server.run_mode.interactive
+            ):
```
liamhuber:
```suggestion
        if master_id is not None and (
            force_update or
            not (
                self.server.run_mode.thread
                or self.server.run_mode.modal
                or self.server.run_mode.interactive
            )
        ):
```

@jan-janssen @samwaseda what about something like this? I share Jan's concern about the artificial extra if statement, but I agree with Sam that we should work on readability. In particular I personally would like to avoid the repeated `and not`. In general I'd like to use `not` as sparingly as possible as it adds a bit of cognitive load each time we do it.

jan-janssen (approved):
Looks good to me

PR #456, opened 2021-09-27
Integration tests

niklassiemer:
These changes trigger the integration workflows 
- if the integration label is set
- for any push to the PR if the integration label is set
- on push to master
- (if any label is added to a PR, where the integration label is also present - could be filtered with a longer if statement, however I consider this a really minor case)

### DIFF
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -4,12 +4,18 @@
 name: Compatibility with pyiron_atomistics
 
 on:
+  push:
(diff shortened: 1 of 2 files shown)


pmrv (approved):
Nice!

niklassiemer:
The x is since I just triggered the integration run by setting the corresponding label and canceled the workflow directly after it has been started. The integration tests currently also run on the master commit. Thus, this setting seems to be working as expected and I will make the same changes to `pyiron_atomistics`.

PR #459, opened 2021-09-29
Remove HDF5Content

pmrv:
As far as I can tell this is not used anywhere, so I just dropped it.


pmrv:
Ah, from the tests I suppose the point of `HDF5Access` is that you can do

```python
job = pr.inspect('...')
job.content.foo == job['foo']
```

I'll move it to `JobPath` then with some docs later.

PR #460, opened 2021-10-05
search function for DataContainer

freyso:
Implement the search_tree function to DataContainer as suggested in pyiron_atomistics issue #382.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -331,7 +331,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
                 res = self.search (key[1])
                 if (len(key)== 2):
                    res = val
-                else
+                else:
                    res[key[2:]] = val
                 return
             if key[0] not in self.keys():
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -288,7 +288,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
 
         if isinstance(key, tuple):
             if (key[0] == "..." and len(key)>1):
-                res = self.search (key[1])
+                res = self.search (key[1], False)
                 return res if (len(key)== 2) else res[key[2:]]
             return self[key[0]][key[1:]]
 
@@ -328,11 +328,8 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
 
         if isinstance(key, tuple):
             if (key[0] == "..." and len(key)>1):
-                res = self.search (key[1])
-                if (len(key)== 2):
-                   res = val
-                else:
-                   res[key[2:]] = val
+                res = self._search_parent (key[1], False)
+                res[key[1:]] = val
                 return
             if key[0] not in self.keys():
                 self[key[0]] = type(self)()
@@ -520,13 +517,14 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
         else:
             return super().get(key, default=default)
 
-    def search(self, key, fail_if_not_found=True):
+    def search(self, key, stop_on_first_hit=True):
         """
         Search for ``key`` in the Container hierarchy.
 
-        This should be used if there is only one item in the hierarchy.
-        If the key appears multiple times, there is no control over which
-        item is found.
+        This should be used if there is only one such item in the hierarchy.
+
+        If stop_on_first_hit is True the first item found is taken.
+        Otherwise, a ValueError is raised if the key appears multiple times.
 
         Args:
             key (str):                the key to look for
@@ -535,7 +533,8 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
                                       False => return None
         Raise:
             TypeError:  if key is not str
-            IndexError: if key is not found and fail_if_not_found is True.
+            KeyError:   if key is not found
+            ValueError: if stop_on_first_hit is False and key is found twice
 
         Returns:
             object: element at ``key``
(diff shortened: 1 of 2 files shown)


pyiron_base/generic/datacontainer.py
```diff
@@ -510,6 +510,43 @@ def get(self, key, default=None, create=False):
...
+        if not isinstance (key,str):
+            raise TypeError ("Cannot search for non-string key.")
+
+        # search within current level
+        if (key in self): return self.get(key)
+
+        # descend into subgroups
+        for it in self.groups ():
+            pick = search_tree(self[it], key, False)
+            if (not pick is None): return pick
```
pmrv: Did you add `fail` just for this check or is that something we want to expose to users?  If the former I think always trowing an exception/catching it and not having the argument feels a bit cleaner.

pyiron_base/generic/datacontainer.py
```diff
@@ -510,6 +510,43 @@ def get(self, key, default=None, create=False):
...
+
+        if not isinstance (key,str):
+            raise TypeError ("Cannot search for non-string key.")
+
+        # search within current level
+        if (key in self): return self.get(key)
+
+        # descend into subgroups
+        for it in self.groups ():
+            pick = search_tree(self[it], key, False)
```
pmrv: Thinking about it again, I guess for the small containers depth-breadth first doesn't make much of a performance difference, so depth-first is fine with me.

pyiron_base/generic/datacontainer.py
```diff
@@ -510,6 +510,43 @@ def get(self, key, default=None, create=False):
         else:
             return super().get(key, default=default)
 
+    def search_tree(self, key, fail=True):
```
pmrv: I'm thinking to just call it `search`, since in general the `DataContainer` docs don't talk about "trees".  If that sounds to general, we could go with 'recursive' or 'nested' which is more consistent with the normal terminology in the docs.

pmrv (commented):
Looks pretty good generally.  I have some PEP8 nits, but I can take care of those once we discuss everything else.

pyiron_base/generic/datacontainer.py
pmrv:
```suggestion
            pick = self[it].search_tree(key, False)
```
;P

pyiron_base/generic/datacontainer.py
freyso: I do a recursion here. Inside the recursion, not finding the key is not an error - it just means I didn't search the right part yet. Only if the key is nowhere to be found, there is a true problem.
There is different alternatives to the current solution:
- simply return None if key is not found (no fail needed). Disadvantage - you don't get a very useful error message if you had expected the key to be found. 
- always raise an IndexError, but catch it inside the recursion. I had once learned that one should use exceptions for exceptional cases, because exception handling is not designed to be performant. I don't know if this is true for python. That's why I decided against this solution

I went for the current solution because it allows the user to decide whether or not he/she prefers an error, or None return value.

pyiron_base/generic/datacontainer.py
freyso: sounds good. I'll do it.

pyiron_base/generic/datacontainer.py
freyso: Agreed. I had developed this outside the class and forgot to do it right.

pyiron_base/generic/datacontainer.py
freyso: renamed "fail" into "fail_if_not_found" to make more clear what this arg is about.

freyso:
I just realized that DataContainer supports slash notation, i.e.,
``job.input.sphinx["main/ricTS/scfDiag/dEnergy"]``
we could add search into this like this:
``job.input.sphinx["main/.../scfDiag/dEnergy"]`` and ``job.input.sphinx[".../scfDiag/dEnergy"]``
assuming that "..." is a very unlikely key.

pmrv:
> I just realized that DataContainer supports slash notation, i.e., `job.input.sphinx["main/ricTS/scfDiag/dEnergy"]` we could add search into this like this: `job.input.sphinx["main/.../scfDiag/dEnergy"]` and `job.input.sphinx[".../scfDiag/dEnergy"]` assuming that "..." is a very unlikely key.

I like it, I'll see where to best add it.

freyso:
> > I just realized that DataContainer supports slash notation, i.e., `job.input.sphinx["main/ricTS/scfDiag/dEnergy"]` we could add search into this like this: `job.input.sphinx["main/.../scfDiag/dEnergy"]` and `job.input.sphinx[".../scfDiag/dEnergy"]` assuming that "..." is a very unlikely key.
> 
> I like it, I'll see where to best add it.

I have an implementation at hand, would be a fix to __getitem__ . Shall I add it?

pmrv:
> > > I just realized that DataContainer supports slash notation, i.e., `job.input.sphinx["main/ricTS/scfDiag/dEnergy"]` we could add search into this like this: `job.input.sphinx["main/.../scfDiag/dEnergy"]` and `job.input.sphinx[".../scfDiag/dEnergy"]` assuming that "..." is a very unlikely key.
> > 
> > 
> > I like it, I'll see where to best add it.
> 
> I have an implementation at hand, would be a fix to **getitem** . Shall I add it?

Sure, sounds good!

pyiron_base/generic/datacontainer.py
niklassiemer: I do not like to rely on `None` here. `None` might be a valid value for a given key! How about `return`ing the error instead of raising?

niklassiemer (commented):
I really like the functionality! We should make sure that `'...'` is a protected key, though!
One thing which concerns me a bit is the danger of having multiple matches for a key using the `__getitem__` and even more for the `__setitem__` methods. In these cases, I would prefer a full search and only get/set if there is only one occurrence.

pyiron_base/generic/datacontainer.py
pmrv: I did quick benchmark suggesting that raising errors instead of returning `None` gives a performance penalty of ~0.5%.  I'd therefore say we just go with exceptions, but I'll do profile and more thorough benchmark once I'm in office.

freyso:
> I really like the functionality! We should make sure that `'...'` is a protected key, though! 

this is not really necessary. First, '...' is an insane key to begin with. Second, the current implementation takes '...' as implicit search only if it occurs in a tuple index (or slash notation). So you are able to use '...' as a regular key, but then you cannot access it in a tuple. This is such a small limitation that I believe one can risk a surprise for the innocent(?) user.

> One thing which concerns me a bit is the danger of having multiple matches for a key using the `__getitem__` and even more for the `__setitem__` methods. In these cases, I would prefer a full search and only get/set if there is only one occurrence.

I don't like it performance-wise, but I agree in terms of detecting surprises. I'll adapt.

freyso:
About returning None: Agreed, I'll adapt when I fix the current bug in setitem for the final position. Solution is to have an internal function that returns not the searched item, but the parent group. If that is None, the key was not found.

pmrv:
> > One thing which concerns me a bit is the danger of having multiple matches for a key using the `__getitem__` and even more for the `__setitem__` methods. In these cases, I would prefer a full search and only get/set if there is only one occurrence.
> 
> I don't like it performance-wise, but I agree in terms of detecting surprises. I'll adapt.

From the benchmark there's about a factor 1000 between the best (recursion never fails) and the worst case (recursion always fails) due to the depth-first searching. So from my side I'm ok with picking only the first match in both `__setitem__` and `__getitem__`, as long as it's consistent (which it is right now).  Then again for a three level nested container with 26 groups at each level even the worst case search is only 18ms, which is still ok for interactive use.  @niklassiemer thoughts?

pmrv:
> > I really like the functionality! We should make sure that `'...'` is a protected key, though!
> 
> this is not really necessary. First, '...' is an insane key to begin with. Second, the current implementation takes '...' as implicit search only if it occurs in a tuple index (or slash notation). So you are able to use '...' as a regular key, but then you cannot access it in a tuple. This is such a small limitation that I believe one can risk a surprise for the innocent(?) user.

Agreed.

niklassiemer:
With 18ms I lean towards paying that cost to be on the save side. For the search function itself, I am actually happy with both solutions, since we have at least a waring in the docstring for someone calling this function. Indeed, if we want to use the same function for a full search, we would also have another bool, e.g. `save_search=False` which could trigger a full search?

freyso:
@niklassiemer, @pmrv : I have implemented the duplicate check for the .../ notation, as suggest by @niklassiemer , mostly because I think that balance between performance and avoiding surprises must be on the latter side. Most every-day DataContainers will not be arbitrarily complex, so the actual timing will be below human threshold. The entire feature is meant to be convenience for simple things; complicated use cases need problem-specific search strategies.

If performance become an issue, the "search" function allows to do it on a first-hit basis (there is an argument ``stop_on_first_hit` with a default=True). Or people better remember where they put their data... 

The full search is not yet implemented. I am not sure there is an obvious use case. If there is, I probably would go for an iterator generator type of implementation:
```
for it in container.search_all ("key"):
    ...
```
which would then allow to do all kinds of reasonable stuff. The implementation could be copy&pasted from _search_parent and replacing the return statements by ``yield``s.

But I suggest to postpone this until someone has a use case.

niklassiemer (approved):
I am very happy with the current implementation! I opened PR #466 to this branch with some nits, mostly formatting and one docstring issue I found.

freyso:
Thanks, Niklas, for the cleanup!

pmrv (approved):
Looks good to me.  Thanks for the contribution @freyso !

PR #462, opened 2021-10-07
Remove JobCore.load_object and clean up around it

pmrv:
It was only used in a few places and only convoluted the logic around loading jobs from job core objects.  With this the only way to instantiate a job from an inspect object is `to_object`.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -541,14 +541,11 @@ class JobCore(HasGroups):
             convert_to_object=False
         )
 
-    def load_object(self, convert_to_object=True, project=None):
+    def load_object(self, project=None):
         """
         Load object to convert a JobPath to an GenericJob object.
 
         Args:
-            convert_to_object (bool): convert the object to an pyiron object or only access the HDF5 file - default=True
-                                      accessing only the HDF5 file is about an order of magnitude faster, but only
-                                      provides limited functionality. Compare the GenericJob object to JobCore object.
             project (ProjectHDFio): ProjectHDFio to load the object with - optional
 
         Returns:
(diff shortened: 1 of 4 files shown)


niklassiemer (commented):
I actually not fully understand the difference between `to_object` and `load_object`. If `convert_to_object=False` it just returns self and otherwise it calls `to_object`? 
I think this PR does what it claims and looks good to me (and the tests pass 👍 )

pmrv:
> I actually not fully understand the difference between `to_object` and `load_object`.

Me neither, and there's no observable difference, hence I removed it.

niklassiemer:
> > I actually not fully understand the difference between `to_object` and `load_object`.
> 
> Me neither, and there's no observable difference, hence I removed it.

Perfect, then let's float this around for one or two days such that everybody can have a quick glance at it.

liamhuber:
I just snooped real quick, but is there the opportunity to do a bit of refactoring to avoid some of the code duplication in `load_from_jobpath_string` and `load_from_jobpath`? 

> then let's float this around for one or two days such that everybody can have a quick glance at it.

Otherwise it all sounds good to me.

pmrv:
> I just snooped real quick, but is there the opportunity to do a bit of refactoring to avoid some of the code duplication in `load_from_jobpath_string` and `load_from_jobpath`?

`load_from_jobpath_string` is on my hit list.  As far as I can tell it's never used and if it was it should be a class method on `JobCore` and not a method on `Project`.

liamhuber:
> > I just snooped real quick, but is there the opportunity to do a bit of refactoring to avoid some of the code duplication in `load_from_jobpath_string` and `load_from_jobpath`?
> 
> `load_from_jobpath_string` is on my hit list. As far as I can tell it's never used and if it was it should be a class method on `JobCore` and not a method on `Project`.

Super, no need to mess with it here if you're already aware and scheming 😉

PR #463, opened 2021-10-07
Save pyiron tables via DataContainer instead of csv

### LINKED ISSUE #353: Loading pyiron table from HDF doesn't preserve some types
I have a pyiron table that stored some ~lists~ numpy arrays, on reloading them those entries have been turned into strings.

pmrv:
All data in the table are now saved via datacontainer and when loaded
from HDF5 the data is taken from there instead of from the csv file.
This makes sure complex data are correctly instantiated when reloading a
table, since writing to csv forces a text representation, which cannot
be converted back on loading.

This tests the new behavior for numpy
arrays, but structures, etc. are probably also affected.


niklassiemer (commented):
I like this new behavior! Are there any problems with backwards compatibility? Do we write the csv file after all?

liamhuber (approved):
My perspective is pretty naive here, but I like it. At the cost of introducing `_save_output` (which is anyhow quite readable), the `to/from_hdf` see nice readability improvements.

pmrv:
> I like this new behavior! Are there any problems with backwards compatibility? Do we write the csv file after all?

The csv file is still written, but not read anymore by our code.  I am unaware how many people actually look at it for their workflows and I guess it's nice to have, so I think we should keep it around.

Backwards compatibility is actually an issue that I hadn't thought about.  The table content was always written to HDF, just never used, so I could either put back the code that reads the csv or load from HDF in case it's reading an old job.

jan-janssen:
Somehow this broke my calculation - the worst part is it just hangs without any error message.

PR #467, opened 2021-10-11
All include header in job_table even if empty

pmrv:
Previously calling `job_table` on an empty project gives an empty data frame with no header.  This breaks indexing and some of our related methods of `Project` (e.g. `get_jobs_status()`).  This PR just makes sure the columns are always defined.


niklassiemer (approved):
Looks good to me!

PR #468, opened 2021-10-11
Empty queue fix

jan-janssen:


pmrv:
Is this still an issue after #467?

PR #469, opened 2021-10-12
Make FileHDFio a MutableMapping

pmrv:
With this `FileHDFio` and `ProjectHDFio` behave like normal python mappings.  There is still a small bug in `DataContainer` that doesn't allow wrapping of HDF5 files. I'll fix that in a separate PR.

PR #471, opened 2021-10-13
Use a blacklist from wrapping values

pmrv:
When initializing `DataContainer`s, previously only values that are list,
dict, or tuples were recursively wrapped into `DataContainer`s.  Now
`__init__` and `update` take an explicit black list of types that are
mappings but should not be wrapped.  This changes the behavior such that
more values are now wrapped, unless explicitly black listed.

This allows to wrap e.g. HDF5 files as returned from h5py.  For our `FileHDFio` objects #469 is also needed.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -274,7 +274,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
         self.table_name = table_name
         self._lazy = lazy
         if init is not None:
-            self.update(init, wrap=True, blacklist=wrap_blacklist)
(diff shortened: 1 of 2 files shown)


max-hassani:
@pmrv, 👍. The code looks good to me.  Will you also add a test to load a hdf5 file into a `DataContainer` to ensure the functionality?

pmrv:
Added tests and found some bugs.  Force-push to include #469 .

pmrv:
`DataContainer.__init__` keeps becoming more complicated, so sooner or later we should break the arguments related to `init` into its own method called `from_mapping(init, wrap_blacklist)` or something like that.

PR #473, opened 2021-10-14
Test against more versions for backwards compatibility

pmrv:
We somehow forgot to update the backwards tests after bumping minor level version.  I rewrote the script to save jobs from version 0.1.48 (the last 0.1 release), 0.2.24 (the last 0.2) and 0.3.0.  

Generally the backwards tests could use some more love and harnessing to make it easier to write them, but I don't have the time right now.

### DIFF
--- a/.ci_support/test_backwards.sh
+++ b/.ci_support/test_backwards.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
 
-pip install --no-deps pyiron_base==0.1.21
-echo "Before save";
-for t in tests/backwards/*save.py; do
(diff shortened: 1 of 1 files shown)


niklassiemer:
The workflow seems fine, however, the test breaks due to some `DataContainer` read error. Was it possible to save the input via `input['value']` in the 0.1.x version? Or did we break some behavior which was not intended?

niklassiemer:
We decided to restructure and enhance our backwards testing in a different way. Therefore, I finally close this here.

PR #476, opened 2021-10-18
Unittesting template with filled projects

sudarsan-surendralal:
I think it is useful in many cases (Edit:  For example https://github.com/pyiron/pyiron_base/pull/476) to have test case templates with a `cls.project` class attribute which already has a few jobs and sub projects. Here's a minimal implementation.

### DIFF
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -101,4 +101,40 @@ class ToyJob(PythonTemplateJob):
             h5out["energy_tot"] = self.input["input_energy"]
         self.status.finished = True
 
-_TO_SKIP = [PyironTestCase, TestWithProject, TestWithCleanProject]
+
+class TestWithFilledProject(PyironTestCase, ABC):
+
+    """
+    Tests that creates a projects, creates jobs and sub jobs in it, and at the end of unit testing,
+    removes the project.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.project_path = getfile(cls)[:-3].replace("\\", "/")
+        cls.file_location, cls.project_name = split(cls.project_path)
+        cls.project = Project(cls.project_path)
+        job = cls.project.create_job(job_type=ToyJob, job_name="toy_1")
+        job.run()
+        job = cls.project.create_job(job_type=ToyJob, job_name="toy_2")
+        job.run()
+        job.status.aborted = True
+
+        with cls.project.open("sub_project") as pr_sub:
+            job = pr_sub.create_job(job_type=ToyJob, job_name="toy_1")
+            job.run()
+            job = pr_sub.create_job(job_type=ToyJob, job_name="toy_2")
+            job.run()
+            job.status.aborted = True
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.project.remove(enable=True)
+        try:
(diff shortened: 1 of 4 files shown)


pyiron_base/_tests.py
```diff
@@ -88,4 +89,52 @@ def tearDown(self):
         self.project.remove_jobs_silently(recursive=True)
 
 
-_TO_SKIP = [PyironTestCase, TestWithProject, TestWithCleanProject]
+class ToyJob(PythonTemplateJob):
+    def __init__(self, project, job_name):
+        """A toyjob to test export/import functionalities."""
+        super(ToyJob, self).__init__(project, job_name)
+        self.input['input_energy'] = 100
```
```suggestion
        self.input.in_data = 100
```

`PythonTemplateJob` uses `DataContainer` for IO now. Also while 'energy' is pretty generic, since this is in base let's go all the way with our generality.

pyiron_base/_tests.py
```diff
@@ -88,4 +89,52 @@ def tearDown(self):
...
+
+    # This function is executed
+    def run_static(self):
+        with self.project_hdf5.open("output/generic") as h5out:
+            h5out["energy_tot"] = self.input["input_energy"]
```
```suggestion
        self.status.running = True
        self.output.data_out = self.input.data_in + 1
```

The `running` is just trivial to set a good example. The benefit of leaning on the `DataContainer` interface is clear though, since we never need to talk about hdf.

Aha, no, wait. I still need to add a final `to_hdf()`. In the future hopefully this will be handled by the `.run()` flow too, but for now we'll live with it.

pyiron_base/_tests.py
```diff
@@ -88,4 +89,52 @@ def tearDown(self):
...
+        self.status.finished = True
+
```
```suggestion
        self.to_hdf()
        
```

pyiron_base/_tests.py
```diff
@@ -88,4 +89,52 @@ def tearDown(self):
...
+    Tests that creates a projects, creates jobs and sub jobs in it, and at the end of unit testing,
+    removes the project.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.project_path = getfile(cls)[:-3].replace("\\", "/")
+        cls.file_location, cls.project_name = split(cls.project_path)
+        cls.project = Project(cls.project_path)
```
```suggestion
```

This is all taken care of in the super call, or?

pyiron_base/_tests.py
```diff
@@ -88,4 +89,52 @@ def tearDown(self):
...
+            job.run()
+            job.status.aborted = True
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.project.remove(enable=True)
+        try:
+            remove(join(cls.file_location, "pyiron.log"))
+        except FileNotFoundError:
+            pass
```
```suggestion
```

This has no change to the parent method.

liamhuber (changes requested):
Yeah, I agree! Nice idea!

Some streamlining changes suggested.

niklassiemer (commented):
I like the streamlining of `ToyJob` and the `FilledProject` test case! With Liams suggestions this lgtm.

sudarsan-surendralal:
@liamhuber thanks for the suggestions. I've streamlined the `ToyJob` class originally used and defined by @pmrv separately for each module. I've now replaced all these class definitions with the common one. @pmrv hope this is okay?

liamhuber (approved):

PR #477, opened 2021-10-18
More filtering options in Project.iter_jobs

### LINKED ISSUE #399: More filtering options in Project.iter_jobs would be cool
Currently it's possible to filter jobs by the `status` and `hamilton` columns of the job table, but more would be sweet and easy to implement, see discussion [here](https://github.com/pyiron/pyiron_base/pull/398).

sudarsan-surendralal:
This PR extends `pr.iter_jobs()` to include filtering based on matching column values (eg. `pr.iter_jobs(hamilton="Vasp")` or `pr.iter_jobs(parentid=None)`

### DIFF
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -101,4 +101,40 @@ class ToyJob(PythonTemplateJob):
             h5out["energy_tot"] = self.input["input_energy"]
         self.status.finished = True
 
-_TO_SKIP = [PyironTestCase, TestWithProject, TestWithCleanProject]
+
+class TestWithFilledProject(PyironTestCase, ABC):
+
+    """
+    Tests that creates a projects, creates jobs and sub jobs in it, and at the end of unit testing,
+    removes the project.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.project_path = getfile(cls)[:-3].replace("\\", "/")
+        cls.file_location, cls.project_name = split(cls.project_path)
+        cls.project = Project(cls.project_path)
+        job = cls.project.create_job(job_type=ToyJob, job_name="toy_1")
+        job.run()
+        job = cls.project.create_job(job_type=ToyJob, job_name="toy_2")
+        job.run()
+        job.status.aborted = True
+
+        with cls.project.open("sub_project") as pr_sub:
+            job = pr_sub.create_job(job_type=ToyJob, job_name="toy_1")
+            job.run()
+            job = pr_sub.create_job(job_type=ToyJob, job_name="toy_2")
+            job.run()
+            job.status.aborted = True
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.project.remove(enable=True)
+        try:
+            remove(join(cls.file_location, "pyiron.log"))
+        except FileNotFoundError:
+            pass
+
+
+_TO_SKIP = [PyironTestCase, TestWithProject, TestWithCleanProject, TestWithFilledProject]
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -10,6 +10,7 @@ from io import StringIO
 import unittest
 from os.path import split, join
 from os import remove
+from pyiron_base import PythonTemplateJob
 from pyiron_base.project.generic import Project
 from abc import ABC
 from inspect import getfile
@@ -87,5 +88,17 @@ class TestWithCleanProject(TestWithProject, ABC):
     def tearDown(self):
         self.project.remove_jobs_silently(recursive=True)
 
+
+class ToyJob(PythonTemplateJob):
+    def __init__(self, project, job_name):
+        """A toyjob to test export/import functionalities."""
+        super(ToyJob, self).__init__(project, job_name)
+        self.input['input_energy'] = 100
+
+    # This function is executed
+    def run_static(self):
+        with self.project_hdf5.open("output/generic") as h5out:
+            h5out["energy_tot"] = self.input["input_energy"]
+        self.status.finished = True
 
 _TO_SKIP = [PyironTestCase, TestWithProject, TestWithCleanProject]
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -126,7 +126,7 @@ class TestWithFilledProject(TestWithProject, ABC):
             job.run()
             job = pr_sub.create_job(job_type=ToyJob, job_name="toy_2")
             job.run()
-            job.status.aborted = True
+            job.status.suspended = True
 
     @classmethod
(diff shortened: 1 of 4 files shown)


pyiron_base/project/generic.py
```diff
@@ -516,34 +516,46 @@ def inspect(self, job_specifier):
...
 
         Args:
             path (str): HDF5 path inside each job object
             recursive (bool): search subprojects [True/False] - True by default
             convert_to_object (bool): load the full GenericJob object (default) or just the HDF5 / JobCore object
-            status (str/None): status of the jobs to filter for - ['finished', 'aborted', 'submitted', ...]
-            job_type (str/None): job type to filter for, corresponds to the 'hamilton' column in the job table
             progress (bool): if True (default), add an interactive progress bar to the iteration
-
+            **kwargs (dict): Can be
```
liamhuber: ?

pyiron_base/_tests.py
```diff
@@ -88,4 +89,52 @@ def tearDown(self):
         self.project.remove_jobs_silently(recursive=True)
 
 
-_TO_SKIP = [PyironTestCase, TestWithProject, TestWithCleanProject]
```
liamhuber: Hm, this is already handled in #476 . I suggested resolving that PR first, merging master in here, and handling the merge conflict. Should be 'theirs' across the board for `_tests.py` so at least it shouldn't be hard.

pyiron_base/project/generic.py
```diff
@@ -516,34 +516,46 @@ def inspect(self, job_specifier):
...
+            return []
+        mask = np.ones_like(df.index, dtype=bool)
+        for key, val in kwargs.items():
+            if val is None:
+                mask &= df[key].isnull()
+            else:
+                mask &= df[key] == val
+        if not mask.any():
+            return []
+        return df[mask]["id"].to_list()
```
liamhuber, line 545: Is there a special reason to force it to return as list instead of np.ndarray?

tests/project/test_project.py
```diff
@@ -57,6 +57,40 @@ def test_iterjobs(self):
...
+        self.assertEqual(len(self.project.get_filtered_job_ids(recursive=True, status="finished")), 2)
+        self.assertEqual(len(self.project.get_filtered_job_ids(recursive=False, status="finished")), 1)
+        self.assertEqual(len(self.project.get_filtered_job_ids(recursive=True, status="suspended")), 1)
+        self.assertEqual(len(self.project.get_filtered_job_ids(recursive=True, status="aborted")), 1)
+        self.assertEqual(len(self.project.get_filtered_job_ids(recursive=False, status="suspended")), 0)
+        self.assertEqual(len(self.project.get_filtered_job_ids(recursive=False, hamilton="ToyJob")), 2)
+        self.assertEqual(len(self.project.get_filtered_job_ids(recursive=True, parentid=None)), 4)
+
+    def test_get_iter_jobs(self):
+        self.assertEqual([val["output/generic/energy_tot"] for val in self.project.iter_jobs(recursive=True)],
```
liamhuber: My suggested changes to the generic test class will mess this part up, but I think it will just be `"output/generic/energy_tot"` --> `"output/storage/output/data_out"` or some other little string change.

liamhuber (commented):
lgtm. A couple of nits, but my only real concern is avoiding conflict with #476

pyiron_base/project/generic.py
sudarsan-surendralal: Oops! Fixed.

pyiron_base/project/generic.py
sudarsan-surendralal: I don't mind either way. But since we really don't expect to perform numerical operations with this (we are always going to iterate over this) I stuck to lists.

pyiron_base/_tests.py
sudarsan-surendralal: This should be fixed now.

tests/project/test_project.py
sudarsan-surendralal: Fixed now. I'd rather load the job rather than use string paths (should be fast either way)

liamhuber (approved):

pyiron_base/project/generic.py
pmrv: I like lists because it's what you'd expect from a simple function like this and it shouldn't be very expensive.

pyiron_base/project/generic.py
```diff
@@ -512,34 +512,53 @@ def inspect(self, job_specifier):
...
+
+    def iter_jobs(self, path=None, recursive=True, convert_to_object=True, progress=True, **kwargs):
```
```suggestion
    def iter_jobs(self, path=None, recursive: bool=True, convert_to_object: bool=True, progress: bool=True, **kwargs) -> Iterator[GenericJob]:
```
~Did we loose the `path` parameter somewhere?~ Nevermind, github hid the rest of the method and I was not smart enough to click on expand.

pyiron_base/project/generic.py
```diff
@@ -512,34 +512,53 @@ def inspect(self, job_specifier):
         """
         return self.load(job_specifier=job_specifier, convert_to_object=False)
 
-    def iter_jobs(self, path=None, recursive=True, convert_to_object=True, status=None, job_type=None, progress=True):
+    def get_filtered_job_ids(self, recursive: bool = True, **kwargs: dict) -> list:
```
```suggestion
    def get_filtered_job_ids(self, recursive: bool = True, **kwargs: dict) -> List[int]:
```
`List` comes from `typing`.

jan-janssen:
I like this extension, I am just wondering if we could use the same filter functions we use for the pyiron table, that would improve the user experience for new users.

jan-janssen:
Sorry I closed the pull request by accident

pmrv (approved):
Aside from the typing nits, lgtm.

sudarsan-surendralal:
> 
> 
> I like this extension, I am just wondering if we could use the same filter functions we use for the pyiron table, that would improve the user experience for new users.

The current filters work at the table level. I guess the filters in pyiron tables need access to the job or job path. But maybe this can improve the filtering in pyiron tables?

pyiron_base/project/generic.py
sudarsan-surendralal: I'm a python type hinting n00b so thanks for this.

pmrv:
> > I like this extension, I am just wondering if we could use the same filter functions we use for the pyiron table, that would improve the user experience for new users.
> 
> The current filters work at the table level. I guess the filters in pyiron tables need access to the job or job path. But maybe this can improve the filtering in pyiron tables?

Agree, the current implementation is ok and probably more future-proof once @tnecnivkcots database restructuring takes place.  At that point I see us querying the database directly with these constraints so there would be no way to convert job ids to jobs/job paths first to apply the filters.

PR #483, opened 2021-10-21
FileTable hotfix 

### LINKED ISSUE #482: Project wait_for_jobs( ) TypeError
Since I update my pyiron installation today the Project wait_for_jobs method throws the following error message after successfully fetching the first job from our remote cluster. My guess is that some changes introduced in #336 cause this issue.

I think the FileTable job_table() is completely different compared to an sql database. At least the it takes completely different arguments than the job_table() of IsDatabase. The job_table() method is also used in the get_jobs() method of FileTable, so I assume it is also broken.
Maybe it woud be good to only use IsDatabase for SQL databases or have something like IsSQLDB and IsFileDB,
maybe this can be fixed by setting some default values for the arguments in _get_job_table(), but I don't know much about how these work internally, so this could be completely off. Would be great if you could have a look at this @pmrv 

```
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_4170575/3846198617.py in <module>
      1 pr = Project(".")
----> 2 pr.wait_for_jobs(max_iterations=1, interval_in_s=0)

~/git_projects/pyiron_base/pyiron_base/project/generic.py in wait_for_jobs(self, interval_in_s, max_iterations, recursive)
   1339             ValueError: max_iterations reached, but jobs still running
   1340         """
-> 1341         wait_for_jobs(
   1342             project=self,
   1343             interval_in_s=interval_in_s,

~/git_projects/pyiron_base/pyiron_base/server/queuestatus.py in wait_for_jobs(project, interval_in_s, max_iterations, recursive)
    220     finished = False
    221     for _ in range(max_iterations):
--> 222         project.update_from_remote(recursive=True)
    223         df = project.job_table(recursive=recursive)
    224         if all(df.status.isin(job_status_finished_lst)):

~/git_projects/pyiron_base/pyiron_base/project/generic.py in update_from_remote(self, recursive)
    608             recursive (bool): search subprojects [True/False] - default=True
    609         """
--> 610         update_from_remote(
    611             project=self,
    612             recursive=recursive

~/git_projects/pyiron_base/pyiron_base/server/queuestatus.py in update_from_remote(project, recursive)
    271                 if status_hdf5 in job_status_finished_lst:
    272                     job_object = job.to_object()
--> 273                     job_object.transfer_from_remote()
    274 
    275 

~/git_projects/pyiron_base/pyiron_base/job/generic.py in transfer_from_remote(self)
    782         else:
    783             ft = FileTable(project=self.project_hdf5.path + "_hdf5/")
--> 784             df = ft.job_table(all_columns=True)
    785             db_dict_lst = []
    786             for j, st, sj, p, h, hv, c, ts, tp, tc in zip(

TypeError: job_table() missing 3 required positional arguments: 'sql_query', 'user', and 'project_path'
```

pmrv:
Fixes #482.

In #336 I tried to match the methods getting the job_table from both `DatabaseAccess` and `FileTable` and as part of that introduced `user` and `sql_query` to the method on `FileTable`.  When calling `get_job_table` from a `Project` these values are always supplied and then ignored, but there were two sites that called `FileTable.get_job_table` directly that I had missed earlier.

### DIFF
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -153,13 +153,13 @@ class FileTable(IsDatabase, metaclass=Singleton):
         element_lst=None
     ):
         self.update()
-        if project is None:
-            project = self._project
+        if project_path is None:
+            project_path = self._project
         if len(self._job_table) != 0:
             if recursive:
-                df = self._job_table[self._job_table.project.str.contains(project)]
+                df = self._job_table[self._job_table.project.str.contains(project_path)]
             else:
-                df = self._job_table[self._job_table.project == project]
+                df = self._job_table[self._job_table.project == project_path]
         else:
             return self._job_table
 
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -168,7 +168,7 @@ class FileTable(IsDatabase, metaclass=Singleton):
             project = self._project
         if columns is None:
             columns = ["id", "project"]
-        df = self.job_table(project=project, recursive=recursive, columns=columns)
+        df = self.job_table(sql_query=None, user=s.login_user, project_path=project, recursive=recursive, columns=columns)
         if len(df) == 0:
             dictionary = {}
             for key in columns:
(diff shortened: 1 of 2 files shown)


pmrv:
@Leimeroth Give it a try with this one.  I also found an unrelated typo, so hopefully it's ok now.

Leimeroth:
This fixes the problem

PR #484, opened 2021-10-21
Settings promises

liamhuber:
Define and test the promises of the `Settings` class.

@pmrv and I have been talking about getting spec written. Re-reading [Joel's take on spec](https://www.joelonsoftware.com/2000/10/02/painless-functional-specifications-part-1-why-bother/) I wonder whether we aren't being too fine grained by having a special spec document for each (important?) class? With that said, once I sat down to write something spec-like for `Settings` I realised (a) that I *still* don't really grok what settings is doing, and (b) having this sort of high-level description of what the point is is absolutely critical.

So what should we do with this stuff? I am not totally against putting it in separate documents (cf. Marvin's spec/had_hdf.md), and I could also see this sort of thing just being in the class docstring. At the moment I put it at the module level docstring, although I am really seriously wondering if the class level is better.

Also, at the moment, the spec and tests are pretty much written to match the code and not exactly to match the designed behaviour I'd want. I might play around with it in the future.

@pmrv sorry, but I don't know if I'm going to get all the way to database access by the end of the week 😛

### DIFF
--- a/pyiron_base/settings/generic.py
+++ b/pyiron_base/settings/generic.py
@@ -12,7 +12,7 @@ instantiated.
 It is possible to run pyiron only with default behaviour from the `Settings` class itself, but standard behaviour is to
 update the configuration by reading information stored on the system.
 First, `Settings` tries to read these updates from a file the operating system environment keys are searched for
-`'PYIRONCONFIG'`, and if this doesn't exist then a default location (~/.pyiron_config) is used.
+`'PYIRONCONFIG'`, and if this doesn't exist then a default location (~/.pyiron) is used.
 Second, if no file can be found in either of these locations, `Settings` tries to update the configuration by looking
 for system environment variables containing `'PYIRON'`.
 Additionally, if either of the conda flags `'CONDA_PREFIX'` or `'CONDA_DIR'` are system environment variables, they get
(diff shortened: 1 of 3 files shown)


liamhuber:
EDIT: Resolved.

Below is the test that's failing on the Windows build but passes on Ubuntu (and doesn't run here but passes on MacOS on my local machine). Does anyone understand why it would fail only for windows? If I understand correctly it's complaining the that first list, `[self._niceify_path(p1), self._niceify_path(p2)]`, only has one element. I would be confused by could see how it's possible that the second guy winds up with only one element, but I completely fail to see how it's possible for the first one...

```python
    def test_path_conversion(self):
        local = Path(self.cwd + "/.pyiron")
        p1 = '~/here/is/a/path/'
        p2 = 'here\\is\\another'  # TODO: Is this really good enough? Is it really windowsy enough?
        local.write_text(f"[DEFAULT]\nRESOURCE_PATHS = {p1}, {p2}\n")
        self.env["PYIRONCONFIG"] = str(local)
        self.s._update_configuration(self.s._configuration)
        self.assertListEqual(
            [self._niceify_path(p1), self._niceify_path(p2)],
            self.s._configuration['resource_paths']
        )
        local.unlink()
```

liamhuber:
>  (cf. Marvin's spec/had_hdf.md)

correction: has_groups.md

PR #485, opened 2021-10-22
give entries indices

samwaseda:
There are some input entries in `GenericMaster` (e.g. `child_id_func`, `generic_dict` (?), `job_list`) which are not enumerated and therefore give an error when split by `__index_` is effectuated. I don't whether these entries should come first or last, because I don't really understand in the first place why the enumeration really matters. Anyway I put them to the top - let me know what you think.


pmrv:
Can you describe where this problem comes up?  DataContainers `from_hdf` should never try to read from a HDF group where the job itself stores it's data in.  It sounds like somewhere a DataContainer is written to a group that already has a job inside and that would be a bug.

pmrv:
@samwaseda Is this still relevant?

PR #486, opened 2021-10-22
Refactor global state

liamhuber:
~Awaiting #484.~

I want to break apart the settings object so its responsibility is clearer and more limited in scope. This was already done by breaking database access stuff off it (I forget the PR but it's `DatabaseManager`). If we *just* do this, it means more and more imports, cf. `Project` which now depends on both the `Settings` *and* `DatabaseManager`. My solution is to make a new ~`IDE`~ `state` singlet helper class that acts as a single point of entry to all the other singlets (except maybe `JobTypeChoice`, which is anyhow on the chopping block).

Still TODO in this PR:
- ~Fix `ScriptJob` error I introduced (at least that test is failing on my machine)~
- ~Further rationalisation of `Settings`~
  - ~Under-the-hood improvements to read and overwrite configuration in a clear way~ User input > env > specified config > default config, all XOR merging on top of the code defaults
  - ~Streamline user-facing stuff to a `configuration` attribute and `update` method~
  - ~Shift responsibility for constructing database framework-specific connection strings from the `Settings` over to `DatabaseManager`~
- ~Extract logger from `Settings`~
- ~Extract queue adapter from `Settings`~
- ~Extract publications from `Settings`~
- ~Add an `update` method to `state` which updates the configuration *and* reinitializes the database and queue adapter stuff.~

Non-objectives for this PR:
- Any further work on `DatabaseManager` or database/filetable stuff. For now that the manager singleton can stay where it is.
- Advancing or polishing the publications stuff


niklassiemer:
I like the idea! One thing I am not so sure about is the handling of publications. IMHO this should rather be a project thing than a singleton... Or at least not be accessible via `Project.get_publications` (is this the name?...) but `Project.get_session_publications` or the like.

liamhuber:
> I like the idea! One thing I am not so sure about is the handling of publications. IMHO this should rather be a project thing than a singleton... Or at least not be accessible via `Project.get_publications` (is this the name?...) but `Project.get_session_publications` or the like.

Yeah, I could definitely get on board with that. Having both per-project and per-session methods makes sense to me. For this PR I'm just refactoring*, but at least decoupling it from settings gives us more freedom to make such changes.

*ok, almost. I add a little functionality here and there. In this case I added `publications.reset`, which I guess would actually be an important tool for the paradigm you're talking about anyhow 🤓

pyiron_base/project/generic.py
```diff
@@ -1302,6 +1300,7 @@ def wait_for_jobs(self, interval_in_s=5, max_iterations=100, recursive=True):
         )
 
     @staticmethod
+    @deprecate(message="Use IDE.logger.set_logging_level instead.")
```
niklassiemer: should this be `state.logger` now?

niklassiemer (commented):
I did not check the moved parts (yet - just skimmed over the 'new' parts).

liamhuber:
> I did not check the moved parts (yet - just skimmed over the 'new' parts).

Super, thanks! It has all the functionality I want, so as soon as I can get to it I'll make a final refactoring/proofreading pass then set the PR active 😄

pyiron_base/database/manager.py
```diff
@@ -143,10 +192,10 @@ def switch_to_user_mode(self):
             if self._database.view_mode:
                 self.close_connection()
                 self._database = DatabaseAccess(
-                    s.configuration["sql_connection_string"],
-                    s.configuration["sql_table_name"],
+                    self.sql_connection_string,
+                    self.sql_table_name,
                 )
-                self._database.view_mode = True
+                self._database.view_mode = False
```
max-hassani, line 198: shouldn't it be `True` like the original?

pyiron_base/database/performance.py
```diff
@@ -79,9 +78,8 @@ class DatabaseStatistics:
     """
 
     def __init__(self):
-        s = Settings()
-        connection_string = s._configuration['sql_connection_string']
-        self._job_table = s._configuration['sql_view_table_name']
+        connection_string = state.settings.configuration['sql_connection_string']
```
max-hassani: It is not related to this line per se, and it might be a misunderstanding from my side, but I thought we are separating setting and database, so why do we have database configuration in setting? Instead of `state.setting.configuration[]`, we can use directly `state.database.sql_connection_string`.

pyiron_base/project/generic.py
```diff
@@ -126,15 +120,19 @@ def __init__(self, path="", user=None, sql_query=None, default_working_directory
...
+            state.database.open_connection()
+            self.db = state.database.database
         else:
             self.db = FileTable(project=path)
         self.job_type = JobTypeChoice()
 
         self._maintenance = None
 
+    @property
+    def ide(self):
```
max-hassani: Shouldn't we use here `state` instead of `ide` as well?

pyiron_base/project/generic.py
```diff
@@ -1573,8 +1572,7 @@ def __init__(self):
         initialize the flag self._check_postgres, to control whether pyiron is
         set to communicate with a postgres database.
         """
-        s = Settings()
-        connection_string = s._configuration['sql_connection_string']
+        connection_string = state.settings._configuration['sql_connection_string']
```
max-hassani: I suggest to use `state.database.sql_connection_string`, but if `state.setting` should be kept, I noticed there are both `state.setting.configuration` and `state.setting._configuration`. I think we should use the former.

max-hassani (commented):
I could not go through all now, but I made some comments on the parts I viewed. I will look in more details later in the afternoon.

pyiron_base/database/manager.py
liamhuber: I'm 99% sure the original was a bug. Inside, `switch_to_viewer_mode`, `if not self._database.view_mode:` then we do some stuff and set `self._database.view_mode = True`. By symmetry, therefore, in `switch_to_user_mode`, `if self._database.view_mode:`, after doing our thing we need to set `self._database.view_mode = False`.

This stuff was never covered by tests and didn't have spec so I can't be 100% sure, but the context and deductive reasoning do make me quite confident.

If I'm right, it's also a rather damning sign of how (in)frequently this tool gets used...

pyiron_base/database/performance.py
liamhuber: That is an *extremely* good catch, thank you Muhammad! Not only did we separate them, but the on stored in settings now stays as the raw input the user gave, and the one in the database is fancified, so this would have actually broken. I did string searching for all the keys that `database` is stealing from `settings.configuration` and now reference the database directly (although the `sql_connnection_string` is the only one that is not identical)

pyiron_base/project/generic.py
liamhuber: Yep, I agree with your entire chain of logic/preferences. It got fixed dealing with your earlier comment though so no further action here.

liamhuber:
Last call for reviews. I had to resolve some merge conflicts earlier this week to keep this guy up-to-date...it didn't take long but I still don't want to have to do that again next week 😝 @niklassiemer @muh-hassani if you have concerns or really want some extra time to look at it lmk, otherwise I'll merge this evening.

pyiron_base/project/generic.py
```diff
@@ -245,7 +243,7 @@ def create_from_job(self, job_old, new_job_name):
         """
         job_id = self.get_job_id(new_job_name)
         if job_id is not None:
-            s.logger.info("create_from_job has already job_id {}!".format(job_id))
+            state.logger.info("create_from_job has already job_id {}!".format(job_id))
```
niklassiemer: Made the info more informative.
```suggestion
            state.logger.info(f"create_from_job: {new_job_name} has already job_id {job_id}!")
```

pyiron_base/state/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Liam Huber"
+__email__ = "huber@mpie.de"
+__status__ = "production"
+__date__ = "Oct 22, 2021"
+
+
+class State(metaclass=Singleton):
```
niklassiemer, line 37: Actually, the `State` is not required to be a singleton itself since it does not hold any data and all referenced classes are singletons. However, I am not opposed to leave it a singleton to avoid surprises in the future :)

pyiron_base/state/publications.py
```diff
@@ -0,0 +1,174 @@
...
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Liam Huber"
+__email__ = "huber@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+
+class Publications(metaclass=Singleton):
```
niklassiemer, line 28: If we are able to channel all access to this class via the `State` class, I would suggest to make this here the non-singleton and remain the `State` as the singleton. This way, exactly this class could hold the project related publications in a later stage.

pyiron_base/state/settings.py
```diff
@@ -0,0 +1,394 @@
...
+
+        Returns:
+            str: absolute path in POSIX format
+        """
+        return (Path(path.strip())
+                .expanduser()
+                .resolve()
+                .absolute()
+                .as_posix()
+                .replace("\\", "/")
```
niklassiemer, line 222: I did not know this is required in the `as_posix()` version!

pyiron_base/state/settings.py
```diff
@@ -0,0 +1,394 @@
...
+
+    @property
+    # @deprecate("Use pyiron_base.state.state.queue_adapter")
+    def queue_adapter(self):
+        from pyiron_base.state import state
+        return state.queue_adapter
+
+    @property
+    # @deprecate("Use pyiron_base.state.state.publications.list()")
+    def publication_lst(self):
```
niklassiemer, line 369: 😞 As long as we need this for our backwards compatibility, we cannot un-singleton the publications...

niklassiemer (approved):
Only some small comments/ general ideas for the future. LGTM!

pyiron_base/state/publications.py
liamhuber: I agree, but that's a behaviour change. I guess previously the publications were a dictionary living on the settings singleton. I think the most-equivalent  thing is to put the new class as a singleton. Going forward we will indeed need to think about how this thing should behave though.

pyiron_base/state/settings.py
liamhuber: Unfortunately it is.

```python
from pathlib import Path
p = Path("~/foo\\bar")
p.expanduser().resolve().absolute().as_posix()
> '/Users/liamhuber/foo\\bar'
```

Silly behaviour, IMO, but c'est la vie.

liamhuber:
Build windows latest for py 3.8 hung indefinitely -- >2hrs, the rest of the builds, including windows on 3.9, finished in <<15 minutes. I'm restarting the build tests.

pyiron_base/job/generic.py
```diff
@@ -952,7 +950,7 @@ def run_if_scheduler(self):
             self._logger.warning("Job aborted")
             self.status.aborted = True
             raise ValueError("run_queue.sh crashed")
-        s.logger.debug("submitted %s", self.job_name)
+        state.logger.debug("submitted %s", self.job_name)
```
max-hassani, line 953: Here `state.logger` and `self._logger` are basically the same!

liamhuber:
> Build windows latest for py 3.8 hung indefinitely -- >2hrs, the rest of the builds, including windows on 3.9, finished in <<15 minutes. I'm restarting the build tests.

Must have been cosmic rays; it ran fine the second time.

pyiron_base/job/generic.py
max-hassani: I guess `self._logger` is now redundant.

max-hassani (approved):
Sorry @liamhuber for the delay. Thanks for this careful refactory of the state. I just had one comment

pyiron_base/job/generic.py
liamhuber: Ah yeah, up above we have `self._logger = state.logger`. I guess I was a little too much on autopilot doing the refactor. I would actually suggest to eliminate `self._logger` in a housekeeping PR later. The extra layer of misdirection is not necessary in the code base, and it's a private attribute anyhow so no one's notebooks should be accessing the logger this way anyhow.

PR #488, opened 2021-10-24
Bugfix in case of None type values

sudarsan-surendralal:
This is a hotfix. So I'll refine this later

Edit: Turns out the Hotfix could be sufficient. This is necessary to handle `NoneType` values


jan-janssen:
When you say nonetype values, are these entries in the job? I do not really like the `try` and `except` part.

sudarsan-surendralal:
> 
> 
> When you say nonetype values, are these entries in the job? I do not really like the `try` and `except` part.

Yes they are in the job. The `np.concatenate` function does not work on such values.

jan-janssen (approved):
Looks good to me

PR #492, opened 2021-10-25
Filtering

samwaseda:
I did the extension of the filtering function as we discussed today. In short, @sudarsan-surendralal implemented:

```python
for job in pr.iter_jobs(job='md_something'):
    ....
````

And with this PR, we will be able to do the following as well:

```python
for job in pr.iter_jobs(job='md*'):
    ....
````

It should be self-explanatory but it means all jobs with their job names starting with `md`.

In addition to `Project.iter_jobs`, the same functionality is available to `Project.job_table` now as well.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -92,6 +92,7 @@ class IsDatabase(ABC):
             full_table=False,
             element_lst=None,
             job_name_contains='',
+            **kwargs,
     ):
         """
         Access the job_table.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by pmrv — APPROVED** [review]
Tests with @sudarsan-surendralal new test class would be nice, but lgtm othwise.

File: pyiron_base/database/generic.py [hunk 1]
```diff
@@ -79,6 +80,38 @@ def _get_job_table(
     ):
         pass
 
+    @staticmethod
+    def _get_filtered_job_table(df, **kwargs: dict):
```
sudarsan-surendralal:
```suggestion
    def _get_filtered_job_table(df: pd.DataFrame, **kwargs: dict) -> pd.DataFrame:
```

pyiron_base/database/generic.py
```diff
@@ -79,6 +80,38 @@ def _get_job_table(
...
+        """
+        Get a list of job ids in a project based on matching values from any column in the project database
```
sudarsan-surendralal: It returns a job table and not a list of ids

sudarsan-surendralal (changes requested):
I think you need to fix the tests

```
ERROR: test_get_filtered_job_ids (project.test_project.TestProjectOperations)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_project.py", line 68, in test_get_filtered_job_ids
    self.assertEqual(len(self.project.get_filtered_job_ids(recursive=False)), 2)
AttributeError: 'Project' object has no attribute 'get_filtered_job_ids'

```

sudarsan-surendralal (approved):
Looks good now!

PR #474, opened 2021-10-15
Add worker mode

jan-janssen:
It is primarily a cosmetic change - a current demonstration is available https://github.com/jan-janssen/pyiron-worker

PR #497, opened 2021-11-05
Worker Job Class

jan-janssen:
Follow up to https://github.com/pyiron/pyiron_base/pull/474 

Implementing the workers as separate job class using multiprocessing

### DIFF
--- a/pyiron_base/job/worker.py
+++ b/pyiron_base/job/worker.py
@@ -25,17 +25,20 @@ __date__ = "Nov 5, 2021"
 
 def worker_function(queue):
     while True:
-        working_directory, job_id, _, submit_on_remote, debug = queue.get(
+        status, working_directory, job_id, _, submit_on_remote, debug = queue.get(
             block=True,
             timeout=None
         )
-        JobWrapper(
-            working_directory=working_directory,
-            job_id=job_id,
-            submit_on_remote=submit_on_remote,
-            debug=debug
-        ).run()
-
+        if status:
+            job_wrap = JobWrapper(
+                working_directory=working_directory,
+                job_id=job_id,
+                submit_on_remote=submit_on_remote,
+                debug=debug,
+            )
+            job_wrap.run()
+        else:
+            break
 
 class WorkerJob(PythonTemplateJob):
     def __init__(self, project, job_name):
@@ -83,7 +86,7 @@ class WorkerJob(PythonTemplateJob):
             processes=int(self.server.cores/self.cores_per_job),
             initializer=worker_function,
             initargs=(queue,)
-        ):
+        ) as pool:
             while True:
                 df = pr.job_table()
                 df_sub = df[
@@ -100,12 +103,20 @@ class WorkerJob(PythonTemplateJob):
                             df_sub["id"].values
                         ) if job_id not in active_job_ids]
                     job_lst = [
-                        [p, job_id, None, False, False]
+                        [True, p, job_id, None, False, False]
                         if pp is None else
-                        [os.path.join(pp, p), job_id, None, False, False]
+                        [True, os.path.join(pp, p), job_id, None, False, False]
                         for pp, p, job_id in path_lst
                     ]
-                    active_job_ids += [j[1] for j in job_lst]
+                    active_job_ids += [j[2] for j in job_lst]
                     _ = [queue.put(j) for j in job_lst]
+                elif self.status.collect:
+                    break
                 else:
                     time.sleep(self.input.sleep_interval)
+            for i in range(int(self.server.cores / self.cores_per_job)):
+                queue.put([False, False, False, False, False, False])
+            pool.close()
+            pool.join()
+            pool.terminate()
+        self.status.finished = True
--- a/pyiron_base/job/worker.py
+++ b/pyiron_base/job/worker.py
@@ -6,9 +6,8 @@ Worker Class to execute calculation in an asynchronous way
 """
 import os
 import time
-import multiprocessing
+from aproc import Pool
 from pyiron_base.job.template import PythonTemplateJob
-from pyiron_base.job.wrapper import JobWrapper
 
 
 __author__ = "Jan Janssen"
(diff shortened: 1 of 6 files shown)


jan-janssen:
Example: 
```
from pyiron import Project
pr_worker = Project("worker")
pr_calc = Project("calc")

# Setup Worker
job_worker = pr_worker.create.job.WorkerJob("runner")
job_worker.server.run_mode.non_modal = True
job_worker.server.cores = 4
job_worker.run()

# Submit Calculation
structure = pr_calc.create.structure.ase.bulk("Al", cubic=True)
structure.set_repeat([10, 10, 10])
for i in range(10):
    job = pr_calc.create.job.Lammps("lmp_" + str(i))
    job.structure = structure
    job.server.run_mode.worker = True
    job.master_id = job_worker.job_id
    job.run()

# Monitoring 
pr_calc.job_table()
```

pyiron_base/job/worker.py
```diff
@@ -0,0 +1,113 @@
...
+            job_id=job_id,
+            submit_on_remote=submit_on_remote,
+            debug=debug
+        ).run()
+
+
+class WorkerJob(PythonTemplateJob):
+    def __init__(self, project, job_name):
+        super(WorkerJob, self).__init__(project, job_name)
+        self.input['project'] = None
```
liamhuber: `PythonTemplateJob` facilitates dot notation for input now

jan-janssen:
An mybinder example for this class is available https://github.com/jan-janssen/pyiron-worker

jan-janssen:
Waiting for https://github.com/conda-forge/staged-recipes/pull/16822

pyiron_base/job/worker.py
```diff
@@ -0,0 +1,129 @@
...
+                            df_sub["id"].values
+                        ) if job_id not in active_job_ids]
+                    job_lst = [
+                        [True, p, job_id, None, False, False]
+                        if pp is None else
+                        [True, os.path.join(pp, p), job_id, None, False, False]
+                        for pp, p, job_id in path_lst
+                    ]
+                    active_job_ids += [j[2] for j in job_lst]
+                    _ = [pool.put(j) for j in job_lst]
```
pmrv: What's the reason not to use a plain `multiprocessing.Pool` & `map_async` here?  I'm a bit wary because I see that `aproc` uses `inspect.getsource` to transfer the function to the worker instead of pickle or dill.

jan-janssen: The list is growing while the worker is already executing it, so a simple map does not work. The current combination in the background uses a Multiprocessing pool plus a queue. In addition I also tried `dill` and `cloudpickle` but both failed, this might be related to the use of decorators.

pmrv: I don't understand why the growing list is a problem.

Inside the while loop the worker checks the job table for submitted jobs are assigned to it via the `master_id` and puts them in a list.  It could them `map_async` the list and mark the job ids as 'running' or 'queued' or whatever.  Then in the next iteration of the while loop there may be new submitted jobs which can be handled the same way or there may be submitted jobs that were already found in the last iteration, but since we marked them in the last iteration we can exclude them from the map in this iteration.

jan-janssen: The limitation of multiple `map_async` calls is that they can not be executed concurrently. So the second call to `map_async` is only executed once the first one is finished. The `async` only means you get control of the main process, it does not mean that the workers are asynchronous. That is why I created the `aproc` interface.

pmrv: From my testing this is not the case.  I can call `map_async` multiple times just fine and as long as there are available workers in the pool.  Once the pool is completely used you have to wait naturally, but I assume `aproc` is the same in this regard.

Here's my test script
```python
import multiprocessing as mp

def pow2(x):
    import time
    print(x)
    time.sleep(10)
    return x**2

with mp.Pool(4) as p:
    r1 = p.map_async(pow2, range(2))
    r2 = p.map_async(pow2, range(4, 8))

    r1.wait()
    print(r1.get())
    r2.wait()
    print(r2.get())
```
with output
```
0
1
4
5
<nothing happens here for 10s because we're blocked on r1.wait()>
6
7
[0, 1]
[16, 25, 36, 49]
```
That shows that the two first jobs from the second map call are run straight away and then the last two once the jobs from the first map are finished.  This sounds like exactly what we need to me.

jan-janssen: I thought I tried that before but it is working now, so I reverted to `map_async` - thanks again for pointing me in this direction.

jan-janssen:
@pmrv and @liamhuber Any other feedback? Otherwise I would like to merge this soon to be included in the next pyiron_base release.

pmrv:
I would like to have a small example in the class or module docstring on how to use the worker.  The snippet above would do it already.  Otherwise it looks good to me.

liamhuber:
> @pmrv and @liamhuber Any other feedback? Otherwise I would like to merge this soon to be included in the next pyiron_base release.

I would like to see the class appear in the unit tests. If I understand correctly, as long as the test inherits from `pyiron_base._tests.PyironTestCase` and the `docstring_module` attribute is set then example in the docstring will get run -- this is probably already sufficient. @pmrv let me know if I misspoke or if you think any additional test is necessary.

jan-janssen:
> I would like to see the class appear in the unit tests. If I understand correctly, as long as the test inherits from `pyiron_base._tests.PyironTestCase` and the `docstring_module` attribute is set then example in the docstring will get run -- this is probably already sufficient. @pmrv let me know if I misspoke or if you think any additional test is necessary.

As we no longer have the example job I am not exactly sure what job to test it with. The ScriptJob which is used for many other tests is not really sufficient.

liamhuber:
Why would script job be insufficient?

niklassiemer:
> As we no longer have the example job I am not exactly sure what job to test it with. The ScriptJob which is used for many other tests is not really sufficient.

However, we have 
https://github.com/pyiron/pyiron_base/blob/184671119d072288c5bf5934e7819e90d57810a9/pyiron_base/_tests.py#L92-L103
Maybe this is sufficient?

jan-janssen:
> Maybe this is sufficient?

No because the class defined in the test can not be reloaded in the subprocess on the worker. But I got it working with the script job instead.

PR #503, opened 2021-11-08
Correctly check for ragged array

### LINKED ISSUE #502: Performance issue with h5io
From my perspective this comes from https://github.com/pyiron/pyiron_base/pull/233

pmrv:
Fixes #502 in my testing, but haven't tested yet if something else depended on this weird behavior.

### DIFF
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -14,6 +14,7 @@ import posixpath
 import h5io
 import numpy as np
 import sys
+from typing import Union
 from pyiron_base.interfaces.has_groups import HasGroups
 
 __author__ = "Joerg Neugebauer, Jan Janssen"
@@ -26,6 +27,19 @@ __maintainer__ = "Jan Janssen"
 __email__ = "janssen@mpie.de"
 __status__ = "production"
 __date__ = "Sep 1, 2017"
+
+def _is_ragged_array(value: Union[np.ndarray, list]) -> bool:
+    """
+    Checks whether array or list of lists is ragged.
+
+    Args:
+        value (ndarray/list): array to check
+
+    Returns:
+        bool: True if elements of value are not all of the same shape
+    """
+    shape_lst = [np.shape(sub) for sub in value]
+    return len(set(shape_lst)) > 1
 
 def open_hdf5(filename, mode="r", swmr=False):
     if swmr and mode != "r":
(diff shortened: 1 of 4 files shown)


jan-janssen:
@pmrv Great, can you maybe also add a quick performance test? Just so we do not forget about it.

niklassiemer:
Thanks, indeed! LGTM! However, I am wondering why the compatibility workflows (already) ran?! 

Edit: Made a new issue for this #504

pmrv:
Looking at the diff again, I see that the checks are not actually equivalent.  The original code checks whether the shapes of arrays in `value` are all equal except the first dimension.  It still sounds to me like that check wants to check whether `value` is ragged only in the first dimension, but it's still a bit weird to me.

pyiron_base/generic/hdfio.py
```diff
@@ -27,6 +28,19 @@
...
+    Checks whether array or list of lists is ragged.
+
+    Args:
+        value (ndarray/list): array to check
+
+    Returns:
+        bool: True if elements of value are not all of the same shape
+    """
+    shape_lst = [np.shape(sub) for sub in value]
+    return len(set(shape_lst)) > 1
```
liamhuber: I was curious what was faster: `set`, `numpy.unique`, or `pandas.unique`. Operating on `[np.random.randint(50) for _ in range(2000)]` it was `set` *by an order of magnitude!*

Anyhow, no change here, I just thought this was interesting.

pyiron_base/generic/hdfio.py
```diff
@@ -165,8 +179,7 @@ def __setitem__(self, key, value):
             and len(value[0]) > 0
             and not isinstance(value[0][0], str)
         ):
-            shape_lst = [np.shape(sub) for sub in value]
-            if all([shape_lst[0][1:] == t[1:] for t in shape_lst]):
+            if _is_ragged_array(value):
```
liamhuber, line 185: I appreciate this refactor. Makes things very readable.

liamhuber (approved):
lgtm. I also agree the check has changed, but actually I don't understand exactly what the previous check was trying to accomplish and the current check is very clear so...go for it?

liamhuber:
@raynol-dsouza keep an eye on this -- I guess it will probably be merged quickly enough to give you some QoL improvement for the calculations we talked about yesterday!

pmrv:
I've copied the benchmark workflow over from `pyiron_atomistics`.

pmrv:
Whoops, turns out the default number of repetitions for `timeit` is a tiny bit crazy. Using a lower number now.

niklassiemer (approved):
LGTM!

niklassiemer:
It seems like, we actually depended on a bug in h5io to return float arrays instead of object arrays to allow this bug to work! (see https://github.com/pyiron/pyiron_atomistics/issues/409)

PR #505, opened 2021-11-09
Debug compat ghwf - Do not merge!

niklassiemer:


niklassiemer:
Well I do not at all get this! I print the result of 
```
      ${{ github.event.type == 'PushEvent' ||
          ( github.event.type == 'PullRequestEvent'  && contains(github.event.pull_request.labels.*.name, 'integration' ))
       }}
```
in the setup step. Since this is actually what is also used in the 
```
jobs:
  build:
    if: |
      ${{ github.event.type == 'PushEvent' ||
          ( github.event.type == 'PullRequestEvent'  && contains(github.event.pull_request.labels.*.name, 'integration' ))
       }}
```
block this should not be possible to print `false` because in this case the job should not be running! Nonetheless, I clearly see a `run = false` printed out... Any ideas what is going wrong here? @pmrv, @liamhuber, @jan-janssen

niklassiemer:
Anyone any ideas? I am rather lost...

niklassiemer:
found https://github.community/t/how-to-write-multi-line-condition-in-if/128477/8 will test without `${{ }}`

niklassiemer:
Yes, got skipped. I will update the workflows accordingly.

PR #507, opened 2021-11-09
Avoid compressing only if zipped file name has the job name in it

sudarsan-surendralal:
This is done using an existing function.


jan-janssen (approved):
Looks good to me

PR #508, opened 2021-11-10
Test for type conservation

niklassiemer:
This should cause pyiron/pyiron_atomistics#409 to pop up on our `pyiron_base` unittests already instead of in the compatibility with `pyiron_atomistics` integration tests.


tests/generic/test_fileHDFio.py
```diff
@@ -45,43 +46,79 @@ def tearDownClass(cls):
...
+        with self.subTest('content/dict_numpy'):
+            content_dict = hdf["content/dict_numpy"]
+            self.assertEqual(content_dict["key_1"], 1)
+            self.assertTrue(
+                all(
+                    np.equal(
+                        content_dict["key_2"],
+                        np.array([1, 2, 3, 4, 5, 6]),
+                    )
+                )
```
pmrv:
```suggestion
                np.array_equal(
                     content_dict["key_2"],
                     np.array([1, 2, 3, 4, 5, 6]),
                )
```

pmrv: That pattern comes up a few times, but `array_equal` should be correct for all of them.

pmrv (approved):
lgtm, except for the `array_equal` nit.

PR #349, opened 2021-06-16
Copy files in copy_to

pmrv:
`GenericJob` could previously copy files from the working directory, but didn't by default.  I exposed `copy_files` as an argument and change the default to `True`, same as in `JobCore`.

Fixes #348.

PR #511, opened 2021-11-11
Do not copy files on restart

### LINKED ISSUE #510: Compressed files are copied when restarting jobs
When restarting from an existing calculation, the new job copies ALL the files from the original one by default. I feel this would become extremely costly in terms of disk space. An example code:

```python
from pyiron import Project
pr = Project("vasp_testing")
pr.remove_jobs_silently(True)
basis = pr.create_ase_bulk("Al")
job = pr.create_job("Vasp", "test")
job.structure = basis
job.run()
job_new = job.restart("test_restart")
job_new.run()
job_new.list_files()
```
 produces the output:

```
>>> ['CHGCAR', 'CONTCAR', 'test.tar.bz2', 'test_restart.tar.bz2']
```

'test.tar.bz2' should not be copied IMO

sudarsan-surendralal:
See linked issue

### DIFF
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -8,8 +8,7 @@ from contextlib import redirect_stdout
 import doctest
 from io import StringIO
 import unittest
-from os import remove
-import posixpath
+import os
 from pyiron_base import PythonTemplateJob
 from pyiron_base.project.generic import Project
 from abc import ABC
@@ -69,14 +68,14 @@ class TestWithProject(PyironTestCase, ABC):
         super().setUpClass()
         print("TestWithProject: Setting up test project")
         cls.project_path = getfile(cls)[:-3].replace("\\", "/")
-        cls.file_location, cls.project_name = posixpath.split(cls.project_path)
+        cls.file_location, cls.project_name = os.path.split(cls.project_path)
         cls.project = Project(cls.project_path)
 
     @classmethod
     def tearDownClass(cls):
         cls.project.remove(enable=True)
         try:
-            remove(posixpath.join(cls.file_location, "pyiron.log"))
+            os.remove(os.path.join(cls.file_location, "pyiron.log"))
         except FileNotFoundError:
             pass
 
@@ -96,7 +95,7 @@ class ToyJob(PythonTemplateJob):
         self.input.data_in = 100
 
     def write_input(self):
-        self.input.write(posixpath.join(self.working_directory, "input.yml"))
+        self.input.write(os.path.join(self.working_directory, "input.yml"))
 
     # Allow writing of the input file
     def _check_if_input_should_be_written(self):
--- a/pyiron_base/_tests.py
+++ b/pyiron_base/_tests.py
@@ -10,6 +10,7 @@ from io import StringIO
 import unittest
 from os.path import split, join
 from os import remove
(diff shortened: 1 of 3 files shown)


pmrv:
We still want to copy certain files, though, like `CONTCAR`, right?  Is that taken care of separately?

sudarsan-surendralal:
> 
> 
> We still want to copy certain files, though, like `CONTCAR`, right? Is that taken care of separately?

Yes, for that we need to use `restart_file_list` and/or  `restart_file_dict`.

pyiron_base/_tests.py
```diff
@@ -10,6 +10,7 @@
 import unittest
 from os.path import split, join
 from os import remove
+import posixpath
```
jan-janssen: Just for consistency is there a reason why `os.path.join()` is not sufficient?

jan-janssen:
Just for me to understand this correctly, this behavior was introduced in https://github.com/pyiron/pyiron_base/pull/349 right?

pyiron_base/_tests.py
sudarsan-surendralal: > 
> 
> Just for consistency is there a reason why `os.path.join()` is not sufficient?

The `posixpath.join` method is [platform agnostic](https://www.georgevreilly.com/blog/2016/07/12/PythonJoiningUrlsWithPosixpathJoin.html) which is important esp for windows.

sudarsan-surendralal: I guess `posixpath.join` is platform independent ([source](https://www.georgevreilly.com/blog/2016/07/12/PythonJoiningUrlsWithPosixpathJoin.html)) so we should use this.

sudarsan-surendralal:
> 
> 
> Just for me to understand this correctly, this behavior was introduced in #349 right?

Yes and it worked well. But I guess having the new parameter `copy_files` with `True` as the default meant that the `restart` function which called this automatically copied all files (once again emphasizing the need for more tests on `pyiron_base`).

pyiron_base/_tests.py
pmrv: `os.path.join` should be doing the correct™ thing depending on the platform.

jan-janssen: > I guess `posixpath.join` is platform independent ([source](https://www.georgevreilly.com/blog/2016/07/12/PythonJoiningUrlsWithPosixpathJoin.html)) so we should use this.

The link talks about URL not file paths. `posixpath` always gives you `/` while `os.path` gives `/` on linux and Mac OS X and `\` on windows. We want to use `posixpath` for all HDF5 internal path but for file path I would suggest to use `os.path.join` to be more consistent with the python standards.

sudarsan-surendralal:
It seems that the output also seems to be copied during the restart. So I'm converting this to draft till I figure out what's going on. 

Edit: False alarm, everything seems okay! Do let me know if any further changes are required @jan-janssen @pmrv

sudarsan-surendralal:
@niklassiemer @muh-hassani Can you make a new release with these changes (also for the cluster)?

niklassiemer:
Sure, we will trigger a new release with minor (not only patch level) version bump once we fixed the current problems with the hdf and db connection bugs we have in our master.

PR #512, opened 2021-11-11
Faster hdf reading

pmrv:
I'm waiting for pyiron to submit a large number of jobs right now, so I thought I have a look at how to make it faster.

One of the *major* bottlenecks is calling `list_all()`/`list_groups()`/`list_nodes()` to check what datasets/groups are inside the HDF5 files.  In fact roughly 75%(!) of loading a small lammps job is spent in `FileHDFio.list_all()`.

This makes two changes:
1. Eagerly try load read data directly with `h5io.read_hdf5` instead of checking first whether it is there and then reading it.  This makes a simple read like `job['output/generic/energy_pot']` faster by about a factor 5, so that it takes roughly the same amount of time as calling `h5io.read_hdf5` directly on a file.
2. Use `list_all()` instead of `list_nodes()` and `list_groups` together.  This saves opening the HDF5 file once.

Both together make loading a lammps jobs about 10% faster.

I want to mention that this
```python
h5io.read_hdf5(j.project_hdf5.file_name, title=j.name + '/output/generic/energy_pot')
```
is still about twice as slow as directly doing
```python
with h5py.File(j.project_hdf5.file_name) as f:
    a = f[j.name + '/output/generic/energy_pot']
```
so there's still room for improvement.

### DIFF
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -116,7 +116,7 @@ class FileHDFio(HasGroups, MutableMapping):
                 # try to read it here and one more time to verify it's not a group below).
                 obj = h5io.read_hdf5(self.file_name, title=self._get_h5_path(item))
                 return obj
-            except ValueError:
+            except (ValueError, OSError):
                 # h5io couldn't find a dataset with name item, but there still might be a group with that name, which we
                 # check in the rest of the method
                 pass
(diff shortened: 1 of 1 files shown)


jan-janssen (approved):
Looks good to me

liamhuber (approved):

Issue #504, opened 2021-11-08
Compatibility workflow always triggered.


niklassiemer:
Right now, the compatibility workflows are always running, not only on push to master and if the integration label is set.

niklassiemer:
Testing the behavior in #505
It seems like github is running the workflow although the if statement is evaluated as false. I am clueless...

niklassiemer:
I will test this first in an external repo (https://github.com/niklassiemer/test_checkout), where I can merge to master without problems etc.

niklassiemer:
Solved in #535

PR #519, opened 2021-11-15
State hotfixes

### LINKED ISSUE #517: job_table() returning empty dataframe
After updating conda packages and pyiron versions pr.job_table() returns an empty dataframe. I have absolutely no clue what causes this problem.
The database seems to be completely fine. I tested it using sqlite3 PRAGMA integrity_check, by connecting to my database backup and trying pr.job_table() (still empty df) and had a look at the database using sqlitebrowser. 
I also tried to downgrade sqlalchemy from 1.4.27 to 1.4.26 and checked if the connection in the Settings is correct.
If I create new jobs they get a correct id (starting with a number over 300000, not with 0) and the new jobs are shown correctly in the dataframe, but all jobs from before updating are not shown.

@jan-janssen @pmrv  mentioning you to get you notified, if you have any idea what causes this / how to fix this please let me know, this really messes with my work

liamhuber:
I discovered a bug where calling `state.update()` decoupled two instances of `DatabaseAccess`: `state.database.database` (the canonical instance that nicely gets updated), and `pr.db` which winds up being a dead access point. This is perfectly sensible as the assignment `pr.db = state.database.database` *at project instantiation*.

To fix this I moved the remaining responsibility for `open_connection` calls over to `state.database`, and now just directly access it through a property on `Project`. Before the introduction of `state`, the project DB was anyhow pointing to an attribute of the singleton database manager, so I don't anticipate that this introduces any functional change.

One area of concern is that this will be less performant when someone is using the `FileTable` -- right now this class is still woefully disconnected from the other database code, so I simply re-instantiate the file table at each database call. I am a little bit worried that the `PyFileIndex` instantiation in that might be seriously less performant, but unless we get a complaint I'm willing to make a temporary sacrifice in an under-used feature for the sake of stability.

EDIT: I also now revert the default change in `Settings.configuration['project_check_enabled']` to `True`, as using `False` was causing trouble over in #517.

### DIFF
--- a/tests/database/test_manager.py
+++ b/tests/database/test_manager.py
@@ -41,3 +41,10 @@ class TestDatabaseManager(TestWithProject):
         self.s.configuration["project_check_enabled"] = check_before
         self.s.configuration["project_paths"] = paths_before
         self.s.configuration["disable_database"] = disable_before
+
+    def test_update_project_coupling(self):
+        self.dbm.update()
+        self.assertEqual(
+            self.dbm.database, self.project.db,
+            msg="Expected the database access instance to stay coupled between state and project."
+        )
--- a/tests/database/test_manager.py
+++ b/tests/database/test_manager.py
@@ -4,7 +4,7 @@
 
 from pyiron_base._tests import TestWithProject
 from pyiron_base import state
-from os import getcwd
+import os
 
 
 class TestDatabaseManager(TestWithProject):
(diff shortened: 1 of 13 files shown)


liamhuber:
After the 'project_check_enabled' reversion, compatibility with contrib fails. This is because the contrib CI doesn't set the project root path. I'll deal with it.

niklassiemer:
However, whatever the CI on contrib does is not important here, since we just run 
```
    - name: Setup
      shell: bash -l {0}
      run: |
        cd ..
        git clone https://github.com/pyiron/pyiron_contrib
        cd pyiron_contrib
        conda env update --name test --file .ci_support/environment.yml
        pip install --no-deps .
        cd ../pyiron_base
        pip install --no-deps .
    - name: Test
      shell: bash -l {0}
      run: |
        cd ../pyiron_contrib
        python -m unittest discover tests/
```
in our workflow. Comparing with the atomistics compat workflow, we should probably also 
```
        grep -v "pyiron_base" .ci_support/environment.yml > environment.yml
        conda env update --name test --file environment.yml
```
I do not know if `conda install pyiron_base` sets already a `root_path`?

EDIT:
Actually, contrib also installs `pyiron_base` via `pyiron_atomistics`, thus this probably would not help.

liamhuber:
I think we just need to run the config file in the CI script. We'll see 😉 

Anyhow, I guess it doesn't hurt to have the config script over in contrib too...

tests/database/test_manager.py
```diff
@@ -29,15 +29,35 @@ def test_file_top_path(self):
...
+                                              'sub_project').replace("\\", '/') + '/')
+        finally:
+            # Put things back the way you found them
+            self.s.configuration["project_check_enabled"] = check_before
+            self.s.configuration["project_paths"] = paths_before
+            self.s.configuration["disable_database"] = disable_before
+
+    def test_update_project_coupling(self):
+        self.dbm.update()
+        self.assertEqual(
```
niklassiemer: Should this not even be the same thing?
```suggestion
        self.assertIs(
```

niklassiemer (approved):
As far as I can tell, this lgtm.

tests/database/test_manager.py
liamhuber: Yep, you're spot on

liamhuber:
We passed compatibility earlier and the latest commit was just an adjustment to the unit tests right here --  as soon as windows passes (or more likely times out...) I'd say we're good to go?

niklassiemer:
I reverted my broken fix to just run if the label is present. Thus we will again run multiple compatibility tests... I think that is better than running none :) Reopened #504

niklassiemer:
Windows-3.8 the currently usual
```
Ran 385 tests in 55.476s

OK (skipped=46)
```
Thus passing.

PR #520, opened 2021-11-16
Test top path

niklassiemer:
As seen in #517, we need more tests and here they are. This will currently fail, but I hope this passes with #519.


niklassiemer:
I only expected two of the tests to fail. Will look into this again. Actually, I assume we should merge this into #519 once I am happy with it, since then it should pass :)

niklassiemer:
Now it behaves like I want it: Two failing tests due to a current bug causing a wrong `Project.root_path` (which is `None` due to `project_check_enabled = False`). I close this here and open a PR to #519 where it should run fine.

PR #522, opened 2021-11-16
Fix compatibility workflows run condition.

### LINKED ISSUE #504: Compatibility workflow always triggered.

niklassiemer:
Restrict compatibility workflows to really run only if the `integration` label is present (and on push to master). Tested in #505


max-hassani (approved):
Looks good to me. It is nicely tested in the other PR 👍

PR #523, opened 2021-11-16
Limit the timeout for windows workflows

max-hassani:
The purpose is to limit the timeout for windows workflows. The default value is 360 minutes. I first wanted to limit only windows timeout, but does it make sense that any of the tests take six hours to finish?!
So I limited all the tests to 30 minutes at maximum. As far as I checked, our tests finish in one or two minutes, so this timeout is even too long, but I wanted to stay on the safe side.

### DIFF
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -47,7 +47,7 @@ jobs:
         pip install --no-deps .
     - name: Test
       shell: bash -l {0}
-      timeout-minutes: ${{matrix.test_timeout}}
+      timeout-minutes: ${{ fromJSON(matrix.test_timeout) }}
       run: |
         cd ../pyiron_atomistics
         python -m unittest discover tests/
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -47,7 +47,7 @@ jobs:
         pip install --no-deps .
     - name: Test
       shell: bash -l {0}
-      timeout-minutes: ${{ fromJSON(matrix.test_timeout) }}
+      timeout-minutes: 30
       run: |
         cd ../pyiron_atomistics
         python -m unittest discover tests/
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -22,7 +22,6 @@ jobs:
       matrix:
         operating-system: [ ubuntu-latest, windows-latest, macos-latest ]
         python-version: [ 3.9 ]
-        test_timeout: 30 # this allows only 10 minutes for the test step to be completed.
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


max-hassani:
The timeout is proven to be working, as it correctly ended some windows workflows after 30 minutes.
I could not use `${{ matrix.test_timeout }}`, since it returns a string rather than an integer. So I used directly the value in the test step. I also tried `${{ fromJSON(matrix.test_timeout )}}`, but was not successful.

.github/workflows/atomistics-compat.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       matrix:
         operating-system: [ ubuntu-latest, windows-latest, macos-latest ]
         python-version: [ 3.9 ]
+        test_timeout: 30 # this allows only 10 minutes for the test step to be completed.
```
niklassiemer: I do not think we need this line, since it is not used?

niklassiemer (approved):
Thank you for searching for the timeout option! LGTM.
However, we after all should try to identify the underlying issue with the windows testing and fix that as well...

niklassiemer:
Interestingly enough, we again get a finished unittest output with passing tests:
```
OK (skipped=46)
TestWithProject: Setting up test project
TestWithProject: Setting up test project
```
Thus, I would assume the cleanup of the unittest suite is causing the problems...

.github/workflows/atomistics-compat.yml
max-hassani: I removed the extra line :)

max-hassani:
> 
> 
> Interestingly enough, we again get a finished unittest output with passing tests:
> 
> ```
> OK (skipped=46)
> TestWithProject: Setting up test project
> TestWithProject: Setting up test project
> Error: The action has timed out.
> ```
> 
> Thus, I would assume the cleanup of the unittest suite is causing the problems...

@niklassiemer, thanks for the review. I think the timeout should be there, whether we fix the issue or not, since we should be aware if a test run so long and take an action.
Another problem, although I have used integration label, the compatibility workflow with atomistics module does not start. Do you think https://github.com/pyiron/pyiron_base/pull/522 will fix this?

niklassiemer:
I will have a look after lunch, since right now the workflows are usually always started?! The new pr should hopefully fix the weird behavior altogether :)

max-hassani:
> 
> 
> I will have a look after lunch, since right now the workflows are usually always started?! The new pr should hopefully fix the weird behavior altogether :)

It started with the new commit :)

niklassiemer:
Well in the meantime it ran, thus, I assume everything is fine :)

niklassiemer:
I might have been wrong. In #527 I just tried to trigger the integration with the label and it also did not work??...

max-hassani:
> 
> 
> I might have been wrong. In #527 I just tried to trigger the integration with the label and it also did not work??...

I can create a branch out of Jan's in #527 with a draft PR, to see what is the reason? I guess only #522 and #523 have changed. Am I right?

niklassiemer:
Yes, probably I after all did not test good enough in #505

PR #526, opened 2021-11-17
Update HasGroups spec

niklassiemer:


specs/has_groups.md
```diff
@@ -12,7 +12,7 @@ also provide a way to actually access the contained objects.  This is done via `
 distinguished from a node by containing other objects it also represents a vertex "containing" or "linked-to" other
 vertices.  From this we formulate the requirements for this interface
 
-1. `__getitem__()` must return a non-`None` object for each string returned by `list_nodes()` and `list_groups()`.
+1. `__getitem__()` must return any or a non-`None` object for each string returned by `list_nodes()` or `list_groups()`, repectively.
```
pmrv: > must return any or a non-`None` object

Not sure what that means.

liamhuber: "any of"?

niklassiemer: Sorry,
'any object or a non-None type object' since getitem of a node might return `None`...

niklassiemer: Is it better now?

niklassiemer:
I just realized we are violating our spec! `Project.list_nodes()` returns a list of jobs. However, any job isinstance of `HasGroups`!

PR #527, opened 2021-11-18
Do we need pytables?

jan-janssen:
pandas needs it as optional dependency to read the HDF5 files but apart from this we do not need it inside pyiron.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -12,8 +12,6 @@ import posixpath
 import shutil
 import warnings
 
-from tables import NoSuchNodeError
-
 from pyiron_base.interfaces.has_groups import HasGroups
 from pyiron_base.job.util import \
     _get_project_for_copy, \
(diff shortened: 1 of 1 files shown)


pmrv:
I think it's no longer necessary since #346, but I must have forgotten to remove the dependency back then.

jan-janssen:
So the issue can still be raised if pandas tries to access the HDF file but the corresponding pandas table is not available, but by relying on the underlying standard errors we do not need to import pytables.

pmrv:
I've checked the code 

https://github.com/pyiron/pyiron_base/blob/15470a312df4f4a88de39391e9eb6e5bb387a0dd/pyiron_base/generic/hdfio.py#L242

and exceptions are already caught there, so I think we can probably get rid of the whole try/except here.  Interestingly `h5py` does not document any `__delitem__` method for their groups, so I'm not sure what the official exception types are, but supposedly a `KeyError` might be enough.

EDIT: [Docs claim 'standard' dictionary API](https://docs.h5py.org/en/stable/high/group.html#dict-interface-and-links), so just `KeyError` should be good.

jan-janssen:
So the `OSError` can happen when the file is accessed by another process at the same time, resulting in a file lock. So I would at least keep those two, but then I am not sure if we can handle the case when pandas fails because they have different errors. But maybe that is something we could test with unit tests as well. I am going to take a look at it.

PR #530, opened 2021-11-18
Fix read-conversion of dtype=object arrays

niklassiemer:


niklassiemer:
Great... mamba seems broken right now
```
Error: Unable to locate executable file: /usr/share/miniconda/condabin/mamba. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```
EDIT: Not the only one with this problem: https://github.com/mamba-org/mamba/issues/1278

PR #533, opened 2021-11-19
Fix writing ragged arrays & small update to update script

### LINKED ISSUE #532: dtype=object considered annoying or rewrite_hdf5 fails on Vasp jobs
The update script right now fails on all Vasp calculations, because the numpy conversion to dtype=object

https://github.com/pyiron/pyiron_base/blob/15470a312df4f4a88de39391e9eb6e5bb387a0dd/pyiron_base/generic/hdfio.py#L230

fails, if the arrays in the list are of shape [(n, k), (n,), (n,),...] (this pattern of shapes comes up for the irreducible kpoints in Vasp calculations).  Why this doesn't work for this pattern of shapes, but for others is something only the numpy developers may divine.  In any case I'll try to find a solution without this conversion today, but before that there's no point in running the update script yet.

pmrv:
Fixes #532.

Currently testing the updater again on all my data, but looking good so far.  The only error now are ones that come from missing HDF5 files.

### DIFF
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -18,6 +18,7 @@ import sys
 from typing import Union
 from pyiron_base.interfaces.has_groups import HasGroups
 from pyiron_base.state import state
+import warnings
 
 __author__ = "Joerg Neugebauer, Jan Janssen"
 __copyright__ = (
@@ -46,7 +47,9 @@ def _is_ragged_in_1st_dim_only(value: Union[np.ndarray, list]) -> bool:
         return False
     else:
         def extract_dims(v):
-            s = np.shape(v)
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                s = np.shape(v)
             return s[0], s[1:]
         dim1, dim_other = zip(*map(extract_dims, value))
         return len(set(dim1)) > 1 and len(set(dim_other)) == 1
--- a/pyiron_base/generic/hdfio.py
+++ b/pyiron_base/generic/hdfio.py
@@ -30,9 +30,11 @@ __email__ = "janssen@mpie.de"
 __status__ = "production"
 __date__ = "Sep 1, 2017"
 
-def _is_ragged_array(value: Union[np.ndarray, list]) -> bool:
+def _is_ragged_in_1st_dim_only(value: Union[np.ndarray, list]) -> bool:
     """
-    Checks whether array or list of lists is ragged.
+    Checks whether array or list of lists is ragged in the first dimension.
+
+    That means all other dimensions (except the first one) still have to match.
 
     Args:
         value (ndarray/list): array to check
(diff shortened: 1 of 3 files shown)


pmrv:
It's not finished yet, but no further errors either, so from my side it's good to go.

pyiron_base/generic/hdfio.py
```diff
@@ -185,10 +190,10 @@ def __getitem__(self, item):
     @staticmethod
     def _is_convertable_dtype_object_array(obj):
         if isinstance(obj, np.ndarray) and obj.dtype == np.dtype(object):
-            first_element = obj[tuple([0 for _ in range(obj.ndim)])]
-            last_element = obj[tuple([-1 for _ in range(obj.ndim)])]
+            first_element = obj[(0,) * obj.ndim]
+            last_element = obj[(-1,) * obj.ndim]
             if isinstance(first_element, numbers.Number) and isinstance(last_element, numbers.Number) \
-                    and not _is_ragged_array(obj):
+                    and not _is_ragged_in_1st_dim_only(obj):
```
niklassiemer, line 199: Here, I would like to stick to the old behavior/function. Indeed it would be cool to really check if numpy will complain about the `np.array(obj.tolist())` in the next call.

pyiron_base/generic/hdfio.py
```diff
@@ -43,8 +45,11 @@ def _is_ragged_array(value: Union[np.ndarray, list]) -> bool:
     if isinstance(value, np.ndarray) and value.dtype != np.dtype("O"):
         return False
     else:
-        shape_lst = [np.shape(sub) for sub in value]
-        return len(set(shape_lst)) > 1
+        def extract_dims(v):
+            s = np.shape(v)
```
niklassiemer: For this call, I already get the numpy deprecation warning... Do we want to catch that or what to do if they really take the automated conversion to dtype=object out?...

niklassiemer (commented):
For the writing the new change is a good change, for the reading I might actually want to define 
```python
def _is_ragged_array(array: np.ndarray) -> bool:
  with warnings.catch_warnings(record=True) as w:
    # Cause all warnings to always be triggered.
    warnings.simplefilter("always")
    np.asarray(array)
    if len(w) > 0 and isinstance(w[0].message, np.VisibleDeprecationWarning): 
      return True
    else:
     return False
```

niklassiemer:
@pmrv what is the status, here? I would like to have this in the next release.

pmrv:
> For the writing the new change is a good change, for the reading I might actually want to define
> 
> ```python
> def _is_ragged_array(array: np.ndarray) -> bool:
>   with warnings.catch_warnings(record=True) as w:
>     # Cause all warnings to always be triggered.
>     warnings.simplefilter("always")
>     np.asarray(array)
>     if len(w) > 0 and isinstance(w[0].message, np.VisibleDeprecationWarning): 
>       return True
>     else:
>      return False
> ```

But converting fully ragged arrays is exactly what caused the issue in #532, so I'd like to keep it this way.  

I will try to come up with a solution for the `np.shape` warning though.

niklassiemer:
> But converting fully ragged arrays is exactly what caused the issue in #532, so I'd like to keep it this way.

Sure, then I did not fully understand the implications 👍 
Then I am fine with this PR, silencing the warning would be the only thing left.

PR #535, opened 2021-11-21
Fix integration workflows to run if needed

niklassiemer:
I hope I finally fixed all the bugs on when the workflows are executed! Some testing in https://github.com/niklassiemer/test_checkout/pull/6

Compatibility workflows correctly
- [x]  skipped if no integration label (a67d5db)
- [x] started on setting integration label
- [x] started on pushing to the PR if integration label is there (c6adc1f)
- [x] skipped on pushing to the PR if the integration label was removed (023aa26)
- [x] started on reopen of the PR if the integration label is there (2fb2f72)
- [x] started on merge with master, although the integration label was removed (6640f11)

### DIFF
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -8,7 +8,7 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-    types: [labeled, opened, synchronize]
+    types: [labeled, opened, synchronize, reopened]
 
 jobs:
   build:
--- a/.github/workflows/atomistics-compat.yml
+++ b/.github/workflows/atomistics-compat.yml
@@ -13,10 +13,9 @@ on:
 jobs:
   build:
     if: |
-      ${{ github.event.type == 'PushEvent' ||
-          ( github.event.type == 'PullRequestEvent'  && contains(github.event.pull_request.labels.*.name, 'integration' ))
-       }}
-
+      github.event_name == 'push' ||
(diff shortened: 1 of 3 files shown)


.github/workflows/atomistics-compat.yml
```diff
@@ -13,10 +13,9 @@ on:
 jobs:
   build:
     if: |
-      ${{ github.event.type == 'PushEvent' ||
-          ( github.event.type == 'PullRequestEvent'  && contains(github.event.pull_request.labels.*.name, 'integration' ))
-       }}
-
+      github.event_name == 'push' ||
```
max-hassani, line 16: Doesn't this `or` condition cause the integration tests to run on all the pushes to the master branch?

niklassiemer: Yes, and I think this should be the intended behavior. In general, we only push the merge commit to the master branch. And every time the master is changed I would run the full test suite with all integration tests to validate the new state. If someone did not use the `integration` label and a merge to the master is failing on the integration tests we are directly alerted.

niklassiemer:
Last test missing is the actual push to master - I think everything is finally fine and working as intended! I will add an integration workflow to the https://github.com/pyiron/pyiron_module_template to avoid future headaches like this.

max-hassani (approved):
LGTM! Thanks for the verifying that the test work as intended in the check boxes :)

niklassiemer:
Well then I merge and let's hope the last box gets a check 😅

niklassiemer:
🎉 they started 😃

PR #536, opened 2021-11-22
Use Subprocess rather than Wrapper Class to prevent memory issues

jan-janssen:


pyiron_base/job/worker.py
```diff
@@ -37,15 +37,25 @@ def worker_function(args):
...
+        "-p", working_directory,
+        "-j", str(job_id)
+    ]
+    try:
+        _ = subprocess.run(
+            executable,
+            cwd=working_directory,
+            shell=False,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
```
pmrv:
```suggestion
            stdout=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL,
```
since we're not reading the output and otherwise a lot output can clog the pipe buffer according to python docs.

pmrv (approved):
Looks good to me.

PR #537, opened 2021-11-22
Allow custom return codes

### LINKED ISSUE #496: Allow custom 'successful' exit codes for executables run by custom job class
I wrote a small example wrapper for an external simulation tool I'm trying to integrate with pyiron based on [this example](https://github.com/materialdigital/PMD-workflow-workshop/blob/master/2_customized_job/2_1_custom_bash_job.ipynb).

This seems to work fine so far, however, the external tool returns non-zero exit codes in some cases. Those aren't failure states but useful return values. Therefore, I would like pyiron to not interpret those as errors.

pmrv:
Some executables may use more then just return code 0, to signal
a successful run.  This change allows set additional return codes that
should be accepted as successful.

@s4b7r Please give this a try if it resolves #496 for you.  You should be able to set the custom return codes simply by
```python
class ToyBashJob(GenericJob):
        self.input = GenericParameters(table_name="input")
        self.input['input_energy'] = 100
        self.executable = "cat input > output"
        self.executable.accepted_return_codes += [1,2,3]
...
```

### DIFF
--- a/pyiron_base/job/executable.py
+++ b/pyiron_base/job/executable.py
@@ -85,7 +85,7 @@ class Executable(object):
     @accepted_return_codes.setter
     def accepted_return_codes(self, value):
         if not isinstance(value, list) or any(not isinstance(c, int) or c > 255 for c in value):
-            raise ValueError(f"accepted_return_codes must be a list of integers <= 255!")
+            raise ValueError("accepted_return_codes must be a list of integers <= 255!")
         self._accepted_return_codes = value
 
     @property
--- a/pyiron_base/job/executable.py
+++ b/pyiron_base/job/executable.py
@@ -73,6 +73,20 @@ class Executable(object):
         self._mpi = False
         if self._executable_lst:
             self.version = self.default_version
+        self._accepted_return_codes = [0]
+
+    @property
+    def accepted_return_codes(self):
+        """
+        list of int: accept all of the return codes in this list as the result of a successful run
+        """
+        return self._accepted_return_codes
+
+    @accepted_return_codes.setter
+    def accepted_return_codes(self, value):
+        if not isinstance(value, list) or any(not isinstance(c, int) or c > 255 for c in value):
+            raise ValueError(f"accepted_return_codes must be a list of integers <= 255!")
+        self._accepted_return_codes = value
 
     @property
     def version(self):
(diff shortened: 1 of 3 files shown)


niklassiemer (approved):
lgtm.

pyiron_base/job/executable.py
```diff
@@ -73,6 +73,20 @@ def __init__(
...
+    def accepted_return_codes(self):
+        """
+        list of int: accept all of the return codes in this list as the result of a successful run
+        """
+        return self._accepted_return_codes
+
+    @accepted_return_codes.setter
+    def accepted_return_codes(self, value):
+        if not isinstance(value, list) or any(not isinstance(c, int) or c > 255 for c in value):
+            raise ValueError(f"accepted_return_codes must be a list of integers <= 255!")
```
niklassiemer: Actually, codacy got a point here:
```suggestion
            raise ValueError("accepted_return_codes must be a list of integers <= 255!")
```

s4b7r:
I updated pyiron recently (to pyiron 0.4.5, pyiron_base 0.4.1) to work around another issue (pyiron/pyiron#1380). In this environment I can not reproduce my original issue of pyiron failing due to non-zero exit codes. Was there some change, that could affect this behaviour? When going back to my previous environment (pyiron 0.4.5, pyiron_base 0.3.2) I can confirm pyiron still aborts the job due to the non-zero exit code.

niklassiemer:
There was some changes on the sub-process handling (https://github.com/pyiron/pyiron_base/pull/498) @jan-janssen made to allow for really large scale calculations. This might be related but otherwise I do not know! However, assuming your code after all gives a non-zero exit code and we are not recognizing this any more - this is a new bug!

pmrv:
> There was some changes on the sub-process handling (#498) @jan-janssen made to allow for really large scale calculations. This might be related but otherwise I do not know! However, assuming your code after all gives a non-zero exit code and we are not recognizing this any more - this is a new bug!

Indeed #498 introduced the bug by switching from `subprocess.check_output` to `subprocess.run` and not specifying `check=True`.  According to the [docs](https://docs.python.org/3/library/subprocess.html?highlight=check_output#subprocess.check_output), `check_output` is equivalent to `run(..., stdout=PIPE, check=True)`.  I've pushed the fix and tests here.

pmrv:
Since this is a bug in the return handling and blocks using master on the cluster, I'll merge this after the pyiron meeting, if no one objects.

PR #541, opened 2021-11-22
Disable `project_check_enable` and retain `root_path`/`project_path`

### LINKED ISSUE #538: Changed behavior: Current path not included in `project_paths`
Right now quite some of out CI are failing (atomistics Notebooks, integration tests for contrib was failing before) since we did not properly set up the `.pyiron` configuration. 
```python
ValueError: the current path /home/runner/work/pyiron_atomistics/pyiron_atomistics/tip3p_water/ is not included in the .pyiron configuration 'project_paths': []
```
Before our state updates, we must have had something like `project_check_enable = False if project_paths = []` or the like.

### LINKED ISSUE #539: `project_check_enabled` is misleading
In my opinion, our setting `project_check_enabled` is somewhat misleading or at least changing to much on the behavior side. 
Right now if I run 
```python
from pyiron_base import state, Project

state.settings.update({'project_paths': ['/mnt/c/Users/Siemer/pyiron/projects/'],
                      'project_check_enabled': True})
pr = Project('dummy')
print(pr.root_path, pr.project_path)
>>> /mnt/c/Users/Siemer/pyiron/projects/ dummy/

state.settings.update({'project_paths': ['/mnt/c/Users/Siemer/pyiron/projects/'],
                      'project_check_enabled': False})
pr = Project('dummy')
print(pr.root_path, pr.project_path)
>>> None /mnt/c/Users/Siemer/pyiron/projects/dummy/
```

However, I do not see a reason why the output of  `pr.root_path` and `pr.project_path` should change if `project_check_enabled` is changed! The behavior of restricting access to not known paths is the only expected change in my opinion.

niklassiemer:
Following our discussions in the meeting, I now removed the setup of our `.pyiron` in the setup. This also tests the intended use case `conda install pyiron` and running notebooks.

### DIFF
--- a/tests/database/test_manager.py
+++ b/tests/database/test_manager.py
@@ -19,9 +19,11 @@ class TestDatabaseManager(TestWithProject):
         self.assertEqual(self.s.configuration["disable_database"], self.dbm.database_is_disabled,
                          msg="Database manager should be initialized by settings.")
         self.dbm._database_is_disabled = True
-        self.assertNotEqual(self.s.configuration["disable_database"], self.dbm.database_is_disabled,
-                            msg="But after that it should be independent from the settings")
-        self.dbm._database_is_disabled = False  # Re-enable it at the end of the test
+        try:
+            self.assertNotEqual(self.s.configuration["disable_database"], self.dbm.database_is_disabled,
+                                msg="But after that it should be independent from the settings")
+        finally:
+            self.dbm._database_is_disabled = False  # Re-enable it at the end of the test
 
     def test_file_top_path(self):
         # Store settings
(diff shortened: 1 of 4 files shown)


liamhuber:
This else clause seems like a super shady way to update the default, or?

niklassiemer:
Well, yes and no, we just have (and had) different defaults in case we did not have any configuration. This might have been a second reason to change the default in the first place: 
https://github.com/pyiron/pyiron_base/blob/a71f560a8b26d4306d4cbde10f079c8ff5dd600b/pyiron_base/settings/generic.py#L76-L79
https://github.com/pyiron/pyiron_base/blob/a71f560a8b26d4306d4cbde10f079c8ff5dd600b/pyiron_base/settings/generic.py#L106-L115

niklassiemer:
I could of course refactor into something like, `self._update_defaults_if_no_config_found` to make it more readable.

liamhuber:
@niklassiemer yes, those are both the old behaviour. My goal recently was to purge this sort of misdirection. Why should the default silently differ when I do/don't provide any overriding config values? I'm sorry I missed the discussion today -- maybe someone can post the minutes so I can get up to speed? I am 100% behind setting `project_check_enabled = False` as the default behaviour (and tried and failed to do this recently) -- but silently changing the default value for this parameter depending whether other (unrelated!) config values were set is an extremely dangerous solution and I strongly oppose it. Let's rather change the default and fix whatever other stuff needs to be fixed so we can do that.

liamhuber:
> I could of course refactor into something like, `self._update_defaults_if_no_config_found` to make it more readable.

It's not the readability from the developer side that worries me -- for this the else clause is already sufficiently clear. It's the user side. Right now the behaviour is:

- Get defaults
- Overwrite them from zero or one source

With this the behaviour is:

- Get defaults
- Overwrite them from zero or one source
- If it was zero sources, silently update one of the values to a new default

I get that this resolves the bad symptom of needing .pyiron files in the other repos, but the side effect is not pleasant and it does nothing to treat the underlying disease (which is the weird and poorly encapsulated behaviour of project checking)

niklassiemer:
Ok, I am absolutely on board with this! Thus, #539 is the underlying issue. I would suggest to change the default for the `project_check_enable` again to False (like you had in the first state PR), but retain the behavior of splitting the full path into `root_path` and `project_path`. Currently, this is silently changed to `None` and Full-path, respectively, if the check is disabled. This causes problems in finding old calculations in the database even if the configuration is simply not changed at all. Obviously a no-go. 

With my suggested change to disentangle the 'this path is not allowed' error and the split into `root_path` and `project_path` depending on the current `project_paths` we affect the least people. Only those who were running with `project_check_enable=False` and `project_paths!=[]`. However, since `project_path` is to my knowledge only used for the `project_check_enable` these people could restore their old behavior by setting `project_paths=[]`.

liamhuber:
That sounds super, Niklas, thanks!

niklassiemer:
Ok, I see we all agree on the same thing 👍 I will try to fix #539 by also splitting the paths if `project_check_enable is False` and make this the default, again. If I can't get it running till the next release I would suggest to re-introduce the awful `else` as a quick and dirty fix and revert that with a working version of `project_check_enable`.

liamhuber:
> Ok, I see we all agree on the same thing 👍 I will try to fix #539 by also splitting the paths if `project_check_enable is False` and make this the default, again. If I can't get it running till the next release I would suggest to re-introduce the awful `else` as a quick and dirty fix and revert that with a working version of `project_check_enable`.

Aha, I see the original `else` now. Lines 207-209 from pyiron_base/settings/generic.py:

```python
        else:
             self._configuration["sql_file"] = "~/pyiron.db"
             self._configuration["project_check_enabled"] = False
```

I definitely biffed that in my refactor. I don't like the behaviour though.

I really hope we can get away without the dirty fix, but indeed this sounds like a good plan 👍

niklassiemer:
To change just the order of operations in the `DatabaseManager.top_path` method seems to be enough already 🎉

liamhuber:
> To change just the order of operations in the `DatabaseManager.top_path` method seems to be enough already 🎉

Stellar! I'll be a bit delayed in reviewing, but an extremely quick skim looked like everything is in exactly the right direction 🥳

tests/database/test_manager.py
```diff
@@ -19,9 +19,11 @@ def test_database_is_disabled(self):
         self.assertEqual(self.s.configuration["disable_database"], self.dbm.database_is_disabled,
                          msg="Database manager should be initialized by settings.")
         self.dbm._database_is_disabled = True
-        self.assertNotEqual(self.s.configuration["disable_database"], self.dbm.database_is_disabled,
-                            msg="But after that it should be independent from the settings")
-        self.dbm._database_is_disabled = False  # Re-enable it at the end of the test
+        try:
+            self.assertNotEqual(self.s.configuration["disable_database"], self.dbm.database_is_disabled,
+                                msg="But after that it should be independent from the settings")
+        finally:
```
liamhuber, line 25: I like this `finally` flow, I will adopt it more in my own work

tests/database/test_manager.py
```diff
@@ -30,24 +32,46 @@ def test_file_top_path(self):
...
-            with self.subTest('enable project_check_enabled'):
-                new_root_path = self.s.convert_path_to_abs_posix(os.getcwd())
-                self.s.configuration["project_check_enabled"] = True
-                self.s.configuration["project_paths"] = [new_root_path]
+            new_root_path = self.s.convert_path_to_abs_posix(os.getcwd())
+            self.s.configuration["project_check_enabled"] = True
+            self.s.configuration["project_paths"] = [new_root_path]
+            self.s.configuration["disable_database"] = False
+
+            with self.subTest('enabled'):
```
liamhuber, line 40: Also these subtests. I've been learning lots from reviews lately 😄

liamhuber (approved):
I got a quiet moment to take a deeper look sooner than anticipated. Lgtm! I'm glad the actual fix is so localized.

niklassiemer:
If nobody complains, I will merge this one and the #547 PR this afternoon and release a new base and also finally a hopefully well working atomistics with all the tests passing.

max-hassani (approved):
LGTM! As Liam mentioned, I also liked the changes to the test! Thanks, Niklas 👍

PR #542, opened 2021-11-22
set `project_check_enabled` to False if no configuration found

niklassiemer:


niklassiemer:
This was also the behavior before the state PRs:
https://github.com/pyiron/pyiron_base/blob/a71f560a8b26d4306d4cbde10f079c8ff5dd600b/pyiron_base/settings/generic.py#L76-L79
https://github.com/pyiron/pyiron_base/blob/a71f560a8b26d4306d4cbde10f079c8ff5dd600b/pyiron_base/settings/generic.py#L106-L115

PR #547, opened 2021-11-22
Fix publication

jan-janssen:
The issue first appeared in https://github.com/pyiron/pyiron_gui/pull/93

### DIFF
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -19,3 +19,4 @@ dependencies:
 - pytables =3.6.1
 - sqlalchemy =1.4.27
 - tqdm =4.62.3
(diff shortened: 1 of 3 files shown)


jan-janssen:
And in https://github.com/pyiron/pyiron_contrib/pull/270

niklassiemer:
Thanks a lot! Is this related to the python versions we are checking? If it is, we probably should include the oldest supported version in some tests...

jan-janssen:
```
from typing import Literal
```
is available for python 3.8 and later - but the backwards compatibility works for all versions.

For more Details: https://stackoverflow.com/questions/61206437/importerror-cannot-import-name-literal-from-typing

PR #548, opened 2021-11-23
Extending tests to run one unittest suite for the oldest supported python

niklassiemer:
In https://github.com/pyiron/pyiron_base/pull/547 we need to fix broken python3.6 support. We have to do this since we were only alerted on the unittests of other repositories running older python version:
- pyiron/pyiron_gui#93
- pyiron/pyiron_contrib#270

Thus, I now add one unit test run with python3.6 on a linux machine to check for this type of errors.

### DIFF
--- a/.github/workflows/old_python.yml
+++ b/.github/workflows/old_python.yml
@@ -1,0 +1,34 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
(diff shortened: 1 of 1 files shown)


niklassiemer:
Sadly, due to some of our dependencies dropping older python versions support, I cannot install the same packages under python3.6. For now, I test at least python3.7, however, I would like to think about how to check our own compatibility to python3.6 (or drop the support).

niklassiemer:
For python3.7 we have the ability to build the same environment and it is correctly failing now, thus, I merge this now.

PR #550, opened 2021-11-23
convert missing_ok to try/except

niklassiemer:
The option missing_ok is a python-3.8 feature internally using try/except.


niklassiemer:
Unittests / build (windows-latest, 3.9): 
```
Ran 389 tests in 59.029s

OK (skipped=46)
```

PR #555, opened 2021-11-23
Do not show job remove progress bar during unit tests

pmrv:
I like the feature for interactive use, but it clutters the output of the unit tests a lot, so I disable it in `TestWithCleanProject`.


niklassiemer (approved):
Agreed!

pmrv:
I've added another change that doesn't show the progress bar when there are no jobs to be removed, just because I thought it looks weird to get a print out like `0it [00:00, ?it/s]` from `tqdm`.

Issue #559, opened 2021-11-24
table.run() fails, no `key` attribute for `Dataset` object


max-hassani:
I am trying to run the examples from the phase diagram workshop with our latest releases of base and atomistics, and I received the following error:
```
'Dataset' object has no attribute 'keys'
```
when I run:
```python
def filter_jobs(job):
    return job.status == "finished" and "lmp_" in job.job_name
def get_lattice_constant(job): 
    return job["output/generic/cells"][-1,0,0]
def get_job_name(job):
    return job.job_name
table.filter_function = filter_jobs
table.add["job_name"] = get_job_name
table.add["lattice_constant"] = get_lattice_constant
table.run()
```
the detailed erorr is as followed:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_14522/119277569.py in <module>
      1 # Execute the pyiron table just like a pyiron job object
----> 2 table.run()

~/miniconda3/envs/test_phase/lib/python3.8/site-packages/pyiron_base/generic/util.py in decorated(*args, **kwargs)
    212                         stacklevel=2
    213                     )
--> 214             return function(*args, **kwargs)
    215         return decorated
    216 

~/miniconda3/envs/test_phase/lib/python3.8/site-packages/pyiron_base/job/generic.py in run(self, delete_existing_job, repair, debug, run_mode, run_again)
    663                 self._run_if_repair()
    664             elif status == "initialized":
--> 665                 self._run_if_new(debug=debug)
    666             elif status == "created":
    667                 self._run_if_created()

~/miniconda3/envs/test_phase/lib/python3.8/site-packages/pyiron_base/job/generic.py in _run_if_new(self, debug)
   1317         else:
   1318             self.save()
-> 1319             self.run()
   1320 
   1321     def _run_if_created(self):

~/miniconda3/envs/test_phase/lib/python3.8/site-packages/pyiron_base/generic/util.py in decorated(*args, **kwargs)
    212                         stacklevel=2
    213                     )
--> 214             return function(*args, **kwargs)
    215         return decorated
    216 

~/miniconda3/envs/test_phase/lib/python3.8/site-packages/pyiron_base/job/generic.py in run(self, delete_existing_job, repair, debug, run_mode, run_again)
    665                 self._run_if_new(debug=debug)
    666             elif status == "created":
--> 667                 self._run_if_created()
    668             elif status == "submitted":
    669                 self._run_if_submitted()

[...]

PR #560, opened 2021-11-25
Store pandas dataframe as single object in data frame 

jan-janssen:
Converting to dictionary is terribly slow: 
```
import pandas
with open("test.csv", "w") as f:
    f.write(",a,b\n")
    for i in range(100000): 
        f.write(str(i) + ",1,2\n")
df = pandas.read_csv("test.csv", index_col=0)
pr = Project(".")
hdf = pr.create_hdf(".", "test")
dc = DataContainer(df.to_dict())
dc.to_hdf(hdf)
```
This takes forever, while with a small change it is in the order of ms: 
```
dc = DataContainer(df)
```

### DIFF
--- a/pyiron_base/table/datamining.py
+++ b/pyiron_base/table/datamining.py
@@ -614,7 +614,7 @@ class TableJob(GenericJob):
     def _save_output(self):
         with self.project_hdf5.open("output") as hdf5_output:
             df = self.pyiron_table._df
-            cont = DataContainer(df.to_dict())
+            cont = DataContainer(df)
             cont.to_hdf(hdf=hdf5_output, group_name="table")
 
     def to_hdf(self, hdf=None, group_name=None):
--- a/pyiron_base/table/datamining.py
+++ b/pyiron_base/table/datamining.py
@@ -613,9 +613,7 @@ class TableJob(GenericJob):
 
     def _save_output(self):
         with self.project_hdf5.open("output") as hdf5_output:
-            df = self.pyiron_table._df
-            cont = DataContainer(df)
-            cont.to_hdf(hdf=hdf5_output, group_name="table")
+            self.pyiron_table._df.to_hdf(hdf5_output, "table")
 
     def to_hdf(self, hdf=None, group_name=None):
         """
@@ -692,8 +690,7 @@ class TableJob(GenericJob):
         if hdf_version=="0.3.0":
             with self.project_hdf5.open("output") as hdf5_output:
                 if "table" in hdf5_output.list_groups():
-                    data = hdf5_output["table"].to_object().to_builtin()
-                    self._pyiron_table._df = pandas.DataFrame(data)
+                    self._pyiron_table._df = pandas.read_hdf(hdf5_output, "table")
         else:
             pyiron_table = os.path.join(self.working_directory, "pyirontable.csv")
             if os.path.exists(pyiron_table):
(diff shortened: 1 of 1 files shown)


jan-janssen:
This fixes my issues with https://github.com/pyiron/pyiron_base/pull/463/

pmrv:
Have you checked that this actually stores the data?  In my testing conversion from pandas is broken.

```python
df = DataFrame({'a': range(4), 'b': range(4,8), 'c': range(8,12)})
dc = DataContainer(df)
print(dc)
> DataContainer([])
dc = DataContainer(df.to_dict())
print(dc)
> DataContainer({'a': DataContainer([0, 1, 2, 3]), 'b': DataContainer([4, 5, 6, 7]), 'c': DataContainer([8, 9, 10, 11])})
```

The reason I didn't save it via pandas is that I expect people to put data into a pyiron_table that is HDF-serializable to us, but not to pandas (e.g. structures).  I'll have a look today at #463 and #559.

pyiron_base/table/datamining.py
```diff
@@ -692,8 +690,7 @@ def from_hdf(self, hdf=None, group_name=None):
         if hdf_version=="0.3.0":
             with self.project_hdf5.open("output") as hdf5_output:
                 if "table" in hdf5_output.list_groups():
-                    data = hdf5_output["table"].to_object().to_builtin()
-                    self._pyiron_table._df = pandas.DataFrame(data)
+                    self._pyiron_table._df = pandas.read_hdf(hdf5_output.file_name, hdf5_output.h5_path + "/table")
```
pmrv, line 693: This will break without `pytables` or not?

jan-janssen:
> The reason I didn't save it via pandas is that I expect people to put data into a pyiron_table that is HDF-serializable to us, but not to pandas (e.g. structures). I'll have a look today at #463 and #559.

That is definitely a limitation of the current solution, currently I sure the structures as JSON objects before inserting them into the pyiron table, but that requires extra work on the user side. The advantage is the size difference. Already if I only load the first 10000 entries of the file in the example above the old code resulted in a 124MB HDF5 file. So the conversion to dictionary and storing in the Datacontainer is definitely not scalable for large pyiron tables.

pyiron_base/table/datamining.py
jan-janssen: Yes, I rely on pytables - but as long as that is the recommended solution by pandas I guess that is the only option. Alternatively we could also pickle the object and store the pickled string, but I do not think that is a better alternative.

pmrv: Then there's no need to put the `DataFrame` into the data container at all and we can just `DataFrame.to_hdf`/`read_hdf`, I suppose.  Then we also need to put `tables` back into our `environment.yml`.

jan-janssen: That I exactly what I do in the latest commit.

pmrv:
I'm looking into the file size right now and it seems like an issue with how we write to HDF5 in general.  When I do a `h5repack` on the file in the example, I directly get a factor ~20 decrease in file size.

pyiron_base/table/datamining.py
pmrv: Ah, I missed that.  Looks good to me then, we can always come back to this if someone really wants to store "complex" data in their pyiron tables.

EDIT: The dependency still should be added, though.

pmrv (approved):

jan-janssen:
The dependency is still included https://github.com/pyiron/pyiron_base/blob/pyiron_table_store_using_pandas/setup.py#L45

PR #561, opened 2021-11-25
Replace automatically special symbols in job names

### LINKED ISSUE #549: Create job name automatically
It would be nice to replace `create_job` by:

```python
def create_job(name, *args, delete_existing_jobs=..., ...):
    return name + "_".join(map(some_sanitizer(args))
```

Function suggested by @pmrv  

`some_sanitizer` should do
- Replacement of special symbols (`+`, `.` etc.)
- Rounding up of floating points `5.00000000000000004`

So in total, it should be like

```python
job = pr.create.job.SomeJob('run', lattice_parameter, temperature, pressure, box_repeat, index)
```

samwaseda:
Closes #549 

This wraps up the discussion of the thread above. Now you can set something like:

```python
job = pr.create.job.RandomJob(('job', temperature, 'K', pressure, 'GPa', *structure.cell.diagonal()))
```

This will give a job name like `'job_300d0_K_0_GPa_3d52_3d52_3d52'`


samwaseda:
Aha this doesn't work for paths like `/cmmc/u/samsstud/dev_sam/pyiron_base/tests/generic/../static/dft`. Maybe I'm gonna simply split the string by `/` and do the replacement for the last one.

tests/generic/test_fileHDFio.py
```diff
@@ -375,6 +376,13 @@ def test_remove_group(self):
         # This should not raise an error, albeit the group of hdf is removed
         hdf.remove_group()
 
+    def test_get_safe_job_name(self):
+        T = 300.8000000000002
+        p = 0.30000000000000004
+        self.assertEqual(_get_safe_job_name(('job', T, 'K', p, 'GPa')), 'job_300d8_K_0d3_GPa')
+        self.assertEqual(_get_safe_job_name('file_name.h5', extension='.h5'), 'file_name.h5')
+        self.assertEqual(_get_safe_job_name('file_name', extension='.h5'), 'file_name.h5')
```
niklassiemer:
```suggestion
        with self.subTest("Round floats"):
            self.assertEqual(_get_safe_job_name(('job', T, 'K', p, 'GPa')), 'job_300d8_K_0d3_GPa')
        with self.subTest("File name with extention"):    
            self.assertEqual(_get_safe_job_name('file_name.h5', extension='.h5'), 'file_name.h5')
        with self.subTest("File name with extention"):
            self.assertEqual(_get_safe_job_name('file_name', extension='.h5'), 'file_name.h5')
        with self.subTest("unsafe string"):
            self.assertEqual(_get_safe_job_name("2-be the.most+awful,string"), "2mbe_thedmostpawfulcstring")
```

pyiron_base/generic/hdfio.py
```diff
@@ -31,6 +31,35 @@
 __status__ = "production"
 __date__ = "Sep 1, 2017"
 
+_special_symbol_replacements = {
+    '.': 'd',
+    '-': 'm',
+    '+': 'p',
+    ',': 'c',
```
niklassiemer:
```suggestion
    ',': 'c',
    ' ': '_',
```

pyiron_base/generic/hdfio.py
```diff
@@ -31,6 +31,35 @@
...
+            return round(value, ndigits=ndigits)
+        return value
+    if not isinstance(name, str):
+        name_rounded = [round_(nn) for nn in name]
+        job_name = '_'.join([str(nn) for nn in name_rounded])
+    else:
+        job_name = name
+    if extension is not None and job_name.endswith(extension):
+        job_name = job_name[:job_name.rfind(extension)]
+    job_name = job_name.split('/')
```
niklassiemer: Thus, we require the first member of `name` to be a path, not one of the others?  Should we support paths at all in this kind of name sanitizer?

niklassiemer (commented):
I am not sure if we actually should handle full paths with this function or just get a safe job name.

tests/generic/test_fileHDFio.py
samwaseda: XD

pyiron_base/generic/hdfio.py
samwaseda: No this is just a security measure in order not to change the path. So if we get something like the one above: `/cmmc/u/samsstud/dev_sam/pyiron_base/tests/generic/../static/dft`, we should make sure that it doesn't erroneously change the path. Otherwise we could simply skip the replacement altogether as soon as `/` is detected.

samwaseda: Actually we shouldn't need it in the first place in `ProjectHDFio`, but I saw that the job name is directly passed there if it's created via `pr.create_job` and not via `pr.create.job`.

niklassiemer: Then I would be in favor of sanitizing the job_name in `create_job` and exclude any paths from conversion in this function, i.e. ValueError on `/`, `\`, `\\`.

samwaseda: ok I'm gonna do it :)

pyiron_base/generic/hdfio.py
```diff
@@ -31,6 +31,36 @@
...
+    for k, v in _special_symbol_replacements.items():
+        job_name[-1] = job_name[-1].replace(k, v)
+    job_name = '/'.join(job_name)
+    if extension is not None:
+        job_name += extension
+    return job_name
```
niklassiemer: I just found that we only allow valid python identifiers with a length<50 for job names and that we have a function to check this. Thus, I suggest to use `from pyiron_base.job.util import _is_valid_job_name` and end the function with 

```suggestion
    try:
        _is_valid_job_name(job_name)
    except ValueError as e:
        raise ValueError(f"Could not convert {name} to valid job name"} from e
    else:
        return job_name
```

samwaseda: Aaaaaaah good point

pyiron_base/job/util.py
```diff
@@ -113,12 +113,36 @@ def _get_project_for_copy(job, project, new_job_name):
     return file_project, hdf5_project
 
 
+_special_symbol_replacements = {
+    '.': 'd',
```
max-hassani, line 117: I suggest that instead of having single characters replacing minus or plus sign, we use `_plus_` and `_minus_`, since `p` and `c` can be confusing with pressure and concentration. `.` can also be replaced by `_point_`. This is just a suggestion from my side. Other than that it looks good to me.

max-hassani (approved):

pyiron_base/job/util.py
niklassiemer: In principle I would agree, however, we have a 50 character limit 😅

samwaseda: I have exactly the same concern as @niklassiemer (and indeed in principle I agree that the replacement should be more explicit). But maybe rather than focusing on making the names more explicit, we can make both this and the inverse conversion available to the user, i.e. also a function that turns `'job_300d0_K_0_GPa_3d52_3d52_3d52'` into `['job', 300.0, 'K', 0, 'GPa', 3.52, 3.52, 3.52]`, which I think is feasible since `d`, `m` etc. appear only next to a number of an underscore.

niklassiemer: There, I would rather bring up the idea of metadata for all kinds of pyiron objects. This way we just store the initial list instead of trying to back-convert.

max-hassani: You are right :+1:

samwaseda: > There, I would rather bring up the idea of metadata for all kinds of pyiron objects. This way we just store the initial list instead of trying to back-convert.

yeah in principle I'd agree, but then often there's the reality that comes in between that it's much faster to screen the job names than loading the meta data, or sometimes it's a value that the user defines (like strain, which wouldn't be stored in pyiron), in which case there's the double work of defining the name and storing the data in the metadata for the user.

niklassiemer: I meant to automatically store the provided data for the name generation and return this on request. Otherwise, I fear we would have a hard time to convert back properly, e.g. I would expect some 'failures' like here:
```python
_get_safe_job_name(['some very', 300.5, 'T', 'bad.example'])
>>> 'some_very_300d5_T_baddexample' 
_revert_get_safe_job_name('some_very_500d5_T_baddexample')
>>> ['some', 'very', 300.5, 'T', 'baddexample']
```

samwaseda: ah I see. That's a good idea :)

pmrv: On the other hand, if you're back converting data like this, it means we're not storing enough data in the proper places.  E.g. temperatures, k-points, number of iterations, etc. should all be present in `job['input']` and therefore accessible e.g. via pyiron tables.

samwaseda: The problem is that there are some time human concepts in the input, which don't exist in atomistic simulations as such, e.g. grain boundaries, dislocations, vacancies etc. So when they are input values, like the sigma value of a grain boundary or the dislocation structure, we have to rely on the input that we provide ourselves.

samwaseda:
Since we already talked about the interface in #549 I'm gonna merge this one now. Extensions to this one in farther PRs.

PR #563, opened 2021-11-25
Worker: Set job to aborted after child runtime expires

jan-janssen:


pmrv:
Isn't a timeout the expected outcome for a `Worker` job?  My understanding of the use case so far was, that I create a bunch of jobs with `mode=worker` and then start a `Worker` that runs them till the cluster allocation runs out.  I don't feel strongly about this.

Technically I'm ok with the solution here to track the runtime of the children, but mdi-term it would be cool to be able to query `pysqa` for the remaining run time.

jan-janssen:
> Isn't a timeout the expected outcome for a `Worker` job? My understanding of the use case so far was, that I create a bunch of jobs with `mode=worker` and then start a `Worker` that runs them till the cluster allocation runs out. I don't feel strongly about this.

The time out here is for the execution of one of the child jobs. I am working with a rather unstable simulation code as child and while most of the jobs are correctly set to `aborted` when they fail, some remain in status `running`. This limitation is addressed by the `child_runtime` if the job is not finished within this time the master job overwrites the status to `aborted`.

pmrv:
> > Isn't a timeout the expected outcome for a `Worker` job? My understanding of the use case so far was, that I create a bunch of jobs with `mode=worker` and then start a `Worker` that runs them till the cluster allocation runs out. I don't feel strongly about this.
> 
> The time out here is for the execution of one of the child jobs. I am working with a rather unstable simulation code as child and while most of the jobs are correctly set to `aborted` when they fail, some remain in status `running`. This limitation is addressed by the `child_runtime` if the job is not finished within this time the master job overwrites the status to `aborted`.

Ah, I completely misunderstood then.

PR #565, opened 2021-11-26
Add interactive with statement

jan-janssen:
```
pr = Project("interactive")
structure = pr.create.structure.ase.bulk("Al", cubic=True)
structure_copy = structure.copy()
structure_copy.positions[0] = [0.1, 0.1, 0.1]
lmp = pr.create.job.Lammps("lmp")
lmp.structure = structure
with lmp as lmp_int:
    for s in [structure, structure_copy]:
        lmp_int.structure = s
        lmp_int.run()
pr.job_table()
lmp.output.energy_tot
```


liamhuber:
The bare with is very speedy, but I find it hard to read. To be explicit rather than implicit I think we need something like 'with lmp.open() as lmp_int' or 'with pr.interactive(lmp) as lmp_int'.

These might even just wrap the exact functionality you show, but I think our canonical call should be more explicit about what's happening

jan-janssen:
The updated code is:
```
pr = Project("interactive")
structure = pr.create.structure.ase.bulk("Al", cubic=True)
structure_copy = structure.copy()
structure_copy.positions[0] = [0.1, 0.1, 0.1]
lmp = pr.create.job.Lammps("lmp")
lmp.structure = structure
with lmp.interactive_open() as lmp_int:
    for s in [structure, structure_copy]:
        lmp_int.structure = s
        lmp_int.run()
pr.job_table()
lmp.output.energy_tot
```

pyiron_base/job/generic.py
```diff
@@ -1629,6 +1629,12 @@ def _reload_update_master(self, project, master_id):
                 "busy master: {} {}".format(master_id, self.get_job_id())
             )
             del self
+            
+    def __enter__(self):
```
samwaseda: Since `interactive_open` is now required I find this error message redundant.

jan-janssen: @samwaseda I agree - removed it.

samwaseda:
Ah wait I'm now confused. What happens now if you do `with lmp as lmp_int`? According to the definition of `__enter__`, regardless of `run_mode`, it still returns `self` doesn't it? So what happens with the following code?

```python
lmp = pr.create.job.Lammps('test')
lmp.structure = some_structure
with lmp as lmp_int:
    lmp.run()
```

Doesn't it call `interactive_close` even though the job is already completed? Doesn't it raise an error?

jan-janssen:
> Ah wait I'm now confused. What happens now if you do `with lmp as lmp_int`? According to the definition of `__enter__`, regardless of `run_mode`, it still returns `self` doesn't it?

Then you open the HDF5 file and get a `jobpath` object in return.

samwaseda:
Sorry I edited my comment above to add an example.

jan-janssen:
> Doesn't it call `interactive_close` even though the job is already completed? Doesn't it raise an error?

Calling `interactive_close()` twice does not matter it is only executed when the job is running.

samwaseda:
Ok but I still have three minor concerns:

1. There could be a class in which `interactive_close` does more stuff
2. By looking at the code it's not really clear how `__enter__` is meant to be used (I guess there should be an example in the DocString of `interactive_close`)
3. It is often the case (at least in my notebooks) that a cell ends with `interactive_open` to set up a simulation, in which case I see now the job output instead of just `None`, which would look super weird because it would appear to me like I should assign it to a variable.

With this in mind, we can also think about writing it this way:

```python
class JobHolder:
    def __init__(self, job):
        self.job = job

    def __repr__(self):
        return ''

    def __enter__(self):
        return self.job
    
    def __exit__(self, exc_type, exc_val, exc_tb):
        return self.job.interactive_close()


class InteractiveBase:
    def interactive_open(self):
        self.server.run_mode.interactive = True
        return JobHolder(self)
```

We can also think about putting some text in `__repr__` which explains how it works if that's helpful.

jan-janssen:
> Ok but I still have three minor concerns:
> 
> 1. There could be a class in which `interactive_close` does more stuff

All `interactive_close` methods look like this: 
```
def interactive_close(self):
        if self.interactive_is_activated():
           # do something
```
I am happy to hide this in a separate function like: 
```
def _interactive_close(self):
        if self.interactive_is_activated():
           self.interactive_close()  # This is what the developer defines for a specific code
```
Or maybe the other way around: 
```
def interactive_close(self):
        if self.interactive_is_activated():
           self._interactive_close()  # This is what the developer defines for a specific code
```

> 2. By looking at the code it's not really clear how `__enter__` is meant to be used (I guess there should be an example in the DocString of `interactive_close`)

I agree. 

> 3. It is often the case (at least in my notebooks) that a cell ends with `interactive_open` to set up a simulation, in which case I see now the job output instead of just `None`, which would look super weird because it would appear to me like I should assign it to a variable.

I have the feeling this only makes it more confusing. `open()` always returns an object so `interactive_open()` not returning an object is not intuitive, so that should be changed.

samwaseda:
> I have the feeling this only makes it more confusing. `open()` always returns an object so `interactive_open()` not returning an object is not intuitive, so that should be changed.

I totally disagree. The reason why `open` returns an object is because the object starts existing only from the moment `open()` is called. Object returning itself for no reason (and here, it is literally **no** reason for the users using Notebooks) is super confusing.

The more I think about it, the less I like the idea that it's possible to do `with lmp as lmp_int` with no effect. In my opinion this must raise an error.

jan-janssen:
> The more I think about it, the less I like the idea that it's possible to do `with lmp as lmp_int` with no effect. In my opinion this must raise an error.

That is what I had previously and then you said: 
> Since interactive_open is now required I find this error message redundant.

For me either way is fine.

samwaseda:
> > The more I think about it, the less I like the idea that it's possible to do `with lmp as lmp_int` with no effect. In my opinion this must raise an error.
> 
> That is what I had previously and then you said:
> 
> > Since interactive_open is now required I find this error message redundant.
> 
> For me either way is fine.

No this doesn't solve the problem, because `with lmp as lmp_int` will work anyway.

jan-janssen:
> No this doesn't solve the problem, because `with lmp as lmp_int` will work anyway.

We can use: 
```
    def __enter__(self):
         if self.server.run_mode.interactive:
             return self
         else: 
             super(InteractiveBase, self).__enter__()
```
As that is exactly what interactive open is triggering.

samwaseda:
I can live with this, but I still prefer not to leave the user with two jobs with `interactive_open` (the job itself and the returned job) with no obvious reason.

What's the problem with my job holder idea anyway?

jan-janssen:
> What's the problem with my job holder idea anyway?

The `job holder` does not work like default python: 
```
job_int = job.interactive_open()
job_int.run()
job_int.interactive_close()
```

samwaseda:
> > What's the problem with my job holder idea anyway?
> 
> The `job holder` does not work like default python:
> 
> ```
> job_int = job.interactive_open()
> job_int.run()
> job_int.interactive_close()
> ```

And this is the code that I don't want to offer. I don't think default python has such a behaviour anyway.

```python
job.structure = structure_one
job_int = job.interactive_open()
job_int.structure = structure_two
```

And all of a sudden both `job` and `job_int` have the same structure. I really don't want this.

liamhuber (changes requested):
Implementation and docs look good! Quick unit tests to ensure the interactive guy can be opened and the regular guy throws an error and I'd say it's done.

jan-janssen:
> ```python
> job.structure = structure_one
> job_int = job.interactive_open()
> job_int.structure = structure_two
> ```

But you could write: 
```
job.structure = structure_one
job = job.interactive_open()
job.structure = structure_two
```

Or alternatively: 
```
job.structure = structure_one
job_int = job
job_int.structure = structure_two
```

I still believe it is the best solution we have so far.

liamhuber:
> And all of a sudden both job and job_int have the same structure. I really don't want this.

> Object returning itself for no reason (and here, it is literally no reason for the users using Notebooks) is super confusing.

> No this doesn't solve the problem, because `with lmp as lmp_int` will work anyway.

After my review I saw all the discussion. I think this is largely my fault as I'm the one who pushed for the syntax `with lmp.interactive_open() as lmp_int:`. 

The reason for is is that the syntax `with lmp as lmp_int` was extremely jarring to me. I don't mind that it still works, I just don't think it should be canonical. 

> I still believe it is the best solution we have so far.

If I'm reading correctly and "it" is the current implementation, I'm inclined to agree -- I'm open to the idea of a holder class as long as we're super careful about implementation, as Jan's current concerns seem pretty valid to me.

samwaseda:
> I'm open to the idea of a holder class as long as we're super careful about implementation, as Jan's current concerns seem pretty valid to me.

With Jan's concerns you mean the fact that `job.interactive_open` does not return anything? On the contrary, I find it only confusing if it returns itself because the returned job has zero utility for the Notebook users.

> After my review I saw all the discussion. I think this is largely my fault as I'm the one who pushed for the syntax `with lmp.interactive_open() as lmp_int` :

By the way here I anyway agree that it should be `with lmp.interactive_open()` and not `with lmp`.

samwaseda:
> But you could write:
> ```python
> job.structure = structure_one
> job = job.interactive_open()
> job.structure = structure_two
> ```

I know I can do that. I'm talking about users who start wondering why there's suddenly another job, because there's no obvious reason that you should have two identical jobs at this stage.

All I want is to make sure that people don't start playing with two objects pointing to the same thing without realising what they are doing. I don't think it's harmful to have a class whose sole purpose is to define `__enter__` and `__exit__` and forbids anything else.

jan-janssen:
> Implementation and docs look good! Quick unit tests to ensure the interactive guy can be opened and the regular guy throws an error and I'd say it's done.

I added the unit tests

jan-janssen:
> I know I can do that. I'm talking about users who start wondering why there's suddenly another job, because there's no obvious reason that you should have two identical jobs at this stage.

They can continue running it as before:
```
job.structure = structure_one
job.interactive_open()
job.structure = structure_two
```
There is no need to use the return variable of interactive open. 

> All I want is to make sure that people don't start playing with two objects pointing to the same thing without realising what they are doing. I don't think it's harmful to have a class whose sole purpose is to define `__enter__` and `__exit__` and forbids anything else.

As I pointed out above the disadvantage of this is that it is a different object which to me is more harm than good.

tests/job/test_job_interactive.py
```diff
@@ -0,0 +1,22 @@
...
+        with self.assertRaises(NotImplementedError):
+            with job as _:
+                pass
+                
+    def test_job_interactive_with(self):
+        job = self.project.create_job(InteractiveBase, "job_interactive")
+        job.project.db.add_item_dict(job.db_entry())
+        with job.interactive_open() as job_int:
+            job_int.to_hdf()
+        self.assertTrue(job.server.run_mode.interactive)
```
liamhuber, line 22:
```suggestion
        self.assertFalse(job.server.run_mode.interactive, msg="Job failed to interactively close on exiting with")
```

Right?

liamhuber (commented):
I'm happy except for the fact that I don't understand why the last test passes...

liamhuber:
>> I'm open to the idea of a holder class as long as we're super careful about implementation, as Jan's current concerns seem pretty valid to me.

> With Jan's concerns you mean the fact that job.interactive_open does not return anything? 

No, with his concern that we can't `holder = job.interactive_open(); holder.run()` because `Holder` isn't a job. This is what I meant about being super careful with implementation -- I think we probably can manage to spoof all the behaviour we want without having `Holder(GenericJob)`, but the basics weren't in your toy demo (I mean, fair, it was a toy!) and more generally I worry it will be hard to be sure we've captured everything. Having `interactive_open()` return `self` is, as Jan put it, "the best solution we have so far".

I thought about it a bit more over breakfast, and I'm really not concerned with `interactive_open()` returning `self`. Jan's example a couple comments above rightly points out that anyone determined can get a second tag on the same object anyhow with `job2 = job`. The reason I'm 100% ok with the `return self` behaviour is that in the canonical case of `with job.interactive_open() as job_int:`, `job_int` *clearly* is only intended to exist in the *context* of the `with` clause. It's true that it persists as a variable afterwards, but no pythonista in their right mind goes around using with-clause variables later. At the same time, in the "old-school" way of `job.interactive_open(); ## stuff; job.interactive_close()`, no one is capturing the return value anyhow. 

So overall, I agree that this implementation can be abused to get some weird behaviour, *but* both the "sensible" ways to use it behave exactly as you'd expect.

samwaseda:
> No, with his concern that we can't `holder = job.interactive_open(); holder.run()` because `Holder` isn't a job. This is what I meant about being super careful with implementation -- I think we probably can manage to spoof all the behaviour we want without having `Holder(GenericJob)`, but the basics weren't in your toy demo (I mean, fair, it was a toy!) and more generally I worry it will be hard to be sure we've captured everything. Having `interactive_open()` return `self` is, as Jan put it, "the best solution we have so far".

I'm extremely surprised that you both want to allow `job_int = job.interactive_open(); job_int.run()`, because that's exactly the kind of behaviour that I want to forbid. In particular, `job_int` must not be able to do anything other than being used with `with`. In my opinion it should be like this:

Forbidden:
```python
job_int = job.interactive_open()
job_int.run()
# and anything else GenericJob would otherwise allow
```

Allowed:
```python
with job.interactive_open() as job_int:
    job_int.run()
    # and anything else GenericJob allows
```

> Jan's example a couple comments above rightly points out that anyone determined can get a second tag on the same object anyhow with `job2 = job`.

In this case it's clear what the line means, while it is impossible to understand what kind of job `job.interactive_open()` is returning and *why* it's returning one in the first place.

> The reason I'm 100% ok with the `return self` behaviour is that in the canonical case of `with job.interactive_open() as job_int:`, `job_int` _clearly_ is only intended to exist in the _context_ of the `with` clause. It's true that it persists as a variable afterwards, but no pythonista in their right mind goes around using with-clause variables later.

Here I agree but I don't think our opinions didn't differ here anyway.

samwaseda:
Even better would be to make every attribute of the holder class a private function and implement `__getattr__`, which explains that the object returned via `interactive_open` can be used only with `with`.

pmrv (approved):
I'd rather not define `__enter__`/`__exit__` on `GenericJob`, since it really is interactive-only.

I'm also leaning towards @samwaseda suggestion of returning a separate object from `interactive_open`, because I think it gives a nice opportunity to move all the interactive functionality off to a different object, thus making the normal job interface leaner.  But I'm late to this discussion, so I'm ok merging as is and working on this later.

pmrv:
Btw. until/when we add more functionality, I think what we want is really covered by something like the examples in [PEP0343](https://www.python.org/dev/peps/pep-0343/#examples), e.g. `closing`.

liamhuber:
> Forbidden...
> Allowed...


I don't think this is possible? The only way I see to get the allowed case is `job_int is job`. I am not aware how to implement that without simply `return self`.

A way out would be

```python
with job.interactive_open() as int_holder:
    int_holder.job.run()
    # and anything else GenericJob allows
```

But I don't see `return JobHolder(self)` as so much better than `return self`.

> ...can be used only with with.

Ultimately I'd be perfectly happy with that. I'm not an enter/exit expert, but the problem I see is there's no way to filter by the context in this way. I'd be very happy to be wrong, I hope the issue is merely my ignorance.

liamhuber:
> I'd rather not define `__enter__/__exit__` on `GenericJob`, since it really is interactive-only.

This was my gut response as well, but I guess the current implementation *only* serves to return a more informative error message -- and I'm OK with that.

liamhuber:
> Btw. until/when we add more functionality, I think what we want is really covered by something like the examples in PEP0343, e.g. closing.

Just quickly peeked, but there is literally a `@contextmanager` decorator, so I'm cautiously optimistic that, as hinted two posts above, my issue stems from my ignorance!!!

samwaseda:
> > Forbidden...
> > Allowed...
> 
> I don't think this is possible? The only way I see to get the allowed case is `job_int is job`. I am not aware how to implement that without simply `return self`.

It is possible, and that's exactly why I suggested `JobHolder`.

samwaseda:
Here's **the** perfect class for me:

```python
class _WithInteractiveOpen:
    def __init__(self, job):
        self._job = job

    def __repr__(self):
        return 'Interactive ready'

    def __enter__(self):
        return self._job
    
    def __exit__(self, exc_type, exc_val, exc_tb):
        return self._job.interactive_close()
    
    def __getattr__(self, attr):
        error_message = (
            'Syntax:\n'
            + '`your_job.interactive_open()`\n'
            + '`your_job.run()`\n'
            + 'Alternatively you can use the `with`-statement:\n'
            + '`with your_job.interactive_open() as job_int:`\n'
            + '`    job_int.run()`\n'
        )
        raise ValueError(error_message)

class InteractiveBase:
    def interactive_open(self):
        self.server.run_mode.interactive = True
        return _WithInteractiveOpen(self)
```

Here's what happens:

Case I:

```python
job.interactive_open()
job.run() # -> ok
job.some_other_method_defined_in_class() # -> ok
job.interactive_close()
```

Case II:

```python
with job.interactive_open() as job_int:
    job_int.run() # -> ok
    job_int.some_other_method_defined_in_class() # -> ok
```

Case III:

```python
job_int = job.interactive_open()
job_int.run() # -> error + error message explaining what should be done
```

tests/job/test_job_interactive.py
samwaseda: `run_mode` doesn't change so what @jan-janssen wrote is correct.

liamhuber:
> It is possible, and that's exactly why I suggested `JobHolder`.

Perfect. I misunderstood the behaviour of `__enter__`. Your extended example looks good to me.

samwaseda:
@jan-janssen if you want (and especially if you agree with my suggestion), I can take over this PR and implement what I suggested. What do you think?

jan-janssen:
> @jan-janssen if you want (and especially if you agree with my suggestion), I can take over this PR and implement what I suggested. What do you think?

I added your suggestion.

samwaseda (approved):
Brilliant!!

jan-janssen:
@liamhuber Can you take another look?

tests/job/test_job_interactive.py
liamhuber: Wait, interactive close leaves the run mode as interactive? That seems bad

liamhuber:
> @liamhuber Can you take another look? 

I was pretty much ok with it before so as long as San is happy I'm good to go.

Only outstanding thing for me is the one test assertion. Not really a core part of this pr, but if interactive close leaves the run mode as interactive that seems weird to me. But we can definitely deal with that separately unless there is immediate consensus and an easy solution

tests/job/test_job_interactive.py
samwaseda: Well, as long as `job.status` is set to `finished` I think it's fine.

jan-janssen: The `run_mode` queue also remains when the job is finished, just like `non_modal`. Basically the run mode is only internal for pyiron to understand how to execute the job. The option that changes is `job.interactive_is_activated()` but this should never be touched by the user. As we only test the `wrapper` without any explicit interface to an interactive job we can not test this option.

niklassiemer:
Since this PR, the windows compatibility tests with atomistics are failing.

PR #567, opened 2021-11-29
Projectpath tests

liamhuber:
Actually the problem that prompted me to start messing with this file was just an issue with local config file, so this wound up being purely cosmetic polish.


niklassiemer (approved):
lgtm!

PR #568, opened 2021-11-29
Create job name including rounding and special character replacement

samwaseda:
Following today's discussion, I implemented this function `pr.create.job_name` that does the job name replacement as in `pr.create.job.MyJob(job_name=blablabla)` including all the options, i.e. number of digits and job replacement dictionary

### DIFF
--- a/pyiron_base/job/util.py
+++ b/pyiron_base/job/util.py
@@ -122,9 +122,10 @@ _special_symbol_replacements = {
 }
 
 
-def _get_safe_job_name(name, ndigits=8, special_symbols={}):
+def _get_safe_job_name(name, ndigits=8, special_symbols=None):
     d_special_symbols = _special_symbol_replacements.copy()
-    d_special_symbols.update(special_symbols)
+    if special_symbols is not None:
+        d_special_symbols.update(special_symbols)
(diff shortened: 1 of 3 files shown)


pyiron_base/job/util.py
```diff
@@ -122,7 +122,10 @@ def _get_project_for_copy(job, project, new_job_name):
 }
 
 
-def _get_safe_job_name(name, ndigits=8):
+def _get_safe_job_name(name, ndigits=8, special_symbols=None):
```
liamhuber:
```suggestion
def _get_safe_job_name(name: str, ndigits: Union[int, None] = 8, special_symbols: Union[Dict, None] = None):
```

You will probably also need ```from typing import Union, Dict```

liamhuber (approved):
lgtm! But of course I can alway find a nit somewhere 😉 I would like to start leveraging type hinting more and more strongly, and gave an example of how to do so here. If you hate it then just leave it out, I don't think we've reached a consensus yet on whether we should fully adopt this -- I find it super useful when working in my IDE though, so my vote is a solid "aye"

samwaseda:
What's actually the advantage of type hinting except for the IDE? It's not enough to have it in the DocString?

PR #571, opened 2021-12-03
Add copy, split and merge methods to FlattenedStorage

pmrv:
I've added some methods to split and merge FlattenedStorages that share the same chunk structure.  For now I'm deep-copying all the arrays involved for simplicity.  If we find later that this is a performance bottleneck we can address that separately.

### DIFF
--- a/pyiron_base/generic/flattenedstorage.py
+++ b/pyiron_base/generic/flattenedstorage.py
@@ -550,29 +550,29 @@ class FlattenedStorage(HasHDF):
                 split._per_chunk_arrays[k] = np.copy(split._per_chunk_arrays[k])
         return split
 
-    def merge(self, store: "FlattenedStorage") -> "FlattenedStorage":
+    def join(self, store: "FlattenedStorage") -> "FlattenedStorage":
         """
         Merge given storage into this one.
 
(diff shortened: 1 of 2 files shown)


Leimeroth:
I like this, but I am not really sure if it will help me with my data structure. I use different containers for each fit property, but within each container the arrays have the same name, f.e. "target_val", "weight", "fit", ... So if I understand it correctly I can't merge them.

I don't know if it is worth the effort to generalize over my specific use case, since I expect machine learning potential fitting codes to have more similarities with each other than with the classic potential fitting atomicrex does. Maybe an option to prepend a name in front of every array name would help with this. F.e. merging would give array names like "atomic-energy_target_val", "atomic-forces_target_val", etc. But as already said I am not sure if trying to generalize over this is worth it, when every other code only needs a fraction of the arrays and therefore explicit names are the goto solution anyway.

pmrv:
I think automatic renaming is nice to have.  I had actually already planned to model this after [pandas.merge](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.merge.html?highlight=merge#pandas.DataFrame.merge), but then forgotten about it again. :/  Patch might only come on Monday though.

PR #575, opened 2021-12-09
Extension of the Worker Class 

jan-janssen:


pyiron_base/job/worker.py
```diff
@@ -207,3 +207,36 @@ def run_static(self):
...
+        master_id = self.job_id
+        while not finished:
+            df = pr.job_table()
+            df_sub = df[
+                ((df["status"] == "submitted") | (df.status == "running")) &
+                (df["masterid"] == master_id)
+            ]
+            if len(df_sub) == 0:
+                j += 1
+                if j > 10:
```
liamhuber: If I read right this is a safety feature so you need consecutive blank calls before really considering yourself done. In this case 10 should be a kwarg, and it should probably be in units of seconds and related to the other new kwarg

pyiron_base/job/worker.py
```diff
@@ -207,3 +207,36 @@ def run_static(self):
...
+                    finished = True
+            else:
+                j = 0
+            with open(log_file, 'a') as f:
+                f.write(
+                    str(datetime.today()) + " j: " + str(j) +
+                    "   submitted: " + str(len(df[df.status == "submitted"])) +
```
liamhuber: Nit: this might read better in loop form over the statuses

pyiron_base/job/worker.py
```diff
@@ -207,3 +207,36 @@ def run_static(self):
...
+                    "   running: " + str(len(df[df.status == "running"])) +
+                    "   finished: " + str(len(df[df.status == "finished"])) +
+                    "   aborted: " + str(len(df[df.status == "aborted"])) + "\n"
+                )
+            time.sleep(60)
```
liamhuber: 60 should be a kwarg. Default of 60 is of course fine.

liamhuber (commented):
Would be nice to offer some flexibility in how patient the user wants to be

pmrv:
Could this be implemented in terms of `Project.wait_for_jobs`? If not then the arguments @liamhuber suggested should match that function.  I'm fine with having this extra function for convenience, but we should avoid duplicated logic or interfaces.

jan-janssen:
> Could this be implemented in terms of `Project.wait_for_jobs`? If not then the arguments @liamhuber suggested should match that function. I'm fine with having this extra function for convenience, but we should avoid duplicated logic or interfaces.

The difference is basically that it searches for jobs based on the `master_id` rather than the path, which allows you to have multiple workers inside the same project and it sets the status of the worker to `collect` once all jobs are finished to stop the worker. So while I foresee in the long run the `worker` jobs just like `pyirontable` being meta jobs the user never actively works with, at the current stage I prefer to have a separate function.

liamhuber:
I see the merit of having such `wait` code directly on the worker master, but Marvin's point about not duplicating things is also a very good one. Can we have our cake and eat it too? 

What if we extract the waiting code, such that the source, times, etc. are all taken as arguments, then exploit this extracted behaviour at both the worker and project levels. The additional behaviour that the worker sets itself to collect should be easy enough to append after.

I'm sitting at 30k ft on this, so forgive me if I miss some impossible architecture obstacle. Seems like it should work though..

jan-janssen:
> I'm sitting at 30k ft on this, so forgive me if I miss some impossible architecture obstacle. Seems like it should work though..

While the functions have a similar effect for the user, their technical implementation is different e.g. filter by master id, write log file and only finish after time out, for the case the tasks for the worker are submitted from a different notebook. So I currently do not see an option to combine them. Internally they both use the `job_table()` to get the data from the database but the filtering and the reaction are different.

PR #576, opened 2021-12-09
Fix top path for pyiron installation without database

jan-janssen:


jan-janssen:
```
291
ERROR: test_file_top_path (database.test_manager.TestDatabaseManager) [disabled]
292
293
294
  File "/home/runner/work/pyiron_base/pyiron_base/tests/database/test_manager.py", line 57, in test_file_top_path
295
    self.assertTrue(self.dbm.top_path(self.project_path + "/test") in self.project_path)
296
TypeError: 'in <string>' requires string as left operand, not NoneType
297

298
299
FAIL: test_file_top_path (database.test_manager.TestDatabaseManager) [disabled: test Project.root_path and Project.project_path for a new sub-Project]
300
301
302
  File "/home/runner/work/pyiron_base/pyiron_base/tests/database/test_manager.py", line 61, in test_file_top_path
303
    self.assertEqual(sub_pr.root_path, new_root_path + '/')
304
AssertionError: None != '/home/runner/work/pyiron_base/pyiron_base/'
305

306
307
FAIL: test_project_path (project.test_projectPath.TestProjectPath)
308
309
310
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_projectPath.py", line 61, in test_project_path
311
    self.assertIn(
312
AssertionError: '/home/runner/work/pyiron_base/pyiron_base/tests/project/test_project_path/' not found in ['/home/runner/work/pyiron_base/pyiron_base//home/runner/work/pyiron_base/pyiron_base/tests/project/test_project_path/'] : project.project_path not properly set by default. Check if `project_check_enabled`.
313

314
315
FAIL: test_root_path (project.test_projectPath.TestProjectPath)
316
317
318
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_projectPath.py", line 54, in test_root_path
319
    self.assertIn(
320
AssertionError: None not found in ['/home/runner/work/pyiron_base/pyiron_base/'] : root project.root_path not properly set by default. Check if `project_check_enabled`.
321

322
```

niklassiemer:
We intentionally changed the behavior of the root_path / project_path split to not depend on project_check_enabled such that we are able to default to false for the project check. However, the linked issue is indeed a problem! Since you do not define any project_paths in the config that should be an empty list...

liamhuber:
> We intentionally changed the behavior of the root_path / project_path split to not depend on project_check_enabled such that we are able to default to false for the project check. However, the linked issue is indeed a problem! 

Ah, I thought this was more sensible behaviour when the check was disabled. Darnit. This is just more fuel on the fire of "let's get rid of project checking entirely" IMO

> Since you do not define any project_paths in the config that should be an empty list...

Aha aha aha! My comment over on the issue is also a little off then -- I thought the user path was in `Settings.default_configuration`, but it's not! Indeed, then I am also confused why it's not an empty list. (The database default is still acting sensibly though.)

niklassiemer:
IMO, it is after all more sensible to consistently split the full path into root_path and project_path irrespective of the value of project_check_enabled! 
Moreover, since all our look-up on the database depends on the split into root_path and project_path and this difference might even become useful if the root_path for some projects is actually pointing to an external storage I am in favor of keeping this function the way it is. (And otherwise we have to change the default of project_check_enabled back to True to not to break the job_table)
@jan-janssen how/where does the root_path/project_path split kill the installation without database? Due to the not empty list of root_paths?

jan-janssen:
> @jan-janssen how/where does the root_path/project_path split kill the installation without database? Due to the not empty list of root_paths?

The code without database expects `None` as return value, but this is never reached because the config is not empty as currently the default `project_path` is not cleared.

pmrv:
See our [minutes](https://github.com/orgs/pyiron/teams/pyiron/discussions/97/comments/6).

PR #579, opened 2021-12-09
Filetable: job status return None if the file does not exist

jan-janssen:


pmrv:
If it doesn't break anything immediately I would greatly prefer this to throw an error instead.

jan-janssen:
> If it doesn't break anything immediately I would greatly prefer this to throw an error instead.

Without this the deletion of jobs derived from generic master fails for pyiron running without database, because the child jobs are already deleted when the master is deleted but the file table is not updated immediately. So by returning none it is possible to refresh the file table and once the specific child is about to be deleted then the file table realizes that the job is already deleted and passes along.

liamhuber:
Can we get that case codified as a test? In the long run it sounds like the issue should be resolved upstream in generic master and/or the file table, but if a test comes with this patch proving what it does then we can make sure a future, more involved solution behaves as desired

jan-janssen:
@liamhuber  I added the test

PR #584, opened 2021-12-13
Fix compat tests 

### LINKED ISSUE #506: Compatibility workflow do not check dependency updates.
I was wondering why our unittests on https://github.com/pyiron/pyiron_atomistics/pull/410 fail while the ones for #500 passed also for the compatibility workflow. The reason is
https://github.com/pyiron/pyiron_base/blob/e22dbb86bd4f0d1f1b3a360cb5fd3f9707f64b57/.github/workflows/atomistics-compat.yml#L43-L44

where we update the environment with all packages as specified in atomistics except pyiron_base. To check dependency updates, we only should update the environment with packages not yet installed.

jan-janssen:


.github/workflows/atomistics-compat.yml
```diff
@@ -23,24 +23,34 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: pyiron_atomistic
+      run: |
+        git clone https://github.com/pyiron/pyiron_atomistics ../pyiron_atomistics
+        grep -v "pyiron_base" ../pyiron_atomistics/.ci_support/environment.yml > environment.yml
+        tail -n +4 .ci_support/environment.yml >> environment.yml
```
niklassiemer: What happens if we update a dependency in base but not (yet) in atomistics? would this work or throw an error due to twice the same dependency with different versions?

jan-janssen: To my understanding the last entry for the same package overwrites all previous ones

jan-janssen: It does not work ... here is an example [Binder](https://mybinder.org/v2/gh/jan-janssen/multi-def-conda/HEAD)

jan-janssen: I wrote a small python script to merge the environments https://github.com/pyiron/pyiron_base/blob/compatfix/.ci_support/condamerge.py

pyiron_base/state/settings.py
```diff
@@ -284,12 +284,12 @@ def _validate_viewer_configuration(config: Dict) -> None:
     @staticmethod
     def _validate_no_database_configuration(config: Dict) -> None:
         if "disable_database" in config.keys() and config["disable_database"]:
-            if config["project_check_enabled"]:
+            if "project_check_enabled" in config.keys() and config["project_check_enabled"]:
```
niklassiemer, line 287: I am a bit surprised to see this in this PR, but the changes are ok to me. Did they pop up here or just sneak in?

niklassiemer (approved):
LGTM! We could only think about a common script/class to handle environments - we also have the script to update an environment.yml by providing package, old, and new version.

pyiron_base/state/settings.py
jan-janssen: They sneaked in

PR #586, opened 2021-12-13
Fix: DeprecationWarning: pyiron_base.project.generic.view_mode is deprecated: use db.view_mode.

jan-janssen:


jan-janssen:
When we add a `DepreciationWaring` then we should also fix the code at least in pyiron base https://github.com/pyiron/pyiron_base/pull/336

pmrv (approved):

PR #593, opened 2021-12-14
Use notebook version of tqdm if possible

pmrv:
I just found out that tqdm has a really cute widget for the progress bars if running in notebooks, so I've added it.

EDIT: See [here](https://pypi.org/project/tqdm/#ipython-jupyter-integration).


jan-janssen (approved):
Looks good to me

PR #594, opened 2021-12-14
Remove get_pandas

### LINKED ISSUE #592: Is JobCore.get_pandas really needed?
This function 

https://github.com/pyiron/pyiron_base/blob/7003bd1c03d4e6632d56a336fd91e0951687ea2d/pyiron_base/job/core.py#L384

just delegates to `job.project_hdf5.get_pandas` and does nothing else. I've also never seen it used anywhere.

Since it's on `JobCore` almost every object in pyiron has that method, so it takes up a good amount of tab-complete real estate and is sometimes even actively confusing, e.g. on pyiron tables.  There you need to call `get_dataframe` instead, but `get_pandas` also pops up, so users need to double check with the docstring to get the correct one.

pmrv:
Fixes #592.

Just to see if any compatibility tests break.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -381,18 +381,6 @@ class JobCore(HasGroups):
         """
         return self.project_hdf5.get_from_table(path, name)
 
-    def get_pandas(self, name):
-        """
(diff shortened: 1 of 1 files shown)


jan-janssen (approved):
Looks good to me

PR #598, opened 2021-12-15
Use Black formating

jan-janssen:


jan-janssen:
https://black.readthedocs.io/en/stable/

PR #606, opened 2021-12-21
Refactor maintenance into own module

niklassiemer:
First part of #525 - just the refactoring into new module. The only thing changed is where I import `get_database_statistics` from. Now, I import it from the full path instead directly `from pyiron_base` to avoid circular imports.


max-hassani (approved):
Looks good to me!

PR #607, opened 2021-12-21
Add update functionality to maintenance

niklassiemer:
pending #606

### DIFF
--- a/pyiron_base/project/maintenance.py
+++ b/pyiron_base/project/maintenance.py
@@ -6,8 +6,7 @@ from _warnings import warn
 import pandas
 from git import Repo, InvalidGitRepositoryError
 
-from pyiron_base import state
-from pyiron_base.database.performance import get_database_statistics
+from pyiron_base import state, get_database_statistics
 
 
 class Maintenance:
@@ -16,17 +15,22 @@ class Maintenance:
     some measures of perfomance for pyiron, whether local to the project
     or global (describing the status of pyiron on the running machine)
     """
-
-    def __init__(self):
+    def __init__(self, project):
         """
         initialize the local and global attributes
         """
+        self._project = project
         self._global = GlobalMaintenance()
+        self._update = UpdateMaintenance(self._project)
         self._local = None
 
     @property
     def global_status(self):
         return self._global
+
+    @property
+    def update(self):
+        return self._update
 
     @staticmethod
     def get_repository_status():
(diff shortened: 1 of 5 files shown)


pyiron_base/project/maintenance.py
```diff
@@ -60,6 +67,42 @@ def get_repository_status():
...
+        such arrays, but does not automatically save them.  This conversion script
+        simply goes over all jobs and rewrites their HDF5 files, since it's read with
+        the correct dtype, this then writes this correct dtype.
+
+        Args:
+            project(None/pyiron_project/"all"): The project to be converted from 0.3 to 0.4 ; default: current project
+                if "all" is provided, pyiron tries to find all projects using the PROJECT_PATHS defined in the
+                configuration.
+        """
+        if project is None:
```
max-hassani, line 110: I like the option that the user can pass a list of projects or even choose all to be updated, but a clearer docstring, which explains different options, is missing.

pyiron_base/project/maintenance.py
```diff
@@ -60,6 +67,42 @@ def get_repository_status():
         return report
 
 
+class UpdateMaintenance:
+    def __init__(self, project):
+        self._project = project
+
+    def base_v0_3_to_v0_4(self, project=None):
```
max-hassani, line 94: I suggest we have a separate function called `base_to_latest()`, which upgrade the project to be compatible with the latest version. At the moment, such a function would refer to this function.

max-hassani (approved):
I liked very much the refactoring you made to the maintenance object and addition of the update function to it. I had two comments, which might be helpful to improve the usability.

pyiron_base/project/maintenance.py
niklassiemer: Fair idea, right now this would work fine. However, we would need something like
```python
def base_to_latest(self, start_version, project=None):
  mayor, minor, patch = start_version.split('.')
  if mayor != 0:
    raise ValueError

  if minor < 4:
    base_v0_3_to_v0_4(project)
  # Hope we will not need this, though :)
  if minor < 7:
    base_v0_6_to_v07(project)
```

niklassiemer: and actually if we call it 'latest' we should also check if there is a more recent `pyiron_base` release available.

pmrv: > and actually if we call it 'latest' we should also check if there is a more recent `pyiron_base` release available.

If that bloats the logic to much, just name to `*_to_current` or something similar.  People already have conda/pip to check for new versions, no need for us to re-implement that imho.

pyiron_base/project/maintenance.py
max-hassani: Have you changed the docstring, or I missed the part above the Args. If it is the later, sorry for this nonsense comment.

niklassiemer: Well, the part above the Args was there already. After all I improved the docstring to list all possibilities to define a project :)

niklassiemer:
Todo: update 
https://github.com/pyiron/pyiron_base/blob/0a602b80437dba1e07f87de9e5fded5b035f301e/pyiron_base/generic/hdfio.py#L221-L222

pyiron_base/project/maintenance.py
```diff
@@ -17,17 +18,23 @@ class Maintenance:
     or global (describing the status of pyiron on the running machine)
     """
 
-    def __init__(self):
+    def __init__(self, project):
         """
         initialize the local and global attributes
         """
```
pmrv:
```suggestion
        """
        Args:
            (project): pyiron project to do maintenance on
        """
```

pyiron_base/project/maintenance.py
```diff
@@ -17,17 +18,23 @@ class Maintenance:
...
         self._global = GlobalMaintenance()
+        self._update = UpdateMaintenance(self._project)
         self._local = None
 
     @property
     def global_status(self):
         return self._global
 
+    @property
+    def update(self):
```
pmrv, line 37: Should document at least the type, so users can follow in the online docs.  (How do type annotations for properties work? :S)

pyiron_base/project/maintenance.py
```diff
@@ -60,6 +67,75 @@ def get_repository_status():
...
+                raise ValueError(
+                    f"{project} is a str but neither 'all' nor a directory."
+                )
+        else:
+            projects = [project]
+
+        if len(projects) == 0:
+            raise ValueError(
+                f"Provided project {project} lead to 0 projects to be converted."
+            )
```
pmrv: This can only when providing `project='all'`, right? In that case a warning might be enough.

pyiron_base/project/update/pyiron_base_03x_to_04x.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    pyiron_base<=0.3.10 has a bug that writes all arrays with dtype=object even
+    numeric ones.  As a fix pyiron_base=0.4.1 introduces a conversion when reading
+    such arrays, but does not automatically save them.  This conversion script
+    simply goes over all jobs and rewrites their HDF5 files, since it's read with
+    the correct dtype, this then writes this correct dtype.
+    """
+    total_size = 0
+    for l in subprocess.getoutput(
+        f"find {project.path} -regex \".*\.h5\" -exec wc -c '{{}}' \;"
```
pmrv: This was my quick hack to get a progress bar working, but it feels a bit hacky now.  I suppose it would work on all linux distros, but might break on windows/mac.

OTOH, this really only counts the number of bytes in all HDF5 files, so maybe you can whip something up with `pyfileindex`/`os.walk` and `os.stat`.

pmrv (changes requested):
Looks mostly good, the biggest point is about replacing the `find` logic to give an accurate progress bar.  If that's too difficult, just drop it and enable the normal project-based progress in L66.

pyiron_base/project/maintenance.py
niklassiemer: Yes, you are right!

pyiron_base/project/update/pyiron_base_03x_to_04x.py
niklassiemer: I will think about it. Getting rid of the `subprocess` call would be an added benefit.

pmrv (approved):

niklassiemer:
@pmrv since you did run some tests on your data previously - could you have a quick check that the this after all does what it should?

niklassiemer:
For me it seems to work now - with the script and also via the maintenance call. The only annoying thing is the warning which tells you to do what you already did... 
@pyiron/release I will merge this after the ADIS meeting and release base-0.5.0.

PR #614, opened 2022-01-05
Use psf/black action directly

niklassiemer:


jan-janssen:
I like the suggestion, can you also update https://github.com/pyiron/pyfileindex/blob/master/.github/workflows/black.yml , https://github.com/pyiron/pylammpsmpi/blob/master/.github/workflows/black.yml and https://github.com/pyiron/pysqa/blob/master/.github/workflows/black.yml

PR #617, opened 2022-01-10
job name conversion in `pr.load` same as `pr.create`

### LINKED ISSUE #616: Why is the conversion of `.` -> `_` done only when the job is loaded?
https://github.com/pyiron/pyiron_base/blob/adb7b978ea1636b76701e5c464d070aa27847ae4/pyiron_base/database/jobtable.py#L138

I don't understand why this conversion is done only when the job is loaded. Is it meant to be applied on the job name that the user gives, or something that pyiron appends internally?

samwaseda:
`pr.load(('job', variable_1, vaiable_2))` works with this PR.

Closes #616 (although the point raised there must be clarified)


jan-janssen (approved):
Looks good to me

PR #618, opened 2022-01-10
Introduce black formatting workflow triggered by `format_black` label

niklassiemer:


niklassiemer:
The Format black workflow fails if there is nothing to do (And I cannot change this). However, one probably adds the label if there is a failing Black test :) Testing if does the correct thing (and not fail) in #619

niklassiemer:
Seems to work fine. One thing: The commit with the changes due to black formatting does not retrigger CI (if we stick to the GitHub Token) - I would be fine with this since black does not change behavior. However, one has to check for other than 'black' failed tests on the previous commit! We could also use another token (e.g. Dependabot token) to push the commit - this should retrigger CI.

jan-janssen (approved):
Looks good to me

niklassiemer:
Let's see if it works: #619

niklassiemer:
Not yet...

niklassiemer:
It seems to need the same token on checkout?! Now it should work according to my tests :)

PR #626, opened 2022-01-12
Support secure sql passwords

jan-janssen:
For example: 
```
erk^ka%bCm%;4VF
```

### DIFF
--- a/pyiron_base/state/settings.py
+++ b/pyiron_base/state/settings.py
@@ -33,7 +33,7 @@ cleaning and consistency checks.
 """
 
 import os
-from configparser import RawConfigParser
+from configparser import ConfigParser
 from pyiron_base.state.logger import logger
 from pyiron_base.state.publications import publications
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


pyiron_base/state/settings.py
```diff
@@ -333,7 +333,7 @@ def _get_config_from_file(self) -> Union[Dict, None]:
             config_file = os.path.expanduser(os.path.join("~", ".pyiron"))
 
         if os.path.isfile(config_file):
-            parser = ConfigParser(inline_comment_prefixes=(";",))
+            parser = RawConfigParser(inline_comment_prefixes=(";",))
```
pmrv: Is the issue that arbitrary passwords can trigger the interpolation by accident or does `ConfigParser` generally refuse "weird" characters for option values?  If it's only the former I would be more comfortable with
 ```suggestion
            parser = ConfigParser(inline_comment_prefixes=(";",), interpolation=None)
```
since `RawConfigParser` is marked as legacy in the docs.

PR #629, opened 2022-01-15
Convert boolean config keys

jan-janssen:


liamhuber (commented):
A bit of code duplication. Could we just extract it to a private routine with the same name as the pr?

liamhuber (approved):

PR #631, opened 2022-01-16
Get filesize of Project

### LINKED ISSUE #630: Is there a way to see (disk space) size of project?
It sometimes depends on the size of a project (or its jobs) whether I launch new calculations while keeping the old ones or by erasing them, it would be super helpful to know it inside pyiron. Is there a way to do so?

jan-janssen:
Closes https://github.com/pyiron/pyiron_base/issues/630

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -660,13 +660,12 @@ class Project(ProjectPath, HasGroups):
             factor = 1024 * 1024 * 1024
         else:
             factor = 1
-        size = sum([
-            sum([
-                os.path.getsize(os.path.join(path, f))
-                for f in files
-            ])
-            for path, dirs, files in os.walk(self.path)
-        ])
+        size = sum(
+            [
+                sum([os.path.getsize(os.path.join(path, f)) for f in files])
+                for path, dirs, files in os.walk(self.path)
+            ]
+        )
         return size * factor
 
     def keys(self):
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -666,7 +666,7 @@ class Project(ProjectPath, HasGroups):
                 for path, dirs, files in os.walk(self.path)
             ]
         )
-        return size * factor
+        return size / factor
 
     def keys(self):
         """
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -643,6 +643,32 @@ class Project(ProjectPath, HasGroups):
         )
         return df["status"].value_counts()
 
+    def get_size_of_project(self, kilobytes=False, megabytes=False, gigabytes=False):
+        """
+        Gives the size of the project in bytes(B), kilobytes(KB), megabytes(MB) or gigabytes(GB).
+
+        Args:
+            kilobytes (bool): [True/False] - default=False
+            megabytes (bool): [True/False] - default=False
+            gigabytes (bool): [True/False] - default=False
+        """
+        if kilobytes:
+            factor = 1024
+        elif megabytes:
+            factor = 1024 * 1024
+        elif gigabytes:
+            factor = 1024 * 1024 * 1024
+        else:
+            factor = 1
+        size = sum([
+            sum([
+                os.path.getsize(os.path.join(path, f))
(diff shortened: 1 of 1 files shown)


jan-janssen:
I thought this kind of pull request is the perfect example for the GitHub Copilot https://copilot.github.com . So I recorded the following video to demonstrate the usage of Github Copilot in PyCharm.
test

samwaseda:
What do you think of `size` as a property and returning the unit automatically? That would be something like:

```python
@property
def size(self):
    size = sum(blablabla)
    units = ['B', 'KB', 'MB', 'GB', 'TB']
    counter = 0
    for ii in range(len(units)):
        if size > 1024:
            size /= 1024
            counter += 1
        else:
            break
return size, units[counter]
```

And if one day we want to have a possibility to compare different projects in size, in which case we want a function that returns the value with the same units, then we can also add `get_size` or so.

niklassiemer:
Yes, I then I would even go one step further and use `pint`
```python
@property
def size(self):
    size = sum(blablabla) * pint.UnitRegistry().byte
    return size.to_compact()
```
This is human readable and automatically allows to compare with different prefixes.

samwaseda:
> Yes, I then I would even go one step further and use `pint`
> 
> ```python
> @property
> def size(self):
>     size = sum(blablabla) * pint.UnitRegistry().byte
>     return size.to_compact()
> ```
> 
> This is human readable and automatically allows to compare with different prefixes.

I like the idea, but I just realised that for 1000 bytes, it shows 1 kilobyte... Is there a bug in pint?

samwaseda:
I opened an [issue on their page](https://github.com/hgrecco/pint/issues/1452).

jan-janssen:
@samwaseda and @niklassiemer Is it ok if we merge this pull request and update it to use `pint` later?

niklassiemer:
The remaining question would be - as `@property` returning simply bytes or as `get_`.. function. I would prefer the first solution and convert to `pint` afterwards.

niklassiemer:
Their last answer pointed out that they have `Ki`, `Mi`, etc. prefix being `2**10`, `2**20`, etc. Thus 
```python 
@property
def size(self):
    size = sum(blablabla) * pint.UnitRegistry().byte
    
    prefix_index= math.floor(math.log2(size) / 10) - 1
    prefix= ['Ki', 'Mi', 'Gi', 'Ti', 'Pi']

    if prefix_index< 0:
        return size
    elif prefix_index< 5:
        return size.to(f'{prefix[prefix_index]}byte')
    else:
        return size.to(f'{prefix[-1]}byte')
```
would return a useful and rather nice (kibibyte for kilo-binary-byte?) version to work with.

samwaseda:
I'm not really sure whether we should be using `pint` in the end, because a normal user familiar with `pint` would expect something like this to work as well: `pr.size.to('gigabyte')`, but from what I understand, it doesn't, correct? As long as the byte convention is not properly implemented, I'm afraid it makes more confusion.

niklassiemer:
Actually, there are two ways of using the prefixes in the case of byes, and in the general picture kilo is the prefix for 1000. Why actually should an unit prefix differ for different units? That's the underlying source of confusion and the reason you have to explain someone buying a 16GB stick why they only see 14.90GB in the OS. The prefix kibi is for the binary equivalent of kilo, thus, 1024. Therefore, `pr.size.to('gigabyte')` provides something in `10**9` and not `2**30` while `pr.size.to('gibibyte')` does. In this respect, I would rather claim `pint` is doing a good job. The only thing I would like is to use the binary units by default for 'bytes', i.e. make `bytes.to_compact()` use binary instead of base10 units.

samwaseda:
I would say, whatever the scientifically correct way to call it, as long as we conventionally call it megabyte, gigabyte etc. and mean a power of 1024, it's only confusing to have a different system. So I'd go for my solution of not offering pint conversion at all or give it in byte. In both cases I'd prefer a property solution to the current function, but I don't have a very strong feeling about it.

jan-janssen:
We already have a function named `get_project_size()`: 
https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/project/generic.py#L477

PR #635, opened 2022-01-17
Specify units for memory limit

pmrv:
Previously this is undocumented.  `pysqa` documentation labels the limit to be in GB, but I've found that this is not the case for our current setup and created another PR to fix this.


jan-janssen (approved):
Looks good to me

PR #470, opened 2021-10-12
Fix Database Timeout config

jan-janssen:
Convert from string to int

PR #636, opened 2022-01-17
Grab watchdog lock before checking connection status

pmrv:
Before we first check whether the connection is alive, then kick the
watchdog and only then grab the lock to make the SQL query.  There's a
race condition in this as sketched below, where the watchdog acquires
the lock and closes the connection after we check that it's still alive,
but before we acquire the lock.

```
Main Thread                     Watchdog Thread
-----------                     ---------------

                                TIMER RUNS OUT
                                        |
                                        v
CHECK CONNECTION        ~       ACQUIRE LOCK
    |                                   |
    |                                   v
    |                           CLOSE CONNECTION
    |                                   |
    |                                   v
    |                           RELEASE LOCK
    |
    v
ACQUIRE LOCK
    |
    v
EXECUTE SQL QUERY
    |
    v
RELEASE LOCK
```

The solution is to protect also the checking of the connection with the
same lock, so that closing of the connection happens only before or
after we checked the connection *and* executed our query.

@prince-mathews Please check whether your issue from the meeting persist with this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Unreleased
+- Fix race condition when talking to the database, this could result in cryptic error messages "connection closed" when
+  performing a lot of database queries in a row ([#636](https://github.com/pyiron/pyiron_base/pull/636))
 
 # pyiron_base-0.5.0
 - Create CHANGELOG.md ([#623](https://github.com/pyiron/pyiron_base/pull/623))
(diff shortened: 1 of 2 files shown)


jan-janssen:
That might also fix the issue I worked around in https://github.com/pyiron/pyiron_base/pull/470

prince-mathews:
I used the branch and I don't get the error anymore. Thank you :)

pmrv:
> I used the branch and I don't get the error anymore. Thank you :)

Great!  This is ready to merge then imo.

Issue #640, opened 2022-01-19
`decompress` doesn't work after job name change


samwaseda:
It's because [this line](https://github.com/pyiron/pyiron_base/blob/7ba3283c0926234829ba78a7eedd541b72056e3f/pyiron_base/job/util.py#L280) in `def decompress` requires the job name, even though the job name change doesn't change the tar file.

niklassiemer:
Would it be enough to add
```python
os.rename(
    os.path.join(job.working_directory, old_job_name + ".tar.bz2"), 
    os.path.join(job.working_directory, job.job_name + ".tar.bz2")
)
```
[here](https://github.com/pyiron/pyiron_base/blob/7ba3283c0926234829ba78a7eedd541b72056e3f/pyiron_base/job/util.py#L168)? Or do we use other functions to rename a job somewhere?

samwaseda:
No what the user would see is:

```python
job.job_name = new_job_name
job.decompress() # -> nothing happens
```

So in the first line when the job name is changed the tar file must get a new name.

niklassiemer:
Yes, that what my [here](https://github.com/pyiron/pyiron_base/blob/7ba3283c0926234829ba78a7eedd541b72056e3f/pyiron_base/job/util.py#L168) pointed to: 
https://github.com/pyiron/pyiron_base/blob/7ba3283c0926234829ba78a7eedd541b72056e3f/pyiron_base/job/util.py#L168

and `job.job_name = new_job_name` calls that function.

samwaseda:
Oh right here we go :) yes that's what we'd need

PR #641, opened 2022-01-19
Rename compressed files on job rename

niklassiemer:
closes #640

PR #642, opened 2022-01-20
Add tests for correct renaming of the working directory and archive

niklassiemer:
Exposed bug #640 in the tests


niklassiemer:
Now included in the fixing PR #641

PR #644, opened 2022-01-20
Always write error.out

pmrv:
Previously error.out was written when the job was successful and
error.msg when not.  In case the job returned an accepted non zero
return nothing was written due to a bug.  This commit writes all the
output of the job to error.out in all cases, plus error.msg when the job
errors out for backwards compatibility.

I introduced the bug in #537.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -775,7 +775,9 @@ class GenericJob(JobCore):
             else:
                 job_crashed = True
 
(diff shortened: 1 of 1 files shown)


pmrv:
@niklassiemer It seems the macos tests are broken right now, did that happen for other PRs as well?

niklassiemer:
Seems to be related to #646. Skimming through recent PRs, this problem pops up in a few places.

niklassiemer:
I now merged the PR from Jan and retriggered the tests.

PR #645, opened 2022-01-20
Set default project for WorkerJob

pmrv:
The `WorkerJob` needs an input project assigned, but the default is `None` and it's not checked before the worker job starts running, which leads to silent death of the worker if you forget to assign one.  I've set the default now to be the project of the worker itself, because that seems to me the common use case.  If that's too specific we should instead add a `validate_ready_to_run` method that does this check.

### DIFF
--- a/pyiron_base/job/worker.py
+++ b/pyiron_base/job/worker.py
@@ -128,7 +128,7 @@ class WorkerJob(PythonTemplateJob):
         super(WorkerJob, self).__init__(project, job_name)
         if not state.database.database_is_disabled:
             self.input.project = project.path
-        else: 
+        else:
             self.input.project = self.working_directory
         self.input.cores_per_job = 1
         self.input.sleep_interval = 10
--- a/pyiron_base/job/worker.py
+++ b/pyiron_base/job/worker.py
@@ -126,7 +126,10 @@ class WorkerJob(PythonTemplateJob):
 
     def __init__(self, project, job_name):
         super(WorkerJob, self).__init__(project, job_name)
-        self.input.project = project.path
+        if not state.database.database_is_disabled:
+            self.input.project = project.path
+        else: 
+            self.input.project = self.working_directory
         self.input.cores_per_job = 1
         self.input.sleep_interval = 10
         self.input.child_runtime = 0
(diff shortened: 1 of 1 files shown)


pmrv:
Mh, I've noticed now that this creates the problem that the worker is waiting for itself to finish, which is also a bit borked...

jan-janssen:
That is strange - I initially had an implementation where the worker was watching a specific project, but by now it should only filter based on the `job_id`. So the worker should not wait for itself.

pmrv:
> That is strange - I initially had an implementation where the worker was watching a specific project, but by now it should only filter based on the `job_id`. So the worker should not wait for itself.

Ah, I think I just got confused by the worker log.  I had tried to create & submit my worker job, submit all the smaller calculations and then immediately set the worker status to collect, hoping that this would shut down the worker once all the jobs are finished.  Alas, when the worker started running on the node it changed its status away from `collect` to `running` and therefore did not stop once it was done. This is a little bit of a pain point, but I'll make a separate issue for this + some refactors.  

The new default works for me, if you don't have objections I'll merge this later today.

jan-janssen:
I guess what you are looking for is the `wait_for_worker()` function. It stops the worker once all calculation are finished and no new jobs are submitted plus the timeout.

jan-janssen (approved):
Looks good to me

pyiron_base/job/worker.py
```diff
@@ -126,7 +126,7 @@ class WorkerJob(PythonTemplateJob):
 
     def __init__(self, project, job_name):
         super(WorkerJob, self).__init__(project, job_name)
-        self.input.project = None
+        self.input.project = project.path
```
jan-janssen:
```suggestion
        if not state.database.database_is_disabled:
            self.input.project = project.path
        else: 
            self.input.project = self.working_directory
```
I just noticed that we also have to consider the case without database.

pyiron_base/job/worker.py
```diff
@@ -126,7 +126,10 @@ class WorkerJob(PythonTemplateJob):
 
     def __init__(self, project, job_name):
         super(WorkerJob, self).__init__(project, job_name)
-        self.input.project = None
+        if not state.database.database_is_disabled:
+            self.input.project = project.path
+        else:
```
jan-janssen:
```suggestion
        else:
```
Fix black ~

pmrv:
> I guess what you are looking for is the `wait_for_worker()` function. It stops the worker once all calculation are finished and no new jobs are submitted plus the timeout.

The issue was that I didn't want to block the notebook, but continue working on other things, but I haven't come up with anything better yet.

PR #646, opened 2022-01-21
Restrict openssl  <2 

jan-janssen:


jan-janssen:
Until https://github.com/conda-forge/cryptography-feedstock/pull/80 is fixed

PR #651, opened 2022-01-24
Refactor job submission in worker

pmrv:
I simply rewrote the logic to run jobs inside the main worker loop to be a little clearer, using a simple for loop instead of multiple loop comprehensions.

### DIFF
--- a/pyiron_base/job/worker.py
+++ b/pyiron_base/job/worker.py
@@ -188,21 +188,13 @@ class WorkerJob(PythonTemplateJob):
                     & (~df["id"].isin(active_job_ids))
                 ]
                 if len(df_sub) > 0:  # Check if there are jobs to execute
-                    path_lst = [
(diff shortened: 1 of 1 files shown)


pyiron_base/job/worker.py
```diff
@@ -188,21 +188,13 @@ def run_static_with_database(self):
...
-                            df_sub["project"].values,
-                            df_sub["id"].values,
-                        )
-                        if job_id not in active_job_ids
-                    ]
-                    job_lst = [
-                        [p, job_id] if pp is None else [os.path.join(pp, p), job_id]
-                        for pp, p, job_id in path_lst
-                    ]
-                    active_job_ids += [j[1] for j in job_lst]
```
jan-janssen, line 204: Are you missing the `active_job_ids` list? basically the jobs are not added to it, right?

PR #658, opened 2022-01-26
Draft automatic conversion to DataContainer

pmrv:
Here's a draft of lazy-loading HDF groups with data containers inside them automatically.  I probably won't get to fleshing it out until next week, but maybe it helps already.

### DIFF
--- a/pyiron_base/job/core.py
+++ b/pyiron_base/job/core.py
@@ -891,7 +891,11 @@ class JobCore(HasGroups):
         # first try to access HDF5 directly to make the common case fast
         try:
             group = self._hdf5[item]
-            if isinstance(group, ProjectHDFio) and "NAME" in group and group["NAME"] == "DataContainer":
+            if (
+                isinstance(group, ProjectHDFio)
+                and "NAME" in group
+                and group["NAME"] == "DataContainer"
+            ):
                 return group.to_object(lazy=True)
             else:
                 return group
@@ -899,6 +903,7 @@ class JobCore(HasGroups):
             pass
 
         name_lst = item.split("/")
+
         def successive_path_splits(name_lst):
             """
             Yield successive split/joins of a path, i.e.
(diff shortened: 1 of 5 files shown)


pmrv:
@samwaseda This was supposed to be for the sphinx output rewrite (I think) you mentioned in a pyiron meeting.  Did you get to try it?

pmrv:
I had to make a small change in a test for the script jobs.  Since the input/output is already a `DataContainer` this no longer needs to be loaded explicitly.  I hope this preserves the original intent @jan-janssen .

pmrv:
I've had to drop the `JobPath.__getitem__` implementation to make this work (not sure why it had one before even).  Also this changes the semantics a little bit in this load any datacontainer you might access through `job[]`, whereas before you'd need to explicitly call `to_object`, but I think it's a change for better and shouldn't affect too many things.

While working on this I've realized that it's actually a very trivial change now to load any object that supports from `to_object` this way.  Simply drop the check that looks for `DataContainer`.  Any thought whether that is desirable?

niklassiemer:
Well, if they are not lazy, loading each supported object could result in severe performance issues on pyiron tables/inspect mode.

pmrv:
> Well, if they are not lazy, loading each supported object could result in severe performance issues on pyiron tables/inspect mode.

Ah, naturally they would only be loaded when accessed.  But e.g. `GenericParameter` is extremely cumbersome to access from plain HDF in tables, so I actually would like them to be loaded directly.  Structures might be a problem, because in a table you would only use positions or indices directly, so loading it would be expensive.  Naja, it's mostly something to think about not in scope for this PR.

pmrv:
Actually now that I think about it, this wouldn't be a problem if you want keep access to say `job['output/structure/positions']` fast, since it always first tries to load directly the data from HDF first and only then tries to `to_object` things.

pmrv:
I've verified that this fixes the [Murnaghan/Sphinx issue](https://github.com/pyiron/pyiron_atomistics/issues/647).  So if you don't see other issues, I'll merge this this afternoon.

PR #669, opened 2022-02-01
Add timeout for coverage test

niklassiemer:


niklassiemer:
It does what it is supposed to do

PR #671, opened 2022-02-01
Remote job error handling

Leimeroth:
Make fetching jobs from a remote server more robust by simply catching all exceptions and printing them as a warning. Right now wait_for_jobs() aborts completely when some error occured that f.e. does not leave the hdf5 file in a valid state, so if you sent 1000 jobs to the cluster and try to get them back it may occur that you need to hack your way around errors by manually fetching the failed job/deleting them completely/whatever to avoid it being in front of the other jobs on the list. Also prevents that you come back after lunch just to notice that all jobs are still on the cluster instead of you computer.
This simply catches all exceptions and prints them as a warning together with the job id. Not sure if this is really a smart way to handle this but at least it is really robust.

Additionally I changed the interval_in_s and max_iterations defaults to only try to get jobs once. This makes more sense in my opinion since asking the login node for a list of all your jobs every few seconds just puts a lot of really unnecessary load on them,
especially if I consider the typical runtime of a job that is sent to a cluster to be > 1h. So at least my typical use case is to sent a lot of jobs and than get them back a few hours or a day later for analysis.

### DIFF
--- a/pyiron_base/server/queuestatus.py
+++ b/pyiron_base/server/queuestatus.py
@@ -211,7 +211,7 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
 
 
 def wait_for_jobs(
-    project, interval_in_s=0, max_iterations=1, recursive=True, ignore_exceptions=False
+    project, interval_in_s=5, max_iterations=100, recursive=True, ignore_exceptions=False
 ):
     """
     Wait for the calculation in the project to be finished
--- a/pyiron_base/server/queuestatus.py
+++ b/pyiron_base/server/queuestatus.py
@@ -6,7 +6,6 @@ Set of functions to interact with the queuing system directly from within pyiron
 """
 
 import warnings
-from logging import warning
 import pandas
 import time
 from pyiron_base.state import state
--- a/pyiron_base/server/queuestatus.py
+++ b/pyiron_base/server/queuestatus.py
@@ -5,6 +5,8 @@
 Set of functions to interact with the queuing system directly from within pyiron - optimized for the Sun grid engine.
 """
 
+import warnings
+from logging import warning
 import pandas
 import time
 from pyiron_base.state import state
@@ -210,7 +212,7 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                 )
 
 
-def wait_for_jobs(project, interval_in_s=5, max_iterations=100, recursive=True):
+def wait_for_jobs(project, interval_in_s=0, max_iterations=1, recursive=True):
     """
     Wait for the calculation in the project to be finished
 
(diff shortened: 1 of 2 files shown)


pyiron_base/server/queuestatus.py
```diff
@@ -269,17 +270,23 @@ def update_from_remote(project, recursive=True):
...
+                        file=job.project_hdf5.file_name,
+                        transfer_back=True,
+                        delete_remote=False,
+                    )
+                    status_hdf5 = job.project_hdf5["status"]
+                    project.set_job_status(job_specifier=job.job_id, status=status_hdf5)
+                    if status_hdf5 in job_status_finished_lst:
+                        job_object = job.to_object()
+                        job_object.transfer_from_remote()
+                except Exception as e:
```
jan-janssen, line 297: Can you be more explicit? What kind of error is raised here?

pyiron_base/server/queuestatus.py
```diff
@@ -210,7 +211,7 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                 )
 
 
-def wait_for_jobs(project, interval_in_s=5, max_iterations=100, recursive=True):
+def wait_for_jobs(project, interval_in_s=0, max_iterations=1, recursive=True):
```
jan-janssen: Is it necessary to change the defaults?

pyiron_base/server/queuestatus.py
Leimeroth: The problem is that I had all kinds of errors already and don't remember all of them. But since there is tons of possible failure mechanisms I don't know if it really makes sense to explicitly list them.
Last error I encountered was a missing hdf5 file and my ugly workaround was this
jobs = pr.job_table()
jobs = jobs[(jobs.status=="running") | (jobs.status=="submitted")]
jobs
for ID in jobs.id:
    try:
        job = super_pr.load(ID)
        job.transfer_from_remote()
        job.status.collect = True
        job.run()
    except Exception as e:
        errors[ID] = e

jan-janssen: The issue with the except all errors part is that it makes it even harder for the next user to debug.

pyiron_base/server/queuestatus.py
Leimeroth: Not really necessary, but I really think the default doesn't make much sense, because submitting a job takes like 20-30 seconds, so everything that runs in 5 seconds is multiple times faster when run locally anyway. Also at least on our cluster the HPC team is not a fan of putting high loads on login nodes by stuff like this.

pyiron_base/server/queuestatus.py
Leimeroth: Maybe as an option? 
if robust_fetch:
    catch all errors
else:
    current behavior

or I can add all errors I encounter one after another, but it will still be a list with like 5 different errors I guess.

niklassiemer: I would like to have the switch to change the behavior. Furthermore, I would suggest to use `state.logger.warning` instead of `warnings.warn` such that the warnings enter our pyiron log.
We could also think about returning a list of failed job_ids in the end (keeping None return value if everything was fine).

Leimeroth: I added an ignore_exceptions option and changed the warning

niklassiemer (commented):
Except for missing docstring updates this LGTM.

PR #661, opened 2022-01-28
Easy rewrite

sudarsan-surendralal:
Currently the way we rewrite the hdf5 file of a job is:

```python 
job.project_hdf5.rewrite_hdf5(job.job_name)
```
I simplified this to

```python 
job.rewrite_hdf5()
```

PR #673, opened 2022-02-01
Add local maintenance with defragment_storage

niklassiemer:
Substitute for #661 from @sudarsan-surendralal 
This adds a local maintenance to the project, where one can trigger a rewrite_hdf for all jobs.

### DIFF
--- a/pyiron_base/project/maintenance.py
+++ b/pyiron_base/project/maintenance.py
@@ -76,9 +76,6 @@ class LocalMaintenance:
     def __init__(self, project):
         self._project = project
 
-        from pyiron_base import Project
-
-        self._project = Project()
 
     def defragment_storage(
         self,
--- a/pyiron_base/project/maintenance.py
+++ b/pyiron_base/project/maintenance.py
@@ -2,7 +2,6 @@ import importlib
 import os
 import pkgutil
 import warnings
-from _warnings import warn
 
 import pandas
 from git import Repo, InvalidGitRepositoryError
@@ -27,7 +26,7 @@ class Maintenance:
         self._project = project
         self._global = GlobalMaintenance()
         self._update = UpdateMaintenance(self._project)
-        self._local = None
+        self._local = LocalMaintenance(self._project)
 
     @property
     def global_status(self):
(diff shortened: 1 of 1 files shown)


pyiron_base/project/maintenance.py
```diff
@@ -69,6 +72,37 @@ def get_repository_status():
...
+            progress (bool): if True (default), add an interactive progress bar to the iteration
+            **kwargs (dict): Optional arguments for filtering with keys matching the project database column name
+                            (eg. status="finished"). Asterisk can be used to denote a wildcard, for zero or more
+                            instances of any character
+        """
+        for job in self._project.iter_jobs(
+            recursive=recursive, progress=progress, convert_to_object=False, **kwargs
+        ):
+            hdf = job.project_hdf5
+            hdf.rewrite_hdf(job.name)
```
sudarsan-surendralal, line 100: Would it be a good idea to remove the `job_name` argument for `rewrite_hdf` in this PR itself?

pyiron_base/project/maintenance.py
```diff
@@ -69,6 +72,37 @@ def get_repository_status():
         return report
 
 
+class LocalMaintenance:
+    def __init__(self, project):
+        self._project = project
+
+        from pyiron_base import Project
+
+        self._project = Project()
```
sudarsan-surendralal: Why do we need to do this?

niklassiemer: Thanks! This should not be in here... I just used it such that pycharm knows how to autocomplete... Typing hints would be of course better but result in circular import 
```suggestion
```

pyiron_base/project/maintenance.py
niklassiemer: I am not really sure. Looking at `rewrite_hdf` I sort of had the feeling that this _only_ rewrites the job itself and not (necessarily) the overall file. I wanted to do (and probably include some tests into base) to see if it actually correctly rewrites the whole hdf file if multiple jobs are in there or has a latent bug for such cases.

sudarsan-surendralal (approved):
LGTM! I think the changes with the `rewrite_hdf` function can be done in a separate PR

PR #688, opened 2022-02-08
Test project size

jan-janssen:
A very basic test, as we had two bug fixes https://github.com/pyiron/pyiron_base/pull/687 and https://github.com/pyiron/pyiron_base/pull/685 because never actually tested the code and the error only appeared in https://github.com/pyiron/pyiron_gui/pull/118


jan-janssen:
The bugs were originally introduced in https://github.com/pyiron/pyiron_base/pull/631

Issue #152, opened 2020-11-27
Removing jobs takes absurd amount of time


caglayanoaras:
**Summary**
remove_jobs_silently(recursive=True) method takes too much time. I am not an expert in the field but I believe deleting jobs recursively one by one is not the best algorithm.
**Detailed Description**
I did some dummy tests to see the pattern. Figures below show the number of phonopy jobs and calc_minimize jobs in a project and the time it took to delete the project.
![1](https://user-images.githubusercontent.com/56139345/100401007-844fa700-3058-11eb-9a3b-8103980b888f.PNG)
![2](https://user-images.githubusercontent.com/56139345/100401089-c973d900-3058-11eb-9f07-165d9df040fc.PNG)

I do not think that this is super critical but I wanted to open an issue because as the number of dimensions increase, job numbers might increase a lot. For instance, in my case I have around 6000 jobs an it takes around 75 minutes to delete the project.

I also have a question. Since there is some sql db working in the background, I do not delete the files directly. But I do not know if it will make a problem. What will happen if I directly delete the project folder?

pmrv:
This came up again for me, so I've collected a profile, but I don't have time right now to dig into it.

Looking superficially at it, it seems were spending
- ~20% opening HDF5 files
- ~50% talking to the database
- only ~10% actually removing files

[remove_profiles.zip](https://github.com/pyiron/pyiron_base/files/6961418/remove_profiles.zip)

jan-janssen:
One thing I typically do is: 
```
df = pr.job_table()
for job_id in df[np.isnan(df.masterid)].id.values[::-1]:
    pr.remove_job(job_id)
```
This is already faster than `pr.remove_jobs(recursive=True, silently=True)`

jan-janssen:
The ugly direct deletion code:
```
df = pr.job_table()
for job_id, project_path, project, job in zip(df.id, df.projectpath, df.project, df.job):
    pr.db.delete_item(item_id=job_id)
    shutil.rmtree(os.path.join(project_path, project, job + "_hdf5"))
    # shutil.rmtree(os.path.dirname(pr.inspect(job_id).working_directory))
```

samwaseda:
Discussed in #844

PR #690, opened 2022-02-10
Do not recursively remove project

pmrv:
Previously on removing a project we first query the database for jobs in
each subproject recursively and separately, then remove them, then
remove empty folders.  This scales (in terms of database queries)
    O(number of recursive subprojects) + O(number of recursive jobs)

Instead of doing this by recursively calling .remove() on sub projects
this new implementation simply queries for all jobs in one database
call, removes them, then finds and removes empty subfolders.  This
scales
    O(1) + O(number of recursive jobs)

(The O(number of recursive jobs) is due to us removing each job in a separate query, this is of course also not ideal)

I'm not exactly sure how this interacts with the `enforce` flag or what is meant by 'delete jobs even though they are used in other project'.  If someone can tell me how to construct a project that contains jobs (at the file level) that are actually contained in a different project (on the database level), I'll add a test case for that.  So far I think that this should be strictly equivalent since the `enforce` does not enter the `_remove_jobs_helper` method and only concerns the removal of the project directory.

PS: I caught this because removing a project with 40 empty projects inside took ~40s and it's now down to ~0.5s.  This touches #152, but does not completely fix it.

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -1093,10 +1093,6 @@ class Project(ProjectPath, HasGroups):
                 "To prevent users from accidentally deleting files - enable has to be set to True."
             )
         if not self.db.view_mode:
-            for sub_project_name in self.list_groups():
-                if "_hdf5" not in sub_project_name:
-                    sub_project = self.open(sub_project_name)
-                    sub_project.remove(enable=enable, enforce=enforce)
             self._remove_jobs_helper(recursive=True)
             for file in self.list_files():
                 os.remove(os.path.join(self.path, file))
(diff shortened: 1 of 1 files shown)


niklassiemer (approved):
If I did not oversee anything, your implementation does exactly the same as the old one! And about 'not deleting jobs used in other projects' I also do not see any connection... The removal of jobs with the `self._remove_jobs_helper(recursive=True)` is exactly the same and independent of this flag (and with recursive, it should not matter if I use it on the top, or on all sub-projects). And I also do not know how to have a job not belonging to a project in there...

PR #681, opened 2022-02-07
pyiron without database: Reduce need to build file table 

jan-janssen:

PR #694, opened 2022-02-15
Worker - Execute jobs directly in Multiprocessing 

jan-janssen:
This reduces the levels of python instances and hopefully the memory footprint

Based on https://github.com/pyiron/pyiron_base/pull/681


jan-janssen:
With this new implementation, the individual nodes use 50% less memory on the master node. Basically because we remove one layer of subprocesses. I am currently checking how this scales up to 512 nodes.

PR #702, opened 2022-02-21
Store Job_id in HDF5 file

### LINKED ISSUE #700: pi_None job name on remote cluster
Since a few day all jobs I submit to our cluster get the job name pi_None, instead of pi_SomeNumber. This leads to an error when trying to retrieve the jobs while some are still waiting in the queue because the table is filtered using the job names and None can not be converted to an int. I suspect that this is related to the changes in #681.

jan-janssen:


Leimeroth:
ID seems correct on the cluster. The only problem I see is that the standard squeue format does not use enough characters for the job name even with my job IDs for < 1e6 jobs. I am not sure which output formatting is used when this is accessed.

jan-janssen:
The code we use is located in https://github.com/pyiron/pysqa/blob/master/pysqa/wrapper/slurm.py#L35 Can we extend the job_name on the slurm side?

jan-janssen:
Can you try https://github.com/pyiron/pysqa/pull/116 ?

jan-janssen:
If it does not work out of the box then it might be necessary to replace `startswith` with `contains` in https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/server/queuestatus.py#L266 and https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/server/queuestatus.py#L268  - I am just not sure if the empty spaces are ignored or not

jan-janssen:
The white spaces are not ignored.

jan-janssen:
@Leimeroth Can you check if the updates in this branch and the modified version of `pysqa` solve your issue?

Leimeroth:
Sorry for the late reply, I was in france with a terrible internet connection. 

Now jobs try to update their status and fail doing that after running successfully, but before collecting the output.
Exemplary Backtrace from error file:

```
2022-02-27 09:47:05,013 - pyiron_log - INFO - run job: ConstrainedMishinStart105508 id: 618454, status: collect
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 716, in run
    self._run_if_collect()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 1534, in _run_if_collect
    self.project.db.item_update(self._runtime(), self.job_id)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 1654, in _runtime
    start_time = self.project.db.get_item_by_id(self.job_id)["timestart"]
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/database/filetable.py", line 118, in get_item_by_id
    return {
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/database/filetable.py", line 119, in <dictcomp>
    k: list(v.values())[0]
IndexError: list index out of range

During handling of the above exception, another exception occurred:

  File "/home/nl77pupy/miniconda3/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/nl77pupy/miniconda3/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/cli/__main__.py", line 3, in <module>
    main()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/cli/__init__.py", line 63, in main
    args.cli(args)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/cli/wrapper.py", line 31, in main
    job_wrapper_function(
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/wrapper.py", line 171, in job_wrapper_function
    job.run()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/wrapper.py", line 124, in run
    self.job.run_static()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 813, in run_static
    self.run()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/generic/util.py", line 213, in decorated
    return function(*args, **kwargs)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 726, in run
    self.drop_status_to_aborted()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 1393, in drop_status_to_aborted
    self.refresh_job_status()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 439, in refresh_job_status
    initial_status=self.project.db.get_job_status(self.job_id),
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/database/filetable.py", line 317, in get_job_status
    return self._job_table[self._job_table.id == job_id].status.values[0]
IndexError: index 0 is out of bounds for axis 0 with size 0
```

Leimeroth (changes requested):
See comment with error backtrace

Leimeroth:
@jan-janssen do you have any update on this? I have a few workarounds for this problem but they are quite clunky and can cause other problems

Leimeroth:
When locally disabling the database Master jobs will also fail, because they don't run child jobs and instead directly try to run they analysis (tested for ElasticTensor and Murnaghan jobs). In the job_table they still show up as initialized, but the jobs have the status aborted if directly checking via job.status.

Currently the job_id when database is disabled is None. I am not sure if this is actually wanted, since jobs are stored with an ID in the filetable and also searched based on the ID in get_item_by_id().

On the remote cluster they must have an ID but I am not sure if is one that is consistent with entries in the filetable, because the error thrown above shows up when trying get_item_by_id() with an id that is not in the table,
but they do not throw the error that occurs when trying int(None)
Is the filetable on the remote cluster stored somewhere so that I could check which IDs job have in there?

Also in the code there are tons of checks which ask if database_is_disabled, but recent reworks on the database and filtetable interfaces were trying to unify them, so are they still necessary? ( @pmrv I think you implemented a lot of the interface changes, maybe you know more about this?)

pmrv:
> When locally disabling the database Master jobs will also fail, because they don't run child jobs and instead directly try to run they analysis (tested for ElasticTensor and Murnaghan jobs). In the job_table they still show up as initialized, but the jobs have the status aborted if directly checking via job.status.
> 
> Currently the job_id when database is disabled is None. I am not sure if this is actually wanted, since jobs are stored with an ID in the filetable and also searched based on the ID in get_item_by_id().
> 
> On the remote cluster they must have an ID but I am not sure if is one that is consistent with entries in the filetable, because the error thrown above shows up when trying get_item_by_id() with an id that is not in the table, but they do not throw the error that occurs when trying int(None) Is the filetable on the remote cluster stored somewhere so that I could check which IDs job have in there?
> 
> Also in the code there are tons of checks which ask if database_is_disabled, but recent reworks on the database and filtetable interfaces were trying to unify them, so are they still necessary? ( @pmrv I think you implemented a lot of the interface changes, maybe you know more about this?)

The changes in #336 should only affect code paths where we run with a database.  All logic still flows through the methods defined [here](https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/database/jobtable.py) and the cases where `isinstance(database, FileTable)` should be unchanged unless I've made a mistake somewhere.

Leimeroth:
@pmrv is the filetable stored somehow to check how ids look in there on the remote cluster?
I think with @jan-janssen changes they should have the same IDs as on my local computer, but I can't really verify that and the ID is often set to the job_name when database is disabled.

pmrv:
> @pmrv is the filetable stored somehow to check how ids look in there on the remote cluster? I think with @jan-janssen changes they should have the same IDs as on my local computer, but I can't really verify that and the ID is often set to the job_name when database is disabled.

Quickly browsing `FileTable` that does not seem to be the case, am I missing something @jan-janssen?

jan-janssen:
Initially the filetable was numbering jobs by counting the jobs in a given project. I the accelerated the pyiron version without the database by only initializing the fileable when necessary, so that in particular the job wrapper does not need to build the whole file table if it is only executing a single calculation, as that was slowing down workers without the database. Now in the latest changes we reuse the job_id from the local database, which guarantees that the job_id is always unique, but when we build a filetable on the remote cluster this ID is currently not considered. I plan to look into this but I can not guarantee I get it done before the weekend.

Leimeroth:
Any update on this?

liamhuber:
How will this interact with `Project.pack` and `unpack`? Sorry to butt in with a probably naive question, but at first glance it looks like we'd run into trouble where the stored job id would overwrite the (safe and unique) auto-generated job id from unpack... or?

liamhuber:
How will this interact with `Project.pack` and `unpack`? Sorry to butt in with a probably naive question, but at first glance it looks like we'd run into trouble where the stored job id would overwrite the (safe and unique) auto-generated job id from unpack... or?

Leimeroth:
Bump

Also
> How will this interact with `Project.pack` and `unpack`? Sorry to butt in with a probably naive question, but at first glance it looks like we'd run into trouble where the stored job id would overwrite the (safe and unique) auto-generated job id from unpack... or?

I don't know but currently pack and unpack also allow to add a job with the same name to a project multiple times, so I guess this will mess with it even more.

Leimeroth:
Bump

Leimeroth:
Could we merge this and fix leftover bugs in another pr? I need to sit on this super old branch on the cluster, causing more and more trouble with incompatibilities, f.e. with the executable inheriting from has_storage

niklassiemer:
If you are more or less happy with this implementation (and you are the only one using this feature in production to the best of my knowledge) feel free to merge and open/update an issue with the remaining bugs concerning this feature!

Leimeroth (approved):

PR #701, opened 2022-02-21
Updates for Scriptjob

jan-janssen:
```
import os
from pyiron_base import Project

script_py = """\
from pyiron_base import load, dump
input_dict = load()
output_dict = input_dict.copy()
dump(output_dict)
"""

file_name = "test.py"
with open(file_name, "w") as f:
    f.writelines(script_py)

pr = Project("test")
job = pr.create.job.ScriptJob("script")
job.script_path = os.path.abspath(file_name)
job.input["a"] = 1
job.input["b"] = [1,2,3]
job.run()

data_container = job["output"].to_object().to_builtin()
```

In combination with https://github.com/pyiron/pyiron_base/pull/637 this should cover the implementation of new classes.

PR #705, opened 2022-02-22
Dynamic loading of classes

jan-janssen:
Based on these changes I created a "low-code" example https://github.com/jan-janssen/pyiron-dynamic . In the example the developer only needs to create a `input.json` file for the default inputs and a `script.py` file for interfacing with the simulation code. The rest is handled by the functionality in `pyiron_dynamic`. That is less than 50 lines which include all the magic of dynamically generating classes on-the-fly. Currently these examples do not work on the queuing system, but if we integrate this functionality inside `pyiron_base` that should be no issue. @niklassiemer and @muh-hassani maybe the two of you can collect some feedback if this would address the needs of users from other disciplines.

### DIFF
--- a/pyiron_base/generic/dynamic.py
+++ b/pyiron_base/generic/dynamic.py
@@ -12,8 +12,11 @@ def _get_class_path_lst():
     class_path_lst = []
     for path in path_lst:
         class_path_lst += [
-            cp for cp, files in
-            [[cp, os.listdir(cp)] for cp in [os.path.join(path, c) for c in os.listdir(path)]]
+            cp
+            for cp, files in [
+                [cp, os.listdir(cp)]
+                for cp in [os.path.join(path, c) for c in os.listdir(path)]
+            ]
             if "script.py" in files and "input.json" in files
         ]
     return class_path_lst
@@ -36,13 +39,22 @@ def _init_constructor(class_name, script_path, input_dict):
 
 def _class_constructor(cp):
     from pyiron_base.job.script import ScriptJob
+
     class_name = os.path.basename(cp)
     script_path = os.path.join(cp, "script.py")
     input_dict = _load_input(path=cp)
-    return type(class_name, (ScriptJob,), {
-        "__init__": _init_constructor(class_name=class_name, script_path=script_path, input_dict=input_dict)
-    })
+    return type(
+        class_name,
+        (ScriptJob,),
+        {
+            "__init__": _init_constructor(
+                class_name=class_name, script_path=script_path, input_dict=input_dict
+            )
+        },
+    )
 
 
 def get_template_classes():
-    return {os.path.basename(cp): _class_constructor(cp=cp) for cp in _get_class_path_lst()}
+    return {
+        os.path.basename(cp): _class_constructor(cp=cp) for cp in _get_class_path_lst()
+    }
--- a/pyiron_base/generic/dynamic.py
+++ b/pyiron_base/generic/dynamic.py
@@ -55,9 +55,7 @@ def class_constructor(cp):
 
 
 def _get_template_classes():
-    return {
-        os.path.basename(cp): cp for cp in _get_class_path_lst()
-    }
+    return {os.path.basename(cp): cp for cp in _get_class_path_lst()}
 
 
 JOB_DYN_DICT = _get_template_classes()
--- a/pyiron_base/generic/dynamic.py
+++ b/pyiron_base/generic/dynamic.py
@@ -1,0 +1,48 @@
+import os
+import json
+from pyiron_base.state import state
+
+
+def _get_class_path_lst():
+    path_lst = [
+        os.path.abspath(os.path.expanduser(os.path.join(p, "template")))
+        for p in state.settings.resource_paths
+        if "template" in os.listdir(p)
+    ]
+    class_path_lst = []
+    for path in path_lst:
+        class_path_lst += [
+            cp for cp, files in
+            [[cp, os.listdir(cp)] for cp in [os.path.join(path, c) for c in os.listdir(path)]]
+            if "script.py" in files and "input.json" in files
+        ]
+    return class_path_lst
+
+
+def _load_input(path):
+    with open(os.path.join(path, "input.json")) as f:
+        return json.load(f)
+
+
+def _init_constructor(class_name, script_path, input_dict):
+    def __init__(self, project, job_name):
+        super(self.__class__, self).__init__(project, job_name)
+        self.__name__ = class_name
+        self.input.update(input_dict)
+        self.script_path = script_path
+
+    return __init__
+
+
+def _class_constructor(cp):
+    from pyiron_base.job.script import ScriptJob
+    class_name = os.path.basename(cp)
+    script_path = os.path.join(cp, "script.py")
+    input_dict = _load_input(path=cp)
+    return type(class_name, (ScriptJob,), {
+        "__init__": _init_constructor(class_name=class_name, script_path=script_path, input_dict=input_dict)
+    })
+
+
+def get_template_classes():
+    return {os.path.basename(cp): _class_constructor(cp=cp) for cp in _get_class_path_lst()}
--- a/pyiron_base/generic/dynamic.py
+++ b/pyiron_base/generic/dynamic.py
@@ -5,9 +5,9 @@ from pyiron_base.state import state
 
 def _get_class_path_lst():
     path_lst = [
-        os.path.abspath(os.path.expanduser(os.path.join(p, "template")))
+        os.path.abspath(os.path.expanduser(os.path.join(p, "templates")))
         for p in state.settings.resource_paths
-        if os.path.exists(p) and "template" in os.listdir(p)
+        if os.path.exists(p) and "templates" in os.listdir(p)
     ]
     class_path_lst = []
     for path in path_lst:
--- a/pyiron_base/generic/dynamic.py
+++ b/pyiron_base/generic/dynamic.py
@@ -37,7 +37,7 @@ def _init_constructor(class_name, script_path, input_dict):
     return __init__
 
 
-def _class_constructor(cp):
+def class_constructor(cp):
     from pyiron_base.job.script import ScriptJob
 
     class_name = os.path.basename(cp)
@@ -54,7 +54,10 @@ def _class_constructor(cp):
     )
 
 
-def get_template_classes():
+def _get_template_classes():
     return {
-        os.path.basename(cp): _class_constructor(cp=cp) for cp in _get_class_path_lst()
+        os.path.basename(cp): cp for cp in _get_class_path_lst()
     }
+
+
+JOB_DYN_DICT = _get_template_classes()
--- a/pyiron_base/generic/dynamic.py
+++ b/pyiron_base/generic/dynamic.py
@@ -7,7 +7,7 @@ def _get_class_path_lst():
     path_lst = [
         os.path.abspath(os.path.expanduser(os.path.join(p, "template")))
         for p in state.settings.resource_paths
-        if "template" in os.listdir(p)
+        if os.path.exists(p) and "template" in os.listdir(p)
     ]
     class_path_lst = []
     for path in path_lst:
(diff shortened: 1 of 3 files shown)


jan-janssen:
Waiting for https://github.com/pyiron/pyiron_base/pull/701

jan-janssen:
Waiting for https://github.com/pyiron/pyiron_base/pull/706

pyiron_base/generic/dynamic.py
```diff
@@ -0,0 +1,60 @@
...
+        super(self.__class__, self).__init__(project, job_name)
+        self.__name__ = class_name
+        self.input.update(input_dict)
+        self.script_path = script_path
+
+    return __init__
+
+
+def _class_constructor(cp):
+    from pyiron_base.job.script import ScriptJob
```
jan-janssen, line 41: The local import inside the function is not so nice - but otherwise I always got infinite import loops.

pyiron_base/generic/dynamic.py
```diff
@@ -0,0 +1,60 @@
...
+    ]
+    class_path_lst = []
+    for path in path_lst:
+        class_path_lst += [
+            cp
+            for cp, files in [
+                [cp, os.listdir(cp)]
+                for cp in [os.path.join(path, c) for c in os.listdir(path)]
+            ]
+            if "script.py" in files and "input.json" in files
```
max-hassani, line 20: This assumes that inside the directory, the name of the script is `script.py` and by default there is a `input.json` file available. I see both reasonable as we need some default keys for the input, but if the script does not accepts any input, no input.json is needed. However, one can leave it empty.

max-hassani (approved):
@jan-janssen thanks for these 3 PRs. I have just a minor comment on necessaty of having an input.json file in the template directory. But other than that LGTM!

max-hassani:
On Friday, I will communicate this recent development to our PMD partners and comment here the feedback.

pyiron_base/generic/dynamic.py
```diff
@@ -0,0 +1,60 @@
+import os
+import json
+from pyiron_base.state import state
+
+
+def _get_class_path_lst():
+    path_lst = [
+        os.path.abspath(os.path.expanduser(os.path.join(p, "template")))
+        for p in state.settings.resource_paths
+        if os.path.exists(p) and "template" in os.listdir(p)
```
pmrv: 'template' is not a very descriptive name, I'd suggest 'scriptjob_templates' or similar.
Generally this kind of snippet of searching resources comes up very often all over pyiron and we should make a function for it, but maybe in a different PR.

pyiron_base/job/jobtype.py
```diff
@@ -176,6 +177,42 @@ def wrapper(job_name, delete_existing_job=False, delete_aborted_job=False):
 
             return wrapper
         else:
+            job_dyn_dict = get_template_classes()
+            if name in job_dyn_dict.keys():
+
+                def wrapper(
+                    job_name, delete_existing_job=False, delete_aborted_job=False
+                ):
```
pmrv, line 185: This does not need to be copied from above, just re-indented or am I missing a subtle change?

pmrv (changes requested):
If I understand this correctly this only creates classes dynamically to set `self.script_path` from the `_init_constructor`.  In that case is it not much clearer to have 
1. function that gets the name/path of available scripts from the template directory
2. in jobtype.py:179 define a wrapper that creates a script job and sets the `script_path` before returning the newly created job?

Unless this is really how we want to structure the interface, I'd much prefer a solution that simply adds a property on `ScriptJobs` that allows (tab-completable) setting of the script_path, because it's a much more localized change of the code. 
```python
class ScriptJob:
  ...
  @property
  def set_template(self):
    return TemplateSetter(self)

class TemplateSetter:
  def __init__(self, job):
    self._job = job
  def __dir__(self):
    return get_script_templates() # same as current get_template_classes, but just returns the paths
  def __getattr__(self, name):
    def f():
      self._job.script_path = name
    return f
```
then users do something like
```python
job = pr.create.job.ScriptJob('name')
job.set_template.my_fancy_script()
```

pyiron_base/job/jobtype.py
jan-janssen: The only thing that changes is the dictionary with the available classes - but I agree we could just have a separate function for this.

pyiron_base/generic/dynamic.py
jan-janssen: If the script does not accept inputs, then it does not really make sense to execute it multiple time - maybe to sample a statistical distribution, but this can be done once and does not need a pyiron job. Based on this my assumption is that each script requires some inputs and generates some outputs and this can be as one or two variables.

pyiron_base/generic/dynamic.py
jan-janssen: I agree that `template` is not very descriptive and I start to like `templates` more than just `template`. Still I find `scriptjob_templates` too long, as I imagine in future we might have users who only use these template jobs and run `pyiron_base` as a general workflow manager. From their perspective they put their own classes under `pyiron/resources/templates` right next to the configuration of the queuing system in `pyiron/resources/queues`.

jan-janssen:
> then users do something like
> 
> ```python
> job = pr.create.job.ScriptJob('name')
> job.set_template.my_fancy_script()
> ```

I guess the use case I envision is that people start to use `pyiron_base` as a workflow manager and just use these kind of script job based jobs. Then it would be confusing if all jobs have the same name even though they might have used different scripts. As I demonstrate in the example notebook https://github.com/jan-janssen/pyiron-dynamic/blob/master/example.ipynb with my current implementation the jobs get the name of the folder the `script.py` and `input.json` file are included in. So there is no limit to the number of script job based job classes.

pyiron_base/job/jobtype.py
jan-janssen: Fixed

jan-janssen:
> On Friday, I will communicate this recent development to our PMD partners and comment here the feedback.

Great - just be careful to use the latest changes of this branch, as I made one change I had to revert.

pyiron_base/job/jobtype.py
jan-janssen: I tried to wrap it into a separate function, but that did not work. https://github.com/pyiron/pyiron_base/blob/dynamic2/pyiron_base/job/jobtype.py

max-hassani:
I tried this branch for [a fenics example](https://github.com/muh-hassani/pyiron-dynamic) as a demonstrator today. I found out that the auto-completion does not work. 
I know that having this dynamically created job classes are fantastic so far, but I wanted to point out this point, and wonder whether there is a solution to get the auto-completion working?

jan-janssen:
> I know that having this dynamically created job classes are fantastic so far, but I wanted to point out this point, and wonder whether there is a solution to get the auto-completion working?

I just pushed the implementation of the auto-completion, can you test it again?

max-hassani:
> > I know that having this dynamically created job classes are fantastic so far, but I wanted to point out this point, and wonder whether there is a solution to get the auto-completion working?
> 
> I just pushed the implementation of the auto-completion, can you test it again?

@jan-janssen, thanks for your prompt action. I confirm that the auto-completion works. 
By the way, the feedback I got from PMD partners was very positive.

jan-janssen:
> @jan-janssen, thanks for your prompt action. I confirm that the auto-completion works. By the way, the feedback I got from PMD partners was very positive.

Then we should try to get this functionality merged on Monday.

pyiron_base/generic/dynamic.py
```diff
@@ -0,0 +1,61 @@
...
+    script_path = os.path.join(cp, "script.py")
+    input_dict = _load_input(path=cp)
+    return type(
+        class_name,
+        (ScriptJob,),
+        {
+            "__init__": _init_constructor(
+                class_name=class_name, script_path=script_path, input_dict=input_dict
+            )
+        },
```
niklassiemer, line 53: We need to overwrite `write_input` to copy all files (and directories) but the `input.json` . Otherwise scripts calling other local scripts do not work.

pyiron_base/generic/dynamic.py
pmrv: The reverse argument of course is that the majority of our users might not use `ScriptJobs` and for them it would be confusing to have such a specialized feature living in such a generically named folder.   I don't think folder names can be too long since shells tab-complete, but I also don't care very strongly about this.

pyiron_base/job/jobtype.py
pmrv: Bummer. :/

pmrv:
> > then users do something like
> > ```python
> > job = pr.create.job.ScriptJob('name')
> > job.set_template.my_fancy_script()
> > ```
> 
> I guess the use case I envision is that people start to use `pyiron_base` as a workflow manager and just use these kind of script job based jobs. Then it would be confusing if all jobs have the same name even though they might have used different scripts. As I demonstrate in the example notebook https://github.com/jan-janssen/pyiron-dynamic/blob/master/example.ipynb with my current implementation the jobs get the name of the folder the `script.py` and `input.json` file are included in. So there is no limit to the number of script job based job classes.

What I generally don't like about this is that it makes the job creation logic significantly more complex (and harder to maintain for anyone but the two of us), but only benefits a small feature set.   Job names are a good argument, but is there another way of getting that in without creating new classes?  We might mess with the `job.__name__` if a template script is set, but that would require more logic loading the job.

Anyways, since I'm not involved I will not road block anything, but to me these difficulties suggest that we should not tie the job type name and the actual class that implements them to tightly and we might take this up in the new database scheme.

niklassiemer:
Well, the database scheme would only be available if there is a database...
However, I get your point that this is in general more complex this way. What would you think of an additional job factory a la `pr.create.template_job.` which autocompletes to all script-jobs and returns a script-job with the correct template attached? The job_name problems, however, would remain and I agree that we should decouple the job_name and the class doing it.

pyiron_base/generic/dynamic.py
```diff
@@ -0,0 +1,61 @@
+import os
+import json
+from pyiron_base.state import state
+
+
+def _get_class_path_lst():
+    path_lst = [
+        os.path.abspath(os.path.expanduser(os.path.join(p, "template")))
+        for p in state.settings.resource_paths
+        if os.path.exists(p) and "template" in os.listdir(p)
```
niklassiemer:
```suggestion
        os.path.abspath(os.path.expanduser(os.path.join(p, "templates")))
        for p in state.settings.resource_paths
        if os.path.exists(p) and "templates" in os.listdir(p)
```

pyiron_base/generic/dynamic.py
jan-janssen: It might be a bit ambiguous but from my perspective this feature could really help to increase the number of people who just use `pyiron_base` with other all the other functionality. For the users who do not want to use this feature, they do not need to place a `templates` folder in their resource directory, in that case it is less confusing.

pyiron_base/job/jobtype.py
jan-janssen: I guess the issue is that the separate function is evaluated directly, so all the magic breaks down.

pyiron_base/generic/dynamic.py
jan-janssen: I think this is fine - as users should place their scripts in the python path. In the worst case they can do it by setting `sys.path`.

jan-janssen:
The reason why I am pushing for this feature, is that I have the feeling that `pyiron_base` can finally compete with other general workflow managers. 
* https://snakemake.readthedocs.io/
* https://signac.io 
* https://materialsproject.github.io/fireworks/ 
* https://parsl-project.org

Our approach of using Jupyter notebooks, defining templates which are reusable and the ability to scale up to large computer clusters is rather unique. So I have the hope to attract a larger community of users, who then help us to stabilize this functionality. We should maybe even think about a separate name for `pyiron_base` if more and more users start to use it as stand-alone solution.

niklassiemer:
I like the general functionality and idea, maybe we will iterate over some implementation details, however, I will now merge as is and we might refactor/redesign later :)

Issue #627, opened 2022-01-13
Two JobFactories?


niklassiemer:
We currently have two `JobFactories` in 
https://github.com/pyiron/pyiron_base/blob/7101ea78d8c9689c2b8b38cdbb9796cfe4601514/pyiron_base/job/factory.py#L118
and 
https://github.com/pyiron/pyiron_base/blob/7101ea78d8c9689c2b8b38cdbb9796cfe4601514/pyiron_base/job/jobtype.py#L130.

While the first one implements a `call` and is used in base_tools, the second one is actually used in the `Creator` and, thus, responsible to handle `pr.create.job`
Is there a particular reason we have these two different classes?

I really would like to streamline all job creation through one `JobFactory` with one implementation of `JobType` along the lines
```python
class JobType:
    _job_class_dict = {}

   def __new__(cls, ...):
        # same as now

    @classmethod
    def register_job_types(cls, job_class_dict):
        for job_name, job_class in job_class_dict.items():
            if job_name not in cls._job_class_dict:
                cls._job_class_dict[job_name] = job_class
            else:
                raise ValueError(f'Name clash. A job with name "job_name" already defined!')
```
such that we can handle our current JOB_CLASS_DICT as a class attribute on the `JobType`.

liamhuber:
I would also like to only have one (or at least one parent that has the call functionality and gets inherited from). I'm afraid I won't be able to deal with this for the next couple of months though. I think the root of the issue is the Creator vs toolkit mentality. I much prefer toolkits (even if we somehow maintain a creator that gets toolkit elements dynamically loaded onto it) because the Creator path currently leads to reimplementing Project in every pyiron submodule.

liamhuber:
Oh also I'm pretty sure I want tp get rid of JOB_CLASS_DICT altogether as it uses string based links to the objects instead of real object links, which is hell for IDEs and is prone to errors leaking in when we rename or move stuff

liamhuber:
Ping @liamhuber

PR #718, opened 2022-03-09
Use JobType to carry the JOB_CLASS_DICT and add register method

niklassiemer:
Just one idea I have floating around and mentioned in #627 
This could be used to get rid of the `JOB_CLASS_DICT` and ease implementing new classes for new developers, since they only would need to write the class itself without 'by-hand' registering the class.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1486,22 +1486,9 @@ class GenericJob(JobCore):
             self._logger.info("busy master: {} {}".format(master_id, self.get_job_id()))
             del self
 
-    @classmethod
-    def _register_jobtype_name(cls) -> Union[str, None]:
-        """Name of the JobType this class is to be registered with; None for do not register.
-        Behavior is not inherited"""
-        return cls.__name__
-
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
-        # If the _register_jobtype_name is not defined in the new class definition _itself_ it falls back to the
-        # default, i.e. to register the class with the class.__name__
-        if "_register_jobtype_name" not in cls.__dict__.keys():
-            register_name = cls.__name__
-        else:
-            # This one follows the method resolution order
-            register_name = cls._register_jobtype_name()
-        JobType.register_job_type(register_name, cls)
+        JobType.register_job_type(cls.__name__, cls)
 
 
 class GenericError(object):
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -8,11 +8,8 @@ Generic Job class extends the JobCore class with all the functionality to run th
 from datetime import datetime
 import os
 import posixpath
-import multiprocessing
 from typing import Union
-
 import h5io
-from pyiron_base.job.wrapper import JobWrapper
 from pyiron_base.job.jobtype import JobType
 import signal
 import warnings
@@ -1499,7 +1496,7 @@ class GenericJob(JobCore):
         super().__init_subclass__(**kwargs)
         # If the _register_jobtype_name is not defined in the new class definition _itself_ it falls back to the
         # default, i.e. to register the class with the class.__name__
-        if '_register_jobtype_name' not in cls.__dict__.keys():
+        if "_register_jobtype_name" not in cls.__dict__.keys():
             register_name = cls.__name__
         else:
             # This one follows the method resolution order
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1486,6 +1486,7 @@ class GenericJob(JobCore):
             del self
 
     def __init_subclass__(cls, **kwargs):
+        """Auto register all subclasses of GenericJob as available JobType."""
         super().__init_subclass__(**kwargs)
         JobType.register(cls)
 
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -6,9 +6,14 @@ Generic Job class extends the JobCore class with all the functionality to run th
 """
 
 from datetime import datetime
-import h5io
 import os
 import posixpath
+import multiprocessing
+from typing import Union
+
+import h5io
+from pyiron_base.job.wrapper import JobWrapper
+from pyiron_base.job.jobtype import JobType
 import signal
 import warnings
 
@@ -1483,6 +1488,15 @@ class GenericJob(JobCore):
             project.db.set_job_status(job_id=master_id, status="busy")
             self._logger.info("busy master: {} {}".format(master_id, self.get_job_id()))
             del self
+
+    @staticmethod
+    def _register_jobtype_name() -> Union[str, None]:
+        """Name of the JobType this class is to be registered with; None for do not register."""
+        return None
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        JobType.register_job_type(cls._register_jobtype_name(), cls)
 
 
 class GenericError(object):
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1492,7 +1492,7 @@ class GenericJob(JobCore):
     @classmethod
     def _register_jobtype_name(cls) -> Union[str, None]:
         """Name of the JobType this class is to be registered with; None for do not register."""
-        return cls.__name__ 
+        return cls.__name__
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1487,7 +1487,7 @@ class GenericJob(JobCore):
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
-        JobType.register_job_type(cls.__name__, cls)
+        JobType.register(cls)
 
 
 class GenericError(object):
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1491,12 +1491,20 @@ class GenericJob(JobCore):
 
     @classmethod
     def _register_jobtype_name(cls) -> Union[str, None]:
-        """Name of the JobType this class is to be registered with; None for do not register."""
+        """Name of the JobType this class is to be registered with; None for do not register.
+        Behavior is not inherited"""
         return cls.__name__
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
-        JobType.register_job_type(cls._register_jobtype_name(), cls)
+        # If the _register_jobtype_name is not defined in the new class definition _itself_ it falls back to the
+        # default, i.e. to register the class with the class.__name__
+        if '_register_jobtype_name' not in cls.__dict__.keys():
+            register_name = cls.__name__
+        else:
+            # This one follows the method resolution order
+            register_name = cls._register_jobtype_name()
+        JobType.register_job_type(register_name, cls)
 
 
 class GenericError(object):
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1489,10 +1489,10 @@ class GenericJob(JobCore):
             self._logger.info("busy master: {} {}".format(master_id, self.get_job_id()))
             del self
 
-    @staticmethod
-    def _register_jobtype_name() -> Union[str, None]:
+    @classmethod
+    def _register_jobtype_name(cls) -> Union[str, None]:
         """Name of the JobType this class is to be registered with; None for do not register."""
-        return None
+        return cls.__name__ 
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -8,7 +8,6 @@ Generic Job class extends the JobCore class with all the functionality to run th
 from datetime import datetime
 import os
 import posixpath
-from typing import Union
 import h5io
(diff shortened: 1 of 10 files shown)


pyiron_base/job/generic.py
```diff
@@ -1754,6 +1757,15 @@ def _reload_update_master(self, project, master_id):
             self._logger.info("busy master: {} {}".format(master_id, self.get_job_id()))
             del self
 
+    @staticmethod
+    def _register_jobtype_name() -> Union[str, None]:
+        """Name of the JobType this class is to be registered with; None for do not register."""
+        return None
```
pmrv:
```suggestion
    @classmethod
    def _register_jobtype_name(cls) -> Union[str, None]:
        """Name of the JobType this class is to be registered with; None for do not register."""
        return cls.__name__ 
```

I suppose we want to do that everywhere by default.  Jobs that specifically don't want to end up in the job class dict can still override this method to do nothing.

pmrv (approved):
I generally like a mechanism like this, but we can probably not do away with manually registration completely.  [In times of yore](https://github.com/pyiron/pyiron/pull/706) we had already tried something similar and iirc @jan-janssen had done benchmarks that showed that importing all job classes on pyiron import takes way too long.  Hence the current solution of static strings.  

That said I'd much prefer pyiron sub projects (like atomistics/contrib/etc.) to do something like
```python
from pyiron_base import JobType
JobType.register('MyJob', 'pyiron_mine.my_module.my_job.MyJob')
```
than our current solution and this seems like a step in that direction.

pyiron_base/job/generic.py
niklassiemer: I tried that and right now (without any further change) all subclasses of `GenericJob` end up in the attached `JOB_CLASS_DICT`. Therefore, we seem to already import all classes?! In addition, all subclasses of master jobs would need to reimplement that method with other than `return None`, i.e. invoke the method of the super super class...
However, overwriting with 
```
if cls.__name__ == "NotToBeRegisteredJobType":
   return None
else:
   return cls.__name__ 
```
would survive naive subclassing :)

liamhuber:
> I generally like a mechanism like this, but we can probably not do away with manually registration completely. [In times of yore](https://github.com/pyiron/pyiron/pull/706) we had already tried something similar and iirc @jan-janssen had done benchmarks that showed that importing all job classes on pyiron import takes way too long. Hence the current solution of static strings.
> 
> That said I'd much prefer pyiron sub projects (like atomistics/contrib/etc.) to do something like
> 
> ```python
> from pyiron_base import JobType
> JobType.register('MyJob', 'pyiron_mine.my_module.my_job.MyJob')
> ```
> 
> than our current solution and this seems like a step in that direction.

Are we sure times of yore are still relevant? I did similar testing when I was working on the [`Toolkit`](https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/toolkit.py) I experimented with including registration in init and not. For base it's not a big deal, but for [atomistics](https://github.com/pyiron/pyiron_atomistics/blob/22fab20c8c6aa9c9e4ed50f47b4fd21934d96154/pyiron_atomistics/__init__.py#L15), registering the toolkit imports *all* the atomistics jobs. I never noticed an appreciable difference in speed when importing pyiron in a restarted notebook.

niklassiemer:
For some reason, not-string version of the classes in the `JOB_CLASS_DICT` break reloading of jobs. It seems to be caused by our loading of objects in the `ProjectHDFio`, however, I do not have time right now to dig deeper... 
Therefore, I now register all classes using their string representation.

jan-janssen:
@niklassiemer Can you move commit https://github.com/pyiron/pyiron_base/pull/718/commits/06abf36bbb169e24f165138a8892b2bde1ad5aff to a separate pull request? - it is hard to review 65 changed files and the black formatting is not necessary for the proposed changes in this pull request.

jan-janssen:
> I never noticed an appreciable difference in speed when importing pyiron in a restarted notebook.

I guess it strongly depends on your application. For me running 500+ DFT calculation in parallel controlled from one node 256MB for importing pyiron is already terrible. If we want to remain competitive for the upcoming field of Exascale computing we have to be careful with our memory usage.

niklassiemer:
With todays discussion in the pyiron meeting, I made this 'draft' again. We will first have a look at the `JobFactories` to reduce our memory footprint again and then try to straighten our job spawning interface.

liamhuber:
> > I never noticed an appreciable difference in speed when importing pyiron in a restarted notebook.
> 
> 
> 
> I guess it strongly depends on your application. For me running 500+ DFT calculation in parallel controlled from one node 256MB for importing pyiron is already terrible. If we want to remain competitive for the upcoming field of Exascale computing we have to be careful with our memory usage. 

Hmm, I'm afraid I don't know enough to understand the concern. If this is something we want to optimize for, we should definitely add it to the test suite. I can imagine some combination of timeit and module unloading and reloading. Thoughts?

niklassiemer:
Ping @pmrv

pyiron_base/job/generic.py
```diff
@@ -6,9 +6,11 @@
 """
 
 from datetime import datetime
-import h5io
 import os
 import posixpath
+from typing import Union
+import h5io
```
pmrv:
```suggestion
import h5io
```
Doesn't seem to be used anymore.

pyiron_base/job/jobtype.py
```diff
@@ -95,8 +98,50 @@ def __new__(
             job.set_input_to_read_only()
         return job
 
+    @classmethod
+    def un_register_job_type(cls, job_name):
```
pmrv:
```suggestion
    def unregister(cls, job_name_or_class):
      if insinstance(job_name_or_class, type):
        job_name_or_class = job_name_or_class.__name__
```

pmrv:
```python
@JobType.unregister
class Foo:
  pass
```
sounds a little smoother.

pyiron_base/job/jobtype.py
```diff
@@ -95,8 +98,50 @@ def __new__(
...
+        if job_name in cls._job_class_dict:
+            del cls._job_class_dict[job_name]
+        else:
+            raise KeyError(f"No JobType with name '{job_name}' found.")
+
+    @classmethod
+    def register_job_type(cls, cls_name, _cls):
```
pmrv:
```suggestion
    def register(cls, job_class, job_name=None):
```
Since we don't allow custom class names, I would make it optional.

pyiron_base/job/jobtype.py
```diff
@@ -95,8 +98,50 @@ def __new__(
...
+        else:
+            raise ValueError(
+                f"A JobType with name {cls_name} is already defined! New class = {repr(_cls)}, "
+                f"already registered class = {repr(cls._job_class_dict[cls_name])}."
+            )
+
+    @staticmethod
+    def _get_class_path_str(_cls):
+        full_path = repr(_cls).split("<class '")[-1].split("'>")[0]
+        return ".".join(full_path.split(".")[:-1])
```
pmrv: Use `_cls.__name__` and `_cls.__module__` instead of fiddling with strings.

pyiron_base/job/jobtype.py
```diff
@@ -127,6 +172,11 @@ def convert_str_to_class(job_class_dict, class_name):
         )
 
 
+def unregistered_jobtype(cls):
+    JobType.un_register_job_type(cls.__name__)
+    return cls
+
+
```
pmrv:
```suggestion
```
Like I said above, I'd avoid this function and make the class method itself usable as the decorator.

pyiron_base/master/list.py
```diff
@@ -4,12 +4,14 @@
 """
 The ListMaster behaves like a list, just for job objects.
 """
+from typing import Union
```
pmrv:
```suggestion
```

pmrv (approved):
Few nits, but general architecture looks great.

pyiron_base/job/jobtype.py
niklassiemer: That way, unregister has to return the class... sound a bit odd to me?

pyiron_base/job/jobtype.py
niklassiemer: Yes, that full function only returns the value of `cls.__module__` if I am not mistaken 😲 I was not aware of that attribute and copied the function from some place in the `hdf` modules... Will remove the entire function 😄

pyiron_base/job/generic.py
niklassiemer: The h5io is used, typing not. I modified your suggestion to commit only that part.

pyiron_base/job/jobtype.py
niklassiemer: The problem is a bit that it should be possible to register jobs on pure string basis like we do with the `JOB_CLASS_DICT`
```suggestion
    def register(cls, job_class_or_module_str, job_name=None):
```
I will try to make this as clear as possible

pyiron_base/job/jobtype.py
niklassiemer:
```suggestion
            cls._job_class_dict[cls_name] = _cls.__module__
        elif (
            cls_name in cls._job_class_dict
            and not isinstance(_cls, str)
            and cls._job_class_dict[cls_name] == _cls.__module__
        ):
            pass
        else:
            raise ValueError(
                f"A JobType with name {cls_name} is already defined! New class = {repr(_cls)}, "
                f"already registered class = {repr(cls._job_class_dict[cls_name])}."
            )

```

pyiron_base/job/jobtype.py
pmrv: Actually I'd prefer that in case you unregister by name, since that would allow you to check that the right one was removed once we do allow custom names.

pmrv: Also, as the function doesn't return anything else, I don't see a problem.

pyiron_base/job/jobtype.py
```diff
@@ -95,8 +98,48 @@ def __new__(
...
 
+    @classmethod
+    def unregister(cls, job_name_or_class):
+        if isinstance(job_name_or_class, type):
+            job_name_or_class = job_name_or_class.__name__
+        if job_name_or_class in cls._job_class_dict:
+            del cls._job_class_dict[job_name_or_class]
+        else:
+            raise KeyError(f"No JobType with name '{job_name_or_class}' found.")
+        return cls
```
niklassiemer: Should return the correct class...
```suggestion
            raise KeyError(f"No JobType with name '{job_name_or_class}' found.")
        return job_name_or_class
```

pyiron_base/job/jobtype.py
```diff
@@ -95,8 +98,48 @@ def __new__(
...
+    @classmethod
+    def unregister(cls, job_name_or_class):
+        if isinstance(job_name_or_class, type):
+            job_name_or_class = job_name_or_class.__name__
+        if job_name_or_class in cls._job_class_dict:
+            del cls._job_class_dict[job_name_or_class]
+        else:
+            raise KeyError(f"No JobType with name '{job_name_or_class}' found.")
+        return job_name_or_class
+
```
niklassiemer:
```suggestion
    @classmethod
    def unregister(cls, job_name_or_class):
        _cls = None 
        if isinstance(job_name_or_class, type):
            _cls = job_name_or_class
            job_name_or_class = job_name_or_class.__name__
        if job_name_or_class in cls._job_class_dict:
            del cls._job_class_dict[job_name_or_class]
        else:
            raise KeyError(f"No JobType with name '{job_name_or_class}' found.")
        return _cls

```

niklassiemer:
@pmrv the nits caused some rewrites :D I will add some docstrings and done.

niklassiemer:
@jan-janssen are you fine with this new feature? Then, I would like to have it in the next release and to `register` new jobs in our pyiron_module init files.

jan-janssen:
Just for me to understand this functionality, this class requires to import all derived classes for them to be automatically registered, correct? this is in contrast to the previous implementation, where only a list of import paths was stored in the dictionary and classes were only imported when used. As a consequence I am wondering if it increases the memory footprint of `pyrion_atomistics`.

pmrv:
No, this *allows* the automatic registration for classes that derive from `GenericJob`.  Registration can still be "fast-tracked" with the usual "putting strings in `JOB_CLASS_DICT`" method and is in fact still done.

jan-janssen:
> No, this _allows_ the automatic registration for classes that derive from `GenericJob`. Registration can still be "fast-tracked" with the usual "putting strings in `JOB_CLASS_DICT`" method and is in fact still done.

So, it is an optional feature to simplify the development, but the default is to keep everything the way it is, basically using the same dictionary as before?

pmrv:
Yes, the mechanism will stay the same for most of our job classes.  We do want to move `JOB_CLASS_DICT` to class variable and refactor around it, but it will not require importing all modules to register the jobs.

pmrv:
@niklassiemer Would be nice if can add somewhere a little spec, how the register/unregister and automatic registration is supposed to work.  Can also include our plans one namespaces for jobs if it marked as WIP.  Just to have some high level view somewhere in the documentation.  No need at all for this to be in this PR though.

niklassiemer:
The jobs are after all stored as strings in a dictionary, however, I would like to channel the job registration through this class. At some point I want to move the JOB_CLASS_DICT entirely into the JobType class and always use JobType.register to add new jobs. This would then error on attempt to overwrite an existing job type (or allow such with an overwrite=True).
The self registration is merely a nice feature to `pr.create.job.MyJob` after defining MyJob in a notebook :)

niklassiemer:
To be clear, registration would happen on string basis in the init of other pyiron modules, i.e. similar to the current setup.

niklassiemer:
I now take this as no big objections and merge it in.

PR #721, opened 2022-03-10
Fix formatting strings in server debug messages

pmrv:
Previously this was inconsistent (.format-style vs. %-style) and
sometimes completely missing, which lead to errors from the logging
module.


jan-janssen (approved):
Looks good to me

PR #723, opened 2022-03-12
Adding to install dialog

max-hassani:
- I corrected the version of pyrion-resources to be downloaded. Now 0.0.18 is downloaded, instead of 0.0.17.
- I would like to remove the need for user's input in the `install_dialog`, if intended:
```python
pyiron_base.install_dialog(input_by_user=False)
# this does not ask the user for yes or no
pyiron_base.install_dialog()
# this asks the user for a yes/no 
```

### DIFF
--- a/pyiron_base/state/install.py
+++ b/pyiron_base/state/install.py
@@ -115,7 +115,7 @@ def install_dialog(silently=False):
                 config_file_name="~/.pyiron",
                 zip_file="resources.tar.gz",
                 resource_directory="~/pyiron/resources",
-                giturl_for_zip_file="https://github.com/pyiron/pyiron-resources/releases/download/0.0.18/resources-0.0.18.tar.gz",
+                giturl_for_zip_file="https://github.com/pyiron/pyiron-resources/releases/download/latest/resources-0.0.18.tar.gz",
                 git_folder_name="resources",
             )
         else:
--- a/pyiron_base/state/install.py
+++ b/pyiron_base/state/install.py
@@ -99,7 +99,7 @@ def install_dialog(input_by_user=True):
         user_input = None
     else:
         user_input = "yes"
-    
+
     if "PYIRONCONFIG" in os.environ.keys():
         config_file = os.environ["PYIRONCONFIG"]
     else:
--- a/pyiron_base/state/install.py
+++ b/pyiron_base/state/install.py
@@ -98,7 +98,8 @@ def install_dialog(input_by_user=True):
     if input_by_user:
         user_input = None
     else:
-        user_input = 'yes'
+        user_input = "yes"
+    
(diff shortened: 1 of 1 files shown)


pyiron_base/state/install.py
```diff
@@ -111,7 +115,7 @@ def install_dialog():
                 config_file_name="~/.pyiron",
                 zip_file="resources.tar.gz",
                 resource_directory="~/pyiron/resources",
-                giturl_for_zip_file="https://github.com/pyiron/pyiron-resources/releases/download/0.0.17/resources-0.0.17.tar.gz",
+                giturl_for_zip_file="https://github.com/pyiron/pyiron-resources/releases/download/0.0.18/resources-0.0.18.tar.gz",
```
pmrv:
```suggestion
                giturl_for_zip_file="https://github.com/pyiron/pyiron-resources/releases/download/latest/resources-0.0.18.tar.gz",
```
I just saw [here](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases) that we can link to the latest release, I guess that makes sense by default.

pyiron_base/state/install.py
```diff
@@ -94,8 +94,12 @@ def _write_config_file(
             os.makedirs(project_path)
 
 
-def install_dialog():
-    user_input = None
+def install_dialog(input_by_user=True):
+    if input_by_user:
+        user_input = None
+    else:
+        user_input = "yes"
```
pmrv:
```suggestion
def install_dialog(silently=False):
    if not silently:
        user_input = None
    else:
        user_input = "yes"
```

I think silently is slightly more consistent with e.g. `Project.remove_jobs()`.

pyiron_base/state/install.py
pmrv: Actually I think it's better to not pass any arguments here and just use the defaults defined in the `install_pyiron` function.

pmrv (approved):
Added some suggestions but I think the idea to have a non-interactive version is nice.

max-hassani:
@pmrv, using the `latest` key in  the URL [https://github.com/pyiron/pyiron-resources/releases/download/latest](https://github.com/pyiron/pyiron-resources/releases/download/latest/resources-0.0.18.tar.gz) I noticed that it does not exist at all. I only found https://github.com/pyiron/pyiron-resources/releases/latest/ working, but adding the tag does not work. 
Therefore, I reverted the changes and defined a variable `_resource_latest_tag` that requires us to only change it once in this variable.

Testing these changes, I noticed that damask and templates directories are not included in the releases of pyiron-resources. Therefore, I created the PR https://github.com/pyiron/pyiron-resources/pull/53 to fix the issue. After making a new release of the resources, I will change the tag and hopefully merge this PR.

Issue #726, opened 2022-03-14
Implement HasStorage for Executable


pmrv:
We have a few places now where information about the executable needs to survive reloading of the job e.g. whether it's an [MPI executable](https://github.com/pyiron/pyiron_base/pull/713) or what the [accepted return codes](https://github.com/pyiron/pyiron_base/pull/537) are (currently broken because of this).

In the pyiron meeting we discussed to solve this by 
- [x] deriving `Executable` from `HasStorage`, then
- [x] storing everything in the `DataContainer`
- [ ] saving and restoring the `Executable` when saving and restoring `GenericJob`

Leimeroth:
As discussed in there I tested #728 with my remote Setup for a simple VASP job. Back then everything seemed to work, but know I got an error on the cluster because the path to the atomicrex executable was a hardcoded version to the path on my desktop PC
after switching the branch on the cluster from #702 to master.

Right now I am not sure anymore if I changed the branch on both our cluster and my PC or just on the PC when I tested it, so maybe this caused the problem to only show up now. For now I switched back to #702 on the cluster and everything seems to work again, I just fear the point at which incompatibilities between branches break my system.

niklassiemer:
As I understood in #784, this is fixed now. If there are remaining problems - please reopen.

PR #713, opened 2022-03-01
fix mpi support for script job

jan-janssen:

PR #728, opened 2022-03-15
Derive Executable from HasStorage

pmrv:
Implements #726; progress tracked there.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1236,7 +1236,7 @@ class GenericJob(JobCore):
         try:
             self.executable.from_hdf(self._hdf5)
         except ValueError:
-            pass # older versions of pyiron did not save executable information
+            pass  # older versions of pyiron did not save executable information
         with self._hdf5.open("input") as hdf_input:
             if "generic_dict" in hdf_input.list_nodes():
                 generic_dict = hdf_input["generic_dict"]
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1193,6 +1193,7 @@ class GenericJob(JobCore):
         if self._import_directory is not None:
             self._hdf5["import_directory"] = self._import_directory
         self._server.to_hdf(self._hdf5)
+        self.executable.to_hdf(self._hdf5)
         with self._hdf5.open("input") as hdf_input:
             generic_dict = {
                 "restart_file_list": self._restart_file_list,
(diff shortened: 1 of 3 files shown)


pmrv:
I'm a bit surprised that the CI works, because the tests fail for me locally.  Can someone double check this?  Otherwise this should fix #726.

niklassiemer:
I just ran the unittests for this PR locally and it ran through :)

pmrv:
@jan-janssen Can you also check whether this resolves #713?

jan-janssen:
> @jan-janssen Can you also check whether this resolves #713?

I guess the point where this fails is if we use a remote cluster which uses different file paths. To address this previously in the database we only stored the version and whether or not and MPIE enabled executable is needed (this is stored as the number of cores is stored 1 = serial, >1 = parallel).

pmrv:
I think as long as you don't explicitly set the executable path, this should not be an issue, since `_executable_select` still builds it from the `executable_lst` and `version`, but I have no way to check.

pmrv:
I guess it could happen that `executable_lst` is written on the submitting host and then read on the running, remote, host.  In that case we just don't want to store it in HDF.

pmrv:
> I think as long as you don't explicitly set the executable path, this should not be an issue, since `_executable_select` still builds it from the `executable_lst` and `version`, but I have no way to check.

Did you have time to think on this @jan-janssen / @Leimeroth ? Otherwise I'll go ahead and merge and we can sort issues out down the line.

Leimeroth:
did you mean to tag @niklassiemer maybe? Or should I check something specifically related to my setup?

pmrv:
> did you mean to tag @niklassiemer maybe? Or should I check something specifically related to my setup?

Yes, since you are actually using remote jobs, I was hoping you could give it a go with this PR.

Leimeroth:
> > did you mean to tag @niklassiemer maybe? Or should I check something specifically related to my setup?
> 
> Yes, since you are actually using remote jobs, I was hoping you could give it a go with this PR.

I tested if the executable version is recovered for a simple vasp job after fetching it back and reload it using pr.load(). The correct version was used on the cluster and recovered correctly from hdf.
Is there something more specific that I should test @pmrv ?

pmrv:
@jan-janssen concern was that the file path might be messed up, but if you could submit and load a finished job, I suppose that works. Thanks!

I'll merge this later today then.

niklassiemer:
Anything holding this back?

pmrv:
> Anything holding this back?

Nah, just forgot about it.

pyiron_base/job/generic.py
```diff
@@ -1626,12 +1631,7 @@ def _type_from_hdf(self):
         Internal helper function to load type and version from HDF5 file root
         """
         self.__obj_type__ = self._hdf5["TYPE"]
-        if self._executable:
-            try:
-                self.executable.version = self._hdf5["VERSION"]
-            except ValueError:
-                self.executable.executable_path = self._hdf5["VERSION"]
-        else:
```
niklassiemer, line 1634: @jan-janssen this change might be related with not loading the right version?!

PR #730, opened 2022-03-22
Allow inverted matches in job queries

pmrv:
I sometimes want to get jobs that are not matching a certain thing, like all not finished jobs, so with this you can do e.g. `job_table(status='!finished')`.  Works also with the other wildcards.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -89,6 +89,12 @@ class IsDatabase(ABC):
         """
         Get a job table in a project based on matching values from any column in the project database
 
+        The values in `kwargs` can be wildcards, with the following special charaters:
+            - !value matches in the inverse of value
+            - *value matches anything that ends in value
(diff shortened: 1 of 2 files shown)


sudarsan-surendralal (commented):
This looks good from a quick glance. Can you add tests for the inverted matching?

pmrv:
> This looks good from a quick glance. Can you add tests for the inverted matching?

Did we have tests for the wildcards already somewhere?  I didn't find them grepping, but if not I can add those too.

sudarsan-surendralal:
> > This looks good from a quick glance. Can you add tests for the inverted matching?
> 
> Did we have tests for the wildcards already somewhere? I didn't find them grepping, but if not I can add those too.

We have a few here:

https://github.com/pyiron/pyiron_base/blob/2ff89b4a960812e40b03bc08cd1ed253745017ba/tests/project/test_project.py#L97

PR #731, opened 2022-03-24
Add meta DataContainer to GenericJob

pmrv:
I sometimes want to tag jobs with certain extra information, e.g. want kind of atomic structure is used, where it is derived from etc. Right now I usually encode this in subprojects or the job name, but it can get convoluted and error-prone to retrieve it again (due to converting to and from strings), so I'd like a cleaner solution.  

In this PR I've just added an extra DataContainer to GenericJob that saves anything a user might put, but otherwise does not touch it.  There's a bit of danger that people start putting information there that is currently not tracked in pyiron, but really should be, e.g. something like [this](https://github.com/pyiron/pyiron_atomistics/pull/547).  I personally feel it's still a nice thing to have, but we should probably discuss this in the next pyiron meeting.  Any thoughts already?

I'll add tests and such after our discussions.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -26,6 +26,7 @@ from pyiron_base.job.util import (
 from pyiron_base.generic.util import static_isinstance, deprecate
 from pyiron_base.server.generic import Server
 from pyiron_base.database.filetable import FileTable
+from pyiron_base import DataContainer
 import subprocess
 import warnings
 
@@ -179,6 +180,7 @@ class GenericJob(JobCore):
         self._python_only_job = False
         self.interactive_cache = None
         self.error = GenericError(job=self)
+        self._meta = DataContainer(table_name="meta")
 
         for sig in intercepted_signals:
             signal.signal(sig, self.signal_intercept)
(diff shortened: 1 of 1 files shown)


pyiron_base/job/generic.py
```diff
@@ -1232,6 +1247,10 @@ def from_hdf(self, hdf=None, group_name=None):
         if "import_directory" in self._hdf5.list_nodes():
             self._import_directory = self._hdf5["import_directory"]
         self._server.from_hdf(self._hdf5)
+        try:
+            self._meta.from_hdf(self._hdf5)
+        except ValueError:
+            pass # no meta data written for this job previously
```
jan-janssen, line 1253: Can this be more elegant by checking if the corresponding group exists in the HDF5 file

pmrv:
Rename `job.meta` to `job.data` or `project.data` to `project.meta`.  Conflict with `job['user/foo']`?

liamhuber:
@pmrv after recent meetings, shall we wait until `GenericJob` looks more like `TemplateJob`, then simply add a `self.storage._meta = DataContainer(table_name='meta')` and accompanying read-only-access property? I like this idea, but I think that's an easier time to make the change. We could even directly do it over in #758 as I feel it's very on-topic for that PR.

pmrv:
> @pmrv after recent meetings, shall we wait until `GenericJob` looks more like `TemplateJob`, then simply add a `self.storage._meta = DataContainer(table_name='meta')` and accompanying read-only-access property? I like this idea, but I think that's an easier time to make the change. We could even directly do it over in #758 as I feel it's very on-topic for that PR.

Yup, that sounds like a good idea.  I'm not urgently depending on this change anyway.

PR #734, opened 2022-03-28
Check covergence for parallel masters

sudarsan-surendralal:
To solve issues like: https://github.com/pyiron/pyiron_atomistics/issues/528


pmrv (approved):
lgtm

PR #743, opened 2022-04-19
Add clearer error message in DataContainer.from_hdf

pmrv:
When trying to read a DataContainer from a HDf group where none was
stored before (potentially due to an unrelated coding error), this would
usually fail with a cryptic error from normalize_key.  This patch
intercepts that error and re-raises a new one with a clearer error
message.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -815,8 +815,10 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
                     k, i = name.split("__index_", maxsplit=1)
                     i = int(i)
                 except ValueError:
-                    raise ValueError(f"Could not parse item name {name} in HDF group {hdf.h5_path}. "
-                                     "Was this group really written by a DataContainer?") from None
+                    raise ValueError(
+                        f"Could not parse item name {name} in HDF group {hdf.h5_path}. "
+                        "Was this group really written by a DataContainer?"
+                    ) from None
                 if k == "":
                     return i, i
                 else:
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -811,8 +811,12 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
             def normalize_key(name):
                 # split a dataset/group name into the position in the list and
                 # the key
-                k, i = name.split("__index_", maxsplit=1)
-                i = int(i)
+                try:
+                    k, i = name.split("__index_", maxsplit=1)
+                    i = int(i)
+                except ValueError:
+                    raise ValueError(f"Could not parse item name {name} in HDF group {hdf.h5_path}. "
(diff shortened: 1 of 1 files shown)


pyiron_base/generic/datacontainer.py
```diff
@@ -811,8 +811,14 @@ def _from_hdf(self, hdf, version=None):
             def normalize_key(name):
                 # split a dataset/group name into the position in the list and
                 # the key
-                k, i = name.split("__index_", maxsplit=1)
-                i = int(i)
+                try:
+                    k, i = name.split("__index_", maxsplit=1)
+                    i = int(i)
+                except ValueError:
```
jan-janssen, line 817: Do we need the `try except` or can we split into a list and then check the size of the list?

pmrv: We can do that, but we also need to check whether `i` can be parsed as an `int`. Personally I'd think at that point checking explicitly is more convoluted.  In the common case where no exception is thrown `try`/`expect` as supposedly faster also, though I think it hardly matters here.  Anyways, I don't feel strongly about this and can change it.

pyiron_base/generic/datacontainer.py
```diff
@@ -811,8 +811,14 @@ def _from_hdf(self, hdf, version=None):
...
+                    raise ValueError(
+                        f"Could not parse item name {name} in HDF group {hdf.h5_path}. "
+                        "Was this group really written by a DataContainer?"
```
samwaseda, line 820: An error practically asking the user where the error comes from sounds super weird to me. I'd rather say that this element cannot have been written by `DataContainer` because it doesn't have the format "NAME__index_NUMBER".

This being said, can't we have an option to make all elements `DataContainer`-compatible? I mean, the sounds like a horrible hack that shouldn't happen by default, but it might make sense when we want to convert older stuff so having such a function might be useful (although I don't have a strong feeling for this so you can also ignore me on this point)

pmrv: > An error practically asking the user where the error comes from sounds super weird to me. I'd rather say that this element cannot have been written by `DataContainer` because it doesn't have the format "NAME__index_NUMBER".

Fair, I'll reword it.  The motivation was that this error mostly comes up for people using `DataContainer` in new job classes and make small mistakes and which point they are greeted by a weird `ValueError` from deep within their `from_hdf` without any indication what is the root cause.  This happened a few times now and while I can directly see the issue, it's difficult without more experience, so I want to transfer some of that knowledge in this better error message.

> This being said, can't we have an option to make all elements `DataContainer`-compatible? I mean, the sounds like a horrible hack that shouldn't happen by default, but it might make sense when we want to convert older stuff so having such a function might be useful (although I don't have a strong feeling for this so you can also ignore me on this point)

I can see that being useful for converting old classes, but would prefer that to be outside `DataContainer.from_hdf` either in a classmethod constructor or a separate function.  Implementation should be pretty trivial though.

PR #753, opened 2022-05-20
Implement Marvins input field class

samwaseda:
Since @pmrv wrote this super nice class, I wanted to have it as a part of the pyiron standard class. I was wondering whether it would be acceptable to have it directly in `HasStorage` but I decided to make a separate file. Let me know what you think

Tests will follow

### DIFF
--- a/pyiron_base/generic/genericinput.py
+++ b/pyiron_base/generic/genericinput.py
@@ -104,6 +104,18 @@ class InputField:
         fset=lambda x: x,
         default=None,
     ):
+        """
+        Args:
+            name (str): Name of the attribute
+            doc (str): Docstring of the attribute
+            data_type (None/type/list/tuple): Data type. If set, setter function
+                checks the input. If the check should not take place, set `None`.
+                If several types should be specified, use `list` or `tuple` to
+                set all types.
+            fget (function): Getter function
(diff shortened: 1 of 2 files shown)


samwaseda:
I guess the direction is not so bad, but I have the feeling that it has to be `DataContainer` in the end. Anyway, I make it draft for now.

pmrv:
What's the issue with what you have written so far?  It looks quite nice, I think.

samwaseda:
My biggest problem is the fact that the docstring of the attribute itself doesn't work, i.e. "job.input.temperature" gives the docstring of `int` and not what I write in the docstring.

Another problem is it starts becoming difficult to check the input more flexibly, i.e. for example in the case of temperature, I want to check not only the type, but also whether the value is positive. We could add an argument like `conditions=[lambda x: return x > 0]`, but this doesn't allow to check everything (especially it depends on the other inputs). In the worst case, we would have some input parameters as attributes, others as properties, in which case some of them appear in the class docstring and `__repr__` and not in the variable docstring, in the other case exactly other way around. Of course, we can do some checks inside `validate_ready_to_run`, but I'm not really sure how elegant that is...

And finally there was a case where I thought we'd need `DataContainer` in the end but I can't remember where.

PR #758, opened 2022-05-23
Better handling of custom IO for template job

liamhuber:
Following up on a nice discussion of @yuyuanjingxuan's new solvation class at today's meeting. 

To make it easier for new developers to customize IO on children of `TemplateJob`, we now give a friendlier error message if the `input` or `output` fields are (attempted to be) overwritten. New properties allow setting input and output fields with custom children of `DataContainer`.

### DIFF
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -31,7 +31,7 @@ class TemplateJob(GenericJob, HasStorage):
 
     @property
     def _input_class(self) -> Type[DataContainer]:
-        """Children can overwrite this method to return some other child of DataContainer for custom behaviour"""
+        """Children can overwrite this method to return some other child of `DataContainer` for custom behaviour"""
         return DataContainer
 
     @property
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -36,7 +36,7 @@ class TemplateJob(GenericJob, HasStorage):
 
     @property
     def _output_class(self) -> Type[DataContainer]:
-        """Children can overwrite this method to return some other child of DataContainer for custom behaviour"""
+        """Children can overwrite this method to return some other child of `DataContainer` for custom behaviour"""
         return DataContainer
 
     @property
--- a/pyiron_base/job/template.py
+++ b/pyiron_base/job/template.py
@@ -46,8 +46,21 @@ class TemplateJob(GenericJob, HasStorage):
     @input.setter
     def input(self, new_input):
         raise AttributeError(
-            "Input cannot be overwritten; to get custom output behaviour, override the `_input_class` property to "
-            "return your own custom class (which inherits from `DataContainer`)."
+            """
+            Input cannot be overwritten; to get custom input behaviour, override the `_input_class` property to return 
+            your own custom class (which inherits from `DataContainer`).
+            
+            Example:
+            >>> class MyInput(DataContainer):
+            >>>     def __init__(self, init=None, table_name=None, lazy=False, wrap_blacklist=()):
+            >>>         super().__init__(init=init, table_name=table_name, lazy=lazy, wrap_blacklist=wrap_blacklist)
+            >>>         self.foo = 42
+            >>>
+            >>> class MyJob(TemplateJob):
+            >>>     @property
+            >>>     def _input_class(self) -> MyInput:
+            >>>         return MyInput
+            """
         )
 
     @property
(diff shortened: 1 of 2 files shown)


pyiron_base/job/template.py
```diff
@@ -24,17 +26,41 @@ class TemplateJob(GenericJob, HasStorage):
...
         HasStorage.__init__(self)
-        self.storage.create_group("input")
-        self.storage.create_group("output")
+        self.storage.input = self._input_class(table_name='input')
+        self.storage.output = self._output_class(table_name='output')
 
     @property
-    def input(self):
+    def _input_class(self) -> Type[DataContainer]:
+        """Children can overwrite this method to return some other child of DataContainer for custom behaviour"""
```
samwaseda:
```suggestion
        """Children can overwrite this method to return some other child of `DataContainer` for custom behaviour"""
```

pyiron_base/job/template.py
```diff
@@ -24,17 +26,41 @@ class TemplateJob(GenericJob, HasStorage):
...
+        return DataContainer
+
+    @property
+    def _output_class(self) -> Type[DataContainer]:
+        """Children can overwrite this method to return some other child of DataContainer for custom behaviour"""
```
samwaseda:
```suggestion
        """Children can overwrite this method to return some other child of `DataContainer` for custom behaviour"""
```

pyiron_base/job/template.py
```diff
@@ -24,17 +26,41 @@ class TemplateJob(GenericJob, HasStorage):
...
         GenericJob.__init__(self, project, job_name)
         HasStorage.__init__(self)
-        self.storage.create_group("input")
-        self.storage.create_group("output")
+        self.storage.input = self._input_class(table_name='input')
+        self.storage.output = self._output_class(table_name='output')
 
     @property
-    def input(self):
+    def _input_class(self) -> Type[DataContainer]:
```
samwaseda, line 33: Don't we actually need also a setter?

```python
@_input_class.setter
def _input_class(self, new_input):
    if not isinstance(new_input, DataContainer):
        raise ValueError('input must inherit from `DataContainer`')
    self.storage.input = new_input
```

Something like this? If what you wrote does the same thing, just ignore this comment :D

pyiron_base/job/template.py
```diff
@@ -24,17 +26,41 @@ class TemplateJob(GenericJob, HasStorage):
...
+
+    @property
+    def input(self) -> Type[DataContainer]:
         return self.storage.input
 
+    @input.setter
+    def input(self, new_input):
+        raise AttributeError(
+            "Input cannot be overwritten; to get custom output behaviour, override the `_input_class` property to "
+            "return your own custom class (which inherits from `DataContainer`)."
```
samwaseda: Nice! And I guess I'd even give a simple example:

```python
Example:

>>> class MyInput(DataContainer):
>>>     def __init__(self, init=None, table_name=None, lazy=False, wrap_blacklist=()):
>>>         super().__init__(init=init, table_name=table_name, lazy=lazy, wrap_blacklist=wrap_blacklist)
>>>         self.foo = 5
>>>
>>>
>>> job.input._input_class = MyInput()
```

We can maybe also mention `from pyiron_base.generic.datacontainer import DataContainer`

samwaseda (commented):
This one is a really good example of what pyiron should do more and more in the future - it is true that we need more documentation on the one hand, but on the other hand it's important that new developers can simply give it a try and pyiron throws errors on the right places with the right messages (especially also when a new class is developed).

pyiron_base/job/template.py
liamhuber: Actually I think this doesn't work, I think it really needs to be a property defined in the new job's source code. If it's just a property that can be set by the user I don't think it will survive (de)serialization, and will always just revert back to being a standard `DataContainer`

pyiron_base/job/template.py
liamhuber: An example is a stellar idea, but I think we'll need to rework it based on my serialization comment. Adding an example leans into your high-level comment about really helping developers, which I love!

pyiron_base/job/template.py
samwaseda: ah ok then I probably somehow misunderstood the idea

pyiron_base/job/template.py
samwaseda: aaaah nice now with the example I actually understood how it really works XDXDXDXDXD

samwaseda (approved):
Brilliant! I like it a lot

yuyuanjingxuan (approved):

pmrv:
I think it's a cool trick to allow overriding input/output classes, but I might relax the requirement for them to derive from `DataContainer`.  At least for output I'd think we would want to have e.g. plotting methods on them, which wouldn't play nice again with the recursive nature of the `DataContainer`.  Or would we want this on the job itself?  In that case this looks good.

liamhuber:
I'd put a separate plot field right on the job and just have it reference the output, although I'm not married to that setup.

Does the recursion really get angry if you add custom methods to a DataContainer child? I guess I haven't played around with it enough if this is the case! I do think that either the type hint or the docstring needs to ensure that whatever class you put there has to/from hdf methods so that the storage data container doesn't get any headaches when it tries to serialize its contents.

PR #763, opened 2022-06-01
Do not initialize Executable when none saved in HDF

pmrv:
The previous code first try'd to load saved executable information via

self.executable.from_hdf()

and then bailed if there was an error.  The problem is that just
accessing `self.executable` instantiates an `Executable` object on the
job even if it is a job, that does not need it, e.g. because it does not
require an external executable.  Due to the error this left the object
in an inconsistent state, unable to save itself again with `to_hdf()`.

Imo this shows that ideally we'd have at least one more `GenericJob` base class, that does not presume the need for an external executable, but that's for the future.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1233,10 +1233,8 @@ class GenericJob(JobCore):
         if "import_directory" in self._hdf5.list_nodes():
             self._import_directory = self._hdf5["import_directory"]
         self._server.from_hdf(self._hdf5)
-        try:
+        if 'executable' in self._hdf5.list_groups():
             self.executable.from_hdf(self._hdf5)
-        except ValueError:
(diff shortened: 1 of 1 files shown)


jan-janssen:
So these new changes basically add backwards compatibility to the previous pull request https://github.com/pyiron/pyiron_base/pull/728 ? I guess we should see if there is a way to spot this kind of issues before we release a new version in future.

pmrv:
> So these new changes basically add backwards compatibility to the previous pull request #728 ? I guess we should see if there is a way to spot this kind of issues before we release a new version in future.

Yeah, I guess the way to do it would be to be more rigorous with the backwards compatibility testing.  Right now it's very brittle and a bit of hassle to setup, hence we have basically no tests in that workflow.

PR #771, opened 2022-06-13
Perform `to_buildin` when `lazy` before `copy`

samwaseda:
Otherwise it won't copied properly.


pmrv:
Calling `to_builtin` is a good work-around, but I'll add a proper method to force evaluate it.

samwaseda:
> Calling `to_builtin` is a good work-around, but I'll add a proper method to force evaluate it.

ok then I'll just keep this PR open until you implement it

pmrv:
With #772 merged, I think this is not necessary anymore.  Reopen if you feel otherwise.

PR #772, opened 2022-06-15
Add DataContainer.force method load all HDFStubs

pmrv:
I haven't understood yet why the recursion error comes up, but I added a test for it and a method to load all data from HDF before a copy happens.

### DIFF
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -782,6 +782,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
         >>> all(a is not b for a, b in zip(pl.copy().values(), pl.values()))
         True
         """
+        self.force()
         return copy.deepcopy(self)
 
     def _get_hdf_group_name(self):
--- a/pyiron_base/generic/datacontainer.py
+++ b/pyiron_base/generic/datacontainer.py
@@ -782,7 +782,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
         >>> all(a is not b for a, b in zip(pl.copy().values(), pl.values()))
         True
         """
-        self.force()
+        self._force_load()
         return copy.deepcopy(self)
 
     def _get_hdf_group_name(self):
@@ -906,7 +906,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
         """
         write(self.to_builtin(), file_name)
 
-    def force(self, recursive=True):
+    def _force_load(self, recursive=True):
         """
         Load all HDFStubs present in the data container.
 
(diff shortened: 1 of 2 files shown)


samwaseda:
I find both the function name and how it handles loading not explicit enough. I presume that the fact that it goes through all elements via `for v in self.values()` loads the contents behind the scenes, right? I will probably have a hard time understanding how this function works if I look at it later. I don't really have a good suggestion, but at least you can maybe add one line comment in which moment the content is actually loaded. And I guess it can also be a private function, right? Maybe we can call it something like `_force_load`?

pmrv:
I'll rename it and add a comment, but your understanding is correct.  The loading of the stubs happens, when items are accessing via `__getitem__`, which is what `values` does.

I wanted to have it as a public method actually, because I thought it might also be useful in other situations.

samwaseda:
> The loading of the stubs happens, when items are accessing via `__getitem__`, which is what `values` does.

yeah that's my problem - it's not explicit enough. With a comment in the code I'd say it's ok

> I wanted to have it as a public method actually, because I thought it might also be useful in other situations.

Hm not sure if that really happens. Besides I think by adding more and more public functions we start killing the autocomplete feature.

pyiron_base/generic/datacontainer.py
```diff
@@ -905,10 +906,23 @@ def write(self, file_name):
...
         write(self.to_builtin(), file_name)
 
+    def _force_load(self, recursive=True):
+        """
+        Load all HDFStubs present in the data container.
+
+        Args:
+            recursive (bool): force also nested data containers, default True
+        """
+        if not self._lazy: return
```
niklassiemer: Is it guaranteed, that a `DataContainer` containing other containers of which only one is lazy loaded is also set to lazy?

pyiron_base/generic/datacontainer.py
```diff
@@ -905,10 +906,23 @@ def write(self, file_name):
...
+
+        # values are loaded from HDF once they are accessed via __getitem__, which is implicetly called by values()
+        for v in self.values():
+            if recursive and isinstance(v, DataContainer):
```
niklassiemer, line 921: Do we have other things which are lazy? If yes, they should also be `_force_load()`ed.

pmrv: So far only `DataContainers` (and subclasses) do this, so should be ok.

pyiron_base/generic/datacontainer.py
pmrv: In the current setup I would think yes, but you could manually construct a counter example, so might be better to take it out again.

niklassiemer: Ok, we could do
`if not (self._lazy or recursive): return` to spare some iterations if we are neither lazy nor recursive, however, that is probably only an edge case.

PR #774, opened 2022-06-20
Make _system_function_lst a class attr

pmrv:
This will allow future PR in downstream modules to modify it, instead of
inheriting from TableJob just to modify it.


niklassiemer (approved):

jan-janssen (approved):
Looks good to me

PR #786, opened 2022-07-06
Remove the internal _run_manually method

niklassiemer:
I did not find a usage of that function and it is actually doing the same as `run_if_manually`.

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -1229,15 +1229,6 @@ class GenericJob(JobCore):
         if not (self.status.finished or self.status.suspended):
             self.status.aborted = True
 
-    def _run_manually(self, _manually_print=True):
-        """
-        Internal helper function to run a job manually.
-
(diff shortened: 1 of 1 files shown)


jan-janssen:
What about https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/runfunction.py#L95 ?

jan-janssen (changes requested):
We still need this function, even if it is not included in the tests. But I agree that we should add a test to prevent this kind of confusion in future.

niklassiemer:
The line you mentioned calls run_if_manually not the internal _run_manually function. Pycharm told me there is no usage of the second one?

jan-janssen (approved):

pmrv:
I think one source of the confusion is that the function in the `runfunction` module call methods on the job object even though the methods just call back into the original module.  That would actually be a nice refactor for our tutorial workshop.

PR #787, opened 2022-07-06
Raise an error if `run` is called for an aborted job

### LINKED ISSUE #770: run() does not raise an error when called on an aborted job.
Instead it just silently returns.

niklassiemer:
closes #770 
If we want to raise an error in this case. We could also just warn?

### DIFF
--- a/pyiron_base/job/generic.py
+++ b/pyiron_base/job/generic.py
@@ -742,7 +742,9 @@ class GenericJob(JobCore):
                     run_again=run_again,
                 )
             elif status == "aborted":
-                raise ValueError("Running an aborted job with `delete_existing_job=False` is meaningless.")
+                raise ValueError(
+                    "Running an aborted job with `delete_existing_job=False` is meaningless."
+                )
         except Exception:
             self.drop_status_to_aborted()
             raise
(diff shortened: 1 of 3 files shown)


niklassiemer:
I already like to raise an Error:
```
ERROR: test_script_path (job.test_script.TestScriptJob)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/job/test_script.py", line 42, in test_script_path
    self.job.run()
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/generic/util.py", line 213, in decorated
    return function(*args, **kwargs)
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/job/generic.py", line 745, in run
    raise ValueError("Running an aborted job with `delete_existing_job=False` is meaningless.")
ValueError: Running an aborted job with `delete_existing_job=False` is meaningless.
```
is caused by the last `run` in 
https://github.com/pyiron/pyiron_base/blob/d4f57c36affa9d69efc9ea121f582f3da247bc1c/tests/job/test_script.py#L35-L42
which silently returned `None` without performing any test...

niklassiemer:
ping @pmrv

pmrv (approved):
lgtm

PR #792, opened 2022-07-18
Check that server.core is given as an integer

pmrv:


niklassiemer (approved):
Holds even more than it promises :)

pyiron_base/server/generic.py
```diff
@@ -246,6 +247,12 @@ def cores(self, new_cores):
         Args:
             new_cores (int): number of cores
         """
+        if not isinstance(new_cores, numbers.Integral):
```
samwaseda, line 250: What do we need this line for?

niklassiemer: to check if the given `new_cores` is an integer-typed thing. If it is (i.e. `int`, `np.int64`, `np.int32`,...) we just use the value as the number of cores. If `new_cores=2.0` we would cast it to an `int` and accept. However, if it was `new_cores=2.1` we would raise a `ValueError`.

samwaseda: I meant the very line I commented on. I still don’t see what we need that line for

samwaseda:
Wait the discussion is still going on. Merging it now for me violates basic rules.

pyiron_base/server/generic.py
pmrv: If it's already an integer, no need to try a conversion.  We could remove the line and add a check in case the type casting fails.

samwaseda: Ok then I would have done

```python
if new_cores != int(new_coes):
    raise ValueError(f"cores must be an integer number, not {new_cores}!")
new_cores = int(new_cores)
```

This would have spared two lines. But I'm really super unhappy with the fact that it was merged while it was unresolved, especially my comment was pertinent.

Now you don't need to open another PR just to change this.

pmrv: Yup, sorry for that.  I did not think it was a critical issue and wanted to have the fix quickly available.

PR #794, opened 2022-07-18
Fix job type register

niklassiemer:


niklassiemer:
@jan-janssen This should fix the [issue in pyiron](https://github.com/pyiron/pyiron/pull/1425). The checking for auto-added jobs is not causing failures any more. Furthermore, I check for `pyiron_atomistics` registered job types if one for any `pyiron` should be added.
In addition, I added an `overwrite` option - I actually do not like that, however, in the `TableJob` of atomistics I need to `unregister` + `register` right now. Indeed, a concept like 'extending' a present job type would be nice in that case.

niklassiemer:
I will merge before the release, tomorrow morning, if no one objects.

jan-janssen (approved):
Looks good to me

jan-janssen:
@niklassiemer I am just wondering, why we did not see this before - I thought this should be covered by the integration tests.

niklassiemer:
Me as well. However, pyiron_atomistics does not have any problem. The problem was caused on pyiron itself where we alias pyiron=pyiron atomistics and this is not the same if we compare the module of the job... And pyiron itself is not covered by integration tests.

PR #795, opened 2022-07-19
Hotfix for pack/unpack: Add flag to include all files in jobs

pmrv:
Previously `pr.pack` only keeps the HDF5 files and not the files inside the job directory.  I've added a quick hack to allow exporting of all files.  This has the side effect of including all files in a project directory not just job related ones and is a bit sluggish.  However I'll need this functionality quickly and am unwilling to spend a great deal of effort to patch this in, when we are going to rewrite this functionality anyway.

### DIFF
--- a/pyiron_base/archiving/export_archive.py
+++ b/pyiron_base/archiving/export_archive.py
@@ -59,7 +59,16 @@ def compress_dir(archive_directory):
     rmtree(archive_directory)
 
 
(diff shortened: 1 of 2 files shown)


niklassiemer (commented):
I am ok with these changes, especially, since we want to replace the implementation of pack/unpack :)

PR #799, opened 2022-07-26
fix dynamic jobs

jan-janssen:
The automatic registration breaks the functionality of dynamic jobs, this is now overwritten with this pull request. @niklassiemer If you have an idea for a more elegant solution, feel free to modify my pull request.

### DIFF
--- a/pyiron_base/jobs/job/jobtype.py
+++ b/pyiron_base/jobs/job/jobtype.py
@@ -226,7 +226,10 @@ class JobFactory(PyironFactory):
         return list(self._job_class_dict.keys()) + list(self._job_dyn_dict.keys())
 
     def __getattr__(self, name):
-        if name in self._job_class_dict.keys():
+        if (
+            name in self._job_class_dict.keys()
+            and name not in self._job_dyn_dict.keys()
+        ):
 
             def wrapper(job_name, delete_existing_job=False, delete_aborted_job=False):
                 """


pmrv:
Can you briefly summarize what the underlying issue is and how this fixes it?

niklassiemer:
Right now, auto-registration of the dynamic jobs adds them to the `JOB_CLASS_DICT` and then loading them fails since they cannot be loaded from the location they have been generated.... 
The point is, they should not be part of the normal `JOB_CLASS_DICT`... For now, I would either merge this, or just change the lookup order to first look into the `JOB_DYN_DICT`. 

We could also use a `ScriptJob` subclass (`SpecificScriptJob`) with just `__init_subclass(cls): pass`, or in general not register `ScriptJob` subclasses, or unregister all jobs in the `JOB_DYN_DICT` to fix this issue.

pmrv:
In that case wouldn't the correct solution be that the automatically generated subclasses also overwrite the autoregistration logic?

niklassiemer:
Yes, in principle I would like that, however, I just tested 
```python
class Some:
    @classmethod
    def __init_subclass__(cls):
        print('subclass!')
class Any(Some):    
    @classmethod
    def __init_subclass__(cls):
        pass
```
which prints 'subclass'. I.e. the `__init_subclass__` of the _parent_ class is called after the class definition. To exclude the auto-registration we then would need something from the class itself inside the __init_subclass__ call in general.

jan-janssen:
> For now, I would either merge this, or just change the lookup order to first look into the `JOB_DYN_DICT`.

In the interest of time, I would also prefer to merge this.

jan-janssen:
I be able to run the workshop I merge this now, we can potentially redo it later.

PR #800, opened 2022-07-26
Do not construct all dynamic classes and unregister directly.

niklassiemer:


jan-janssen:
If we change the order Python by default assume that most jobs are dynamic jobs, which I do not think is what we have now or in future. I am not sure how much the optimization of the order in the if statement matters, but in comparison to https://github.com/pyiron/pyiron_base/pull/799 I prefer that one.

niklassiemer:
This version always has less or equal number of boolean evaluations. The statement `if (a in b and not a in c):` are two such operations plus one `if a in c` operation in case of dynamic. The current version just does `a in c` and `a in b` operations.

niklassiemer:
Beware, I also change the content of the if statement, here!

niklassiemer:
Close in favor of #801

PR #801, opened 2022-07-27
Refactor #799

pmrv:
Swaps around the if clauses to make the flow simpler.


niklassiemer (approved):
I like it better this way.

jan-janssen (approved):
Looks good to me

PR #805, opened 2022-07-28
Do not construct all dynamic classes all the time.

niklassiemer:


niklassiemer:
Draft until we have tests for the dynamic classes

PR #806, opened 2022-07-28
Outsource credential information to own config file

niklassiemer:
Currently, all credential information like the password to access the database are stored right in the `~/.pyiron` config file. However, this is sometimes shared with or readable by others. With increasing need to store other credential information like e.g. a [CoScInE token](https://help.itc.rwth-aachen.de/en/service/b2b7729fd93f4c7080b475776f6b5d87/article/b0d8a8372fc947749f0e7d15035d17f1/) or S3 secret keys, I deem another credentials file to hold sensitive information (and being checked for only-user permissions) is required.

Here, I added a new option for our configuration being `credentials_file`, `CREDENTIALS_FILE`, `PYRIONCREDENTIALSFILE` for the settings dict, settings file, and settings environment variables, respectively. This has to provide the path to the credentials file which is parsed like the normal config file.

### DIFF
--- a/pyiron_base/state/settings.py
+++ b/pyiron_base/state/settings.py
@@ -363,15 +363,8 @@ class Settings(metaclass=Singleton):
 
         if not os.path.isfile(credential_file):
             raise FileNotFoundError(credential_file)
-        parser = ConfigParser(inline_comment_prefixes=(";",), interpolation=None)
-        parser.read(credential_file)
-        for sec_name, section in parser.items():
-            for k, v in section.items():
-                try:
-                    config[self.file_credential_map[k.upper()]] = v
-                except KeyError:
-                    pass
-        return config
+        config = self._parse_config_file(credential_file, self.file_credential_map)
+        return config or {}
 
     def _get_config_from_file(self) -> Union[Dict, None]:
         if "PYIRONCONFIG" in os.environ.keys():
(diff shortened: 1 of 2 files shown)


jan-janssen:
I am against multiple configuration files and I do not really see a reason why the configuration file should be readable to anybody else than just the user himself. Just like your .ssh files.

tnecnivkcots:
If we want to have more security, one option could be to use an encrypted credentials file. I would apreciate it and I am sure @freyso would do it too.

niklassiemer:
Yes, in my opinion, this would be a good thing to do :) However, I do not have experience with encryption. Furthermore, at which time would you lift the encryption? Since entering a password on pyiron startup is not very convenient...

tnecnivkcots:
We could make it very easy with [an cryptography module available in python](https://www.geeksforgeeks.org/encrypt-and-decrypt-files-using-python/). Everytime pyiron is loaded the decrypted password would be stored as a variable in the memory (if a password is set).
I am also not experienced in encrypting, but the above mentioned link explains the usage of the encryption very easy.
But before using this module, we should look up the kind of encryption.

tnecnivkcots:
And we should of course do a testing of encryption.

pmrv:
The issue with encryption is that it won't work without an additional password, which the user then needs to input manually or is saved again in some file.  That's the original problem again.  Querying once per pyiron process and keep the decrypted passwords in memory is also not an option, since 
1. Pyiron processes don't live long enough or rather users generally have a lot of them open, so you'd still have to input your password on every kernel restart
2. Compute nodes need to be able to use pyiron completely non-interactively.
The only thing that could work is a keychain like setup that a user unlocks the first time the connect to the cluster, however I'd guess existing implementations would be difficult to adapt to make them work through jupyter and across nodes and implementing one ourselves is very much unfeasible (in a secure manner).  If we want to continue this discussion, we should first clarify what threat model we are trying to protect against.

I like to the have configuration split in two because it reduces the chance that someone copies their password somewhere public when they paste their configuration (that's actually my main threat model).
We could have it such that you can either have the passwords in the normal config or an option that specifies the password file.

niklassiemer:
> The issue with encryption is that it won't work without an additional password, which the user then needs to input manually or is saved again in some file. That's the original problem again. Querying once per pyiron process and keep the decrypted passwords in memory is also not an option, since
> 
> 1. Pyiron processes don't live long enough or rather users generally have a lot of them open, so you'd still have to input your password on every kernel restart
> 2. Compute nodes need to be able to use pyiron completely non-interactively.
>    The only thing that could work is a keychain like setup that a user unlocks the first time the connect to the cluster, however I'd guess existing implementations would be difficult to adapt to make them work through jupyter and across nodes and implementing one ourselves is very much unfeasible (in a secure manner).  If we want to continue this discussion, we should first clarify what threat model we are trying to protect against.

I agree here. To get that working conveniently, we would need to tie the password to the user login in a secure way... 

> I like to the have configuration split in two because it reduces the chance that someone copies their password somewhere public when they paste their configuration (that's actually my main threat model). We could have it such that you can either have the passwords in the normal config or an option that specifies the password file.

Yes, that is what I implemented 😄

jan-janssen:
Just to make this clear, the data stored in the database is just an index of the HDF5 files. So the information a potential attacker could get from the password is rather limited. So I would like to keep the option to have the password in the primary configuration file, if you want an optional additional configuration file that is fine to be, but adding an additional file should be optional not mandatory.

niklassiemer:
That is currently true. However we will add other  credential information to the configuration. The additional file is currently optional and has to be specified in the normal configuration to be used.

PR #808, opened 2022-07-29
Fix Importlib bug

jan-janssen:
Previously the following code failed
```
import pyiron_base
import importlib
importlib.reload(pyiron_base)
```

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -14,6 +14,7 @@ import pint
 import importlib
 import math
 import numpy as np
+import warnings
 
 from pyiron_base.project.maintenance import Maintenance
 from pyiron_base.project.path import ProjectPath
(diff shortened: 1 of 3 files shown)


pyiron_base/project/generic.py
```diff
@@ -1609,10 +1610,11 @@ def register_tools(cls, name: str, tools):
             tools (Toolkit): The tools to register.
         """
         if hasattr(cls, name):
```
pmrv, line 1612: Let's try rescue raising the error first.  Maybe something like this (untested though(

```suggestion
        if hasattr(cls, name) or str(type(getattr(cls, name))) == str(type(tools)):
```
and then raise the error inside the if again.

pmrv: Thinking again about it the condition would actually be something like
```python
if hasattr(cls, name) and str(type(getattr(cls, name))) != str(type(tools)):
```
Since in the case where the types match, we actually *want* to overwrite the tools set on the project (otherwise the reloaded version of the project will have the old version of the tools).

pmrv:
Another general hint: `importlib.reload` is not recursive, i.e. it will not reload modules that are imported by it.  You'll need to do it manually are use ipythons `%autoreload` magick.

jan-janssen:
I guess we can close this right now, as the use case of `importlib` is rather rare.

PR #809, opened 2022-07-29
Trust given module path and only fall back to internal class

### LINKED ISSUE #803: loading a job with custom job type may return wrong job type
As I discovered in https://github.com/pyiron/pyiron_gui/pull/158 loading of custom defined job type returns the pyiron internal job type if this had been `JobType.register`ed (i.e. is in the `JOB_CLASS_DICT`):
```python
from pyiron_base import Project, PythonTemplateJob
from pyiron_base._tests import ToyJob

class ToyJob(PythonTemplateJob):
        """A toyjob which can be run() to test functionalities."""
        super(ToyJob, self).__init__(project, job_name)
        self.input['input_energy'] = 100

    def run_static(self):
        self.output["energy_tot"] = self.input["input_energy"]
        self.status.finished = True

pr = Project('some_tmp')
job = pr.create_job(ToyJob, 'name')
job.run()
print(job["TYPE"])
job=pr['name']
print(job['TYPE'])
>> The job name was saved and received the ID: 86
>> <class '__main__.ToyJob'>
>> <class 'pyiron_base._tests.ToyJob'>
```

The conversion happens in 
https://github.com/pyiron/pyiron_base/blob/69fd55229647bf36c4db78555b62d5635f6f3562/pyiron_base/storage/hdfio.py#L1300-L1320

For me this was absolutely unexpected! If I did a calculation using a specific class I want the _same_ class back if I load the job, regardless of which jobs are registered.

I suggest to rewrite the above function to
```python
   def import_class(self, class_name):
        """
        Import given class from fully qualified name and return class object.
        Args:
            class_name (str): fully qualified name of a pyiron class
        Returns:
            type: class object of the given name
        """
        internal_class_name = class_name.split(".")[-1][:-2]
        class_path = class_name.split()[-1].split(".")[:-1]
        class_path[0] = class_path[0][1:]
        module_path = ".".join(class_path)        
        try:
            return getattr(
                importlib.import_module(module_path),
                internal_class_name,
            )
        except ImportError:    
            if internal_class_name in self._project.job_type.job_class_dict:
                module_path = self._project.job_type.job_class_dict[internal_class_name]
                return getattr(
                    importlib.import_module(module_path),
                    internal_class_name,
                )
            else:
                 raise 

```

which changes the behavior to _trust_ the module path we have written to hdf before and fall back to use the class registered as `JobType` if this is not possible.

niklassiemer:
closes #803 
I implemented my suggestion from #803

### DIFF
--- a/pyiron_base/storage/hdfio.py
+++ b/pyiron_base/storage/hdfio.py
@@ -1308,16 +1308,23 @@ class ProjectHDFio(FileHDFio):
             type: class object of the given name
         """
         internal_class_name = class_name.split(".")[-1][:-2]
-        if internal_class_name in self._project.job_type.job_class_dict:
-            module_path = self._project.job_type.job_class_dict[internal_class_name]
-        else:
-            class_path = class_name.split()[-1].split(".")[:-1]
-            class_path[0] = class_path[0][1:]
-            module_path = ".".join(class_path)
-        return getattr(
-            importlib.import_module(module_path),
-            internal_class_name,
-        )
+        class_path = class_name.split()[-1].split(".")[:-1]
+        class_path[0] = class_path[0][1:]
+        module_path = ".".join(class_path)
+        try:
+            return getattr(
+                importlib.import_module(module_path),
+                internal_class_name,
+            )
+        except ImportError:
+            if internal_class_name in self._project.job_type.job_class_dict:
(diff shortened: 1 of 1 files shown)


niklassiemer:
As discussed, this is the intended behavior. Solution, do not allow for custom job types matching a registered job name: #817

PR #810, opened 2022-07-29
Warn if config may be read by others

niklassiemer:
This is an alternative to #806


jan-janssen (approved):
Looks good to me - I guess we should merge this soon

PR #812, opened 2022-07-29
Add Project.symlink method to move and replace a project by symlink elsewhere

pmrv:
This intents to help on clusters with tight file quotas in the home directory.  With this method you can move the project folder to a temporary working stage (to save quota) and leave your pyiron hierarchy unchanged.


pmrv:
Right now the user has to give an explicit path, but we could think about adding a config flag "TEMPORARY_PROJECT_PATH" or so that is used to automatically pick a location.

niklassiemer (commented):
I like the idea. However, I am not sure how newly generated jobs etc. react on this setup right now. They use the pr.path to determine their path as well, right? Then, something like the working directory should be resolved fine... However, on job submission we might want to resolve the correct path before changing and executing the job to exclude possible problems?

Also an un-link would be nice in case someone wants to have the real data in the old place.

pmrv:
> I like the idea. However, I am not sure how newly generated jobs etc. react on this setup right now. They use the pr.path to determine their path as well, right? Then, something like the working directory should be resolved fine... However, on job submission we might want to resolve the correct path before changing and executing the job to exclude possible problems?

It work fine with new jobs as well.  I've been using this setup since last year, it's just that I did the linking manually.  If you link two directories `original -> temp`, then both `original/my/new/job` and `temp/my/new/job` are completely valid paths and jobs that are started with the first path never need to know that the filesystems redirects them to the second path.


> Also an un-link would be nice in case someone wants to have the real data in the old place.

Can do.

jan-janssen:
@pmrv  Can you add a test for this? then we can merge it soon.

pmrv:
I most likely won't get around to it until after MMM in two weeks.

PR #820, opened 2022-08-08
Filetable: Define column names only once

jan-janssen:


pyiron_base/database/filetable.py
```diff
@@ -87,16 +89,13 @@ def add_item_dict(self, par_dict):
...
             "timestart": datetime.datetime.now(),
-            "computer": None,
-            "parentid": None,
-            "username": None,
-            "timestop": None,
-            "totalcputime": None,
-            "masterid": None,
         }
-        for k, v in default_values.items():
+        for k, v in table_columns.items():
```
liamhuber: I am a little bit uncomfortable that we don't do anything to guarantee that `par_dict` is a strict subset of `table_columns` (it should be, right?), and that the `default_values` re-encode their subset of column names as strings. However, what you have here is still a clear improvement on the existing code, and I'm A-OK with leaving further stability improvements to a future PR.

I then find this formulation easier to read (and a couple lines shorter), but if you disagree feel free to leave it as the loop:
```python
par_with_defaults_dict = dict(table_columns)
par_with_defaults_dict.update(default_values)
par_with_defaults_dict.update(par_dict)
```

liamhuber (approved):
Nice. IMO we can make a readability improvement where there is a for-loop to update the dict, but it's a matter of taste so ignore it if you disagree.

pyiron_base/database/filetable.py
jan-janssen: I like the suggestion and updated the pull request accordingly

jan-janssen: It took a moment, but now I understand it is not the same: 
```
test = {"a": 1}
update = {"a": 2, "b":2}
test.update(update)
>>> {"a": 2, "b":2}
```
Keys which already exist are overwritten, while I only want to insert keys which are not already included in the dictionary.

jan-janssen: I made a mistake in the initial implementation - now everything is working as it should.

PR #825, opened 2022-08-12
Remove dangerous mutable argument

niklassiemer:


pyiron_base/jobs/job/extension/server/queuestatus.py
```diff
@@ -37,6 +37,8 @@ def queue_table(job_ids=[], project_only=True, full_table=False):
     Returns:
         pandas.DataFrame: Output from the queuing system - optimized for the Sun grid engine
     """
+    if job_ids is None:
+        job_ids = []
```
liamhuber:
```suggestion
    job_ids = [] if job_ids is None else job_ids
```

I prefer the one-liner, and like Jan's philosphy of putting the most-likely case in the logic clause first. The alternative where the default value of `[]` follows the `else` is also appealing though.

liamhuber (approved):
Nice catch.

One cosmetic suggestion.

PR #828, opened 2022-08-17
Cache the conda environment of the unittest workflow

niklassiemer:


.github/workflows/unittests.yml
```diff
@@ -41,3 +78,12 @@ jobs:
       shell: bash -l {0}
       timeout-minutes: 30
       run: coverage run --omit pyiron/_version.py -m unittest discover tests
```
niklassiemer:
```suggestion
      run: coverage run --omit pyiron_base/_version.py -m unittest discover tests
```
Use correct repo...

niklassiemer:
With the proposed setup, all workflows using a conda environment now cache the environment. To do so, each workflow
1.  Checks the repository out
2.  Performs a setup/merge of the `environment.yml` - I now always construct and use a `environment.yml` in the repo root, i.e. for the simple cases this is just `cp .ci_support/environment.yml .`.
3.  Install Mambaforge 
    - Here, using `miniforge-variant: Mambaforge` is already faster than installing conda and activating mamba.
    - This step does not use the `environment.yml`
4.  Store cache date and number in the `$GITHUB_ENV` to have ensure a new cache each day and upon request. To use a new cache by hand alter the `CACHE_NUMBER` in `.github/variables/cache_number.env`
5. Use the cache with 
    - a path to the conda environment
    - a key =  LABEL-conda-${{ hashFiles('environment.yml') }}-${{ env.DATE }}-${{ env.CACHE_NUMBER }}
6. Update the environment with the `environment.yml` if the cache did not restore the conda environment.
7. Run the subsequent workflow steps performing the actual unit/integration/benchmark/notebook-tests.
8. In the PostRun-cache the result of 6 is stored under the key for this cache. If multiple workflows build the same environment in their run, only the first one attempting to store the environment under the key is allowed to do so. 

To be quickly decided:
Right now I use the same key for the same environment even if they are from different workflows, i.e. bechmarks=unittests=pypicheck and backwards=notebooks (due to same `environment.yml`). The resulting speedup should not be very big, since most of the workflows start simultaneously. Thus, all these workflows will build their environment in the first run and only the fastest one will store this environment in the cache (at least if the `environment.yml` is the same as checked by the hash). A speedup is, thus, only expected if one workflow already wrote the cache before a second one tries to read the cache. 
On the other hand, this keeps the LABEL short and informative ('linux_64-py-3-10', 'linux_64-py-3-9','linux_64-py-3-8', 'win_64-py-3-10', 'osx_64-py-3-10') and not introducing additional 'notebook', 'pypi' or the like in the key. I.e. if it is the same python version and os and the hash of the `environment.yml` is the same it is considered the same environment. (My choice and, thus, implemented this way).

niklassiemer:
As far as I can tell, this is (already on base) a nice improvement in our test timings! The workflows, however, get somewhat longish...

pmrv (approved):
lgtm.  The race to cache problem should only appear for the first workflows to run on a given day, right?  For PRs later on the same day all workflows would already use the cached env right?

niklassiemer:
Yes, all other workflows should use the available cache (except of course, the `environment.yml` or the `CACHE_NUMBER` change). I also checked [here](https://github.com/niklassiemer/test_checkout/pull/11) how the cache behaves if the race happens and it seems to be robust :)

PR #829, opened 2022-08-19
Use RuntimeWarnings instead of debug log for auto-applied changes

niklassiemer:


pmrv (approved):

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -184,15 +184,19 @@ def queue(self, new_scheduler):
                 )
                 if cores != self.cores:
                     self._cores = cores
-                    state.logger.debug("Updated the number of cores to: %i", cores)
+                    state.logger.warning(
+                        "Updated the number of cores to: %i", RuntimeWarning
```
niklassiemer:
```suggestion
                        f"Updated the number of cores to: {corrs}", RuntimeWarning
```
Obviously, these things are not tested... will add tests and fix these errors...

niklassiemer:
One remaining thing - I would like that initializing with a queue results in the same state as initializing without a queue and setting the queue afterwards. Also I would like to preserve the `None` run_time and memory_limit if changing the queue, i.e. only set these on user input. Otherwise one could be surprised by changing to a low resource (e.g. test) queue and afterwards to a high resource queue and the run time is minimal...
Therefore, I want #831 merged in here first if there is no one against these changes.

PR #831, opened 2022-08-22
Use queue setter in init and preserve None in the setter if possible

niklassiemer:


niklassiemer:
ping @pmrv

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -182,15 +182,15 @@ def queue(self, new_scheduler):
                     run_time_max=self.run_time,
                     memory_max=self.memory_limit,
                 )
-                if cores != self.cores:
+                if self.cores is not None and cores != self.cores:
```
pmrv, line 185: If either `self.{cores,run_time,memory_limit}` is `None` then the corresponding branches are not taken and the value cannot be updated, right?  Is that what we want?

pmrv (changes requested):
I'm afraid I don't understand what the changes are for.  Is there an associated bug that would help me understand the issue?

niklassiemer:
Sorry for not providing the setting, I only wrote it over in https://github.com/pyiron/pyiron_base/pull/829... I simply, would like that initializing with a queue results in the same state as initializing without a queue and setting the queue afterwards. Also I would like to preserve the `None` `run_time` and `memory_limit` if changing the queue, i.e. only set these on user input. Otherwise one could be surprised by changing to a low resource (e.g. test) queue and afterwards to a high resource queue and the run time is minimal...

pyiron_base/jobs/job/extension/server/generic.py
niklassiemer: IMO it is what we want :) The updates would be from the `queue_adapter.check_queue_parameters()` call which returns the maximum for the parameter if asked for `None`. However, I would like to preserve the `None` (i.e. no user input) upon changing the queue.

pmrv: But if I do `job.server.cores = ...` in the notebook, this has be reflected, no?

niklassiemer: Yes, in that case `self.cores` would be set and the route would be taken, wouldn't it? The `cores` we compare `self.cores` with comes just from the `queue_adapter.check_queue_parameters()` which was fed by `self.cores` (at least in the case we just change the scheduler, i.e. the queue).

pmrv: Ah, sorry, I had misread the code. This is all in the `queue` setter and the other setters are not changed.  A-ok for me then.

pmrv (approved):

PR #832, opened 2022-08-23
Use executable.version where possible

niklassiemer:
@srmnitc

### DIFF
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -175,6 +175,7 @@ class GenericJob(JobCore):
         self.__name__ = type(self).__name__
         self.__version__ = "0.4"
         self.__hdf_version__ = "0.1.0"
+        self._parsed_version = None
         self._server = Server()
         self._logger = state.logger
         self._executable = None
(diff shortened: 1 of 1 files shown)


niklassiemer:
Currently, the version (hamversion) stored in the database is first obtained from the `class.__version__` attribute (which is inherited!) and only if this fails from the `executable.version`. Moreover, opon switching the version on the job, the `executable.version` is updated, however, `job.version` after all reports `__version__`. I would like to change the priority order in this case. Actually, I would even like an additional (internal) lookup of the actual version used for the computation (i.e. a parsed_version) and check if this matches our version assumption based on a filename in our resources.

jan-janssen:
Does not this break the transfer to jobs on different systems? I guess in the end it comes down to the question, do we want to store the abbreviation to the executable, like `5.4.1` for VASP or the full path where the executable is located. The second would be to store the full path, but that is fragile when we change the path or transfer the calculation to another cluster. Therefore my preference would be to do all the version management directly in the executable object, rather than parts of it in the job class and parts in the executable.

niklassiemer:
I would 'only' want to store the version number like '1.3.2' or the like in the version attribute.
This PR just changes the lookup order. If the executable has a version I would like to report that instead of the `JobClass.__version__` (maybe even inherited from `AtomisticGenericJob` like it is currently the case for lammps...).

The parsed version is something I would like to have in future, i.e. let the parser of the output file also read the version number. The Executable class just takes the file name of the run script for the version number and sometimes this does not match the actual version... in these cases I would like to update the version number and warn that the resources are not reflecting the reality.

PR #833, opened 2022-08-29
Avoid unnecessary copy in FlattenedStorage._from_hdf

pmrv:
Using `np.asarray` and `np.fromiter` instead of `np.array` avoids double allocation of arrays and can save up to a factor 4 in run time for very large storages (>100k chunks). 

NB: It seems that the utf decoding is the next speed limiting factor.  The decoding step is currently needed because we use `dtype('U')` numpy arrays, which are wide unicode characters, whereas HDF5 natively only handles UTF-8 encoded strings.  So a possible optimization is to use `dtype('O')` arrays with python strings inside and put those into HDF5 directly.  Some testing would be required however to make sure this works with `h5io`.

### DIFF
--- a/pyiron_base/storage/flattenedstorage.py
+++ b/pyiron_base/storage/flattenedstorage.py
@@ -910,12 +910,12 @@ class FlattenedStorage(HasHDF):
 
     def _from_hdf(self, hdf, version=None):
         def read_array(name, hdf):
-            a = np.array(hdf[name])
+            a = np.asarray(hdf[name])
             if a.dtype.char == "S":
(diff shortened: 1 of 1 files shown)


niklassiemer:
I am recalling an issue I had when updating to the correct data structures with our hdfio error: 
https://github.com/pyiron/pyiron_base/blob/2031f57d0a25ba261265b8207f230ee9401b40aa/tests/generic/test_fileHDFio.py#L195-L203

pmrv:
At least the `h5py` [docs](https://docs.h5py.org/en/stable/strings.html) mention that you can write utf8 strings into it.  Probably there's complications with the `h5io` wrapper though.

PR #836, opened 2022-09-01
remove unnecessary conda env update from workflow

niklassiemer:


jan-janssen (approved):
Looks good to me - This might be a general CI bug which also appears in other repositories.

PR #852, opened 2022-10-11
Fix ridiculous typo

pmrv:
When loading a path from HDF and trying to load any datacontainers, instead using the split path it repeatedly tried the same full path.


pmrv:
This fixes [this issue](https://github.com/pyiron/pyiron_atomistics/issues/798) in atomistics.

PR #854, opened 2022-10-13
Add subclass of HasStorage for easy input fields

pmrv:
Instead of expanding #753 I've put together what I think is the simplest solution to the "verbosity" problem we've had over in [atomistics](https://github.com/pyiron/pyiron_atomistics/pull/799) (and elsewhere previously).

This defines a new subclass of `HasStorage`, `KeywordInput`, that should make it much easier to create this kind of code specific input classes.  I've opted let developers manually document options in the class docstring, because I've found generation of them a bit wonky.  In #753 we had also discussed type and range checking of values that would be nice to have, but since `GenericParameters` also doesn't have it, I think it's something we can bolt on later.  Especially this the HDF-layout is fixed now by `HasStorage` we'll have a lot of freedom in re-designing the input classes without backwards compatibility issues.

Since I've overridden `__setattr__` and I wanted to forbid modification of attributes after the class is initialized (so that users get an error if they misspell an option, e.g.), I had to introduce a little hack with `_finalize` variable.  It ain't pretty, but it works.

### DIFF
--- a/pyiron_base/__init__.py
+++ b/pyiron_base/__init__.py
@@ -9,6 +9,7 @@ from pyiron_base.interfaces.factory import PyironFactory
 from pyiron_base.storage.flattenedstorage import FlattenedStorage
 from pyiron_base.storage.hdfio import FileHDFio, ProjectHDFio
 from pyiron_base.storage.datacontainer import DataContainer
+from pyiron_base.storage.keyword import KeywordInput
 from pyiron_base.storage.inputlist import InputList
 from pyiron_base.storage.parameters import GenericParameters
 from pyiron_base.storage.filedata import load_file, FileDataTemplate, FileData
--- a/pyiron_base/storage/keyword.py
+++ b/pyiron_base/storage/keyword.py
@@ -1,0 +1,83 @@
+from pyiron_base.interfaces.object import HasStorage
+
+class KeywordInput(HasStorage):
+    """
+    Base class for keyword based input structures, that know how write themselves to HDF5.
+
+    Derive from this class and add a class level attribute `_defaults`, which can be any mapping from allowed keyword
+    names to their default values.
+
+    Example:
+
+    >>> from pyiron_base.storage.datacontainer import DataContainer
+    >>> class MyOptions(KeywordInput):
+    ...     '''
+    ...     Stores the options to my beautiful code.
+    ...
+    ...     Attributes:
+    ...         run_fast (bool): run fast or slow?
+    ...         foo_level (float): amount of foo-iness
+    ...     '''
+    ...
+    ...     _defaults = {
+    ...             'run_fast': False,
+    ...             'foo_level': .7,
+    ...             'suboptions': DataContainer({'bar': 23})
+    ...     }
+    >>> opt = MyOptions()
+    >>> opt.run_fast
+    False
+    >>> opt['foo_level'] = .1
+    >>> opt.foo_level
+    0.1
+    >>> opt['invalid_option']
+    Traceback (most recent call last):
+        ...
+    KeyError: 'invalid_option'
+    >>> opt.yet_another_option
+    Traceback (most recent call last):
+        ...
+    AttributeError: yet_another_option
+    >>> opt.suboptions.bar
+    23
+    >>> opt.suboptions.name = 'Donald'
+    >>> opt.suboptions
+    DataContainer({'bar': 23, 'name': 'Donald'})
+    """
+
+    _defaults = {}
+    _finalized = False
+
+    def __init__(self, *args, **kwargs):
+        if "group_name" not in kwargs:
+            kwargs["group_name"] = self.__class__.__name__.lower()
+        super().__init__(*args, **kwargs)
+        self._finalized = True
+        for k, v in self._defaults.items():
+            self.storage[k] = v
+
+    def __getitem__(self, name):
+        if name in self._defaults:
+            return self.storage[name]
+        else:
+            raise KeyError(name)
+
+    def __setitem__(self, name, value):
+        if name in self._defaults:
+            self.storage[name] = value
+        else:
+            raise KeyError(name)
+
+    def __getattr__(self, name):
+        if name in self._defaults:
+            return self.storage[name]
+        else:
+            raise AttributeError(name)
+
+    def __setattr__(self, name, value):
+        if not self._finalized:
+            object.__setattr__(self, name, value)
+        elif name in self._defaults:
+            self.storage[name] = value
+        else:
+            raise AttributeError(name)


pmrv:
Tests and such when we have decided to keep it.

liamhuber:
Absolutely no complaints about this. The `_finalize` is indeed a little wonky, but hey, whatever gets the job done.

However, as much as I don't have complaints about this, I wonder whether this is the time to bring in `traitlets`.

- 👍 No need to re-invent the wheel (has equivalent infrastructure for `_defaults` and `_finalized` already built in)
- 👍 :star: :star: :star: Trivially easy to bolt on validation (e.g. type checking) (or not! Optional but super easy)
- 👍 Makes it easy to put callbacks on input changes if we ever want that in the future
  - EDIT: 👍 :star: :star: :star: or in the present, e.g. can do sexy things like update some input when other input changes (populating `potential` when `structure` gets set in the example below)
- 👍 Really easy to subclass and use composition
- 👎 Docstring still needs to be manually populated with the fields, but at least that's not any *worse* than the implementation here.

I played around with it for the last hour or two and this is what I got. What do you think?

```python
from __future__ import annotations

from abc import ABC, ABCMeta, abstractmethod
from typing import TYPE_CHECKING, Optional

from pyiron_atomistics import Atoms
from pyiron_atomistics.atomistics.structure.factory import StructureFactory
from pyiron_atomistics.lammps import list_potentials
from pyiron_base import HasStorage
from traitlets import ( 
    Bool,
    Float, 
    HasTraits, 
    Instance,
    Int, 
    MetaHasTraits,
    TraitError, 
    TraitType, 
    Unicode,
    validate, 
)

if TYPE_CHECKING:
    import pyiron_base
    
class ABCTraitsMeta(MetaHasTraits, ABCMeta):
    # Bookkeeping necessity
    pass

class Input(HasTraits, HasStorage, ABC, metaclass=ABCTraitsMeta):
    """
    The traitlets library allows us to quite easily do type checking and help on our IO values.
    
    Since all these traits are accessible under the `traits()` method, we know where to find 
    everything that needs to wind up in `storage`. We can then do copying at the last minute 
    (during read/write).
    """
    
    _locked = Bool(default_value=False)
    
    def to_hdf(
        self, 
        hdf: pyiron_base.storage.hdfio.ProjectHDFio, 
        group_name: Optional[str] = None
    ):
        for k in self.traits().keys():
            setattr(self.storage, k, getattr(self, k))
        super().to_hdf(hdf, group_name=group_name)
    
    def from_hdf(
        self, 
        hdf: pyiron_base.storage.hdfio.ProjectHDFio, 
        group_name: Optional[str] = None
    ):
        for k, v in self.storage.items():
            setattr(self, k, v)
        if self._locked:
            self.lock()
            
    def lock(self):
        for t in self.traits().values():
            t.read_only = True
            
    def unlock(self):
        for t in self.traits().values():
            t.read_only = False
    
class MDInput(Input):
    """
    We can have pretty simply input using built-in trait types with a bit of extra validation
    """
    temperature = Float(default_value=273, help='Temperature in Kelvin (default is 273).')
    n_steps = Int(default_value=1, help='How many MD steps to run for (default is 1).')
    
    @validate('temperature')
    @validate('n_steps')
    def _non_negative(self, proposal):
        if proposal['value'] < 0:
            raise TraitError(
                f'Expected a non-negative value for {proposal["trait"].name} ' \
                f'but got {proposal["value"]}'
            )
        return proposal['value']
    
class StructureTrait(TraitType):
    """
    Power users can make custom trait types for our more complex situations. 
    """
    allow_none = True
    default_value = None
    info_text = 'An instance of pyiron_atomistics.Atoms to represent the atomic structure.'
    
    def validate(self, obj, value):
        if isinstance(value, Atoms):
            return value
        self.error(obj, value)
        
class AtomisticInput(Input):
    """
    Although in this case, our validation is so trivial we can just use the built-in 
    `Instance` type.
    """
    # structure = StructureTrait()
    structure = Instance(
        klass=Atoms, 
        allow_none=True, 
        default_value=None,
        help='An instance of pyiron_atomistics.Atoms to represent the atomic structure',
    )
    
class LammpsInput(AtomisticInput):
    potential = Unicode(
        allow_none=True, 
        default_value=None,
        help='The emprical potential for representing atomic interactions'
    )

    @validate('potential')
    def _valid_potential(self, proposal):
        if self.structure is None:
            raise TraitError('Please set a structure before setting a potential')
        elif proposal['value'] not in list_potentials(self.structure):
            raise TraitError(
                f'{proposal["value"]} is not a valid potential for this structure,' \
                f'please choose from among {list_potentials(self.structure)}'
            )
        return proposal['value']
    
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.observe(self._set_default_potential, names=['structure'])
        
    def _set_default_potential(self, structure_change):
        """
        Set the potential to the 0th available potential when changing the structure from 
        None to Atoms.
        """
        if structure_change['old'] is None and isinstance(structure_change['new'], Atoms):
            self.potential = list_potentials(structure_change['new'])[0]
    
class LammpsMDInput(LammpsInput, MDInput):
    """Sweet, sweet composition"""
    pass

# Let's see how it works

def case_minus1():
    lammps_input = LammpsMDInput()
    print("I find the help really annoying to access, and it doesn't populate the docstring at all")
    print("The actual help:", lammps_input.traits()['temperature'].help)
    print("What you get if you just ? the attribute:", lammps_input.temperature.__doc__)
    print("This is known and seems unlikely to change. Boo. https://github.com/ipython/traitlets/issues/461")
    

def case_0():
    print("Simple properties get validated nicely as expected, with our requested error message")
    lammps_input = LammpsMDInput()
    lammps_input.temperature = -4
    
def case_1():
    lammps_input = LammpsMDInput()
    print("Complains that you need a structure first")
    lammps_input.potential = 'foo'
        
    
def case_2():
    lammps_input = LammpsMDInput()
    print("potential automatically populated once there's a structure")
    lammps_input.structure = StructureFactory().bulk('Al', cubic=True)
    print(lammps_input.potential)
    print("Will complain that this isn't in the list of available structures:")
    lammps_input.potential = 'foo'

def case_3():
    lammps_input = LammpsMDInput()
    print("Traitlets doesn't observe internal changes to mutable traits")
    lammps_input.structure = StructureFactory().bulk('Al', cubic=True)
    lammps_input.structure[:] = 'Mg'
    print(lammps_input.potential)
    print("So the potential is still a pure-Al potential with NO WARNINGS!")

def case_4():
    lammps_input = LammpsMDInput()
    lammps_input.structure = StructureFactory().bulk('Al', cubic=True)
    lammps_input.structure[:] = 'Mg'
    print("""
Do we care? 
Should we just trust that if a user does something like that they should damned well know 
to update the potential?
I'm not convinced this represents and significant weakness over our current implementation
And the valid potentials get updated.
We can see that the list is now valid potentials for Mg:
    """)
    lammps_input.potential = 'foo'
    
def case_5():
    lammps_input = LammpsMDInput()
    lammps_input.temperature = 5
    print("We can easily freeze the values in, e.g. after they've been used to run a job")
    lammps_input.lock()
    lammps_input.temperature = 6
    
# case_minus1()   
# case_0()
# case_1()
# case_2()
# case_3()
# case_4()
# case_5()
```

pmrv:
I'm happy to use the `traitlets`, but I'd take some time to experiment.  How would we best package this so devs don't have to deal with traitlet internals?

pmrv:
So we just discussed that the `Input` class sketched here is probably very nice to build input classes for generic codes.  Thoughts from the other developers @samwaseda @jan-janssen ?  We'll need another approach for line based input (like lammps), but I don't see that as a fundamental issue.

liamhuber:
At the pyiron meeting today the response was favourable, so I'll make a PR with the traitlets based approach later today.

PR #855, opened 2022-10-13
Call super in HasStorage for cooperative inheritance

pmrv:


pyiron_base/interfaces/object.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, *args, group_name=None, **kwargs):
         Args:
             group_name (str): default name of the HDF group where the whole object should be written to.
         """
+        super().__init__(*args, **kwargs)
```
niklassiemer, line 41: Do we need to pop the `group_name` from the `kwargs` then?

pmrv: It will be popped off automatically, since it is a kwarg of `__init__`.

liamhuber (approved):
Nice.

I actually only learned about cooperative inheritance from you in the last meeting, so I'm glad it is already useful knowledge 😆 Previously I was just manually doing `Parent1.__init__(self, *p1_args); Parent2.__init__(self, **some_other_kwargs)`. Sometimes that's actually nicer to avoid having generic `*args` and `**kwargs` floating around absolutely everywhere, but I'm looking forward to going back to my personal project and adjusting it in light of this new (to me) knowlege.

PR #858, opened 2022-10-14
Add flag "auto_refresh_job_status" that allows

ligerzero-ai:
automatic refresh  of job status upon loading of pr.job_table(). To-do:
I will benchmark this with projects of different sizes to see cost, see if it's negligible.

If negligible, set it to True by default
(better for new users, and is more intuitive overall)

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -655,8 +655,14 @@ class Project(ProjectPath, HasGroups):
         full_table=False,
         element_lst=None,
         job_name_contains="",
+        auto_refresh_job_status=False,
         **kwargs: dict,
     ):
+        """
+        auto_refresh_job_status (bool): will automatically reload job status by calling refresh_job_status() upon calling job_table
+        """
+        if auto_refresh_job_status:
+            self.refresh_job_status()
         return self.db.job_table(
             sql_query=self.sql_query,
             user=self.user,


ligerzero-ai:
image

@liamhuber it's an extra flat half second if I have it on by default no matter the project size. So I will merge this and enable it by default, although you are free to turn the flag off, of course ;)

samwaseda:
> @liamhuber it's an extra flat half second if I have it on by default no matter the project size. So I will merge this and enable it by default, although you are free to turn the flag off, of course ;)

aaaah no I disagree. It very much depends on the project size. It also depends how busy the cluster currently is.

On the other hand, I guess we can maybe also think about making a smarter refreshing routine, for example only if there are jobs with `running` or something.

ligerzero-ai:
nvm -got it

ligerzero-ai:
image

After I set all the statuses to "running". But I can see how the wait could get annoying if you have a bunch of jobs running when loading the jobtable, and this is the default behaviour. I'll merge it with the flag off by default, if nobody has any objections.

samwaseda:
ah ok so what I was saying was already implemented. And yes I'm ok with merging it now.

samwaseda (approved):

samwaseda:
One thing we should perhaps do is to stop sending an inquiry for each job. It makes the process probably much faster if all running jobs are evaluated at the same time.

PR #862, opened 2022-10-17
A (HasTraits, HasStorage)-based data class

liamhuber:
As discussed at the meeting the last couple weeks, and in #854, here is a generic input class that leans on `traitlets.HasTraits` for ease of use/validation/callbacks, and `HasStorage`.

* [x] New Input class
* [x] Tests
* [x] Get a per-instance read-only lock working (the traitlets lock applies to the trait on the `HasTraits` child and thus is global)
* [x] Get all the pedagogical stuff in the tests into the docstring
* [x] Update dependencies to include `traitlets`


liamhuber:
I went shotgun-style for reviewers. You don't *all* need to look at it (although you're certainly welcome to), but if we're going to rebase all our input onto this I'd like at least two opinions that aren't mine 😄

tests/storage/test_input.py
```diff
@@ -92,6 +92,9 @@ def _from_hdf(self, hdf, version=None):
 class CaffeinatedTrait(TraitType):
     """
     We can even make our own trait types.
+
+    In this case, we're making a counter-example -- custom traits with mutable defaults are dangerous! If more than one
+    object uses this trait, their default is the *same instance* of the mutable default.
```
niklassiemer: This is true for _all_ class or function definitions with mutable defaults! That's one reason for the importance of 
```
def some(arg0=None):
   arg0 = arg0 or {'default': True}
```
pattern.

liamhuber: Yeah, this is a great point. I think in my head I was so impressed with traitlets that I figured they were doing some sort of magic in `__new__` with `inspect` or somesuch to get the keyword defaults to re-instantiate as fresh objects? But you're right that this is totally expected; I'll re-phrase the documentation to emphasize how the `@default` decorator is the way to get around this expected behaviour.

liamhuber: I'm glad you pushed me on this, because it turns out that you totally can define a custom `TraitType` with a mutable default! It's not covered in their [readthedocs](https://traitlets.readthedocs.io/en/stable/defining_traits.html), but poking around the source I discovered that all you need to do is define a `make_dynamic_default(self)` method.

ligerzero-ai:
Gonna merge this for now so I can play around with it for vasp options! (As per our meeting https://github.com/orgs/pyiron/teams/pyiron/discussions/141?from_comment=9  ) and Jan's suggestion that we just merge it and see if anything breaks when I'm playing with it.

PR #864, opened 2022-10-20
Revert "emergency solution for loading"

samwaseda:
Sorry I realized that it makes another problem (I'll add a unit test at some point), but for now I'm gonna revert it.

### DIFF
--- a/pyiron_base/jobs/master/interactivewrapper.py
+++ b/pyiron_base/jobs/master/interactivewrapper.py
@@ -40,8 +40,7 @@ class InteractiveWrapper(GenericMaster):
             return self._ref_job
         try:
             if isinstance(self[0], GenericJob):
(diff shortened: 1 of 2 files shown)


niklassiemer:
Could you please write what broke? E.g. providing a small code snipped that was broken due to the fix?

samwaseda:
It's essentially the same problem, but appearing in different forms. The first example below fails without the previous PR, and the second one fails with the PR:

```python
spx = pr.create.job.Sphinx('spx')
sxextopt = spx.create_job('SxExtOptInteractive', 'sxextopt')
sxextopt.save()
sxextopt = pr.load('sxextopt')
sxextopt.run()
```

```python
spx = pr.create.job.Sphinx('spx')
sxextopt = spx.create_job('SxExtOptInteractive', 'sxextopt')
sxextopt.run()
```

PR #865, opened 2022-10-22
Convert `np.ndarray` to `list` before storing in `GenericParameters`

samwaseda:
For some reason `GenericParameters` converts `np.ndarray` to `str`.

To be honest, I would prefer to use `ndarray` as such though.


pmrv:
Thanks! Imo, once we have #862 in, `GenericParameters` can be mostly retired.

niklassiemer:
The unittests fail due to some random different oder in the tests - should be fixed in #866...

pyiron_base/storage/parameters.py
```diff
@@ -622,6 +622,8 @@ def __setitem__(self, key, value):
             key (str): key to be set of modified
             value (float/int/str): value to be set
         """
+        if isinstance(value, np.ndarray):
+            value = value.tolist()
         if isinstance(key, int):
             if self.read_only and self._dataset["Value"][key] != value:
                 self._read_only_error()
```
niklassiemer: Currently, i.e. without this modification, the value would be just assigned to the correct key - I do not see type checking or the like here. If the new lines are not in, the array is stored as is?

niklassiemer (commented):
Why do we need this conversion to list in the first place, again?

pyiron_base/storage/parameters.py
samwaseda: I don't know why, but I see that it becomes `str`. There are SOOOOOOOOOOOO many problems with `GenericParameters` that I don't understand. So don't ask me why...

niklassiemer: At which point does it become a `str`? 
For me running 
```
a = GenericParameters(table_name='gp')
a['some']= np.arange(10)
a.to_hdf(hdf)
a.from_hdf(hdf)
a['some']
```
returns the numpy array?... Maybe the writing of file stuff converts it? However, in that case also the list would be converted 😕

samwaseda: is it the same if you assign something that has a higher dimension? For example `np.ones((3, 3))`?

niklassiemer: Yes... The full MWE is: 
```
from pyiron_base import Project, GenericParameters
import numpy as np
pr = Project('.')
hdf = pr.create_hdf('some_test_hdf', 'genericParameter')
a = GenericParameters(table_name='gp')
a['some']= np.ones((3, 3))
a.to_hdf(hdf)
a.from_hdf(hdf)
a['some']
```
returning the plain `np.ndarray`?!

Therefore, the `str` conversion has to be caused by something else?!

samwaseda: ok I'm more than confused now. Indeed the code you pasted gives `ndarray` **if** they are executed in one cell. If I add the following lines in a **different** cell and execute them, I get `str`.

```python
a['some']= np.ones((3, 3))
a.to_hdf(hdf)
a.from_hdf(hdf)
a['some']
```

samwaseda: Ah no sorry, it has nothing to do with different cells but it's about doing the same thing twice, i.e.

```python
from pyiron_base import GenericParameters
gp = GenericParameters(table_name='test')
gp['test'] = np.ones((3, 3))
gp['test'] = np.ones((3, 3))
type(gp['test'])
```
gives `str` while

```python
from pyiron_base import GenericParameters
gp = GenericParameters(table_name='test')
gp['test'] = np.ones((3, 3))
type(gp['test'])
```

gives `ndarray`.

niklassiemer: Funny enough... I see you found the same line in the code causing this behavior... The question now is: Why was this converted to `str` in the first place 😢

samwaseda:
https://github.com/pyiron/pyiron_base/blob/db74f753c2dbe7d42e4905b5b48d9f90aa473e30/pyiron_base/storage/parameters.py#L428

I guess the problem was the `str`-conversion here. I just removed it and added the integration tests, to see whether there's a hidden problem.

pmrv:
I guess there are possible backwards compatibility issues here, so getting from `GenericParameters` should still try to evaluate the string if possible.

samwaseda:
> I guess there are possible backwards compatibility issues here, so getting from `GenericParameters` should still try to evaluate the string if possible.

So we should disable the conversion only when it’s np array?

niklassiemer:
I understood we would on the `__getattr__` after all try to evaluate the strings :)
In addition, we might need to store the `str` representation for some other cases? I do not know if some things we put into the GenericParameters are not hdf-able?!

samwaseda:
I guess we need @jan-janssen's testimony, otherwise we only keep guessing things.

jan-janssen:
> I guess we need @jan-janssen's testimony, otherwise we only keep guessing things.

The use case for `GenericParameters` is to have a more advanced access to a list of strings. Rather than storing the input for LAMMPS or the VASP KPOINTS file as just large strings `GenericParameters` simplifies the access. When talking about test files the difference between a numpy array and a list does not play a role. But as you see with the integration tests failing the modification breaks existing code, so I am against merging it. I feel the underlying issue is deeper, basically that we want to have support for string based input files and at the same time compatibility to python based data types, so for two types which have the same string representation like numpy array and list this always leads to conflicts unless we store the originating datatype. The same applies to the HDF5 interface. The only advantage of the HDF5 interface is that it allows the storage of attributes, so we can save the datatype the data came from.

samwaseda:
> The use case for `GenericParameters` is to have a more advanced access to a list of strings. Rather than storing the input for LAMMPS or the VASP KPOINTS file as just large strings `GenericParameters` simplifies the access. When talking about test files the difference between a numpy array and a list does not play a role. But as you see with the integration tests failing the modification breaks existing code, so I am against merging it. I feel the underlying issue is deeper, basically that we want to have support for string based input files and at the same time compatibility to python based data types, so for two types which have the same string representation like numpy array and list this always leads to conflicts unless we store the originating datatype. The same applies to the HDF5 interface. The only advantage of the HDF5 interface is that it allows the storage of attributes, so we can save the datatype the data came from.

Not sure what you are suggesting. Are you saying we don't allow numpy arrays at all?

jan-janssen:
> Not sure what you are suggesting. Are you saying we don't allow numpy arrays at all?

I am not suggesting anything - I just say it is a big mess right now. From my perspective the `GenericParameters` class is used forty different applications:
* To write input files like it is done in VASP and LAMMPS 
* To store values to reuse them inside python later

For the second use case I think the DataContainer class is much more appropriated. Or do you have an example of the first use case where you need numpy arrays?

niklassiemer:
Well the failing tests in atomistics break since they check if str(dict)==dict. For a dict stored in a Parameters subclass.

But yes, if the generic Parameter is intended to store a list of strings only, then numpy arrays are not intended to be stored as such. For file writibg they are indeed not well suited...? Why do we need them in GenericParameters, again?

samwaseda:
The problem is that when we work we use `list` and `np.array` almost interchangeably in many cases. My current problem is that the output of the Quasi Harmonic job, i.e. the Hessian matrix, cannot be straightforwardly used in the Quasi Newton job. I personally don't have a problem with converting `np.array` into `list` - actually I don't even think that there's a practical case where the distinction of `list` and `np.array` is indispensable (i.e. under normal circumstances either it doesn't matter or the developer must make sure that a conversion takes place in their job). On the other hand, it makes a huge problem if `str` comes out when `np.array` or `list` is expected. So for me it boils down to the following:

- Good: `list` -> `list`, `np.array` -> `np.array`
- Acceptable: `list` -> `list`, `np.array` -> `list`
- Bad: `list` -> `list`, `np.array` -> `str`

jan-janssen:
> My current problem is that the output of the Quasi Harmonic job, i.e. the Hessian matrix, cannot be straightforwardly used in the Quasi Newton job. 

Why does not the Quasi Harmonic job use a data container?

liamhuber:
> Why does not the Quasi Harmonic job use a data container?

Yep, I'd way rather see jobs updated to use a more modern class where it's not an issue than try to touch `GenericParameters`...`GenericParameters` scares me 😝

samwaseda:
that's a recurring problem ... `GenericMaster` is currently not compatible with `DataContainer`. I'm working on it, but it's not like we can straightforwardly solve the issue by exchanging `GenericParameters` by `DataContainer`.

pmrv:
I haven't followed all the details, but simply to get somewhere I would suggest:

1. we try to convert every iterable into a list and store that in `GenericParameters`
2. we issue a warning if it works or an error if it doesn't
3. we return all values as lists

PR #870, opened 2022-10-24
Unify master jobs

samwaseda:
Part of the reason why [this PR](https://github.com/pyiron/pyiron_atomistics/pull/588) is failing is because `InteractiveWrapper` does (not) do something that `ParallelMaster` and `SerialMaster` do (not). So far I failed to see what's exactly the difference, but in the meantime I try to put them closer, because a lot of functions are repeated, even though they all derive from `GenericMaster`.

Main changes

- [x] `input` defined in `GenericMaster`
- [x] `collect_output` defined in `GenericMaster`, which now collects everything in `output/generic`
- [ ] Make `SerialMaster` compatible with interactive jobs
- [x] `SerialMaster` should launch the next calculation only when `convergence_goal` returns a job


pyiron_base/jobs/master/serial.py
```diff
@@ -150,13 +150,7 @@ def start_job(self):
         Returns:
             GenericJob: start job
         """
-        if self._start_job is not None:
-            return self._start_job
-        elif len(self) > 0:
-            self._start_job = self[-1]
```
jan-janssen, line 81: `self[-1]` is not the same as `self[0]`

jan-janssen:
The behavior of the `SerialMaster` and the `ParallelMaster` differ, that is the reason why it is implemented in different classes.

pyiron_base/jobs/master/serial.py
samwaseda: I understand that they don't have the same effect, but I don't see why we should have different definitions for `SerialMaster` and `ParallelMaster` here

samwaseda:
> The behavior of the `SerialMaster` and the `ParallelMaster` differ, that is the reason why it is implemented in different classes.

I know that they are not the same, so I'm also not trying to have one class out of them.

samwaseda:
My biggest problem is that I've been working on [this PR](https://github.com/pyiron/pyiron_atomistics/pull/588) for more than half a year now without being able to figure out what's the right solution. This has been a huge problem for me because I cannot easily move on as it's part of the fundamental SPHInX functionality. Since it looks like the combination of SPHInX and `ParallelMaster` seems to be working, I'm trying to see where the difference between `ParallelMaster` and `InteractiveWrapper` comes from.

samwaseda:
@pmrv and @ligerzero-ai : Here I am adding docstrings to some of the classes we talked about yesterday. Let me know what you think and feel free to make suggestions.

pyiron_base/jobs/master/generic.py
```diff
@@ -263,6 +312,36 @@ def move_to(self, project):
...
+    def collect_output(self):
+        """
+        Collect the output files of the individual jobs and set the output of the last job to be the output of the
+        SerialMaster - so the SerialMaster contains the same output as its last child.
+        """
+        if self.ref_job.server.run_mode.interactive:
+            if (
+                "output" in self.ref_job.list_groups()
+                and "generic" in self.ref_job["output"].list_groups()
+            ):
```
pmrv:
```suggestion
            output = self.ref_job['output/generic']
            if output is not None:
```

pmrv: And then work with output below

samwaseda:
I'm still not perfectly happy, especially with the fact that `convergence_goal` in `SerialMaster` and `job_generator` in `ParallelMaster` coexist, even though they do virtually the same thing for the following reasons:

- `convergence_goal` is a function while `job_generator` is a class
- `convergence_goal` is set by a setter function, while `job_generator` is set by a property setter
- `job_generator` is set to a private function

liamhuber:
@raynol-dsouza Could you please check out this branch on your local environment and then try running a really simple `Protocol` calculation? We depend on these masters over there, but I don't think there's any part of the CI that explicitly checks that base is not breaking contrib.

samwaseda:
Nothing is failing -> Can I merge this one?

pyiron_base/jobs/master/generic.py
```diff
@@ -263,6 +309,47 @@ def move_to(self, project):
...
+            self._ref_job = self.pop(-1)
+            if self._job_id is not None and self._ref_job._master_id is None:
+                self._ref_job.master_id = self.job_id
+
+    def write_input(self):
+        """
+        Write the input files - for the SerialMaster this only contains convergence goal.
+        """
+        super().write_input()
+        self.input.write_file(file_name="input.inp", cwd=self.working_directory)
```
jan-janssen, line 286: Why do you need to write the input ? As the jobs are all python there should be no need for an input file or am I missing something?

pyiron_base/jobs/master/serial.py
```diff
@@ -25,119 +24,37 @@
...
+    )
 
     def __init__(self, project, job_name):
-        self._input = GenericParameters("parameters")  # e.g. convergence goal
 
         super(SerialMasterBase, self).__init__(project, job_name=job_name)
         self.__version__ = "0.3"
 
-        self._output = GenericOutput()
-        self._max_iterations = 100
```
jan-janssen, line 64: Where do you define the maximum number of iterations?

pyiron_base/jobs/master/generic.py
samwaseda: This was something that was in both `ParallelMaster` and `SerialMaster`. So I moved it to `GenericMaster`.

pyiron_base/jobs/master/serial.py
samwaseda: hmmm? I don't ... ? If you are asking why I'm erasing it, it's because I saw that it was not used anywhere.

jan-janssen:
> Nothing is failing -> Can I merge this one?

As it is a rather complex change to a brittle part of the code, I would recommend running the notebooks for Murnaghan and Phonopy at least once. As we discussed in the meeting I would prefer to have a new class and then migrate the individual classes step by step.

samwaseda:
> > Nothing is failing -> Can I merge this one?
> 
> As it is a rather complex change to a brittle part of the code, I would recommend running the notebooks for Murnaghan and Phonopy at least once. As we discussed in the meeting I would prefer to have a new class and then migrate the individual classes step by step.

In this PR I didn't change anything for the user side. I only removed duplications (or at least what looked like duplications)

pyiron_base/jobs/master/generic.py
```diff
@@ -263,6 +309,47 @@ def move_to(self, project):
...
+            nodes = ham_lst[0]["output/generic"].list_nodes()
+            with self.project_hdf5.open("output/generic") as hh:
+                for node in nodes:
+                    if ham_lst[0]["output/generic/{}".format(node)] is not None:
+                        hh[node] = np.concatenate(
+                            [ham["output/generic/{}".format(node)] for ham in ham_lst],
+                            axis=0,
+                        )
+                    else:
+                        hh[node] = None
```
jan-janssen, line 270: This makes no sense for parallel masters

pyiron_base/jobs/master/generic.py
jan-janssen: > This was something that was in both `ParallelMaster` and `SerialMaster`. So I moved it to `GenericMaster`.

As far as I can see it was only in the `SerialMaster` but I do not see why it is necessary.

pyiron_base/jobs/master/serial.py
jan-janssen: By briefly looking over the code I also do not find a use case, so maybe it is good to delete it.

jan-janssen:
> In this PR I didn't change anything for the user side. I only removed duplications (or at least what looked like duplications)

You merge two classes, so now the ParallelMaster has functions it did not have before, for example the collect output part and I just want to make sure everything works. Otherwise I assume that we might soon have users complaining that some functionality that worked before broke. The tests currently do not cover the Serial and Parallel Master sufficiently.

pyiron_base/jobs/master/generic.py
samwaseda: What exactly doesn't make sense?

jan-janssen: It takes the output of only the first child job

samwaseda: No it does the loop over the jobs inside the list comprehension.

jan-janssen: No, you only iterate over the groups of the first job `nodes = ham_lst[0]["output/generic"].list_nodes()`

samwaseda: Yes, I don't think the nodes change for different child jobs, especially for `ParallelMaster`, so it doesn't matter which child gives the nodes.

samwaseda:
So these ones all survived:

```python
lmp = pr.create.job.Lammps('lmp_phono_static')
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
phono = lmp.create_job('PhonopyJob', 'phono_static')
phono.run()
```

```python
lmp = pr.create.job.Lammps('lmp_phono_interactive')
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.interactive_open()
phono = lmp.create_job('PhonopyJob', 'phono_interactive')
phono.run()
```

```python
lmp = pr.create.job.Lammps('lmp_murn_static')
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
murn = lmp.create_job('Murnaghan', 'murn_static')
murn.run()
```

```python
lmp = pr.create.job.Lammps('lmp_murn_interactive')
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.interactive_open()
murn = lmp.create_job('Murnaghan', 'murn_interactive')
murn.run()
```

samwaseda:
Can I merge this one? @jan-janssen

jan-janssen:
> Can I merge this one? @jan-janssen

Have you tested the notebooks in `pyiron_atomistics` ? I mean in particular the phonopy notebook and the pipeline notebook?

pyiron_base/jobs/master/generic.py
jan-janssen: > Yes, I don't think the nodes change for different child jobs, especially for `ParallelMaster`, so it doesn't matter which child gives the nodes.

No, for the `ParallelMaster` you want to iterate over all children.

pyiron_base/jobs/master/generic.py
jan-janssen: This is not resolved !

samwaseda: I just moved `write_input` into `generic.py`, because it was defined in both `ParallelMaster` and `SerialMaster`. The fact that there is `write_input` both in `ParallelMaster` and `SerialMaster` has nothing to do with this PR.

jan-janssen: > As far as I can see it was only in the `SerialMaster` but I do not see why it is necessary.

samwaseda: Why did you repost it?

jan-janssen: Because it is still true, it was only used in `SerialMaster` so I would recommend to keep it there.

samwaseda: No you are wrong. It was defined in both `SerialMaster` and `ParallelMaster`.

jan-janssen: > No you are wrong. It was defined in both `SerialMaster` and `ParallelMaster`.

No https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/jobs/master/parallel.py

samwaseda: https://github.com/pyiron/pyiron_base/blob/7af9a6c070cd3653a6b1f7401804a2b172f092cc/pyiron_base/jobs/master/parallel.py#L250-L257

?????

pyiron_base/jobs/master/generic.py
samwaseda: `ParallelMaster` creates the child jobs from **one** template job. It is by principle not possible to have different nodes.

jan-janssen: > `ParallelMaster` creates the child jobs from **one** template job. It is by principle not possible to have different nodes.

This is not about the nodes, it is about the content of them.

samwaseda: I don't understand. You raised the point that it takes only the nodes of the first job. The iteration over the jobs (which is anyway done) was never the issue here.

pyiron_base/jobs/master/generic.py
jan-janssen: Ok, that is something that should be fixed

jan-janssen (changes requested):
For me this pull request tries to do too many things at once, can we split it in multiple pull request? For example we could have one to unify the `write_input` functions and so on.

pyiron_base/jobs/master/generic.py
samwaseda: So this one is resolved?

samwaseda:
> For example we could have one to unify the `write_input` functions and so on.

This PR just moves the location of `write_input` without changing literally **anything** in functionality. Feel free to open another PR when this is done, but the change you are suggesting has nothing to do with this PR.

samwaseda:
> > Can I merge this one? @jan-janssen
> 
> Have you tested the notebooks in `pyiron_atomistics` ? I mean in particular the phonopy notebook and the pipeline notebook?

They are not included in the integration tests?

jan-janssen:
> They are not included in the integration tests?

Notebooks are not included in integration tests

pyiron_base/jobs/master/generic.py
```diff
@@ -350,14 +439,32 @@ def get_child_cores(self):
...
+        """
+        new_job = super().copy()
+        if self.ref_job is not None:
+            new_job.ref_job = self.ref_job
+        return new_job
+
     def __len__(self):
         """
-        Length of the GenericMaster equal the number of childs appended.
+        Length of the ListMaster equal the number of childs appended.
```
jan-janssen: Why change it from `GenericMaster` to `ListMaster` ?

samwaseda:
> For me this pull request tries to do too many things at once, can we split it in multiple pull request?

Yes I can, but your suggestion makes only sense if the changes are entangled or there are conceptually multiple changes, which is not the case here. From what I am seeing right now, you are pointing out individual changes which have nothing to do with each other. In the end, for the changes I'm proposing here it shouldn't matter whether there are n times small PR, each of which takes a shorter time to review, or one time this PR which takes n times the amount of time.

pyiron_base/jobs/master/generic.py
samwaseda: I vaguely remember that I wanted to write something different but I don't know anymore. Anyway I put back the original description

jan-janssen:
@samwaseda I created three examples how the changes in this pull request can be split in small changes, which are easy to review: https://github.com/pyiron/pyiron_base/pull/914 , https://github.com/pyiron/pyiron_base/pull/915 and https://github.com/pyiron/pyiron_base/pull/916 . I hope this helps to demonstrate what I meant with separate pull requests, rather than one big one.

samwaseda:
Whatever.

PR #871, opened 2022-10-24
No codacy on push to master

niklassiemer:


niklassiemer:
The downside is, that our coverage is also only tracked by the codacy-coverage action...! Is it possible to configure codacy to not check the master/main branch for coding issues?!
And we might consider to switch the https://github.com/codacy/codacy-coverage-reporter instead of the archived solution 😅 They also provide a github action to report.

jan-janssen:
Be careful codacy and coveralls are not the same: 
Screen Shot 2022-10-24 at 3 39 52 PM

niklassiemer:
Ok, then I just need to leave out the very last call to not get a codacy report? Or is the `python-coverage-codacy` also reporting our coverage?

jan-janssen:
No `codacy` is in https://github.com/pyiron/pyiron_base/blob/no_master_codacy/.github/workflows/codeql-analysis.yml

... At least so I thought but that seems to be yet something different ...

niklassiemer:
Yes codacy and coverage are now part of the unittest workflow... And they seam to mangled in that python-codacy-coverage call?!
Maybe we should clearly separate these steps, e.g. by using the coveralls action https://github.com/marketplace/actions/coveralls-github-action

niklassiemer:
As said in the [coveralls-python docu](https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support) the plain `coveralls` call reports the stuff.

liamhuber:
Just to keep you guys in the loop, I'm playing around with this in [pyiron_github](https://github.com/pyiron/pyiron_github) and in a [PR to ironflow](https://github.com/pyiron/ironflow/pull/99). 

I abstracted the coverage call to unittests into a github action on pyiron_github. My understanding is still pretty shaky on all this, but my current understanding is that `coverage run` is actually generating data from each of the matrix runs, and that this data is later is later collated in `coverage combine` and used in `coverage xml`, which feed into coveralls and codacy. Even if I'm wrong that it's collecting stuff across the matrix runs, I am more confident that it is at least seeing the output from the matrix run that the coveralls and codacy actually trigger on (ubuntu with py3.10). To this end, it seems most computationally efficient to leave unittests and coveralls+codacy combined in the same job to avoid re-running the coverage and unittests.

I would very much welcome correction on that understanding if you have it to offer, but I'll keep slugging away at this for a bit so don't feel any pressure to go out and do additional research yet!

liamhuber:
I should also quickly say, splitting the coveralls and codacy bit up and not running codacy on push to main is brilliant. My concern is just that I think they do indeed still need to live in the unittests job.

niklassiemer:
Well, yes and no... I actually only recently moved then in, since they need the data from the testing. However, it is somehow possible to define artifacts and such stuff... probably, the coverage runs already have some files as artifact. 
I.e. it should be fine to run them in different jobs but on the same workflow and thus VM. Or one exposes the artifact and refers to that again. That's however some additional work ..

PR #873, opened 2022-10-25
add docstring to pyiron table

samwaseda:
Just like always, there was no doc string to the pyiron table, so I added a minimum example.


samwaseda:
I see `run_again=True` on [this page](https://pyiron.readthedocs.io/en/latest/source/notebooks/data_mining.html), but should it be actually there? Does pyiron table run from the beginning in this case?

samwaseda:
I didn't get a review, but since I'm not changing the code itself I'm gonna merge it.

pmrv:
Looks good, thanks!

PR #875, opened 2022-10-25
Adds a job.kill() functionality that stops a job in the queueing system without removal of files and updates status. 

### LINKED ISSUE #922: Is there a possibility to stop submitted jobs without deleting the existing results?
I know that `job.remove()` will kill the job BUT it also deletes the folder and therefore all existing results when the jobs was running on the cluster. Is there a possibility to stop the job on the cluster, but not delete the results?

ligerzero-ai:
Resolves https://github.com/pyiron/pyiron_base/issues/922


pyiron_base/project/generic.py
```diff
@@ -1213,6 +1213,21 @@ def queue_delete_job(self, item):
         else:
             raise EnvironmentError("copy_to: is not available in Viewermode !")
 
+    def kill_job(self, job_specifier):
+        """
+        Kills a running job via the queueing system
+
+        Args:
+            job_specifier (int, GenericJob): Provide either the job_ID or the full hamiltonian
```
samwaseda: Shouldn't it be the job name?

ligerzero-ai: > Shouldn't it be the job name?

Actually I don't really understand the nomenclature in the `project` module - from what I can see, these things `job_specifier`, `item` are used somewhat interchangeably.

samwaseda: I mean more the variable type - now it's `int` and `GenericJob`, but if it's the job name that would be `str`. And actually does it work with the job name?

ligerzero-ai: I am not sure, actually. But this is essentially just a copy of the queue_delete_job function of the class. So if it works there, it should work here. Should we just implement this with the ability to take in job_id, job object, and job_name? So we just check for the type?

jan-janssen: The naming is bad - I agree, but sometimes it helps to just look for other places where this function is used: 
https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/jobs/job/core.py#L420

So maybe rather than having duplicated code, the removal of the job from the queuing system can be placed in its own function which is then called in both situations.

ligerzero-ai: I have a preference for `kill_job` to `queue_delete_job` on the user-utilisation side. 

Possible to rename? I would imagine that this feature will be used by many users and `queue_delete_job` is somewhat obscure and not as clear, imho.

pmrv: I find `job.*_job` a bit.  Maybe just `job.kill()` or `job.stop()`?

pmrv: Mentioning `queue` is fair imo, since this doesn't work for `non_modal` jobs, but `kill` sort of implies that.

ligerzero-ai: The problem is that I can see that a `kill` method already exists for the job object in `pyiron_base/pyiron_base/jobs/job/generic.py` in the `GenericJob` class.

Unfortunately, it appears that it calls `remove` which uses `remove_child` and it has a inbuilt deletion of the working directory.

What do? 

Implement a new method which doesn't remove the directories?

Or modify the kill method and extend it with mostly duplicated code? 

Or modify the remove_child and remove methods to have an if branch and a flag which toggles the deletion of the other stuff (super ugly?)

Either way I don't see a way around writing significant amounts of duplicated code or having a really ugly spaghetti solution.

The problem is that `job.kill()` calls `remove_child` and `remove`, and there is too much functionality built into these two functions that isn't consistent with the behaviour that we want.

ligerzero-ai: So; I sketched this out:

Move 
https://github.com/pyiron/pyiron_base/blob/dd27661aa00c135f0b4fcc6506a2fe8e2d9a921f/pyiron_base/jobs/job/core.py#L433-L441

into it's own little function in `pyiron_base/jobs/job/core.py` named `kill`

And then I will change the old `job.kill` method already in `generic.py` to inherit it from the `kill` method in `GenericJob` 

Any objections? I've checked that the kill method isn't currently being used anywhere. It appears to remove the folders but keep a database entry with master_id and parent_id with no job folder.

ligerzero-ai:
Okay, I'm stuck and I need some help.

Consider the following code:


```
from pyiron.project import Project
import pyiron.atomistics.structure.atoms as pi_atoms
from pyiron_atomistics.atomistics.structure.factory import StructureFactory

pr = Project("test_module_idx_map")
pyiron_atoms = StructureFactory().bulk(name = "Fe", crystalstructure = "bcc", cubic=True) * [2,2,2]
pyiron_atoms[1] = "P"
pyiron_atoms[5] = "Ni"
pyiron_atoms
pr.remove_job(["sorted", "sorting_forbidden"])
KPAR = 2; NCORE = 4; NCPU = 40
sorting_behaviour = [True, False]
for i, job_name in enumerate(["sorted", "sorting_forbidden"]):
    with pr.open("%s" % job_name) as pr_sub:
        sub_job = pr_sub.create_job(pr.job_type.Vasp, job_name)
        sub_job.structure = pyiron_atoms 
        sub_job.set_kpoints(mesh=[4, 4, 4], scheme = "GC")
        sub_job.server.queue = "cm"
        sub_job.server.cores = NCPU
        sub_job.executable = "5.4.4_mpi_AutoReconverge"
        sub_job.run()
```

Now, let us cancel them with the new kill functions, which should, in theory, kill the job, and set the status to aborted:

```
sub_job.kill()
sub_job.status
```
which yields output 
`'aborted'`

Let's check the jobtable! It should now read aborted right?

`pr.job_table()`

<html>
<body>


id | status | chemicalformula | job | subjob | projectpath | project | timestart | timestop | totalcputime | computer | hamilton | hamversion | parentid | masterid
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
18327785 | submitted | Fe14NiP | sorted | /sorted | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorted/ | 2022-11-22 12:36:00.541940 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None
18327786 | submitted | Fe14NiP | sorting_forbidden | /sorting_forbidden | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorting_forbidden/ | 2022-11-22 12:36:00.912308 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None


</body>
</html>

Okay, no problem; let's refresh the job status in the project using
```
pr.refresh_job_status()
table_job = pr.job_table()
table_job
```

why is this still yielding the same output?

<html>
<body>


id | status | chemicalformula | job | subjob | projectpath | project | timestart | timestop | totalcputime | computer | hamilton | hamversion | parentid | masterid
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
18327785 | submitted | Fe14NiP | sorted | /sorted | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorted/ | 2022-11-22 12:36:00.541940 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None
18327786 | submitted | Fe14NiP | sorting_forbidden | /sorting_forbidden | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorting_forbidden/ | 2022-11-22 12:36:00.912308 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None


</body>
</html>

pyiron_base/jobs/job/core.py
```diff
@@ -429,6 +431,48 @@ def remove(self, _protect_childs=True):
...
+
+        def base_kill():
+            if "server" in self.project_hdf5.list_nodes():
+                server_hdf_dict = self.project_hdf5["server"]
+                if (
+                    "qid" in server_hdf_dict.keys()
+                    and server_hdf_dict["qid"] is not None
+                ):
+                    self.project.queue_delete_job(server_hdf_dict["qid"])
+                    self._status = "aborted"
```
jan-janssen, line 470: Please use `self.status` then the database is also updated.

jan-janssen: https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/jobs/job/extension/jobstatus.py#L196

pmrv: Since we're operating on `JobCore` that is not automatically called.

pyiron_base/jobs/job/generic.py
```diff
@@ -653,16 +653,19 @@ def remove_child(self):
         _kill_child(job=self)
         super(GenericJob, self).remove_child()
 
-    def kill(self):
-        if self.status.running or self.status.submitted:
-            master_id, parent_id = self.master_id, self.parent_id
-            self.remove()
-            self.reset_job_id()
-            self.master_id, self.parent_id = master_id, parent_id
```
jan-janssen, line 626: Where do you handle the `job_id` reset? It would be great to maintain the old functionality in addition to the new functionality.

pyiron_base/jobs/job/core.py
```diff
@@ -776,6 +815,9 @@ def copy_to(
                         posixpath.join(new_job_core.project_hdf5.h5_path, group)
                     ]
             new_job_core._status = "initialized"
+            # THIS PROBABLY DOESN'T WORK! DATABASE NEEDS TO BE UPDATED - SEE KILL METHOD
+            # self.project.db.item_update({"status": "aborted"}, self._job_id)
```
jan-janssen: At this stage the job is not yet stored in the database, so resenting the internal variable is sufficient.

pyiron_base/jobs/job/core.py
jan-janssen: > Since we're operating on `JobCore` that is not automatically called.

Oh, ok now I see where the trouble is coming from. But then would not it be easier to implement this action on the job level rather than the jobcore level?

ligerzero-ai: I am just seeing that `remove_child` is inherited from the jobcore level (right above this in the generic.py).

`remove_child` is part of the `remove` functionality in core.

`remove_child` now calls the kill functionality as part of it's operation (as per your suggestion to avoid duplicated code)

To me, it is weird to have functions inheriting from one module right above it and have another function right below it do the opposite; i.e. the behaviour being coded in the `job` module and inherited by the `core` module. This doesn't make much sense especially since now it is part of core functionality in the `remove_child` method.

The old kill method also derived from core - by implementing this in the generic level this represents a reversal in architecture.

Design wise, it would be preferable to me to have all the core functionality in one place as opposed to having interdependencies between packages. (even if it exists currently)

I feel that is important for code maintainability and understandability overall.

pyiron_base/jobs/job/generic.py
ligerzero-ai: Why is this necessary? I don't quite understand the code. 

```
    def reset_job_id(self, job_id=None):
        """
        Reset the job id sets the job_id to None in the GenericJob as well as all connected modules like JobStatus.
        """
        if job_id is not None:
            job_id = int(job_id)
        self._job_id = job_id
        self._status = JobStatus(db=self.project.db, job_id=self._job_id)
```

The goal of this implementation of kill is to act exactly as if the run did happen, but the user intervenes, but all generated input and output thus far are preserved. What does resetting the job id achieve exactly in the old implementation?

jan-janssen: If you calculate an energy volume curve with the `Murnaghan` class and one of the calculation fails, then you want to be able to restart the calculation that failed. In the background this is using the existing/ old kill functionality.

ligerzero-ai: Now that I understand the use case for the old method, I would prefer that we can move the old function to it's own thing and let this new method be kill() as I feel quite strongly that this specific use does not warrant occupying this name.

If I understand the use correctly, the removal of the directory is necessary for the restart yes? Which is why the old kill method employed the `remove` method. This is not compatible with the new purpose of kill() and what it intends to achieve. 

Any comment? Can you link me the specific code in pyiron that uses this? Or is this something that you used in your own notebooks?

ligerzero-ai: **tl;dr** Functions which occupy obvious names such as `kill` for the job object should not have niche functionality which is unexpected and exists because of legacy reasons, or backwards compatibility pains.

Just to explain why I feel quite strongly about this, I am viewing this from the perspective of a new user or the general potential pyiron userbase.

I'm looking for functionality that kills the job in the queueing system, while preserving the job data generated so far. Extremely common operation in the terminal, or any HPC user. I should find something in pyiron, since it claims that the need to access the terminal when I use pyiron should be almost nothing. For something as obvious as killing non-converging jobs on the queueing system, something like `qdel` or `scancel` functionality, surely this should exist! To my delight, I find that `job.kill()` exists. If I'm careful, I check the docstring, and I'm not too convinced, but I can't be bothered to see if it actually does what it is claiming through careful inspection of the source, so I call it. 

To my dismay, I find that it removes the entire directory for whatever reason! WTH? This makes no sense to me! Why doesn't it just delete/kill the job via the queueing system commands, like I would expect? Why is this method named `kill` when it doesn't agree with any conventions in HPC terminology?

Then, I go peek into the source, ctrl+click reveals that this function isn't used anywhere in the code. I go digging around in the github, and I find this issue, and I find that it exists for some cryptic use-case that only exists in the Murnaghan job, which I'd argue is not relevant to most of the potential userbase for pyiron. Wow! 

Occupying this method name with code which exists for an extremely specific use-case for legacy puposes might have been acceptable when pyiron was for a niche userbase, but this is no longer acceptable if we're taking the viewpoint of general adoption.

pmrv: I just grepped `kill`  in both base and atomistics and actually nothing relevant pops up.  Either this went out quietly a long time ago or users were somehow expected to call it on the child jobs themselves? In both cases I'd say go ahead with the current direction.  If the "old" functionality is something we want to support again in master jobs, we can do that with cleaner code.

ligerzero-ai:
For posterity purposes, as this kind of thing is not reasonable to set up tests for on real queueing systems in the CI, here is the jupyter notebook, cell by cell that one can use to test this functionality:
```
from pyiron.project import Project
from pyiron_atomistics.atomistics.structure.factory import StructureFactory

pr = Project("test_module_idx_map")
pyiron_atoms = StructureFactory().bulk(name = "Fe", crystalstructure = "bcc", cubic=True) * [2,2,2]
pyiron_atoms[1] = "P"
pyiron_atoms[5] = "Ni"
sorting_behaviour = [True, False]
for i, job_name in enumerate(["sorted", "sorting_forbidden"]):
    with pr.open("%s" % job_name) as pr_sub:
        sub_job = pr_sub.create_job(pr.job_type.Vasp, job_name)
        sub_job.structure = pyiron_atoms 
        sub_job.set_kpoints(mesh=[4, 4, 4], scheme = "GC")
        #sub_job.input.options.allow_structure_reordering = sorting_behaviour[i]
        sub_job.input.incar['EDIFF'] = 1E-7
        sub_job.input.incar['EDIFFG'] = -1E-2
        sub_job.server.queue = "cm"
        sub_job.server.cores = 40
        sub_job.executable = "5.4.4_mpi_AutoReconverge"
        sub_job.run()
```
<html>
<body>


id | status | chemicalformula | job | subjob | projectpath | project | timestart | timestop | totalcputime | computer | hamilton | hamversion | parentid | masterid
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
18334536 | submitted | Fe14NiP | sorted | /sorted | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorted/ | 2022-11-23 15:47:27.083149 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None
18334537 | submitted | Fe14NiP | sorting_forbidden | /sorting_forbidden | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorting_forbidden/ | 2022-11-23 15:47:27.377781 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None


</body>
</html>

```
pr.job_table()
```
```
    JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
           4254461    p.cmfe pi_18334     hmai  R       0:07      1 cmti004
           4254462    p.cmfe pi_18334     hmai  R       0:07      1 cmfe033
```

```
sub_job.kill?
```
```
Signature: sub_job.kill()
Docstring:
Kills a job via queueing system commands (e.g. scancel/qdel in terminal)
It can update the status of the job to "aborted" in the job database.

enable (bool): Attempt a job-kill regardless of status when set to True.
               This will set the status to "aborted" no matter if the job was successfully killed or not!
               Set to False if you don't know what you are doing (unlikely).
               When False, it checks that the job is actually in a queued/running state
               in the database before attempting to remove it from the queue and set the aborted status.

Use case: A python-side equivalent method to call "scancel JOBID" (slurm) or "qdel JOBID" (PBSPro/torque) via pyiron
This command is "dumb" by default, it will attempt to kill anything that it is called on, regardless of job-status.
File:      /cmmc/u/hmai/github_branch_sorting_behaviour/pyiron_base/pyiron_base/jobs/job/generic.py
Type:      method
```
```
sub_job.kill()
```
```
2022-11-23 15:47:41,724 - pyiron_log - WARNING - The job.kill() functionality has changed! It now kills the job via queueing system commands and preserves the files and directory
```
```
pr.job_table()
```
<html>
<body>


id | status | chemicalformula | job | subjob | projectpath | project | timestart | timestop | totalcputime | computer | hamilton | hamversion | parentid | masterid
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
18334536 | running | Fe14NiP | sorted | /sorted | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorted/ | 2022-11-23 15:47:39.530490 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None
18334537 | aborted | Fe14NiP | sorting_forbidden | /sorting_forbidden | /cmmc/u/ | hmai/dev_pyiron/create_jobkill_feature/test_module_idx_map/sorting_forbidden/ | 2022-11-23 15:47:39.158205 | None | None | hmai@cmti001#40#cm | Vasp | 5.4.4_mpi_AutoReconverge | None | None


</body>
</html>

```
! squeue -u hmai
```

```
      JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
           4254461    p.cmfe pi_18334     hmai  R       0:27      1 cmti004
```

ligerzero-ai:
If there are no further comments/suggestions by the end of the week, I'll merge this on Friday.

pyiron_base/jobs/job/generic.py
```diff
@@ -654,15 +654,20 @@ def remove_child(self):
...
+
+        enable (bool): Attempt a job-kill regardless of status when set to True.
+                       This will set the status to "aborted" no matter if the job was successfully killed or not!
+                       Set to False if you don't know what you are doing (unlikely).
+                       When False, it checks that the job is actually in a queued/running state
+                       in the database before attempting to remove it from the queue and set the aborted status.
+
+        Use case: A python-side equivalent method to call "scancel JOBID" (slurm) or "qdel JOBID" (PBSPro/torque) via pyiron
+        This command is "dumb" by default, it will attempt to kill anything that it is called on, regardless of job-status.
+        """
```
pmrv: The docstring can be moved to `JobCore.kill` and then this method removed.

pyiron_base/jobs/job/core.py
```diff
@@ -429,6 +431,50 @@ def remove(self, _protect_childs=True):
...
+
+        if enable:
+            base_kill()
+        else:
+            if self.status.running or self.status.submitted:
+                base_kill()
+            else:
+                raise ValueError(
+                    "The kill() function is only available during the execution of the job."
+                )
```
pmrv: Not sure I follow this logic.  If `enable` is given, we attempt to kill the job regardless of status and if not we only do it during the `running` or `submitted` status.  Given that `kill` only makes sense during those anyway and `enable` is meant only as footgun protection, I'd prefer more some like `if enable and (self.status.running or self.status.submitted)` and `enable=False` by default.

pyiron_base/jobs/job/core.py
```diff
@@ -444,13 +490,8 @@ def remove_child(self):
...
         # Delete job from HPC-computing-queue if it is still running.
         job_status = str(self.status)
-        if (
-            job_status in ["submitted", "running", "collect"]
-            and "server" in self.project_hdf5.list_nodes()
-        ):
-            server_hdf_dict = self.project_hdf5["server"]
-            if "qid" in server_hdf_dict.keys() and server_hdf_dict["qid"] is not None:
-                self.project.queue_delete_job(server_hdf_dict["qid"])
+        if job_status in ["submitted", "running", "collect"]:
```
pmrv, line 506: Synchronize the job status checking with the kill method.  Here it's 'submitted', 'running', and 'collect' (which makes sense imo), but the kill method doesn't check for 'collect'.

pmrv (approved):
Some comments on `kill(enable=...)` and some nits, but generall lgtm.

pyiron_base/jobs/job/core.py
```diff
@@ -429,6 +431,50 @@ def remove(self, _protect_childs=True):
         # After all children are deleted, remove the job itself.
         self.remove_child()
 
+    def kill(self, enable=True):
```
pmrv:
```suggestion
    def kill(self, enable=False):
```
Because the I guess the point was to avoid people accidentally calling it.

pyiron_base/jobs/job/core.py
ligerzero-ai: I guess the point was that the status checking isn't foolproof. In an ideal world job statuses are rock-solid reliable and the force-kill option wouldn't be necessary, but there are some cases where this is not the case (e.g. Leimeroth's experience with the remote setup). Another example would be that the jobtable isn't updated with "aborted" when jobs hit walltime. 

There should be an option to kill the job regardless of the job status until we reach the point that funny business with job statuses is no longer the case.

pyiron_base/jobs/job/core.py
ligerzero-ai: See above - the enable would be a "force" option to attempt the job kill with no safeties regarding job status. I'll switch it to False by default (statuses are checked before job kill is attempted), as you suggest, but the `enable` flag here is not to protect the user, but to safeguard against killing jobs which statuses are not in the list.

pyiron_base/jobs/job/core.py
ligerzero-ai: Resolved in latest commit

pyiron_base/jobs/job/core.py
pmrv: Oh, than I had always misunderstood this, my bad.  With this intention the 'force' or whatever we called it earlier is a better name.

PR #878, opened 2022-10-25
Queue table compatible to File table

jan-janssen:


jan-janssen:
Requires https://github.com/pyiron/pysqa/pull/143

pyiron_base/project/generic.py
```diff
@@ -902,11 +902,18 @@ def queue_table(self, project_only=True, recursive=True, full_table=False):
         Returns:
             pandas.DataFrame: Output from the queuing system - optimized for the Sun grid engine
         """
-        return queue_table(
-            job_ids=self.get_job_ids(recursive=recursive),
-            project_only=project_only,
-            full_table=full_table,
-        )
+        if not isinstance(self.db, FileTable):
```
pmrv, line 905: This pattern of explicit typechecking and then dispatching on the type makes it extremely hard to write generic interfaces and generic code using them.  This is not the only instance (#336 tracked more).  It seems to me that `queue_table` would have been better as a method on the database objects.

Issue #768, opened 2022-06-08
updating pyiron table broken


Leimeroth:
Calling the run() or update_table() methods after adding another function using table.add["otherFunc"] = otherFunc on a pyiron table does not seem to change the table.

pmrv:
Fixed in #436

PR #882, opened 2022-10-30
Pyiron Table Refactor

pmrv:
I got annoyed by the tables again, so I bit the bullet and cleaned up. 

Short list of changes:
1. missing values are no longer represented by `'-'`, but by `None`.  This has the advantage that it doesn't mess up the datatype of the pandas arrays.  (`'-'` forces `dtype=object`, but `None` doesn't)
2. string values are not coerced with `eval`.  `eval` is generally not nice, but I also lost data because of this, see commit messages
3. refactor of `create_table` and bug fixing so that updating the table inplace actually works now.
4. bunch of smaller rearragements and docstrings
5. There was a whole feature set `project_level`, but it was broken/disabled and I couldn't understand what the point was, so I axed it. If someone complains about it, we can re-add it.

### DIFF
--- a/pyiron_base/jobs/datamining.py
+++ b/pyiron_base/jobs/datamining.py
@@ -317,17 +317,12 @@ class PyironTable(HasGroups):
         elif len(df_new_ids) > 0:
             self._df = df_new_ids
 
-    def convert_dict(self, input_dict):
-        return {key: self.str_to_value(value) for key, value in input_dict.items()}
-
     def refill_dict(self, diff_dict_lst):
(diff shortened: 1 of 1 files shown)


pmrv:
- [ ] Add eval as an option
- [ ] ~idea: find differences between jobs and put columns for that~  Moved to [here](https://github.com/pyiron/pyiron_atomistics/issues/878).

jan-janssen:
Maybe we should add a couple of tests to the pyiron table, to test that functionality like updating an existing table actually works.

pmrv:
Fixes #768

PR #888, opened 2022-11-07
Set new job name when copying db entry

pmrv:
Previously the function kept the old job name, because it modifies the db_entry and didn't overwrite it.

That meant that when copying a job inside the same directory to a new name, it would keep the old name in the database which leads to all kinds of funny inconsistencies.

### DIFF
--- a/pyiron_base/jobs/job/util.py
+++ b/pyiron_base/jobs/job/util.py
@@ -37,6 +37,8 @@ def _copy_database_entry(new_job_core, job_copied_id, new_database_entry=True):
     if new_database_entry:
         db_entry = new_job_core.project.db.get_item_by_id(job_copied_id)
         if db_entry is not None:
+            db_entry["job"] = new_job_core.job_name
+            db_entry["subjob"] = new_job_core.project_hdf5.h5_path
             db_entry["project"] = new_job_core.project_hdf5.project_path
             db_entry["projectpath"] = new_job_core.project_hdf5.root_path
             db_entry["subjob"] = new_job_core.project_hdf5.h5_path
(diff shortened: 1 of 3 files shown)


pmrv:
The failing unit test copies a job in the same project with a new name and then checks the there's two jobs with the original name.  ... that's just wrong or am I missing something? @jan-janssen @niklassiemer

pyiron_base/jobs/job/util.py
```diff
@@ -37,6 +37,8 @@ def _copy_database_entry(new_job_core, job_copied_id, new_database_entry=True):
     if new_database_entry:
         db_entry = new_job_core.project.db.get_item_by_id(job_copied_id)
         if db_entry is not None:
+            db_entry["job"] = new_job_core.job_name
+            db_entry["subjob"] = new_job_core.project_hdf5.h5_path
```
ligerzero-ai, line 41: Is this not duplicated with below?

pyiron_base/jobs/job/util.py
```diff
@@ -37,6 +37,8 @@ def _copy_database_entry(new_job_core, job_copied_id, new_database_entry=True):
...
             db_entry["project"] = new_job_core.project_hdf5.project_path
             db_entry["projectpath"] = new_job_core.project_hdf5.root_path
             db_entry["subjob"] = new_job_core.project_hdf5.h5_path
```
ligerzero-ai: see above comment

pyiron_base/jobs/job/util.py
pmrv: Yeah, that's true. 👍

PR #889, opened 2022-11-07
hdfio: If the initial write fails with a BlockIOError try again after 1 second.

jan-janssen:
Rather than adding the try and except statement to every HDF5 file access, a central set of `read_hdf5()` and `write_hdf5()` functions are defined.

### DIFF
--- a/pyiron_base/storage/hdfio.py
+++ b/pyiron_base/storage/hdfio.py
@@ -1473,7 +1473,7 @@ class ProjectHDFio(FileHDFio):
         return self._project.__class__(path=self.file_path)
 
 
-def read_hdf5(fname, title='h5io', slash='ignore')
+def read_hdf5(fname, title='h5io', slash='ignore'):
     try:
         return h5io.read_hdf5(fname=fname, title=title, slash=slash)
     except BlockingIOError:
--- a/pyiron_base/storage/hdfio.py
+++ b/pyiron_base/storage/hdfio.py
@@ -1473,7 +1473,7 @@ class ProjectHDFio(FileHDFio):
         return self._project.__class__(path=self.file_path)
 
 
-def read_hdf5(fname, title='h5io', slash='ignore'):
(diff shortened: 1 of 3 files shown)


pyiron_base/storage/hdfio.py
```diff
@@ -1481,3 +1471,43 @@ def create_project_from_hdf5(self):
...
         """
         return self._project.__class__(path=self.file_path)
+
+
+def read_hdf5(fname, title="h5io", slash="ignore"):
+    try:
+        return h5io.read_hdf5(fname=fname, title=title, slash=slash)
+    except BlockingIOError:
+        time.sleep(1)
+        return read_hdf5(fname=fname, title=title, slash=slash)
```
pmrv:
```suggestion
    delay = 1
    while True:
        try:
            return h5io.read_hdf5(fname=fname, title=title, slash=slash)
        except BlockingIOError:
            time.sleep(delay)
            delay *= 2
```

Presumably, if a lot of processes are contesting a file and they all wait the same amount of time, the second try can also fail.  I wouldn't be worried about the potential endless loop because of the exponential back-off, but we can also put a `max_tries` and then fail with a well defined error.

jan-janssen: I guess the difference is that when 10 processes try to access the same file at the same time, then one would be able to start, while the other 9 take an additional iteration. Then again one starts while the other 8 do an additional iteration and so on. As always one process starts I do not think that we need the exponential wait. In particular, when all jobs wait the same amount of time, the exponential is not better than the linear approach.

pmrv: Ah, your version was recursive anyway, I just missed that.  File access should be pretty quick usually, so I think we can even make the delay less than 1s.

jan-janssen:
@pmrv Do you want to take another look? I would like to merge this for the next release.

pmrv (approved):
Looks good to me.

PR #894, opened 2022-11-10
Adds tail method to check output files & transparent support for compressed jobs

pmrv:
With this you can do, e.g.

```python
job.tail('error.out', lines=50)
...
```

This and `job['error.out']` also work when the job is compressed.


jan-janssen:
To accelerate the backwards reading, there is some work from the `pymatgen` developers in their `monty` package: 
https://pythonhosted.org/monty/_modules/monty/io.html#reverse_readline

pmrv:
I'm not opposed to using `monty` as such, but for a single function it may be a bit much to pull in a full dependency.  As I expect people to use this for the (smaller) log files mostly anyway, I would keep it this way until someone complains they ran out of memory.

jan-janssen:
> I'm not opposed to using `monty` as such, but for a single function it may be a bit much to pull in a full dependency. 

As we have pymatgen as dependency for `pyiron_atomistics` which itself depends on `monty` I think having `monty` as dependency for `pyiron_base` is not too much of an overhead.

pmrv:
> 

Ah, fair.  I didn't know we already have it.

ligerzero-ai (approved):
Nothing stands out to me.

pmrv:
The failure on windows is due to a bug in `monty`.  I have opened an [issue](https://github.com/materialsvirtuallab/monty/issues/471) with them and added a small workaround+comment so this doesn't have to wait on their fix.

PR #900, opened 2022-11-15
pyiron table: allow users to set job_status_list

jan-janssen:


pmrv:
This is also part of the change set in #882.  I'll try to get it in quickly after the we are back from the retreat.

PR #901, opened 2022-11-16
Separate relocate_hdf

samwaseda:
Separate the rewriting of the hdf address, in order for other jobs to be able to intervene (e.g. the new SPHInX implementation that requires an extra loading of lazy mode for the input)

### DIFF
--- a/pyiron_base/jobs/job/core.py
+++ b/pyiron_base/jobs/job/core.py
@@ -297,6 +297,12 @@ class JobCore(HasGroups):
         """
         self._hdf5 = project.copy()
 
+    def rewrite_hdf5(self):
+        self.project_hdf5.remove_group()
+        self.project_hdf5 = self.project_hdf5.__class__(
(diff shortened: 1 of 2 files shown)


pmrv:
If I understand this correctly, the original code works because `job.project_hdf5` points into the HDF file of the master job and this then resets it to point into its own file.  Is this correct?  If so, I would prefer to call this `reset_hdf5` or something like this.   Otherwise lgtm.

samwaseda:
> If I understand this correctly, the original code works because `job.project_hdf5` points into the HDF file of the master job and this then resets it to point into its own file. Is this correct? If so, I would prefer to call this `reset_hdf5` or something like this. Otherwise lgtm.

Well, I would rather say the original one works because the content of the HDF file is (supposed to be) stored in the job attributes (such as `job.input`) when the child job is loaded, and therefore the job doesn't care about the HDF file anymore. Of course this is not the case for the lazy loading.

Issue #848, opened 2022-10-07
Hashing of `DataContainer`


samwaseda:
There was a pyiron meeting (where @pmrv was unfortunately not present), where we talked about hashing `DataContainer`, which would allow the user to set a unique job name for each job containing a unique input data without much thinking about it, i.e. something like `job = pr.create.job.MyJob(('my_job', job.input.get_hash()))`. Now after my quick research, I found [this one](https://www.doc.ic.ac.uk/~nuric/coding/how-to-hash-a-dictionary-in-python.html) very simple and elegant. However, I'm not really sure how to handle the ordering of entries, which doesn't matter in most of the codes (e.g. VASP), but does matter in some codes (e.g. LAMMPS), and the thing becomes even more complicated when it partially matters (e.g. SPHInX). Let me know what you think.

pmrv:
As far as `DataContainer` is concerned order matters and should be reflected in the "hash".

pmrv:
Maybe we can add a little interface `HasDigest` that gives a hash and classes where ordering doesn't matter can appropriately overwrite it.

samwaseda:
Right now I'm thinking about hashing the content of what would be stored in HDF. For this, I'm now trying to create a pseudo-HDF class to get the hdf content, which is then converted to `dict` to hash it. I will probably have to redefine a lot of functions already defined in `hdfio.py`. Do you have a better idea?

samwaseda:
So, here's my first implementation, which works only if the hdf already exists:

```python
import numpy as np
from typing import Dict, Any
import hashlib
import json

def dict_hash(dictionary: Dict[str, Any]) -> str:
    dhash = hashlib.md5()
    encoded = json.dumps(dictionary, sort_keys=True).encode()
    dhash.update(encoded)
    return dhash.hexdigest()

def serialize(v):
    if isinstance(v, np.ndarray):
        return v.tolist()
    return v

def to_dict(s):
    results = {k: serialize(s[k]) for k in s.list_nodes()}
    for k in s.list_groups():
        results[k] = to_dict(s[k])
    return results
```

This gives:

```python
from pyiron_atomistics import Project

pr = Project('HASH')
spx = pr.create.job.Sphinx('spx')
spx.structure = pr.create.structure.bulk('Al', cubic=True)
spx.run()
print(dict_hash(to_dict(spx['input'])))
```

Output: `382a062448b9052727d703074988e860`

samwaseda:
[...]

PR #903, opened 2022-11-17
Create hash from data content

samwaseda:
Based on the discussion on [this page](https://github.com/pyiron/pyiron_base/issues/848) I started creating the hashing function. I created a new file, because it should not be only applicable for HDF, but also for `DataContainer`.


samwaseda:
Following the discussion on the issue & tests given here it should be relatively clear what this does, but here's a small example of how it should be used:

```python
spx = Project('.').create.job.Sphinx('spx')
spx.input.get_hash()
```

Output

```
'f1e1623734363a61b0f636df3d28d3c5'
```

pmrv:
I still have a few basic problems with this approach.

1. `PseudoHDF` is likely to simple for classes with more complicated `to_hdf` methods.  Basically as soon as I fix [this issue](https://github.com/pyiron/pyiron_base/issues/364) it will break the hashing code here.
2. Have you checked a bit the memory consumption of this, especially on larger objects like whole jobs? Conceptually this would at least take double the memory of whatever you're hashing.

If we anyhow adopt this, I think it should go into the `HasHDF` interface instead of being directly on the `DataContainer`, since that is the only thing it depends on.

samwaseda:
> I still have a few basic problems with this approach.
> 
> 1. `PseudoHDF` is likely to simple for classes with more complicated `to_hdf` methods.  Basically as soon as I fix [this issue](https://github.com/pyiron/pyiron_base/issues/364) it will break the hashing code here.

Yep, I'm sure it's true, on the other hand I also think we should move towards unifying everything under `HasHDF`, so those who see the necessity to add more functions can do it.

> 2. Have you checked a bit the memory consumption of this, especially on larger objects like whole jobs? Conceptually this would at least take double the memory of whatever you're hashing.

Now it doesn't dump the entire dictionary anymore - each level is hashed separately. Not sure how much memory it could take, but in general it should be less than the double the memory.

> 
> If we anyhow adopt this, I think it should go into the `HasHDF` interface instead of being directly on the `DataContainer`, since that is the only thing it depends on.

Good point, I'm gonna do it.

samwaseda:
Now the full LAMMPS and VASP job can be hashed:

```python
from pyiron_base.storage.hash import PseudoHDF, get_hash

def hash_job(job):
    hdf = PseudoHDF()
    job.to_hdf(hdf)
    return get_hash(hdf)

lmp = Project('.').create.job.Lammps('lmp')

print(hash_job(lmp))
```

Output: `2b8fb6e54b515cd8043a35209bf0fb5e`

There's just a weird thing: For some reason I have to refresh the notebook when I try to do the same thing for two different jobs (i.e. with different names). I don't really understand why...

samwaseda:
@pmrv ping

pyiron_base/storage/hash.py
```diff
@@ -0,0 +1,78 @@
...
+def digest(h, sort_keys=True):
+    return base64.b32encode(
+        hashlib.sha1(
+            json.dumps(h, sort_keys=sort_keys).encode("utf-8")
+        ).digest()
+    ).decode("utf-8")
+
+
+def get_hash(h, sort_keys=True):
+    if hasattr(h, "items"):
```
pmrv:
```suggestion
    if isinstance(h, collections.abc.Mapping):
```

pyiron_base/storage/hash.py
```diff
@@ -0,0 +1,79 @@
...
+__author__ = "Joerg Neugebauer, Jan Janssen"
+__copyright__ = (
+    "Copyright 2020, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Jan Janssen"
+__email__ = "janssen@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
```
liamhuber: This is all copypastad and out of date?

tests/storage/test_hash.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        self.assertNotEqual(
+            hdf_1.get_hash(sort_keys=False),
+            hdf_2.get_hash(sort_keys=False),
+            msg='Order must matter'
+        )
+
+
+if __name__ == '__main__':
+    unittest.main()
```
liamhuber, line 43: unnecessary.

liamhuber (commented):
I'll review again once #995 is merged into this branch and I can actually play with it a bit.

tests/storage/test_hash.py
samwaseda: This is necessary on terminal

pyiron_base/storage/hash.py
samwaseda: I should do it more discretely next time...

tests/storage/test_hash.py
liamhuber: Fiiiiiine.

😝 but seriously, good catch. I'm curious, does getting rid of it really break things when explicitly running and discovering test from the command line, i.e. `python -m unittest path/to/file.py`, or only when brute-force executing the file, i.e. `python path/to/file.py`? If it only borks the latter....I'm actually still kind of in favour of getting rid of these lines, although I wouldn't absolutely demand it.

pyiron_base/storage/hash.py
```diff
@@ -0,0 +1,79 @@
...
+Create hash for a given dictionary
+"""
+
+import hashlib
+import json
+import numpy as np
+import base64
+import collections
+
+__author__ = "Joerg Neugebauer, Jan Janssen"
```
liamhuber:
```suggestion
__author__ = "Sam Waseda"
```

pyiron_base/storage/hash.py
```diff
@@ -0,0 +1,79 @@
...
+__copyright__ = (
+    "Copyright 2020, Max-Planck-Institut für Eisenforschung GmbH - "
```
liamhuber:
```suggestion
    "Copyright 2023, Max-Planck-Institut für Eisenforschung GmbH - "
```

liamhuber (changes requested):
> Now the full LAMMPS and VASP job can be hashed:
> 
> ```python
> from pyiron_base.storage.hash import PseudoHDF, get_hash
> from pyiron_atomistics import Project
> 
> def hash_job(job):
>     hdf = PseudoHDF()
>     job.to_hdf(hdf)
>     return get_hash(hdf)
> 
> lmp = Project('.').create.job.Lammps('lmp')
> 
> print(hash_job(lmp))
> ```

I'm afraid I disagree with this claim. There are things that I would thus expect to return different hashes that don't. When I play around with `DataContainer` it all seems to work as I expect (and as you test), but when I apply it to the intended use-case of jobs it totally breaks.

I'm not clear if there is a good workaround here for this, or if we need to wait until jobs are rebased to really _extensively_ use `self.input: DataContainer` (`output`, and whatever else) and then introduce a `GenericJob.get_hash()` method that combines the hash of all the relevant data containers.

Header:
```python
from pyiron_base.storage.hash import PseudoHDF, get_hash

pr = Project('scratch')
pr.remove_jobs_silently()

def hash_job(job):
    hdf = PseudoHDF()
    job.to_hdf(hdf)
    return get_hash(hdf)
```

Examples I would expect to return different hashes but return the same hash:
```python
lmp = pr.create.job.Lammps('lmp')
lmp.structure = pr.atomistics.structure.bulk('Al')
print(hash_job(lmp))

lmp2 = pr.create.job.Lammps('lmp2')
lmp2.structure = pr.atomistics.structure.bulk('Fe')
print(hash_job(lmp2))
```

```python
lmp = pr.create.job.Lammps('lmp')
lmp.calc_md(temperature=300)
print(hash_job(lmp))

lmp2 = pr.create.job.Lammps('lmp2')
lmp2.calc_md(temperature=600)
print(hash_job(lmp2))
```

Less important but still maybe should be returning different hashes?

```python
lmp = pr.create.job.Lammps('lmp')
lmp.server.cores = 10
print(hash_job(lmp))

lmp2 = pr.create.job.Lammps('lmp2')
lmp2.server.cores = 20
print(hash_job(lmp2))
```

I am not totally closed to merging this functionality, but at the very least we would need strong warnings that it breaks for the most obvious use case (jobs).

pyiron_base/storage/hash.py
samwaseda: oh no I'm getting so old...

samwaseda:
> I'm afraid I disagree with this claim. There are things that I would thus expect to return different hashes that don't. When I play around with `DataContainer` it all seems to work as I expect (and as you test), but when I apply it to the intended use-case of jobs it totally breaks.

Ahaaa I realized that the changes I made in [this PR](https://github.com/pyiron/pyiron_base/pull/995) do not do what I had expected. In short, the HDF file becomes empty, probably because `job._hdf` is `None`. Let me check if I find a workaround. This thing is a mess XD

pmrv:
What I didn't like about #995 is that it makes a code change to support this PR in a location far away and in manner that is very difficult to understand if you don't know about this PR.  I haven't come up with a good solution yet, but I would rather have this somehow integrated into the `HasHDF` class, because it'll make the code much more localized.

samwaseda:
> What I didn't like about #995 is that it makes a code change to support this PR in a location far away and in manner that is very difficult to understand if you don't know about this PR. I haven't come up with a good solution yet, but I would rather have this somehow integrated into the `HasHDF` class, because it'll make the code much more localized.

The fix in #995 actually didn't fix anything. We found out that there are two fundmental problems here: 1, `to_hdf` does a lot more than just storing data in hdf; 2, there is no clear input in pyiron.

I am going to freeze this PR as long as these problems are not resolved.

liamhuber:
> The fix in https://github.com/pyiron/pyiron_base/pull/995 actually didn't fix anything.

Nit: it did, it was necessary in order to hash two different jobs consecutively. Once this was possible it became easy to see that hashing didn't capture all the input we wanted (due to your point (2)).

samwaseda:
I close this one for now, because rather than creating a new workaround it would be cleaner to make the new unified input structure with [this](https://github.com/pyiron/pyiron_base/issues/1051).

PR #908, opened 2022-11-24
Create new `GenericParent` to unify `GenericMaster`, `ParallelMaster`, `SerialMaster`, `InteractiveWrapper` and `FlexibleMaster`

samwaseda:
Following the recent discussion at the pyiron meeting, I'm writing a new generic master, which I called generic parent for now.

In my naive imagination, it should be used like this:

```python
from pyiron_base import GenericParent

class MyParent(GenericParent):
    def __init__(self, *args, **argv):
        super().__init__(*args, **argv)
        self.input.my_parameters = []

    def job_generator(self, template_job):
        if len(self) == len(self.input.my_parameters):
            return None
        job_lst = []
        for p in self.input.my_parameters:
            job = template_job.copy()
            job.job_name = '_'.join('my_job', str(p))
            job.input.some_parameter = p
            job_lst.append(job)
        return job_lst
    
    def validate_ready_to_run(self):
        super().validate_ready_to_run()
        if len(job.input.my_parameters) == 0:
            raise ValueError('Parameters not set!')
            

pr = Project('DEMO')
child = pr.create.job.MyJob('my_child')
parent = child.create_job(MyParent, 'my_parent')
parent.input.a = [1, 2, 3]
parent.run()
```

- `self.input` and `self.output` are pre-defined, which are automatically stored and loaded, just like in `TemplateJob`
- As @liamhuber suggested, there are three types that can be returned:
  - `list`: List of jobs, which run like `ParallelMaster`
  - `GenericJob` / `GenericParent`:  One job that runs like `SerialMaster`
  - Anything else (e.g. `None`): Stop calculation
- If @liamhuber and I are on the same page, `template_job` in `generate_job` is either the copy of `self.ref_job` for the first step (like in `ParallelMaster`) or the latest job (like in `SerialMaster`).
- Job names are updated internally, i.e. if not a list of jobs is returned, there should be no need to change the name (i.e. what I'm doing in the example above should become redundant)
- It should become compatible with the interactive mode.


samwaseda:
I opened the PR, because I wanted to initiate the discussion, but the code itself is only the copy of `GenericMaster` + `ParellelMaster`, so as long as it's draft it's not worth taking a look at.

samwaseda:
I just realised that `GenericMaster` can currently append several reference jobs, even though the variable `ref_job` returns only the first one. I'm actually extremely uncomfortable with this behaviour.

Anyway, based on this, it's probably wiser to send all the jobs to `job_generator` at once, meaning it can be either a job, or a list of jobs depending on how the user used it.

liamhuber:
> I just realised that `GenericMaster` can currently append several reference jobs, even though the variable `ref_job` returns only the first one. I'm actually extremely uncomfortable with this behaviour.

Yyyyyyup. Saw that when I was looking at the code during the meeting. Super scary 😬

liamhuber:
Intro looks good to me, I think we're on the same page.

The only thing I don't really like is only really tangentially related:
```python
parent = child.create_job...
```
What does this even mean? It's certainly counter-intuitive that the parent springs from the loins of the child. Why not `_ = parent.create_job...` automatically populating `parent.ref_job` with the resulting object? Finally, it messes with our `pr.create.job` syntax. IMO we should either get rid of this bit of syntactic "sugar" altogether, or seriously re-imagine it 

In this context I think those complaints literally only apply to your demo text though, where I'd just request that any example docstrings follow a format more like:
```python
child = pr.create.job...
parent = pr.create.job...
...
parent.ref_job = child
```

samwaseda:
Following [@liamhuber's comment](https://github.com/pyiron/pyiron_base/issues/798https://github.com/pyiron/pyiron_base/issues/798), there should be a job generator setter, which reads a function, just like in `SerialMaster`. so the pseudo-code should look like:

```python
import inspect

class GenericParent(GenericJob):
    ...
    def set_job_generator(self, job_generator):
        self.input._job_generator = inspect.getsource(job_generator)
        
    def job_generator(self, template_job):
        if self.input._job_generator is None:
            # This error message is not shown anyway if subclassed
            raise NotImplementedError('job_generator must be subclassed or set via set_job_generator')
        return self.input._job_generator(self, template_job)
```

With this implementation, we can define a function and launch simulations via:

```python
def my_job_generator(parent, template_job):
    job_lst = []
    for p in parent.input.my_parameters:
        job = template_job.copy()
        job.job_name = '_'.join('my_job', str(p))
        job.input.some_parameter = p
        job_lst.append(job)
    return job_lst

my_parent = pr.create.job.GenericParent('my_parent')
my_parent.ref_job = ref_job_of_my_choice
my_parent.input.my_parameters = [1, 2, 3]
my_parent.set_job_generator(my_job_generator)
my_parent.run()
```

samwaseda:
I close this one anyway because I think it makes more sense to define the Master jobs in terms of the new workflow concept.

PR #914, opened 2022-11-28
Job docstrings

jan-janssen:

### DIFF
--- a/pyiron_base/jobs/job/core.py
+++ b/pyiron_base/jobs/job/core.py
@@ -104,13 +104,19 @@ _doc_str_job_core_attr = """\
 
 
 class JobCore(HasGroups):
-    __doc__ = ("""
+    __doc__ = (
+        """
     The JobCore the most fundamental pyiron job class. From this class the GenericJob as well as the reduced 
     JobPath class are derived. While JobPath only provides access to the HDF5 file it is about one order faster.
     
     Implements :class:`.HasGroups`.  Groups are HDF groups in the HDF file associated with the job and any 
     child jobs, nodes are HDF dataset in the HDF file.
-""" + "\n" + _doc_str_job_core_args + "\n" + _doc_str_job_core_attr)
+"""
+        + "\n"
+        + _doc_str_job_core_args
+        + "\n"
+        + _doc_str_job_core_attr
+    )
 
     def __init__(self, project, job_name):
         job_name = _get_safe_job_name(job_name)
--- a/pyiron_base/jobs/job/core.py
+++ b/pyiron_base/jobs/job/core.py
@@ -44,18 +44,14 @@ __status__ = "production"
 __date__ = "Sep 1, 2017"
 
 
-class JobCore(HasGroups):
-    """
-    The JobCore the most fundamental pyiron job class. From this class the GenericJob as well as the reduced JobPath
-    class are derived. While JobPath only provides access to the HDF5 file it is about one order faster.
-
-    Implements :class:`.HasGroups`.  Groups are HDF groups in the HDF file associated with the job and any child jobs,
-    nodes are HDF dataset in the HDF file.
-
+_job_core_args = """\
     Args:
         project (ProjectHDFio): ProjectHDFio instance which points to the HDF5 file the job is stored in
         job_name (str): name of the job, which has to be unique within the project
 
+"""
+
+_job_core_attr = """\
     Attributes:
 
         .. attribute:: job_name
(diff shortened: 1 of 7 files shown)


pyiron_base/jobs/job/core.py
```diff
@@ -44,18 +44,15 @@
...
+# Modular Docstrings
+_doc_str_job_core_args = """\
     Args:
         project (ProjectHDFio): ProjectHDFio instance which points to the HDF5 file the job is stored in
         job_name (str): name of the job, which has to be unique within the project
 
+"""
+
+_doc_str_job_core_attr = """\
     Attributes:
```
liamhuber, line 57: Attributes can (and for consistency should) also follow the [google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). I really like this centralization of the docstrings, because now we only need to make the change in one place. Let's take the opportunity to make it now though -- google is way easy to read.

pyiron_base/jobs/job/core.py
```diff
@@ -103,7 +100,23 @@ class are derived. While JobPath only provides access to the HDF5 file it is abo
...
+    JobPath class are derived. While JobPath only provides access to the HDF5 file it is about one order faster.
+    
+    Implements :class:`.HasGroups`.  Groups are HDF groups in the HDF file associated with the job and any 
+    child jobs, nodes are HDF dataset in the HDF file.
+"""
+        + "\n"
+        + _doc_str_job_core_args
+        + "\n"
+        + _doc_str_job_core_attr
+    )
```
liamhuber, line 119: So I'm suspicious that there's "sexier" a way to accomplish this with `functools`, but I don't know it by heart. To that end, the perfect shouldn't be the enemy of the good and I'm super happy to see this ~code~ docstring repetition get demolished like this 🎉

pyiron_base/jobs/master/list.py
```diff
@@ -24,108 +25,31 @@
...
 
+# Modular Docstrings
+_doc_str_list_master_attr = (
+    _doc_str_generic_master_attr
+    + "\n"
+    + """\
+        .. attribute:: submission_status
+
+            Monitors how many jobs have been submitted and how many have to be submitted in future
+"""
```
liamhuber, line 36:
```suggestion
    # Attributes:
    + """\
        submission_status (SubmissionStatus): Monitors how many jobs have been submitted and how many have to be submitted in future
"""
```

e.g. for google conversion. Not sure the indentation lines up and black might want to break the line, but you get the idea.

liamhuber (approved):
Super. Code repetition is the devil, so I love this. Now that it's centralized I'd be even happier to see it take the more readable google style. I think now is a good time for that while you're working on it, but if you don't have time this week I'm content to defer that to a later PR.

jan-janssen:
> Super. Code repetition is the devil, so I love this. Now that it's centralized I'd be even happier to see it take the more readable google style. 

If I remember correctly, then sphinx on readthedocs did not support the google style when we tested it before - which is easily five years ago. So I would merge these changes without changing the formatting and then test the transition to Google Style for attributes in a separate pull request.

pmrv:
I suppose it's too late now, but I want to say that I do not like this way of docstrings at all.  

It's very difficult to read and I don't see it as very extendable and/or very easy to mess up for the next person that has to make a change here.

I guess the intent is to not have the strings that document the attributes duplicated.  Imo there's no point in doing that in the first place.  People will either check `job.attribute?` if they are in the notebook and if they are on the online help super classes are linked automatically so you can easily browse attributes defined by the super class as well.

liamhuber:
@pmrv those are absolutely valid concerns. My take would be that if we leave it defined in multiple places it *will* wind up being incorrect, and IMO bad or awkward docstrings are preferable to wrong or inconsistent docstrings.

> I guess the intent is to not have the strings that document the attributes duplicated. Imo there's no point in doing that in the first place. People will either check `job.attribute?` if they are in the notebook

Can you give some pseudocode for that? I can see obviously how it works for properties, but not for attributes generically. On [stackoverflow](https://stackoverflow.com/questions/3051241/how-to-document-class-attributes-in-python) I see a way to do it for class level attributes, but that is still not sufficient and came from a rejected PEP (apparently).

pmrv:
You are right it will work for properties only, but I think that's already most of the attributes users should see.  I can only think of `job.input` and such that are not properties, but they are complex objects already on their own that can carry docstrings.  My point is more that each class should only document the attributes it newly introduces.

liamhuber:
> You are right it will work for properties only, but I think that's already most of the attributes users should see. I can only think of `job.input` and such that are not properties, but they are complex objects already on their own that can carry docstrings. My point is more that each class should only document the attributes it newly introduces.

As long as the attributes all wind up with documentation attached, I am on board with this. My descending hierarchy of concerns would go:
- Documentation is not duplicated
- Documentation exists
- Documentation is elegant

Directly documenting the properties is more elegant, so I'm all for it, but this PR removed duplication and that's more important to me. Let's move towards what you suggest (and actually if/when we use traitlets more, there is explicit fields for supplying trait help anyhow), but use keep this as an intermediate stage.

PR #916, opened 2022-11-29
Implement a job.input for GenericMaster

jan-janssen:

### DIFF
--- a/pyiron_base/jobs/master/generic.py
+++ b/pyiron_base/jobs/master/generic.py
@@ -176,7 +176,7 @@ class GenericMaster(GenericJob):
         if isinstance(new_input, GenericParameters):
             self._input = new_input
         else:
-            raise TypeError(f"Expected a GenericParameters object but got {new_input.__class__})
+            raise TypeError(f"Expected a GenericParameters object but got {new_input.__class__}")
 
     def child_hdf(self, job_name):
         """
--- a/pyiron_base/jobs/master/generic.py
+++ b/pyiron_base/jobs/master/generic.py
@@ -9,6 +9,7 @@ import inspect
 import textwrap
 
 from pyiron_base.jobs.job.generic import GenericJob
+from pyiron_base.storage.parameters import GenericParameters
 from pyiron_base.jobs.job.extension.jobstatus import job_status_finished_lst
 from pyiron_base.jobs.job.jobtype import JobType
 
@@ -123,6 +124,7 @@ class GenericMaster(GenericJob):
 
     def __init__(self, project, job_name):
         super(GenericMaster, self).__init__(project, job_name=job_name)
+        self._input = GenericParameters("parameters")
         self._job_name_lst = []
         self._job_object_dict = {}
         self._child_id_func = None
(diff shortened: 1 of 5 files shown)


pyiron_base/jobs/master/serial.py
```diff
@@ -176,41 +173,6 @@ def ref_job(self):
...
-
-        Args:
-            value (GenericParameters): input of the start job
-        """
-        if self.start_job is not None:
-            self._start_job.input = value
-        else:
-            raise ValueError(
-                "Input can only be set after a start job has been assinged."
-            )
```
jan-janssen, line 133: Here the functionality changes - rather than accessing the input of the start_job - the template other jobs are generated from - we now access the input of the master, still I think this is more clear this way.

pyiron_base/jobs/master/generic.py
```diff
@@ -193,6 +203,11 @@ def job_object_dict(self):
         """
         return self._job_object_dict
 
+    def set_input_to_read_only(self):
+        self._input.read_only = True
+
+    set_input_to_read_only.__doc__ = GenericJob.set_input_to_read_only.__doc__
```
liamhuber:
```suggestion
    from functools import wraps  # TODO: Move import to head
    @wraps(GenericJob.set_input_to_read_only)
    def set_input_to_read_only(self):
        self._input.read_only = True
```

pyiron_base/jobs/master/generic.py
```diff
@@ -164,6 +166,14 @@ def child_project(self):
         else:
             return self.project.open(self.job_name + "_hdf5")
 
+    @property
+    def input(self):
+        return self._input
+
+    @input.setter
+    def input(self, new_input):
+        self._input = new_input
```
liamhuber:
```suggestion
    def input(self, new_input: GenericParameters):
        if isinstance(new_input, GenericParameters):
            self._input = new_input
        else:
            raise TypeError(f"Expected a GenericParameters object but got {new_input.__class__})
```

If we're going to bother making the input variable private, let's exploit that to do some type checking.

pyiron_base/jobs/master/serial.py
liamhuber: TBH I'm a little confused how it could be any other way. Isn't it possible/desirable/*necessary* that the serial master has different input from its reference job??

liamhuber (approved):
Another nice removal of duplicate code. Some minor improvements suggested.

I don't follow the old behaviour for the serial master (accessing input of the ref job?!), and am surprised this (positive) change doesn't have some side effects. I'm not super familiar with `SerialMaster` though, so if the tests pass and you're happy with it then I'm good.

pyiron_base/jobs/master/serial.py
jan-janssen: > TBH I'm a little confused how it could be any other way. Isn't it possible/desirable/_necessary_ that the serial master has different input from its reference job??

I guess the initial idea was that you could still access the start job from the serial master, so when you get the input you get the input of the start job and correspondingly when you get the output of the serial master you get the output of the last calculation. But I agree that separating both is more clean, so that is the way I designed this pull request.

jan-janssen:
> TBH I'm a little confused how it could be any other way. Isn't it possible/desirable/necessary that the serial master has different input from its reference job??

I agree, this part was confusing before and is more clean now.

PR #927, opened 2022-11-30
Move __getitem__() to GenericMaster

jan-janssen:

### DIFF
--- a/pyiron_base/jobs/master/flexible.py
+++ b/pyiron_base/jobs/master/flexible.py
@@ -175,16 +175,3 @@ class FlexibleMaster(GenericMaster):
                 for funct_str in funct_str_lst:
                     exec(funct_str)
                     self._step_function_lst.append(eval(funct_str.split("(")[0][4:]))
-
-    def __getitem__(self, item):
-        child_id_lst = self.child_ids
-        child_name_lst = [
-            self.project.db.get_item_by_id(child_id)["job"]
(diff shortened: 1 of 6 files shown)


pyiron_base/jobs/master/flexible.py
```diff
@@ -175,16 +175,3 @@ def from_hdf(self, hdf=None, group_name=None):
...
                     self._step_function_lst.append(eval(funct_str.split("(")[0][4:]))
-
-    def __getitem__(self, item):
-        child_id_lst = self.child_ids
-        child_name_lst = [
-            self.project.db.get_item_by_id(child_id)["job"]
-            for child_id in self.child_ids
-        ]
-        if isinstance(item, int):
-            total_lst = child_name_lst + self._job_name_lst
```
jan-janssen, line 186: The order of the jobs is the critical part, for the `ParallelMaster` is was: `total_lst = self._job_name_lst + child_name_lst` while for the `FlexibleMaster`, `InteractiveWrapper`, `ListMaster` and `SerialMaster` it was `total_lst = child_name_lst + self._job_name_lst`- now it is `total_lst = self._job_name_lst + child_name_lst` for all.

liamhuber (approved):
LGTM. Only concern I would have had is that the change in index order had a downstream side effect, but integration tests are passing so it looks like that's OK too. Another nice removal of duplicate code.

PR #929, opened 2022-12-02
Add a function that retries some operation until no exception is thrown

pmrv:
We used logic like this in the database and in HDF ( #884 ).  I've applied to one more location where the `DatabaseAccess` object tries to create the database tables, because that can cause the database to return an error as well if a lot of jobs try to do it simultaneously.

### DIFF
--- a/pyiron_base/utils/error.py
+++ b/pyiron_base/utils/error.py
@@ -112,7 +112,7 @@ def retry(
         msg: str,
         atmost: Optional[int] = None,
         delay: float = 1.0,
-        delay_factor: float = 1
+        delay_factor: float = 1.0
 ) -> T:
     """
     Try to call `func` until it no longer raises `error`.
--- a/pyiron_base/utils/error.py
+++ b/pyiron_base/utils/error.py
@@ -106,13 +106,15 @@ class ImportAlarm:
 
 
 T = TypeVar("T")
+
+
 def retry(
-        func: Callable[[], T],
-        error: Union[Type[Exception], Tuple[Type[Exception], ...]],
-        msg: str,
-        atmost: Optional[int] = None,
-        delay: float = 1.0,
-        delay_factor: float = 1.0
+    func: Callable[[], T],
+    error: Union[Type[Exception], Tuple[Type[Exception], ...]],
+    msg: str,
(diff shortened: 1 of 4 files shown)


pyiron_base/utils/error.py
```diff
@@ -97,3 +103,51 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+
+T = TypeVar("T")
+def retry(
+        func: Callable[[], T],
+        error: Union[Type[Exception], Tuple[Type[Exception], ...]],
+        msg: str,
+        atmost: Optional[int] = None,
+        delay: float = 1.0,
+        delay_factor: float = 1
```
pmrv:
```suggestion
        delay_factor: float = 1.0
```

pyiron_base/utils/error.py
```diff
@@ -97,3 +102,55 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+
+T = TypeVar("T")
+
+
+def retry(
+    func: Callable[[], T],
+    error: Union[Type[Exception], Tuple[Type[Exception], ...]],
+    msg: str,
+    atmost: Optional[int] = None,
```
liamhuber: I think `at_most` would be PEP8y-er

liamhuber (approved):
My big concern with this is that it's in the wrong place. Namely, this should probably be in one of the standard libraries. I did a quick google search and this is a reasonably common problem, but the best public solutions (i.e. the [top SO thread](https://stackoverflow.com/questions/4606919/in-python-try-until-no-error)) are rubbish compared to this, and don't offer nearly the same flexibility. The best home I can think of off-hand is `itertools`, but I have to confess that in my quick search I didn't find any portals for where to suggest changes for that library.

pyiron_base/utils/error.py
pmrv: Agree, will change it.

pmrv:
I've found [this](https://github.com/litl/backoff) from the SO link, but I don't feel like pulling a full dependency for this.

jan-janssen (approved):
Looks good to me - I guess we can merge it like this and then replace it with a separate library in a different pull request.

PR #934, opened 2022-12-05
refresh_job_status also for submitted jobs

pmrv:


samwaseda:
Shouldn't it be done from the jobs? I'm a bit concerned that it might take too much time each time, especially if a large number of simulations are submitted.

pmrv:
The alternative is for me to do `for j in pr.iter_jobs(status='submitted'): j.refresh_job_status()`, which is the same as this code but I have to type more, so I'm not concerned.  It takes about a minute for a few thousand jobs.  It could be made faster by not checking every job individually, but that's a general problem with the refresh logic.

jan-janssen (approved):
Looks good to me

samwaseda:
Would it be possible to add `include_submitted_jobs=False` in the function argument? I still think the jobs status should be controlled inside the job for submitted->running and I don't want to wait minutes for just refreshing the job table.

jan-janssen:
```
[329](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:330)
FAIL: test_worker (job.test_worker.TestScriptJob)
[330](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:331)
[331](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:332)
[332](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:333)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/job/test_worker.py", line 29, in test_worker
[333](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:334)
    self.assertEqual(len(df[df.status == "finished"]), 1)
[334](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:335)
AssertionError: 0 != 1
[335](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:336)

[336](https://github.com/pyiron/pyiron_base/actions/runs/3622385999/jobs/6107032593#step:9:337)
```

pyiron_base/project/generic.py
```diff
@@ -967,17 +967,20 @@ def get_id_from_job_table(job_table, job_path):
         else:
             return None
 
-    def refresh_job_status(self, *jobs):
+    def refresh_job_status(self, *jobs, by_status=["running", "submitted"]):
```
samwaseda, line 970: I would prefer `by_status=["running"]` for the default behaviour, but I can also live with this.

But actually which code doesn't do the transition from `submitted` to `running` automatically inside the code?

pmrv: It can happend when jobs either crash early inside pyiron or are cancelled by the user via slurm.

samwaseda: ok fine by me. I can live with this.

samwaseda (approved):

PR #935, opened 2022-12-05
Communicate to pypi that we support Python 3.11

jan-janssen:


jan-janssen:
https://github.com/pyiron/pyiron_base/pull/881

PR #938, opened 2022-12-05
On label CI

liamhuber:
Over on ironflow I'm having trouble with the organization-level Codacy secret. This step is a necessary part of getting base using the centralized CI anyhow, so I want to use it as an opportunity to play around with base's secrets to figure out if it's the central secret or ironflow that has the problem -- right now all the other repo's override the organization secret, so I'll delete it here (I have access to add it back) and see if I get the same breakage.


liamhuber:
I came here to break codacy, and now coveralls is complaining. I have to stop now, but will come back to this ASAP.

liamhuber:
With Jan's help [updating pyiron/actions](https://github.com/pyiron/actions/pull/23), I'm [debugging this on ironflow](https://github.com/pyiron/ironflow/pull/145) now. 

@mbruns91 @srmnitc, one thing I did run into here that may be of use to you: Once I started using the centralized CI, coveralls started complaining:

`coveralls.exception.CoverallsException: Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs`

Deleting the `.coveralls.yml` cleared this up. I think they were getting confused by the conflict between the github token passed in the CI and the different token in the config file. This might come up again as we get other repos running pyiron/actions, so heads-up.

PR #942, opened 2022-12-15
Adds a job.kill() functionality that stops a job in the queueing system without removal of files and updates status. 

jan-janssen:
The previous pull request https://github.com/pyiron/pyiron_base/pull/875 was reverted, as it caused the compatibility tests to fail.


ligerzero-ai:
I honestly do not understand what could possibly be causing that specific error in the RDM for pyiron contrib. I will need some help on this one.

PR #943, opened 2022-12-15
Fix the flexible master

jan-janssen:


pyiron_base/jobs/master/flexible.py
```diff
@@ -88,7 +88,9 @@ def run_static(self):
                 job = self.pop(0)
                 job._master_id = self.job_id
                 if ind != 0:
-                    prev_job = self[ind - 1]
+                    prev_job = self[
+                        self.project.db.get_item_by_id(self.child_ids[-1])["job"]
```
pmrv, line 92: Shouldn't this be at least `self.child_ids[ind-1]`?  Otherwise this will always load the same job, right?

jan-janssen: It always loads the last finished calculation, which is what the flexible master wants.

PR #944, opened 2022-12-15
Correct type hinting for job name sanitizer

liamhuber:
@samwaseda could you please also add the docstring for the last argument?

### DIFF
--- a/pyiron_base/jobs/job/util.py
+++ b/pyiron_base/jobs/job/util.py
@@ -119,7 +119,7 @@ _special_symbol_replacements = {
 def _get_safe_job_name(
         name: Union[str, tuple],
         ndigits: Union[int, None] = 8,
-        special_symbols: Union[Dict, None] = None
+        special_symbols: Union[dict, None] = None
 ):
     d_special_symbols = _special_symbol_replacements.copy()
     if special_symbols is not None:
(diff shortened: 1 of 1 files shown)


pyiron_base/jobs/job/util.py
```diff
@@ -117,8 +117,23 @@ def _get_project_for_copy(job, project, new_job_name):
 
 
 def _get_safe_job_name(
-    name: str, ndigits: Union[int, None] = 8, special_symbols: Union[Dict, None] = None
+        name: Union[str, tuple],
+        ndigits: Union[int, None] = 8,
+        special_symbols: Union[dict, None] = None
```
pmrv: Better to use `Optional`, which is just a short cut for this.

pyiron_base/jobs/job/util.py
```diff
@@ -117,8 +117,23 @@ def _get_project_for_copy(job, project, new_job_name):
...
 ):
+    """
+    Sanitize a job name, optionally appending numeric values.
+
+    Args:
+        name (str|tuple): The name to sanitize, or a tuple of the name and any number
+            of numeric values to append with '_' in between.
+        ndigits (int|None): How many digits to round any floating point values in a
+            `name` tuple to. (Default is 8; to not round at all use None.)
+        special_symbols:
```
samwaseda:
```suggestion
        special_symbols (dict|None): Conversions of special symbols to apply. This will be applied to the default conversion dict, which contains: DEFAULT_CONVERSION
```

And `DEFAULT_CONVERSION` should be programmatically replaced by `_special_symbol_replacements`

PR #948, opened 2022-12-19
Disable working directory warining

jan-janssen:


jan-janssen:
The warning was introduced in https://github.com/pyiron/pyiron_base/pull/806 now it is possible to disable the warning with the corresponding flag.

niklassiemer (approved):
Misleading title, otherwise lgtm.

Issue #513, opened 2021-11-12
test_install is super slow


liamhuber:
Base tests take ~1 minute on my local machine, and [this test alone](https://github.com/pyiron/pyiron_base/blob/14b6ea2970480404ed8a23ce78e134fe21da15ce/tests/settings/test_install.py#L11) must be at least half of that. I'd love to get this one test sped up. I might be able to take a look at it later, but I figured it doesn't hurt to make my reminder public in case anyone else is more motivated.

PR #950, opened 2022-12-19
Faster install test

liamhuber:
I added the ability to pass `pyiron_base.state.install.install_pyiron(giturl_for_zip_file=None)`, which still creates an (empty!) `resources/` directory, but doesn't download anything. Downloading the zipped resources directory from github is what was taking so much time. `tests/state/test_install.py` now takes ~3s instead of ~30s.


liamhuber:
Closes #513

PR #951, opened 2022-12-19
Tab completed loading

### LINKED ISSUE #241: Would it make sense to make jobs auto-completable?
We all face the problem that it's super difficult to remember the job names and often they are too long to type. I was wondering if you see a possibility to make it auto-completable, e.g. via `pr.load.job_name`. What do you think?

liamhuber:
`Project.load` and `Project.inspect` are now classes whose `__call__` method performs the old `load` and `inspect` behaviour. Additionally, these classes all for tab-completion of jobs at this project level (i.e. not recursively, otherwise the tab-completed names might not be unique). One might think about allowing tab completion for sub-projects, either as `pr.sub_pr.load.my_job` or `pr.load.sub_pr.my_job`, but that's outside the scope of this PR.

While I was here I did some book-keeping and changed `.load(..., convert_to_object=True)` to `.load(...)` and `.load(..., convert_to_object=False)` to `.inspect(...)`.

Closes #241


samwaseda (approved):
Looks good to me. Especially I liked the cleaning of all the `load`'s with `convert_to_object`.

pyiron_base/project/generic.py
```diff
@@ -1730,3 +1712,114 @@ def table(self, job_name="table", delete_existing_job=False):
...
+
+class JobInspector(_JobByAttribute):
+    """
+    Inspect an existing pyiron object - most commonly a job - from the database
+
+    Args:
+        job_specifier (str, int): name of the job or job ID
+
+    Returns:
+        JobCore: Access to the HDF5 object - not a GenericJob object - use load()
```
```suggestion
        JobCore: Access to the HDF5 object - not a GenericJob object - use :meth:`~.Project.load()`
```

tests/project/test_project.py
```diff
@@ -126,6 +129,30 @@ def test_maintenance_get_repository_status(self):
...
+            JobPath,
+            msg="Expected to load only an HDF interface"
+        )
+
+        inspected = self.project.inspect("toy_1")
+        self.assertIsInstance(
+            inspected,
+            JobPath,
+            msg="Expected to load only an HDF interface"
+        )
```
pmrv: Does it makes sense to test that `__getattr__` actually works, i.e. `self.assertEqual(self.project.inspect.toy_1, self.project.inspect('toy_1'))`?

pmrv (approved):
lgtm, except two small things.

pyiron_base/project/generic.py
```diff
@@ -1730,3 +1712,114 @@ def table(self, job_name="table", delete_existing_job=False):
         )
         table.analysis_project = self._project
         return table
+
+
+class _JobByAttribute(ABC):
```
jan-janssen: Does it make sense to move the new classes to a separate python module? With over 1700 lines the project module is already very lengthy.

jan-janssen:
Did you test this on a pyiron installation without database? Just to be sure we do not break existing functionality.

tests/project/test_project.py
liamhuber: Yeah, good idea. I'm just on my phone but I'll do that next time I'm working.

liamhuber: Turns out it doesn't work because neither the `JobPath` nor `GenericJob` are playing nicely with equality comparison. Instead I'm just testing that the result has the same `.name` and added a `TODO` to compare the objects directly sometime in the future once that's supported.

pyiron_base/project/generic.py
liamhuber: Yep, good call.

liamhuber:
> Did you test this on a pyiron installation without database? Just to be sure we do not break existing functionality.

So this wound up being a really important comment. I thought since I was using all the same methods and only relying on `IsDatabase` functionality that everything would be fine -- and in case you *started* without a database this might be the case, but as it stood the loader would set it's database attribute and then be irresponsive to changes in state that switch the database to a filetable.

In `[ca7d93e](https://github.com/pyiron/pyiron_base/pull/951/commits/ca7d93ecfce054015d73200918a3eefe39bf0d0e)` I address this directly, and it handles database updates just fine. But this got me worried that there may be other state updates which get successfully captured in `Project`, but which the loaders don't know about. To get around this, I just pass the whole project as an argument to the loader and use the (up-to-date) project attributes in [1450a83](https://github.com/pyiron/pyiron_base/pull/951/commits/1450a8383d3ce7d2fbafa38b2b5ec3c10e061bef). @jan-janssen, this results in the garbage-collection anti-pattern you don't like `self._loader = JobLoader(project=self)`. I played around with and without this anti-pattern the other day with toy examples and could never get excess memory usage... it could be that my test was faulty or that my example was too trivial,  but here I'd rather have the security of knowing that we're always synced with `state` rather than optimal performance anyhow.

As I write I see that some of the unit tests are failing, but even when I check out main `project.test_projectPath.TestProjectPath` is throwing errors on my local machine. I don't know if I have time to get that sorted today, but in principle as soon as the tests are passing I'm content with this PR.

pmrv:
Imo the 'anti-pattern' comes purely from our insufficient abstraction of the database.  All the attributes that you passed before are connected to the database, so ideally we would move them there and instantiate the `JobLoaders` with `project.db` instead of the whole project.

liamhuber:
> Imo the 'anti-pattern' comes purely from our insufficient abstraction of the database. All the attributes that you passed before are connected to the database, so ideally we would move them there and instantiate the `JobLoaders` with `project.db` instead of the whole project.

I think that still doesn't do the trick. `project.db` is a property that returns either `state.db` or a *new instance* of `FileTable`, depending on a configuration. The latter then depends on the project path. So we'd need at least a similar property on the loader (this smells like we might need a `HasDatabase` mixin...) and also the project path. I'm not saying it can't be done, it's just a bit of a pain with the current up-stream infrastructure.

liamhuber:
> As I write I see that some of the unit tests are failing

Ok, I think I found the issue. I'm suspicious my call to `state.update` is destroying a lucky coincidence we had on the CI (#953)... I'm going to fiddle with it to test this hunch.

liamhuber:
Ok, I think if #955 gets merged and (if I'm correct there) a PR addressing #953 is created and merged, the tests here will pass.

liamhuber:
The pandas `loc` call is still breaking. It's working fine on my local machine, so I am not going to get it fixed tonight.

```
[306](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:307)
ERROR: test_without_database (project.test_jobloader.TestLoaders.test_without_database) [Works without the database]
[307](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:308)
[308](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:309)
[309](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:310)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 3803, in get_loc
[310](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:311)
    return self._engine.get_loc(casted_key)
[311](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:312)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[312](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:313)
  File "pandas/_libs/index.pyx", line 138, in pandas._libs.index.IndexEngine.get_loc
[313](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:314)
  File "pandas/_libs/index.pyx", line 146, in pandas._libs.index.IndexEngine.get_loc
[314](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:315)
  File "pandas/_libs/index_class_helper.pxi", line 49, in pandas._libs.index.Int64Engine._check_type
[315](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:316)
KeyError: False
[316](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:317)

[317](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:318)
The above exception was the direct cause of the following exception:
[318](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:319)

[319](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:320)
[320](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:321)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader.py", line 53, in test_without_database
[321](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:322)
    self.project.load.toy_1, ToyJob, msg="Expected to load the full object"
[322](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:323)
    ^^^^^^^^^^^^^^^^^^^^^^^
[323](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:324)
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/project/jobloader.py", line 52, in __getattr__
[324](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:325)
    job_id=self._id_from_name(item), convert_to_object=self.convert_to_object
[325](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:326)
           ^^^^^^^^^^^^^^^^^^^^^^^^
[326](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:327)
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/project/jobloader.py", line 48, in _id_from_name
[327](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:328)
    return self._job_table.loc[self._job_names == name, "id"].values[0]
[328](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:329)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[329](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:330)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pandas/core/indexing.py", line 1066, in __getitem__
[330](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:331)
    return self.obj._get_value(*key, takeable=self._takeable)
[331](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:332)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[332](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:333)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pandas/core/frame.py", line 3922, in _get_value
[333](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:334)
    row = self.index.get_loc(index)
[334](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:335)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
[335](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:336)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 3805, in get_loc
[336](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:337)
    raise KeyError(key) from err
[337](https://github.com/pyiron/pyiron_base/actions/runs/3754920737/jobs/6379557380#step:9:338)
KeyError: False
```

liamhuber:
Ugh, I forgot about this over the Christmas break. There are more urgent things at the moment, but debugging the loc error is still on my todo. Probably I'll need to update my environment so I can reproduce the error on my local machine (I'm hoping it's from different versions of pandas). In the meantime I merged in main in case that miraculously resolves it...

liamhuber:
I finally got around to updating my conda environment, and now I can reproduce the pandas error. _But_ I can only reproduce it when I run the full test suite -- running the failed test, the class it belongs to, and indeed all the tests in the tests/project subdirectory fail to show any brokenness.

liamhuber:
Ok, the problem is unrelated. The issue is that `Singleton` is not re-initting stuff correctly. So in `Project` we have

```python
    @property
    def db(self):
        if not state.database.database_is_disabled:
            return state.database.database
        else:
            return FileTable(project=self.path)
```

But if `FileTable` was previously initialized, it uses _that_ value for its `project` kwarg and never updates it. So in this case the project was in the correct test directory, but the when we disable the database it points to whatever project used the table the first time the database was disabled -- thus it runs fine when you only run these tests, but not when that other test gets called.

I'll fix that separately and merge the fix from main into here later.

jan-janssen:
> I'll fix that separately and merge the fix from main into here later.

Somehow this issue sounds familiar. @niklassiemer did not we have something similar with file tables on different project levels? The issue is that the project object is reinitialized several times - for example for every job - but we do not want the file system to be reindexed. So the naive idea was storing the filetable as a singleton. But obviously this only works if the path remains the same. A more clean solution would be to store multiple file tables when working on multiple directories.

liamhuber:
As soon as #1073 is resolved, the tests here should pass and this will be ready for merge.

liamhuber:
Argh, #1073 is resolved, but I am still getting 

```
ERROR: test_without_database (project.test_jobloader.TestLoaders) [Works without the database]
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader.py", line 53, in test_without_database
    self.project.load.toy_1, ToyJob, msg="Expected to load the full object"
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/project/jobloader.py", line 52, in __getattr__
    job_id=self._id_from_name(item), convert_to_object=self.convert_to_object
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/project/jobloader.py", line 48, in _id_from_name
    return self._job_table.loc[self._job_names == name, "id"].values[0]
IndexError: index 0 is out of bounds for axis 0 with size 0
```
 
Previously I understood this to be empty because the sans-database Project was pointing to the wrong location. Since that should now be resolved and it should point to the correct location, I need to dig in.

Kids have no daycare tomorrow, so this might not get resolved until next week.

liamhuber:
I still can't reproduce this error on my local machine since the update to `FileTable`. Looking more carefully at the full unit test, I see that the "Update environment" step is being skipped and a cache is being loaded -- maybe the cache is somehow out of date?

```
Cache restored successfully
Cache restored from key: linux-64-py-3-10-conda-330607d4c8f1d6deda7deb69bc4063e3fee45c61965a7966008c25a177c5707d-20230523-0
```

liamhuber:
> I still can't reproduce this error on my local machine since the update to `FileTable`. Looking more carefully at the full unit test, I see that the "Update environment" step is being skipped and a cache is being loaded -- maybe the cache is somehow out of date?
> 
> ```
> Cache restored successfully
> Cache restored from key: linux-64-py-3-10-conda-330607d4c8f1d6deda7deb69bc4063e3fee45c61965a7966008c25a177c5707d-20230523-0
> ```

No, that's a bit dumb, it's the code in _this package_ that got updated, so the pip install should be taking care of it.....

liamhuber:
Ok, I added some debug warnings in the test:

```python
    def test_without_database(self):
        settings_configuration = self.project.state.settings.configuration.copy()
        import logging
        cols = ["id", "job"]
        logging.warning("With database: " + str(self.project.job_table().loc[:, cols]))
        self.project.state.update({'disable_database': True})
        logging.warning("Without database: " + str(self.project.job_table().loc[:, cols]))
```

And in the CI log I see exactly what I am expecting:

```
WARNING:root:With database:    id    job
0   6  toy_1
1   7  toy_2
2   8  toy_1
3   9  toy_2
WARNING:root:Without database:    id    job
0   1  toy_1
1   2  toy_2
2   3  toy_1
3   4  toy_2
```

(The ID drop for the local database is, of course, expected, since those are the only jobs there on the local scale!)

Still I get the error:

```
ERROR: test_without_database (project.test_jobloader.TestLoaders) [Works without the database]
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader.py", line 56, in test_without_database
    self.project.load.toy_1, ToyJob, msg="Expected to load the full object"
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/project/jobloader.py", line 52, in __getattr__
    job_id=self._id_from_name(item), convert_to_object=self.convert_to_object
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/project/jobloader.py", line 48, in _id_from_name
    return self._job_table.loc[self._job_names == name, "id"].values[0]
IndexError: index 0 is out of bounds for axis 0 with size 0
```

This really shouldn't be, as that final map of `self._job_names == name` is what's doing the check against the job tables I printed above.

There are lots of lines of CI log output _between_ the correct debug warning and the failure though, including stuff like 

```
DEBUG:pyiron_log:sql_query: {'job': 'toy_1', 'project': 'tests/project/test_project/'}
DEBUG:pyiron_log:Remove job with ID 6
```

And 7-9...

The cleanup for this test happens at the class level, is it possible that the CI is parallelizing over the tests, making multiple instances of the class with each one running a single test, and that these other instances are deleting stuff at the same file location? This seems unlikely, but is easy enough to check by removing the sibling tests. If the tests succeed then at least I know there is _some_ interaction between the siblings.

liamhuber:
Without the siblings I get the same problem. Maybe the path being referenced by `TestWithProject` is not what I think it is? I am assuming that `cls.project_path = getfile(cls)[:-3].replace("\\", "/")` is grabbing the path at the local class, in which case it should now be totally unique among the other tests. I realized after starting the CI that this is indeed the case on my local machine, so it would be shocking if it's different here.

liamhuber:
Indeed, `At path/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader/`, comes from the path of the test case's project attribute.

liamhuber:
Well, this is progress at least.

With the following logging:
```python
    def test_without_database(self):
        settings_configuration = self.project.state.settings.configuration.copy()
        import logging
        cols = ["id", "job"]
        logging.warning("With database: " + str(self.project.job_table().loc[:, cols]))
        self.project.state.update({'disable_database': True})
        logging.warning("Without database: " + str(self.project.job_table().loc[:, cols]))
        logging.warning("At path" + str(self.project.path))
        logging.warning("Loader job table: " + str(self.project.load._job_table))
        logging.warning("Loader job name: " + str(self.project.load._job_names))
        logging.warning("Map filter: " + str(self.project.load._job_names == "toy_1"))
```

Everything turns out just fine on my local machine. But on the remote CI it falls apart at "Loader job table", which is an empty dataframe, with all the expected downstream trouble from there.

```
ss....s.......WARNING:root:With database:    id    job
0   6  toy_1
1   7  toy_2
2   8  toy_1
3   9  toy_2
WARNING:root:Without database:    id    job
0   1  toy_1
1   2  toy_2
2   3  toy_1
3   4  toy_2
WARNING:root:At path/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader/
WARNING:root:Loader job table: Empty DataFrame
Columns: [id, status, chemicalformula, job, subjob, projectpath, project, hamilton, hamversion, timestart, computer, parentid, username, timestop, totalcputime, masterid]
Index: []
WARNING:root:Loader job name: []
WARNING:root:Map filter: []
```

So, why is the loader not seeing the job table I expect?

liamhuber:
Debug:

```python
    def test_without_database(self):
        settings_configuration = self.project.state.settings.configuration.copy()
        import logging
        cols = ["id", "job", "status"]
        logging.warning("With database: " + str(self.project.job_table().loc[:, cols]))
        logging.warning(f"Loader project (path) with database: {self.project.load._project} ({self.project.load._project.path})")
        self.project.state.update({'disable_database': True})
        logging.warning("Without database: " + str(self.project.job_table().loc[:, cols]))
        logging.warning(f"Loader project (path) with database: {self.project.load._project} ({self.project.load._project.path})")
        logging.warning("At path" + str(self.project.path))
        logging.warning("Loader job table: " + str(self.project.load._job_table.loc[:, cols]))
        logging.warning(f"Loader raw project job table: {self.project.load._project.job_table().loc[:, cols]}")
        logging.warning("Loader job name: " + str(self.project.load._job_names))
        logging.warning("Map filter: " + str(self.project.load._job_names == "toy_1"))
```

Result:

```
WARNING:root:With database:    id    job     status
0   6  toy_1   finished
1   7  toy_2    aborted
2   8  toy_1   finished
3   9  toy_2  suspended
WARNING:root:Loader project (path) with database: /home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader/ (/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader/)
WARNING:root:Without database:    id    job    status
0   1  toy_1  finished
1   2  toy_2  finished
2   3  toy_1  finished
3   4  toy_2  finished
WARNING:root:Loader project (path) with database: /home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader/ (/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader/)
WARNING:root:At path/home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader/
WARNING:root:Loader job table: Empty DataFrame
Columns: [id, job, status]
Index: []
WARNING:root:Loader raw project job table:    id    job    status
0   1  toy_1  finished
1   2  toy_2  finished
2   3  toy_1  finished
3   4  toy_2  finished
WARNING:root:Loader job name: []
WARNING:root:Map filter: []
```

Again, works perfectly on my local machine, but `self.project.load._job_table` is somehow empty here. It's nice to see that the raw job table off the project of the loader of the project is getting preserved fine.

liamhuber:
Added:

```python
        logging.warning(f"Raw non-recursive: + {self.project.load._project.job_table(recursive=False).loc[:, cols]}")
        logging.warning(f"Raw non-recursive and job colum: + {self.project.load._project.job_table(recursive=False, columns=['job']).loc[:, cols]}")
```

Result:

`self.project.load._project.job_table().loc[:, cols]` is fine, but `self.project.load._project.job_table(recursive=False).loc[:, cols]` is once again an empty dataframe.

Why would this possibly be the case? The default recursive call shows all four jobs, the non-recursive shows two on my local machine but _zero_ here? Since they're showing up on the recursive call, doesn't that imply that they're somehow getting moved down to the sub-project?! This is bizarre.

liamhuber:
> `WARNING:root:self.project.load._project.db._path: /home/runner/work/pyiron_base/pyiron_base/tests/project/test_jobloader`

The job table path looks perfectly fine.

liamhuber:
```python
        from os import listdir
        logging.warning(f"Project contents: {listdir(self.project.load._project.db._path)}")
        logging.warning(f"Sub-project contents: {listdir(self.project.load._project.db._path + '/sub_project')}")
```

```
WARNING:root:Project contents: ['toy_2.h5', 'toy_1.h5', 'toy_1_hdf5', 'sub_project', 'toy_2_hdf5']
WARNING:root:Sub-project contents: ['toy_2.h5', 'toy_1.h5', 'toy_1_hdf5', 'toy_2_hdf5']
```

So the _contents are there_. This is totally bonkers, why would `self.project.load._project.job_table()` then be totally fine but `self.project.load._project.job_table(recursive=False)` be empty?!

`Project.job_table` is calling `self.db.job_table`; When the database is disabled this is calling `FileTable(IsDatabase).job_table`, which is calling `FileTable._get_job_table`. And now we start to make some progress! On my local machine, if I _directly_ invoke `self.project.load._project.db._get_job_table(sql_query=self.project.sql_query, user=self.project.user, recursive=False)`, then I get the same empty dataframe seen on the remote, even though the wrapped  `self.project.load._project.job_table(recursive=False)` is giving the right thing (locally).

I included a sanity check, and the database type really is `FileTable`.

Some more playing shows that the empty dataframe is being returned when `project_path=None` is being passed to `FileTable._get_job_table`, but if the project path gets passed in things work fine.

So what different branch of the wrapping public method(s) is being called that is failing to pass this path on the remote CI where it _does_ get passed on my local machine? As far as I can tell, nowhere. From `Project.job_table` on down, the `Project.project_path` is always getting passed.

liamhuber:
Well, it wasn't the project path. Locally, passing `project_path=None` to `self.project.db._get_job_table` returns the empty table, but all other calls are fine. On the CI, passing `project_path=None` does indeed return an empty dataframe, but so does setting `recursive=False` at _any_ depth of the call.

Is non-recursion also broken for the regular database???

liamhuber:
Ok, I finally found it. The issue is that the CI specifies `TOP_LEVEL_DIRS` in its config, which means that `Project.project_path` is relative to an actual `Project.root_path`, which means that over in `FileTable` an `==` comparison of paths fails because they start with different things (since the `project` column of the job table in `FileTable` is the absolute path).

Raised as an issue in #1103 with a proposed fix over in #1104. This is back to draft until that's fixed...

liamhuber:
Same error as always, but this time only for windows. Since it's just windows, I suspect it's something with path comparisons in `FileTable` not being OS agnostic. Will try to fix it later this week.

```
ERROR: test_without_database (project.test_jobloader.TestLoaders.test_without_database) [Works without the database]
  File "D:\a\pyiron_base\pyiron_base\tests\project\test_jobloader.py", line 54, in test_without_database
    self.project.load.toy_1, ToyJob, msg="Expected to load the full object"
    ^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\pyiron_base\pyiron_base\pyiron_base\project\jobloader.py", line 52, in __getattr__
    job_id=self._id_from_name(item), convert_to_object=self.convert_to_object
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\pyiron_base\pyiron_base\pyiron_base\project\jobloader.py", line 48, in _id_from_name
    return self._job_table.loc[self._job_names == name, "id"].values[0]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: index 0 is out of bounds for axis 0 with size 0
```

liamhuber:
Waiting on #1139

liamhuber:
So after a number of other PRs working on `FileTable`, we can finally load a job after deactivating the database:

```python
    def test_without_database(self):
         settings_configuration = self.project.state.settings.configuration.copy()

         self.project.state.update({'disable_database': True})

         with self.subTest("Works without the database"):
             self.assertIsInstance(
                 self.project.load.toy_1, ToyJob, msg="Expected to load the full object"
             )
```


Black failure is a false negative. We need to update the CI so having nothing to commit does not result in failure.

```
nothing to commit, working tree clean
Error: Process completed with exit code 1.
```

Remaining codacy nits -- an unused import and overridden signature in call -- are helpful (for cross-referencing docstring type hints) and intentional (to remove a kwarg from access), respectively, so that failure is fine too.

IMO, this is ready to go.

pyiron_base/jobs/master/generic.py
```diff
@@ -460,7 +460,7 @@ def _get_item_when_str(self, item, child_id_lst, child_name_lst):
             if len(name_lst) > 1:
                 return self.project.inspect(child_id)["/".join(name_lst[1:])]
             else:
-                return self.project.load(child_id, convert_to_object=True)
+                return self.project.load(child_id)
```
jan-janssen, line 463:
```suggestion
                return self.project.inspect(child_id)
```

jan-janssen, line 463:
```suggestion
                return self.project.inspect(child_id)
```

pyiron_base/jobs/master/serial.py
```diff
@@ -123,7 +123,7 @@ def create_next(self, job_name=None):
             raise ValueError("No job available in job list, please append a job first.")
         if len(self._job_name_lst) > len(self.child_ids):
             return self.pop(-1)
-        ham_old = self.project.load(self.child_ids[-1], convert_to_object=True)
+        ham_old = self.project.load(self.child_ids[-1])
```
jan-janssen, line 126:
```suggestion
        ham_old = self.project.inspect(self.child_ids[-1])
```

pyiron_base/project/jobloader.py
```diff
@@ -0,0 +1,120 @@
...
+    A parent class for accessing project jobs by a call and a job specifier, or by tab
+    completion.
+    """
+
+    def __init__(self, project: Project):
+        self._project = project
+
+    @property
+    def _job_table(self):
+        return self._project.job_table(recursive=False, columns=["job"])
```
jan-janssen: Why is this `recursive=False` ?

liamhuber: It was intended to make sure that the mapping between the tab-completion menu and actual jobs stayed injective. I see now that the load function already handles duplicate job names at different levels by returning the upper-most job. I'm removing the flag specification here so that the tab-completion behaves the same as the string-based loading.

liamhuber:
```
FAIL: test_filtered_job_table (project.test_project.TestProjectOperations.test_filtered_job_table)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_project.py", line 103, in test_filtered_job_table
    self.assertEqual(len(self.project.job_table(recursive=True)), 4)
AssertionError: 5 != 4

FAIL: test_get_iter_jobs (project.test_project.TestProjectOperations.test_get_iter_jobs)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_project.py", line 123, in test_get_iter_jobs
    self.assertEqual([job.output.data_out for job in self.project.iter_jobs(recursive=True,
AssertionError: Lists differ: [101, 101, 101, 101, 101] != [101, 101, 101, 101]

First list contains 1 additional elements.
First extra element 4:
101

- [101, 101, 101, 101, 101]
?                 -----

+ [101, 101, 101, 101]

FAIL: test_job_table (project.test_project.TestProjectOperations.test_job_table)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/project/test_project.py", line 98, in test_job_table
    self.assertEqual(len(df), 4)
AssertionError: 5 != 4

```

liamhuber:
Ok, test failures are just because I added an extra job to `TestWithFilledProject` and some tests thought they knew how many jobs were in there. I'll fix it in a second.

jan-janssen (approved):
Looks good to me. I have not tested it locally without a database, but as this part seems to be covered by the tests, I guess we are fine.

PR #956, opened 2022-12-20
Use the dict that got updated in setUp

### LINKED ISSUE #953: Bug in ProjectPath tests?
Shouldn't [this line](https://github.com/pyiron/pyiron_base/blob/685fa919d25c5e844e4b29548b6291892482f8fe/tests/project/test_projectPath.py#L78) be `root_paths = state.settings.configuration["project_paths"]`?

In `setUpClass` we have `cls.settings_configuration = state.settings.configuration.copy()`, then in `setUp` we have `state.settings.configuration["project_paths"] = [self.current_dir]; state.settings.configuration["project_check_enabled"] = True`, and finally in the test we're setting `root_paths = self.settings_configuration["project_paths"]` and expecting it to be non-empty. Now, this passes on the remote CI, but in `pyironconfig.py` we're explicitly setting `TOP_LEVEL_DIRS` (i.e. `project_paths`), so maybe that allows it to get by even though it's comparing to the wrong thing? On my local machine, where `project_paths = []` and `project_check_enabled = False`, this test fails, exactly because `root_paths == []`.

I get the exact same failure in `test_root_path` for [exactly the same (suspected) reason](https://github.com/pyiron/pyiron_base/blob/685fa919d25c5e844e4b29548b6291892482f8fe/tests/project/test_projectPath.py#L70)

liamhuber:
Closes #953, in case it turns out that does need closing.

### DIFF
--- a/tests/project/test_projectPath.py
+++ b/tests/project/test_projectPath.py
@@ -79,7 +79,7 @@ class TestProjectPath(PyironTestCase):
         root_paths = state.settings.configuration["project_paths"]
         self.assertIn(
             self.current_dir + "/test_project_path/",
-            [root_path + "/" + self.project_path.project_path for root_path in root_paths],
+            [root_path + self.project_path.project_path for root_path in root_paths],
             msg="project.project_path not properly set by default. Check if `project_check_enabled`.",
         )
 
--- a/tests/project/test_projectPath.py
+++ b/tests/project/test_projectPath.py
@@ -70,7 +70,7 @@ class TestProjectPath(PyironTestCase):
     def test_root_path(self):
         root_paths = state.settings.configuration["project_paths"]
         self.assertIn(
-            self.project_path.root_path,
+            self.project_path.root_path[:-1],  # Trim the trailing "/"
             root_paths,
             msg="root project.root_path not properly set by default. Check if `project_check_enabled`.",
         )
--- a/tests/project/test_projectPath.py
+++ b/tests/project/test_projectPath.py
@@ -18,12 +18,13 @@ class TestProjectPath(PyironTestCase):
         else:
             cls.current_dir = os.path.dirname(os.path.abspath(__file__))
         cls.settings_configuration = state.settings.configuration.copy()
-        cls.project_path = ProjectPath(path=cls.current_dir)
-        cls.project_path = cls.project_path.open("test_project_path")
 
     def setUp(self) -> None:
         state.settings.configuration["project_paths"] = [self.current_dir]
         state.settings.configuration["project_check_enabled"] = True
+
+        self.project_path = ProjectPath(path=self.current_dir)
+        self.project_path = self.project_path.open("test_project_path")
 
     def tearDown(self) -> None:
         state.settings.configuration.update(self.settings_configuration)
--- a/tests/project/test_projectPath.py
+++ b/tests/project/test_projectPath.py
@@ -79,7 +79,7 @@ class TestProjectPath(PyironTestCase):
         root_paths = state.settings.configuration["project_paths"]
         self.assertIn(
             self.current_dir + "/test_project_path/",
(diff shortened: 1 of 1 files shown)


tests/project/test_projectPath.py
```diff
@@ -67,18 +68,18 @@ def test_path(self):
         )
 
     def test_root_path(self):
-        root_paths = self.settings_configuration["project_paths"]
+        root_paths = state.settings.configuration["project_paths"]
         self.assertIn(
-            self.project_path.root_path,
+            self.project_path.root_path[:-1],  # Trim the trailing "/"
```
jan-janssen: If there is no tailing `/` then this fails right? So I m not sure if that improves the issue ...

liamhuber: On my local machine all that was needed was to remove the trailing /. Now It's failing only for windows, so I think this was almost the right move....

liamhuber: Ah, no, on windows it looks like the issue was actually with the 'join' command which inserted `//` while pyiron stores everything as unix.

jan-janssen: > On my local machine all that was needed was to remove the trailing /. Now It's failing only for windows, so I think this was almost the right move....

The tailing `/` is just what you defined in your `.pyiron` file, right?

liamhuber: This doesn't rely on the config file any more, but on explicit updates in the test setup:

```python
    def setUp(self) -> None:
        state.settings.configuration["project_paths"] = [self.current_dir]
        state.settings.configuration["project_check_enabled"] = True

        self.project_path = ProjectPath(path=self.current_dir)
        self.project_path = self.project_path.open("test_project_path")
```

`ProjectPath` must be appending the `/` to `self.current_dir`, but give me a minute to find where.

(btw `self.current_dir` is just a unix-ized path in the class setup:
```python
@classmethod
    def setUpClass(cls):
        if os.name == "nt":
            cls.current_dir = os.path.dirname(os.path.abspath(__file__)).replace(
                "\\", "/"
            )
        else:
            cls.current_dir = os.path.dirname(os.path.abspath(__file__))
```)

Comment (liamhuber) [reply tests/project/test_projectPath.py]: Yeah, `ProjectPath.__init__` calls `ProjectPath._convert_str_to_generic_path` which calls `GenericPath._windows_path_to_unix_path`:

```python
        if path is not None:
            linux_path = path.replace("\\", "/")
            if linux_path[-1] != "/":
                linux_path += "/"
            return linux_path
        else:
            return None
```

jan-janssen: From my perspective the issue is that `state.settings.configuration["project_paths"]` depends on the users `.pyiron` config or environment variables. So while some users have their paths ending with `/` others do not. So I would recommend cleaning up the `state.settings.configuration["project_paths"]` used in the tests, rather than modifying the tests because the whole `[:-1]` notation to remove the tailing `/` is very confusing.

tests/project/test_projectPath.py
```diff
@@ -67,18 +68,18 @@ def test_path(self):
...
             root_paths,
             msg="root project.root_path not properly set by default. Check if `project_check_enabled`.",
         )
 
     def test_project_path(self):
-        root_paths = self.settings_configuration["project_paths"]
+        root_paths = state.settings.configuration["project_paths"]
         self.assertIn(
             self.current_dir + "/test_project_path/",
```
jan-janssen, line 81: In your case, what is the ending of `self.current_dir` is it ending with a `/`? In that case it might be better to remove the tailing `/` rather than adding an additional one.

tests/project/test_projectPath.py
liamhuber: Yep, sounds good to me.

jan-janssen (approved):
Looks good to me

PR #958, opened 2022-12-21
Merge JobPath & JobPathCore

### LINKED ISSUE #769: JobPathBase vs. JobPath?
The former latter just adds a more complex `__init__`.  I can't find any code that explicitly uses `JobPathBase` as well. If noone complains, I'll merge them together.

pmrv:
Fixes #769 

@jan-janssen Any idea why this line is necessary 
https://github.com/pyiron/pyiron_base/blob/685fa919d25c5e844e4b29548b6291892482f8fe/pyiron_base/project/generic.py#L838 when loading from id, but not from db_entry?

### DIFF
--- a/pyiron_base/jobs/job/path.py
+++ b/pyiron_base/jobs/job/path.py
@@ -22,18 +22,11 @@ __status__ = "production"
 __date__ = "Sep 1, 2017"
 
 
-class JobPathBase(JobCore):
+class JobPath(JobCore):
     """
     The JobPath class is derived from the JobCore and is used as a lean version of the GenericJob class. Instead of
     loading the full pyiron object the JobPath class only provides access to the HDF5 file, which should be enough
     for most analysis.
-
-    Args:
-        db (DatabaseAccess): database object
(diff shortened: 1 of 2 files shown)


jan-janssen:
> @jan-janssen Any idea why this line is necessary

I have to admit I am not sure why we used import lib rather than a direct import. I remember there were some issues with circular imports between `job_path` and `project` but importing inside the function looks like the more clean solution.

jan-janssen (approved):
Looks good to me

pmrv:
> > @jan-janssen Any idea why this line is necessary
> 
> I have to admit I am not sure why we used import lib rather than a direct import. I remember there were some issues with circular imports between `job_path` and `project` but importing inside the function looks like the more clean solution.

Ah, mispasted, I actually meant 

https://github.com/pyiron/pyiron_base/blob/685fa919d25c5e844e4b29548b6291892482f8fe/pyiron_base/project/generic.py#L841

In one case we reset the job id, but not in the other.

PR #967, opened 2023-01-06
Move signal catching to separate module and refactor

pmrv:
The previous implementation also had the problem that it
1. never dropped the signal intercept, but installed it on first initialization of any job, even if the job then doesn't actually run,
2. didn't propagate the signal after it set the job to 'aborted', i.e. trying to interrupt notebooks when no job is running didn't actually interrupt the process
3. kept re-installing the same handler on every job instantiation.  It's not that expensive, but it can cost a few percent of run time, depending on the size of the jobs.

This new implementation only installs the handler during a functions `run()` method call with a context manager and raises KeyboardInterrupt or calls sys.exit as appropriate.

### DIFF
--- a/pyiron_base/state/signal.py
+++ b/pyiron_base/state/signal.py
@@ -11,6 +11,19 @@ intercepted_signals = [
 
 @contextlib.contextmanager
 def catch_signals(cleanup):
+    """
+    Context manager to catch signals.
+
+    During the `with` statement a signal handler is installed to catch SIGINT,
+    SIGTERM and SIGABRT.  On exit from the statement the default handlers from
+    the operating system are reinstalled.  The handler first calles the given
+    `cleanup` function and then either raises KeyboardInterrupt (on SIGINT) or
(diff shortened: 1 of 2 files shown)


pyiron_base/state/signal.py
```diff
@@ -0,0 +1,26 @@
...
+        elif sig in (signal.SIGTERM, signal.SIGABRT):
+            sys.exit(128 & sig)
+
+    for sig in intercepted_signals:
+        signal.signal(sig, handler)
+
+    try:
+        yield
+    finally:
+        signal.signal(sig, signal.SIG_DFL)
```
pmrv:
```suggestion
        for sig in intercepted_signals:
            signal.signal(sig, signal.SIG_DFL)
```

pmrv:
I've checked now on the cluster that this actually solves the problem of not being able anymore to interrupt notebooks after you've instantiated a job, so I'd like to merge this soon.

PR #969, opened 2023-01-10
Add a simple FileBrowser to JobCore

pmrv:
job[] is overloaded to perform many functions at once:
    1. access to files
    2. access to HDF stored data
    3. access to child jobs

This change adds a new attribute `.files` to `JobCore` to take over
function 1 and deprecates methods on the job itself that are connected
to this: `list_files` and `tail`.

One of the problems we had discussed in the strategy meeting was that indexing notation is inconsistent and I think that this multiplexing of responsibilities is one of the reasons why `job[]` is inconsistent.  
I'm thinking therefor to introduce attributes for all three responsibilities (actually job.content works already well for 2.) and then deprecate `job[]` completely or at least agree on which of the three it should do.

Depends on #894


pmrv:
Eh, I've borking my starting point, there relevant changes are in [this commit](https://github.com/pyiron/pyiron_base/pull/969/commits/8db9284dd99a51d54b96360e646c752522544f79).

Issue #791, opened 2022-07-14
Status and fetching of remote jobs


Leimeroth:
Jobs executed on a remote cluster can not run collect, because they are not found in the filetable. So jobs get the status "initialized", independent if they ran successfully or not.
Also update_from_remote() and wait_for_jobs are not fetching the jobs because of this issue. Instead they have to be manually downloaded.
The manual download also comes with several issues because the jobs status does not work. It is hard to filter the jobs that already finished, when everything has the status "intialized". If a job is fetched that is not finished yet it will delete the working directory of the job on the cluster, so the job will fail completely.

IMO a more robust way to fix the issue of fetching is to simply download everything that can't be found in the remote squeue job table anymore instead of looking at the jobs status, which f.e. also breaks when the job runs into a time out.
Would it be possible to create a db table "is_remote" that simply stores all ids of jobs that have been transferred? When they are fetched back they could be deleted from the table so no complex status checking on the cluster would be required, but instead a simple check if the job is still listed with squeue would suffice. This could probably even be implemented as part of pysqa so that it does not mess with anything pyiron specific.

Error backtrace:
```
2022-07-12 21:19:06,128 - pyiron_log - INFO - run job: Cook1_1000 id: 749443, status: collect
Traceback (most recent call last):
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 735, in run
    self._run_if_collect()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 1303, in _run_if_collect
    run_job_with_status_collect(job=self)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/runfunction.py", line 194, in run_job_with_status_collect
    job.project.db.item_update(job._runtime(), job.job_id)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/generic.py", line 1375, in _runtime
    start_time = self.project.db.get_item_by_id(self.job_id)["timestart"]
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/database/filetable.py", line 118, in get_item_by_id
    return {
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/database/filetable.py", line 119, in <dictcomp>
    k: list(v.values())[0]
IndexError: list index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #971, opened 2023-01-12
Dont fetch waiting jobs

Leimeroth:
As described in #791 fetching remote jobs is currently plagued by several bugs.
This helps to fix my problems because jobs sitting on  the queue in the cluster get the status "initialized".
Consequently running update_from_remote() led to the jobs updating their status to "intialized" in my database,
which than caused 2 problems:
1. Since their status was "initialized" and not "submitted" or running their status was not updated anymore when I ran update_from_remote at a later point again.
2. My workaround for the job status not updating correctly on the cluster itself was to fetch jobs with status "initialized", because that is the status they were given when they ran successfully on the Cluster (since they can't update themselves to collect, finished etc.). Than using job.transfer_from_remote and setting the job.status to collect and running them again basically worked well when the jobs did not sit in the queue forever.

For jobs that were still in the queue and not running my workaround let to unfinished jobs being fetched, than collected and therefore ended up as aborted in my home database, requiring manual work.

This PR should fix above issues by only trying to fetch jobs that are not part of df_queue.
Also it is much faster since a simple array isin() is used to check instead of an endless for loop that fetches the hdf5 for every (!) job just to check its status, instead of using the ones directly available in df_queue.

Additionally I added a try_collecting argument, which if set to True tries to collect jobs that are fetched but do not have a
status that is finished in some way. This is especially helpful in the current state, where no job is able to be collected on the cluster it ran on, but I imagine it also being useful once this is fixed for experimental job classes, debugging etc.


pyiron_base/jobs/job/extension/server/queuestatus.py
```diff
@@ -295,29 +301,36 @@ def update_from_remote(project, recursive=True, ignore_exceptions=False):
...
-                    )
-                    status_hdf5 = job.project_hdf5["status"]
-                    project.set_job_status(job_specifier=job.job_id, status=status_hdf5)
-                    if status_hdf5 in job_status_finished_lst:
-                        job_object = job.to_object()
-                        job_object.transfer_from_remote()
-                except Exception as e:
-                    if ignore_exceptions:
+        fetch_ids = df_combined.id.values[
+            ~np.isin(df_combined.id.values, df_queue.jobid.values)
```
liamhuber:
```suggestion
            np.isin(df_combined.id.values, df_queue.jobid.values, invert=True)
```

If I'm reading [the docs](https://numpy.org/doc/stable/reference/generated/numpy.isin.html) right, this is equivalent but apparently faster.

pyiron_base/jobs/job/extension/server/queuestatus.py
```diff
@@ -295,29 +301,36 @@ def update_from_remote(project, recursive=True, ignore_exceptions=False):
...
+                    except Exception as e:
                         state.logger.warning(
-                            f"An error occured while trying to retrieve job {job_id}\n"
+                            f"An error occured while trying to collect job {job_id}\n"
                             f"Error message: \n{e}"
                         )
-                        failed_jobs.append(job_id)
-                    else:
-                        raise e
+                    failed_jobs.append(job_id)
```
liamhuber: This is my first time reading this file, and I found that some of these chunks are quite long and one really has to sit down and read the whole thing to understand anything. IMO it would be nice to refactor some of these clauses into smaller methods. We don't need to do that for the opening of the function in this PR, but we could do it here for the changes.

e.g.
```
try:
    failure_id = self.update_remote_job(project.load(job_id), try_collecting=try_collecting)
    if failure_id is not None:
        failed_jobs.append(failure_id)
except:
    ....
```

```
def update_remote_job(job: GenericJob, try_collecting: bool = True):
    """
    Transfers a job from a remote source, and if requested, attempts to run collection
    on all unfinished jobs.
    
    Args:
        job (GenericJob): The job to update, must have been run remotely(?)
        try_collect (bool): If the job does not have a valid finished status, try setting the
            status to `collect` and running again. (Default is True.)

    Returns:
        (None | int): The job id if an exception is encountered.
    """
    job.transfer_from_remote()
    if job.status in job_status_finished_lst:
        return

    if try_collecting:
        try:
            job.status.collect = True
            job.run()
            return
        except Exception as e:
            state.logger.warning(f"An error occurred while trying to collect job {job.id}\n")
            return job.id
```

liamhuber:
```suggestion
                        failed_jobs.append(job_id)
```

This needs another level of indentation so that it only happens when the exception is encountered, or?

pyiron_base/jobs/job/extension/server/queuestatus.py
```diff
@@ -295,29 +301,36 @@ def update_from_remote(project, recursive=True, ignore_exceptions=False):
...
+
+            except Exception as e:
+                if ignore_exceptions:
+                    state.logger.warning(
+                        f"An error occured while trying to retrieve job {job_id}\n"
```
liamhuber:
```suggestion
                        f"An error occurred while trying to retrieve job {job_id}\n"
```

typo

liamhuber (changes requested):
So there is a giant disclaimer that I have never touched this corner of the code base, nor have I ever executed remote jobs myself -- I do not feel well qualified to say whether this PR is _wise_.

However, just doing a local review of the implementation, I don't see any big problems. A couple small one-line adjustments (including an important logic-altering indentation), and an optional suggestion for some refactoring. Otherwise lgtm.

Leimeroth:
Made a separate function to retrieve the job and deleted some now unnecessary code.
Is there some way to update the status of multiple jobs without having to loop over them?
Like a set_jobs_status(ids, status) function. (Even if it is necessary to loop I guess it would still be nice to have?)

pmrv:
> Is there some way to update the status of multiple jobs without having to loop over them? Like a set_jobs_status(ids, status) function. (Even if it is necessary to loop I guess it would still be nice to have?)

There is not yet, but we should have one at least for the SQL one.  Feel free to add one, though it's a bit round about.  Here's the steps roughly:

1. Pull the `item_update` methods of `FileTable` and `DatabaseAccess` into `IsDatabase`, like I did [here](https://github.com/pyiron/pyiron_base/pull/336) for other methods. 
2. Define a new method `items_update` or some such on `IsDatabase`, that is defined in terms of `item_update`.  Can be just the naive implementation `for j in jobs: item_update(j)`, but it will give us a starting point to add more efficient SQL queries in the future.

Leimeroth:
> > Is there some way to update the status of multiple jobs without having to loop over them? Like a set_jobs_status(ids, status) function. (Even if it is necessary to loop I guess it would still be nice to have?)
> 
> There is not yet, but we should have one at least for the SQL one. Feel free to add one, though it's a bit round about. Here's the steps roughly:
> 
>     1. Pull the `item_update` methods of `FileTable` and `DatabaseAccess` into `IsDatabase`, like I did [here](https://github.com/pyiron/pyiron_base/pull/336) for other methods.
> 
>     2. Define a new method `items_update` or some such on `IsDatabase`, that is defined in terms of `item_update`.  Can be just the naive implementation `for j in jobs: item_update(j)`, but it will give us a starting point to add more efficient SQL queries in the future.

Tried to implement it following your suggestions. Apparently set_job_status is just a convenience wrapper in DatabaseAcces,
but modifiest the hdf5 file in FileTable, so I can't simply put it into IsDatabase with a generic implementation that relies on item_update(). I don't really understand why it modifies the hdf5 file, but I am not sure whether this is actually a good idea since it is the same function but apparently does 2 different things depending on your setup

Leimeroth:
> > Is there some way to update the status of multiple jobs without having to loop over them? Like a set_jobs_status(ids, status) function. (Even if it is necessary to loop I guess it would still be nice to have?)
> 
> There is not yet, but we should have one at least for the SQL one. Feel free to add one, though it's a bit round about. Here's the steps roughly:
> 
>     1. Pull the `item_update` methods of `FileTable` and `DatabaseAccess` into `IsDatabase`, like I did [here](https://github.com/pyiron/pyiron_base/pull/336) for other methods.
> 
>     2. Define a new method `items_update` or some such on `IsDatabase`, that is defined in terms of `item_update`.  Can be just the naive implementation `for j in jobs: item_update(j)`, but it will give us a starting point to add more efficient SQL queries in the future.

Tried to implement it following your suggestions. Apparently set_job_status is just a convenience wrapper in DatabaseAcces,
but modifiest the hdf5 file in FileTable, so I can't simply put it into IsDatabase with a generic implementation that relies on item_update(). I don't really understand why it modifies the hdf5 file, but I am not sure whether this is actually a good idea since it is the same function but apparently does 2 different things depending on your setup

pmrv:
I'm ok with having also a `_set_job_status` as part of the `IsDatabase` interface.  I'm also confused why it writes the status to HDF in case but not the other.  The status is saved in HDF in both cases, so I guess it must be set somewhere else for sql installations.

pmrv:
Ah, and I know I suggested `items_update`, but looking at it it's a bit hard to spot the difference on a casual read.  Maybe we can have the `IsDatabase.item_update` do the loop automatically if it sees a list?  Alternatively we can change the name to something more visually different or just require `_item_update` to be able to handle scalar and lists.  This is more of stretch goal then required for me though.  If it takes too much of your time, move this into a separate issue.

Leimeroth:
> Ah, and I know I suggested `items_update`, but looking at it it's a bit hard to spot the difference on a casual read. Maybe we can have the `IsDatabase.item_update` do the loop automatically if it sees a list? Alternatively we can change the name to something more visually different or just require `_item_update` to be able to handle scalar and lists. This is more of stretch goal then required for me though. If it takes too much of your time, move this into a separate issue.

My first idea was to just automatically loop over the list when a list is given instead of having a separate function, but
item update in both DatabaseAccess and FileTable check whether a list is given and than just take one of the elements,
so I am afraid this would break something downstream. 
Since item_update on DatabaseAccess uses element  [-1] and FileTable uses element [0] I assume if something actually relies on it it is buggy in a dangerous way anyway though. So I could take it out and see whether complaints arise from it.

pmrv:
> I'm ok with having also a `_set_job_status` as part of the `IsDatabase` interface. I'm also confused why it writes the status to HDF in case but not the other. The status is saved in HDF in both cases, so I guess it must be set somewhere else for sql installations.

So it seems this is set here

https://github.com/pyiron/pyiron_base/blob/bffd6d7feb581b33a5f5512d843e5119e0207ca9/pyiron_base/jobs/job/generic.py#L1025

when the database is active, which means that changes to the job state are only reflected when `to_hdf` is called explicitly.  E.g. for running jobs this is still `initialized` (the status that they are in before they are sent to the queue).

Imo this part of writing the status to HDF should then be part of the generic `IsDatabase.set_job_status` method.

Leimeroth:
I modified it to always write to use _item_update, always write hdf5, and simply inherit the function.

pmrv:
> > Ah, and I know I suggested `items_update`, but looking at it it's a bit hard to spot the difference on a casual read. Maybe we can have the `IsDatabase.item_update` do the loop automatically if it sees a list? Alternatively we can change the name to something more visually different or just require `_item_update` to be able to handle scalar and lists. This is more of stretch goal then required for me though. If it takes too much of your time, move this into a separate issue.
> 
> My first idea was to just automatically loop over the list when a list is given instead of having a separate function, but item update in both DatabaseAccess and FileTable check whether a list is given and than just take one of the elements, so I am afraid this would break something downstream. Since item_update on DatabaseAccess uses element [-1] and FileTable uses element [0] I assume if something actually relies on it it is buggy in a dangerous way anyway though. So I could take it out and see whether complaints arise from it.

I think try it with the integration tests, if those work I'm don't see something right now where it would fail.

Leimeroth:
So apparently unit tests check for the list functionality item update. I will put it back in for now, but this looks like a code smell to me.

pmrv:
If it works, I'd say let's just keep it in and do what you had first planned, but your call if it's too much work right now.

Leimeroth:
I am still not sure if we should delete it and change the test. Using [-1] in DatabaseAccess and [0] in FileTable is bound to bite back at some point. I changed it to at least use [-1] in both cases now.

pmrv:
Ah, I misread your comment.  I understand the unit test check *out* for the change.  If there's a unit test specifically for it, I'll bring it up on Monday, but imo slashing the test is fine.

Leimeroth:
Unittests also fail when I try to put the hdf5 file writing part
```
db_entry = self.get_item_by_id(item_id=job_id)
        write_hdf5(
            db_entry["project"] + db_entry["subjob"] + ".h5",
            status,
            title=db_entry["subjob"][1:] + "/status",
            overwrite="update",
        )
```
into the general set_job_status function. So as far as I can tell unittests absolutely require different behavior between sql and file database, which seems odd.

liamhuber (approved):
Skimmed changes and they look good to me. Marvin covered this at the meeting just now too, and everything else sounds good.

Don't forget to add the `black` label when you're done with your human-made changes.

pyiron_base/database/generic.py
```diff
@@ -222,6 +223,53 @@ def job_table(
...
+        """
+        Set status of multiple jobs
+
+        Args:
+            status (_type_): _description_
+            job_ids (_type_): _description_
+        """
+        for j_id in job_ids:
+            self.set_job_status(status=status, job_id=j_id)
+
```
jan-janssen: Can this be handled by `item_update()`?

Leimeroth: What exactly do you mean? set_job_status was just a convenient wrapper around item_update() in the case of DatabaseAccess anyway and in the case of FileTable the hdf5 writing part is the problem.

As we discussed above it would be nice to have an efficient implementation for set_jobs_status, but as long as that does not exist the for loop serves its purpose

pmrv:
We discussed that removing the unit test which indexes either `[0]` or `[-1]` and just see what breaks, but keep the differing behavior writing the job status always to disk in case of `FileTable` but not in case of `DatabaseAccess`.

Leimeroth:
> We discussed that removing the unit test which indexes either `[0]` or `[-1]` and just see what breaks, but keep the differing behavior writing the job status always to disk in case of `FileTable` but not in case of `DatabaseAccess`.


I think there is a missunderstanding here. The unit tests did not check whether it is [0] or [-1] (Well maybe they do, but I changed it to be -1 for both to at least be consistent and they passed), but they do check whether _item_update() allows to take a list as argument, which is a bad idea imo, because it lead to problems like the different indices in FileTable and DatabaseAccess.

I tried to change the implementation in a way that automatically checks if the input is iterable and if yes modifies multiple entries instead of a separate function for multiple entries, which would be very convenient, but maybe also confusing imo. 
I could not get this to work with the current unit tests, so I created separate functions instead.

pmrv:
> > We discussed that removing the unit test which indexes either `[0]` or `[-1]` and just see what breaks, but keep the differing behavior writing the job status always to disk in case of `FileTable` but not in case of `DatabaseAccess`.
> 
> I think there is a missunderstanding here. The unit tests did not check whether it is [0] or [-1] (Well maybe they do, but I changed it to be -1 for both to at least be consistent and they passed), but they do check whether _item_update() allows to take a list as argument, which is a bad idea imo, because it lead to problems like the different indices in FileTable and DatabaseAccess.
> 
> I tried to change the implementation in a way that automatically checks if the input is iterable and if yes modifies multiple entries instead of a separate function for multiple entries, which would be very convenient, but maybe also confusing imo. I could not get this to work with the current unit tests, so I created separate functions instead.

It's [this test](https://github.com/pyiron/pyiron_base/blob/f9364a3a3b14461bba30643160dd9f07522223e7/tests/database/test_database_access.py#L167), right?  I'm saying to cut that test out, proceed with implementation that you tried earlier and then check whether this will fail the integration tests.  If not it's fine to make that change imo.

Leimeroth:
> > > We discussed that removing the unit test which indexes either `[0]` or `[-1]` and just see what breaks, but keep the differing behavior writing the job status always to disk in case of `FileTable` but not in case of `DatabaseAccess`.
> > 
> > 
> > I think there is a missunderstanding here. The unit tests did not check whether it is [0] or [-1] (Well maybe they do, but I changed it to be -1 for both to at least be consistent and they passed), but they do check whether _item_update() allows to take a list as argument, which is a bad idea imo, because it lead to problems like the different indices in FileTable and DatabaseAccess.
> > I tried to change the implementation in a way that automatically checks if the input is iterable and if yes modifies multiple entries instead of a separate function for multiple entries, which would be very convenient, but maybe also confusing imo. I could not get this to work with the current unit tests, so I created separate functions instead.
> 
> It's [this test](https://github.com/pyiron/pyiron_base/blob/f9364a3a3b14461bba30643160dd9f07522223e7/tests/database/test_database_access.py#L167), right? I'm saying to cut that test out, proceed with implementation that you tried earlier and then check whether this will fail the integration tests. If not it's fine to make that change imo.

[This one](https://github.com/pyiron/pyiron_base/blob/f9364a3a3b14461bba30643160dd9f07522223e7/tests/database/test_database_access.py#L173), a few lines below.

PR #972, opened 2023-01-12
Support lookup of job status for pyiron installations without database

### LINKED ISSUE #957: Sync queue status with database status for pyiron without database
Primarily a reminder for myself, but at some point this snippet should be integrated in the `job_table()` function:
```
df = pr.job_table()
df_queue = state.queue_adapter.get_status_of_my_jobs()

status_lst = []
for working_dir, status in zip(df.project + df.job + "_hdf5/" + df.job, df.status):
    if status == "initialized":
        df_tmp = df_queue[df_queue.working_directory == working_dir]
        if len(df_tmp) > 0:
            status_queue = df_tmp.status.values[0]
            if status_queue == "running":
                status_lst.append(status_queue)
            elif status_queue == "pending":
                status_lst.append("submitted")
            else: 
                status_lst.append(status)
        else: 
            status_lst.append(status)
    else:
        status_lst.append(status)

df["status"] = status_lst
```

jan-janssen:


pyiron_base/project/generic.py
```diff
@@ -661,20 +661,71 @@ def job_table(
...
+                project_path=self.project_path,
+                recursive=recursive,
+                columns=columns,
+                all_columns=all_columns,
+                sort_by=sort_by,
+                full_table=full_table,
+                element_lst=element_lst,
+                **kwargs,
+            )
+        else:
```
pmrv, line 692: This should go into the `job_table` method of `FileTable` if I see this correctly, otherwise the code will be more difficult to refactor later.

jan-janssen: The code in the FileTable currently only contains the database interface but does not depend on the queuing system. I am not sure if we want to mix both. I see the database classes as one level and then we can have a derived class which handles the communication with the queuing system.

jan-janssen:
Waiting for pysqa 0.0.18 to be available on conda-forge .

pyiron_base/project/generic.py
pmrv: Then an alternative would be to add support for `FileTable` in `Project.refresh_job_status` which is implemented purely in terms of `get_item_by_id` and `set_job_status`.  (I'm assuming that this doesn't work right now and that that's the reason for this PR, but I had to guess this). 
If this is done, the `if auto_refresh_job_status` clause can be pulled in front of the `if insinstance(self.db, ...)` clauses.

jan-janssen: The `refresh_job_status` functions iterates over one job at a time, while here I get all jobs and then update them all in the database. I agree that the database is not very clean and we definitely have to rework this in particular also in connection with the support of remote clusters. But unfortunately I currently do not have the time to do a major refactoring of that part.

pmrv: > The `refresh_job_status` functions iterates over one job at a time, while here I get all jobs and then update them all in the database. 

Move it into the `FileTable` then since it is specific to that class.  It doesn't matter much to me whether this mixes queuing and database interactions since they are tangled in practice anyway.

> But unfortunately I currently do not have the time to do a major refactoring of that part.

And no one will have time to rework it in the future.  This is why I was insistent in the strategy meeting that when adding new things we make at least an attempt to have maintainable and well founded abstractions.  It might save a day of work now, but it will cost the next developer a week to understand and refactor it.
If we can at least keep everything contained in either `FileTable` or `DatabaseAccess` then unifying them in the future will be much easier.

jan-janssen: I do not see that moving it to FileTable would make the code more readable, because then we would also have to move all the `refresh_job_status` part to `DatabaseAccess`.

jan-janssen: The other issue with moving it to the `FileTable` is that the `FileTable` is connected to the `state` object just like the `QueueAdapter` so we would have to do a major refactoring of the whole interface.

pyiron_base/project/generic.py
```diff
@@ -1036,6 +1065,41 @@ def refresh_job_status_based_on_job_id(self, job_id, que_mode=True):
...
+        Returns:
+            pandas.DataFrame: updated job table with status from the queuing system
+        """
+
+        def convert_queue_status(queue_status):
+            convert_status_dict = {"pending": "submitted"}
+            if queue_status in convert_status_dict.keys():
+                return convert_status_dict[queue_status]
+            else:
+                return queue_status
```
pmrv: Isn't this just 

```python
if queue_status == "pending":
  return "submitted"
else:
  return queue_status
```

If you want to have a dict in case there will be more status in the future to map
```python
return {"pending": "submitted"}.get(queue_status, default=queue_status)
```
is still easier to read.

pyiron_base/project/generic.py
```diff
@@ -1036,6 +1065,41 @@ def refresh_job_status_based_on_job_id(self, job_id, que_mode=True):
...
+            if status == "initialized":
+                df_tmp = df_queue[df_queue.working_directory == working_dir]
+                if len(df_tmp) > 0:
+                    status_lst[i] = convert_queue_status(
+                        queue_status=df_tmp.status.values[0]
+                    )
+        df["status"] = status_lst
+
+        return df
+
```
pmrv, line 1098: If I understand this correctly, this modifies a dataframe obtained from `FileTable.job_table` and sets the updated job status there and returns it.  But this won't modify the dataframe `FileTable` keeps internally, so the queuing system would be queried every time `Project.job_table` is called, right?  Couldn't you equivalently call `FileTable.set_job_status` in line 1096?

jan-janssen: `set_job_status` writes to the HDF5 files which means we would open the HDF5 files in write access on a shared file system, which can cause errors in high throughput settings. If the job fails and we catch the error then the pyiron process which is executing the job should automatically update the status, that is what I changed in https://github.com/pyiron/pyiron_base/pull/973 so for performance reasons I do not want to write to the HDF5 files here.

PR #974, opened 2023-01-15
Add high-level spec

liamhuber:
@all,
 
So this slipped through the cracks this week, but at the last meeting I promised to write a sort of high-level spec document before the next meeting. 

I think it's important that we have a sort of high-level spec document that clarifies our vision, and outlines some key promises that we make to users. This is then useful not just as advertising, but also as a guiding document to devs when implementing new features.

This is just a rough first draft and there are some bits missing. It gets more detailed and (IMO) less important the further down you go, so if you don't have time to look at all of it please start at the top and get far down as you get!

Practically, I suggest that small changes be made using the github "suggestions" tool, and big changes be made by making new PRs into this PR. I'll follow that method too, then we are able to discuss changes asynchronously and transparently.

### DIFF
--- a/specs/speculative_specification.md
+++ b/specs/speculative_specification.md
@@ -1,7 +1,7 @@
 The purpose of this document is to lay out the high-level view of pyiron's purposes, capabilities, and structure.
 Some what's written here represents our ambitions for 2023 and does not necessarily represent the current state of the code base; these ambitional aspects should be clearly indicated as such.
 
-# Why pyiron
+# Why pyiron -- philosophy for prospective users
 
 pyiron exists to facilitate and accelerate computational research.
 
@@ -12,7 +12,7 @@ In particular, pyiron...
 - abstracts away all (or at least most) of the nitty-gritty to let scientists focus on the science
 
 
-# What pyiron
+# What pyiron -- promises for actual users
 
 Pyiron provides a python framework for designing and managing research calculations.
 It is structured into a network of repositories: `pyiron_base`, which provides the core pyiron infrastructure, and then a variety of `pyiron_X` modules which build all the domain-specific tools for particular fields of research.
@@ -44,7 +44,7 @@ Other customization, e.g. whether to run with or without a database, is handled 
 Pyiron also supports a variety of working environments, including local (e.g. your laptop), HPC clusters, and even accessing a variety of remote systems, doing calculations there, and collecting results back to another central system.
 
 
-# How pyiron
+# How pyiron -- guidance for developers
 
 Database interaction (TODO. Is totally deeply embedded in the pyiron library. It's even mangled so atomistics content lives in base. I am not sure how we want to abstract this. We also don't even serialize what code was used (pyiron versions, called executable versions(this maybe we do store?) to produce the output, much less store it nicely in the database.)
 
--- a/specs/speculative_specification.md
+++ b/specs/speculative_specification.md
@@ -1,0 +1,67 @@
+The purpose of this document is to lay out the high-level view of pyiron's purposes, capabilities, and structure.
+Some what's written here represents our ambitions for 2023 and does not necessarily represent the current state of the code base; these ambitional aspects should be clearly indicated as such.
+
+# Why pyiron
+
+pyiron exists to facilitate and accelerate computational research.
+
+In particular, pyiron...
+- makes easy things easy, and novel things possible
+- provides reliability and reproducibility by coupling results to the input and code(TODO) that generated them 
+- gives users tools to rapidly prototype new ideas, and easily scale those ideas up
+- abstracts away all (or at least most) of the nitty-gritty to let scientists focus on the science
+
+
+# What pyiron
+
+Pyiron provides a python framework for designing and managing research calculations.
+It is structured into a network of repositories: `pyiron_base`, which provides the core pyiron infrastructure, and then a variety of `pyiron_X` modules which build all the domain-specific tools for particular fields of research.
+E.g. `pyiron_atomistics` support for atomistic-scale materials science calculations, allowing users to create atomistic structures and perform a variety of calculations (minimization, molecular dynamics (MD), and more sophisticated routines such as nudged elastic band, variance-constrained semi-grand-canonical MD, etc.) using a variety of different engines to calculate energies, forces, and perhaps electronic state (e.g. LAMMPS, VASP, SPHiNX, QuantumEspresso...).
+Most of these extensions currently focus on materials science, but this is purely historical and `pyiron_base` is designed to support.
+
+Pyiron has a single point of entry, a single import, the `Project` object, from which all other pyiron objects can be created.
+Entire projects can be packaged and shared with colleagues.
+`Project` can either be imported from `pyiron_base` or from one of the other domain specific modules.
+When importing additional `pyiron_X` modules, the `Project` object is _automatically powered-up_ and given access to all the tools from that module.
+By heavy use of tab-completion, users can explore what tools are available from their project without having to immediately resort to reading documentation. 
+
+The core of pyiron, and the most common object created from a project, is a `Job`.
+Jobs are the objects which actually perform a calculation by invoking their `run` command.
+Job objects bring together the core elements of pyiron: they provide an easy to use and easy to extend python interface; they support reproducibility by serializing the details of input and executed code(TODO) with output results; they facilitate analysis and collaboration by populating a database with their most important details; they are scalable by simply specifying which remote resource and how many cores to use.
+Importantly, they do all these things without heavy intervention from the user.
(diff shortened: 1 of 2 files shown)


liamhuber:
Ah, we should include `[skip ci]` in all our commit strings! This should [prevent the CI from running](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs) uselessly for what is just an .MD file.

liamhuber:
> Ah, we should include `[skip ci]` in all our commit strings! This should [prevent the CI from running](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs) uselessly for what is just an .MD file.

And look at me, ignoring my own advice...

samwaseda:
Zen of pyiron

vers. Aug 8, 2019

Liam:

You tab-complete me
Limit imports
Do it with a verb
Don’t reinvent the wheel
But be careful who you have a dependency on

Sudarsan:

Run tests cheaply, but not too simple. (Al is better than Fe).
Parameters as cheap as possible at first.
Developing with shorter parameter lists.
Use Lammps for testing rather than VASP and if you use VASP use the gamma point version rather than the regular version.
Minimum working example.
“Restart + Clear Output” is your friend.
Round of digits before you create job names. np.close is your friend.
When you test code output, always test it with structures with multiple elements.
; after matplotlib commands makes your notebook much prettier.

Eunan:

Balance of readability and speed of code.
No notebooks without proper axis and labels should be accepted.

Sam:

Use explicit names, especially for function names and variable names.
Make it autocompletable.

Jan:

Science first, pyiron second.
Generic first, specific second.
Reduce the technical complexity to a minimum.
Question the status quo.
pyiron developers are just like pyiron users.
The open source code is tested and well documented.
Nobody should be forced to use pyiron, although pyiron developers don’t promise to help you with anything else.
Physical objects should be physical meaningful.
Avoid redundancy, we should try to save data only once not multiple times. (Generate vs. Storing)
Easy to debug. pyiron.log is your friend.

specs/speculative_specification.md
```diff
@@ -0,0 +1,67 @@
...
+Most of these extensions currently focus on materials science, but this is purely historical and `pyiron_base` is designed to support.
+
+Pyiron has a single point of entry, a single import, the `Project` object, from which all other pyiron objects can be created.
+Entire projects can be packaged and shared with colleagues.
+`Project` can either be imported from `pyiron_base` or from one of the other domain specific modules.
+When importing additional `pyiron_X` modules, the `Project` object is _automatically powered-up_ and given access to all the tools from that module.
+By heavy use of tab-completion, users can explore what tools are available from their project without having to immediately resort to reading documentation.
```
samwaseda, line 26: If we are talking about tab-completion, I would like to extend it to the triple pillar that I mention sometimes:

- Tab-completion
- Doc string
- Do and fail properly

In words: User checks all available options via tab-completion, chooses one of them and takes a look at the doc string to see what are the next steps. And they can execute the function according to what they think is right and the error message will guide them how to use it correctly, if it happened to not be the correct way to use it.

samwaseda: From today's meeting, I would like add [a global picture](https://scikit-learn.org/stable/tutorial/machine_learning_map/index.html) that @liamhuber mentioned.

specs/speculative_specification.md
```diff
@@ -0,0 +1,67 @@
...
+Configuration for various queueing systems, as well as access to and resources for various third party codes are all possible in the pyiron resources directory.
+Other customization, e.g. whether to run with or without a database, is handled by the global `State` object.
+Pyiron also supports a variety of working environments, including local (e.g. your laptop), HPC clusters, and even accessing a variety of remote systems, doing calculations there, and collecting results back to another central system.
+
+
+# How pyiron -- guidance for developers
+
+Database interaction (TODO. Is totally deeply embedded in the pyiron library. It's even mangled so atomistics content lives in base. I am not sure how we want to abstract this. We also don't even serialize what code was used (pyiron versions, called executable versions(this maybe we do store?) to produce the output, much less store it nicely in the database.)
+
+For storage, pyiron uses HDF5 and relies on the pyiron/datacontainer library(TODO) which in turn uses the h5io/h5io library.
```
liamhuber, line 51: At the meeting (2023-01-23) we discussed putting *all* the job input into a single `DataContainer`-powered input field. We had to remind ourselves why [@pmrv thought it was unwise](https://github.com/pyiron/pyiron_atomistics/pull/799#discussion_r994964343) to simply subclass `DataContainer`. We played around briefly with the `pyiron_base.HasStoredTraits` class I made, but it doesn't natively support the nice JSON output that `DataContainer` does (the underlying data container only gets populated *after* a `to_hdf` call in my implementation). @samwaseda also had some concerns about how easy (or not!) it would be to create deeply nested input structures in this way, since `HasStoredTraits` also loses the `create_group` method that he likes to use.

jan-janssen:
> Ah, we should include `[skip ci]` in all our commit strings! This should [prevent the CI from running](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs) uselessly for what is just an .MD file.

But there still seems to be an issue in Joerg's notebook:
```
FileNotFoundError: [Errno 2] No such file or directory: 'steps.ipynb'
```

jan-janssen:
As I spend quite some pages of my PhD thesis explaining pyiron on a high level, I just link it here if somebody is interested https://digital.ub.uni-paderborn.de/hs/content/titleinfo/6098963 - pages 55-68 cover pyiron in general

specs/speculative_specification.md
mbruns91: Not adding much to this, but we could also consider [this figure displaying the `LAMMPS` class topology](https://docs.lammps.org/Developer_org.html#class-topology). At least for me, this is kind of easier to understand as I'm much more familiar with `LAMMPS` than `scikit-learn` estimators.  
Also, I guess this actually fits better what we want to have - a visual overview of what `pyiron` (or, better, the project class) offers - rather than a "decision tree" as a help to select routines (actually, both would be great to have?!)

samwaseda:
There's one aspect that I want to emphasise, especially when someone compares pyiron with AiiDA or Galaxy: *Interactivity*, which I would like to clearly distinguish from *automaticity*. Here's why:

In other platforms they try to make tools as automatic as possible, in order for the user to not have to know what's going on. In pyiron, on the other hand, we try to make it easy for the user to interact. These two points are obviously not exclusive to each other, especially because we would like to achieve both at some point. However, in practice, they differ for the development strategy. Let's take Murnaghan as an example. It sometimes fails to detect the minimum lattice parameter. In AiiDA or Galaxy, they would probably try to automatically relaunch the calculations with modified parameters, while in pyiron, we make the visualization tools and output properties easily accessible, so that the user can modify the parameters themselves. At first glance, AiiDA or Galaxy's strategy might sound more sophisticated, but we all know that the reality is usually more complex, because the reason why Murnaghan fails to detect the minimum lattice parameter could have any (combination) of the following reasons:

- Initial lattice parameter was bad
- DFT parameters were not properly chosen
- There was structural and/or magnetic transitions for certain boxes
- Conjugate gradient did not converge for certain boxes
- Strain value was too large/too small

With all these potential sources of problems, it is almost irrational to make the detection fully automatic. However, by looking at the simple E-V curve a well trained theoretician would see where the problem comes from, and therefore can solve the problem more efficiently and effectively.

In addition, it is deeply problematic to claim "you can do calculations without understanding what's happening behind the scenes, because pyiron does everything automatically", because this would imply that we don't need us theoreticians anymore (as well as it's anyway a claim that cannot be fulfilled). It is a lot more attractive for us developers to claim "you can reach a lot more with pyiron with the understanding you have for all the calculations". In short, the claim would look like this:

For automaticity-centric platforms (such as AiiDA and Galaxy): "You can reach as much as before with a lot less effort"
For interactivity-centric platforms (pyiron): "You can reach a lot more with the same effort"

samwaseda:
This really has to be completed before we finalise the workflow implementation.

liamhuber:
@samwaseda, I don't know. IMO we might need to simply throw away 90% of what's here. It is based on a three-pillar concept of Python object+database+file storage. With the new workflow stuff, it seems to me like we're fundamentally shifting to _significantly_ lower the importance of the latter two items.

samwaseda:
> @samwaseda, I don't know. IMO we might need to simply throw away 90% of what's here. It is based on a three-pillar concept of Python object+database+file storage. With the new workflow stuff, it seems to me like we're fundamentally shifting to _significantly_ lower the importance of the latter two items.

One way or other we must have a written document like this one stating our principles and promises. Right now pyiron is like a country with no written constitution. It might work for internal people, but it's very difficult for external people to come in if we keep going like this.

liamhuber:
Yes, with that I'm fully on board. I think we might need to just play around with the workflow paradigm for another month or so before having the intuition/experience to write something sensible. At least that's roughly where I'm at. But our disagreement is then only one of timing and not at all of principle.

PR #984, opened 2023-01-16
Make it easier to store the run time

jan-janssen:
I keep the _run_time() function for the moment as it is still used in randomatomitics and sqs. Once those are updated we can make it an internal function.

### DIFF
--- a/pyiron_base/jobs/datamining.py
+++ b/pyiron_base/jobs/datamining.py
@@ -772,8 +772,7 @@ class TableJob(GenericJob):
             os.path.join(self.working_directory, "pyirontable.csv"), index=False
         )
         self._save_output()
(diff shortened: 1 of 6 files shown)


samwaseda:
Is `run_time_to_db` going to coexist with [the decorator](https://github.com/pyiron/pyiron_base/pull/992)? It sounds a bit confusing to me.

jan-janssen:
> Is `run_time_to_db` going to coexist with [the decorator](https://github.com/pyiron/pyiron_base/pull/992)? It sounds a bit confusing to me.

I am not sure if how soon we can get the decorator integrated into the parallel master and the interactive wrappers, so yes currently I would recommend to have both.

PR #985, opened 2023-01-16
Remove redundant `_check_if_input_should_be_written()`

jan-janssen:
Already in `pyiron_base/jobs/job/generic.py` this case is handled:

```
    def _check_if_input_should_be_written(self):
        if self._python_only_job:
            return False
        else:
            return not (
                self.server.run_mode.interactive
                or self.server.run_mode.interactive_non_modal
            )
```

### DIFF
--- a/pyiron_base/jobs/job/template.py
+++ b/pyiron_base/jobs/job/template.py
@@ -49,6 +49,3 @@ class PythonTemplateJob(TemplateJob):
     def __init__(self, project, job_name):
         super().__init__(project, job_name)
         self._python_only_job = True
-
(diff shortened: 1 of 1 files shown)


pmrv:
lgtm as it is, but I'd suggest that in a separate issue we change how `write_input` works.  

Right now `_check_if_input_should_be_written` is only used in `save` and if it returns `True`, the working directory of the job is created, `write_input` is called and the restart files are copied.  If we moved that first and last step to `GenericJob.write_input` and then overwrote that in `PythonTemplateJob` with a NOP, we could just do away with the whole `_check_if_input_should_be_written` method.

Issue #979, opened 2023-01-16
Job time not automatically tracked in user defined jobs


pmrv:
@JNmpi discovered that the `timestart`/ `timestope` is not automatically tracked when user jobs defined do not explicitly call `self.db.update_item(self._runtime(), self.id)` or some such.

jan-janssen:
As a first step I started to create a simpler `run_time_to_db()` function. I guess the easiest way is to define a decorator which wraps the run() function. https://github.com/pyiron/pyiron_base/pull/984

pmrv:
I was actually thinking about setting `timestart` and `timestop` when entering and leaving status `running`.  That might take some refactoring, so a decorator in the mean sounds fine to me.

jan-janssen:
> I was actually thinking about setting `timestart` and `timestop` when entering and leaving status `running`. That might take some refactoring, so a decorator in the mean sounds fine to me.

The issue is that the interactive jobs as well as the Genericmasters both call run_static multiple times.

jan-janssen:
A decorator to track the run time is implemented in https://github.com/pyiron/pyiron_base/pull/992 still we might want to extend this for the Python Template Job, if and call `to_hdf()` also within the decorator.

samwaseda:
> The issue is that the interactive jobs as well as the Genericmasters both call run_static multiple times.

Don't interactive jobs use `run_interactive`?

samwaseda:
But actually I don't see the problem with calling `run_static` multiple times. Can't we simply update the run time whenever it's called? Basically, the starting time can be set only once (and therefore if something like [run decorator](https://github.com/pyiron/pyiron_base/pull/992) is called multiple times, the starting time wouldn't be updated), while the ending time is updated whenever the job sends the signal. In this regard, when `timestart` is set, it should also reinitialise `timestop` and `totalcputime`, i.e. `job.project.db.item_update({"timestart": datetime.now(), "timestop": None, "totalcputime": None}, job.job_id)`

pmrv:
> > I was actually thinking about setting `timestart` and `timestop` when entering and leaving status `running`. That might take some refactoring, so a decorator in the mean sounds fine to me.
> 
> The issue is that the interactive jobs as well as the Genericmasters both call run_static multiple times.

That would be an issue for decorating `run_static`, but doing it on status change would still work, assuming these jobs do not set their state to `finished` multiple times.

pmrv:
[...]

PR #992, opened 2023-01-17
Add a run decorator

jan-janssen:
Based on https://github.com/pyiron/pyiron_base/pull/987


jan-janssen:
It is a bit surprising that the worker job fails with these changes.

samwaseda:
Can you add doc string describing what this is used for? What it does is relatively clear, but in the pyiron context it's not straightforward to see when it should be applied (or if it's only meant to be used in `GenericJob` then I think it can also be a private function).

pyiron_base/jobs/job/runfunction.py
```diff
@@ -456,6 +454,19 @@ def run_job_with_runmode_srun(job):
     )
 
 
+def run_static_decorator(func):
+    def wrapper(job):
+        if job.job_id is not None:
+            job.project.db.item_update({"timestart": datetime.now()}, job.job_id)
```
samwaseda, line 461: Based on [my comment](https://github.com/pyiron/pyiron_base/issues/979#issuecomment-1384891292), I would suggest to write:

```python
update_dict = {"timestop": None, "totalcputime": None}
if job.project.db.get_item_by_id(job._job_id)["timestart"] is None:
    update_dict["timestart"] = datetime.now()
job.project.db.item_update(update_dict, job.job_id)
```

pyiron_base/jobs/job/runfunction.py
```diff
@@ -456,6 +454,19 @@ def run_job_with_runmode_srun(job):
     )
 
 
+def run_static_decorator(func):
```
pmrv: The name is not very clear, maybe `set_runtime` or something like that.

pyiron_base/jobs/job/runfunction.py
jan-janssen: > Based on [my comment](https://github.com/pyiron/pyiron_base/issues/979#issuecomment-1384891292), I would suggest to write:
> 
> ```python
> update_dict = {"timestop": None, "totalcputime": None}
> if job.project.db.get_item_by_id(job._job_id)["timestart"] is None:
>     update_dict["timestart"] = datetime.now()
> job.project.db.item_update(update_dict, job.job_id)
> ```

Running `run_static()` multiple times currently only works for the `GenericMasters` including the `InteractiveWrapper`. For a regular job this is something we do not support at the moment, for example for VASP we would overwrite the output, so to reduce the overhead I would not add another database request to get the `timestart` currently stored in the database.

pyiron_base/jobs/job/runfunction.py
jan-janssen: I renamed the decorator to `run_time_decorator`.

PR #994, opened 2023-01-19
Update CI to pyiron/actions

jan-janssen:
After fixing the individual actions, I guess it is more efficient to switch to the central set of CI actions:
https://github.com/pyiron/pyiron_gpl/pull/123


jan-janssen:
All tests pass, but currently the `push-pull-main` workflow is implemented again in this repository. The more elegant solution would be to merge these changes to the `actions` repository or find a solution that works for both `pyiron_base` and `ironflow` https://github.com/pyiron/actions/pull/33

PR #995, opened 2023-01-19
set self._hdf5 only when ProjectHDF comes

samwaseda:
For [the hash](https://github.com/pyiron/pyiron_base/pull/903) it's important that `self._hdf5` is not used for the pseudo hdf.


liamhuber:
@samwaseda I was digging into why you can't create jobs after hashing, and it looks to me like there must be some deep property getting overridden at the class level -- i.e. so that all child classes also get this value updated, that's why I can't even create Vasp jobs after hashing a Lammps job!

This fix looks at least plausibly consistent with that.

Instead of merging this into main, can we just merge it directly into your other PR? I want to see it actually fix the problem we're running into before shoving it into main. I would be content to merge it into the other PR immediately, without compatibility tests or anything.

samwaseda:
> Instead of merging this into main, can we just merge it directly into your other PR? I want to see it actually fix the problem we're running into before shoving it into main. I would be content to merge it into the other PR immediately, without compatibility tests or anything.

I do it because you say so XD but @pmrv might not be very happy & it's a fairly different change in my opinion

PR #997, opened 2023-01-20
project - raise error when the directory was not created

### LINKED ISSUE #996: Project creation does not raise error
I just had the case of a user who ran out of space and when he created a new project pyiron did not complain even though the project folder could not be created. It took us quite some time to find this bug, so we should see why now error message is raised. As python `os.mkdir()` does raise an error.

jan-janssen:


pyiron_base/project/path.py
```diff
@@ -391,10 +391,9 @@ def _create_path(self, path, rel_path=None):
         if rel_path:
             rel_path = self._windows_path_to_unix_path(rel_path)
             path = posixpath.join(path, rel_path)
-        try:
-            os.makedirs(path)
-        except os.error:
-            pass
+        os.makedirs(path, exist_ok=True)
+        if not os.path.exists(path):
+            raise os.error("pyiron cannot create directory.")
```
pmrv: Do these lines actually have an effect? Either `os.makedirs` succeeds and doesn't raise an error, but then path exists, or it fails and the path doesn't exist, but it'll raise an error, right?

jan-janssen: No, that is the tricky part. When you are out of space - reached the limit of your quota - then `os.makedir` raises an `OsError` but Python thinks this means the folder already exists and continues. We just had this case recently and it is very tricky to find, as you create a project like `Project('test')` and no folder is created but also no error appears. Only later when you try to write into the project the script fails. Unfortunately, with a tool like pyiron it is rather easy to reach the limits of your quota, so I presume we might have people with similar issues soon.

pmrv: Oh, I see.  I would add a comment to this effect, though.  We can also think about raising this as a bug with python itself.

jan-janssen: I like that idea, let's see what they respond https://github.com/python/cpython/issues/101321

pyiron_base/project/path.py
```diff
@@ -391,10 +391,7 @@ def _create_path(self, path, rel_path=None):
         if rel_path:
             rel_path = self._windows_path_to_unix_path(rel_path)
             path = posixpath.join(path, rel_path)
-        try:
```
jan-janssen, line 394: The issue was this part - rather than using `os.makedirs(exist_ok=True)` we used our own `try except` which ignored the error when the folder was not created.

pyiron_base/project/path.py
jan-janssen: I had to look at the underlying code to realize it was not python's fault. The issue was on our side, as we had: 
```
         try:
             os.makedirs(path)
         except os.error:
             pass
```
This could be resolved by using `os.makedirs(path, exists_ok=True)`.

PR #1001, opened 2023-01-25
Allow external tags in `DataContainer`

samwaseda:
The main reason that the master jobs currently cannot use `DataContainer` is because they contain child jobs in their HDF files. I would like to suggest this change, because maybe it would allow the transition from the current implementation to the unified input structure (cf. [this comment](https://github.com/pyiron/pyiron_base/pull/974#discussion_r1084340423)) easier.

And obviously this should be removed as soon as we get rid of this input/output chaos.

### DIFF
--- a/pyiron_base/storage/datacontainer.py
+++ b/pyiron_base/storage/datacontainer.py
@@ -835,7 +835,11 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
             def normalize_key(name):
                 # split a dataset/group name into the position in the list and
                 # the key
-                k, i = name.split("__index_", maxsplit=1)
+                if "__index_" in name:
+                    k, i = name.split("__index_", maxsplit=1)
+                else:
+                    k = name
+                    i = -1
                 i = int(i)
                 if k == "":
                     return i, i


samwaseda:
I asked the review of @pmrv because he might have a better insight, while setting it to draft, because I haven't tested this with the master jobs.

pmrv:
Iirc, you should be able to do `DataContainer(hdf['path/to/my/normal/mapping'])` and it will automatically turn it into a proper `DataContainer`, so try that first.

EDIT: you might have to use `DataContainer.update(hdf, wrap=True)`, haven't tested it in a while.

samwaseda:
Now we have the following options (one of them **must** be done, because otherwise the current error message cannot be understood)

- Raise an error, as soon as a tag is encountered which is not supported by `DataContainer`
- Return the external entries along with he `DataContainer` entries
- Convert the external entries into `DataContainer`, update HDF and return the output

I guess what you are suggesting is the last option. My preference would be the first two options, because I have the feeling that we would break something if we try to convert everything into `DataContainer` by force.

samwaseda:
Just to make sure: I'm trying to address such a situation, where multiple sources store data in `job['input']`, for example. In this case, the data stored by `DataContainer` cannot be retrieved, because inside the HDF file the tags don't contain `__index_`.

pmrv:
Ah, I misunderstood the problem then.  Is the a reason the `DataContainer` replacement has to live in `input` directly and not in `input/new` or something like that?  I've posted also some code in the [old issue](https://github.com/pyiron/pyiron_base/issues/364) that doesn't need the weird names anymore.  I'd prefer that to be integrated than working around the issue.

samwaseda:
Now there's another argument for this PR -> [The new input/output interface](https://github.com/pyiron/pyiron_base/issues/1051) would currently disrupt the job loading, because if we move to `DataContainer` from dict-type HDF, then we would not be able to load the old jobs anymore. In order to be able to load old HDF, we would need this change. What do you think? @pmrv

pmrv:
Like I said, I don't object to the principle.  I've started some work based on the code I mentioned in the comment above.  I'll post it tomorrow.

samwaseda:
> Like I said, I don't object to the principle. I've started some work based on the code I mentioned in the comment above. I'll post it tomorrow.

Does it mean I should close this PR?

samwaseda:
ping @pmrv

samwaseda:
I merge this one because it wouldn't break anything.

pmrv:
Sorry, I didn't get to it today.  I suppose we can still replace "properly" later.

samwaseda:
ok but it's not necessary to revert this PR, right? I presume you'll just make the changes on top or replace it?

pmrv:
Yes, exactly.

PR #1002, opened 2023-01-25
Use builtin mappings() instead of looping in python

pmrv:
This would take up to 20% of database queries in projects >1k jobs.

On second profile the speed up is not as good as expected because now creating the pandas data frame from the list of dicts is the bottle neck, but that'll be for another day.


jan-janssen (approved):
Looks good to me

PR #1003, opened 2023-01-26
fix black formating

jan-janssen:


jan-janssen:
The last commit in https://github.com/pyiron/pyiron_base/pull/967 broke the black formatting - this is now fixed.

PR #1005, opened 2023-01-29
setting ProjectData to be lazy by default

### LINKED ISSUE #835: ProjectData should be lazy by default
It's probably enough to set the lazy flag in the overridden `__init__` and add some tests.

https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/project/data.py

yasser-sulaiman:
In order to improve performance, avoid unnecessary computations, and decrease memory requirements.
I have set the _lazy flag in ProjectData class to True so it is lazy by default.
closes #835

### DIFF
--- a/pyiron_base/project/data.py
+++ b/pyiron_base/project/data.py
@@ -43,9 +43,7 @@ class ProjectData(DataContainer):
         Args:
             project (pyiron_base.Project): The project instance the storage is attached to.
         """
-        super().__init__(*args, **kwargs)
-        # Projectdata should be lazy by default.
-        self._lazy = lazy
+        super().__init__(*args, lazy=lazy, **kwargs)
         self._project = project
 
     def read(self):
--- a/pyiron_base/project/data.py
+++ b/pyiron_base/project/data.py
@@ -44,6 +44,8 @@ class ProjectData(DataContainer):
             project (pyiron_base.Project): The project instance the storage is attached to.
         """
         super().__init__(*args, **kwargs)
+        # Projectdata should be lazy by default.
+        self._lazy = True
         self._project = project
 
     def read(self):
--- a/pyiron_base/project/data.py
+++ b/pyiron_base/project/data.py
@@ -36,7 +36,7 @@ class ProjectData(DataContainer):
         object.__setattr__(instance, "_project", None)
         return instance
 
-    def __init__(self, *args, project=None, **kwargs):
+    def __init__(self, *args, project=None, lazy=True, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


liamhuber (commented):
DataContainer has a "lazy" kwarg, so I don't see a need to set the variable manually; you can just add lazy=True to the super call. Or, maybe better, reproduce the full signature but change the default value for lazy to true and pass lazy=lazy so we still leave the door open to instantiating it non-lazily

yasser-sulaiman:
> DataContainer has a "lazy" kwarg, so I don't see a need to set the variable manually; you can just add lazy=True to the super call. Or, maybe better, reproduce the full signature but change the default value for lazy to true and pass lazy=lazy so we still leave the door open to instantiating it non-lazily

so I have added lazy with default value of True as a parameter to the ProjectData's constructor, is that what you meant?

pyiron_base/project/data.py
```diff
@@ -36,14 +36,16 @@ def __new__(cls, *args, **kwargs):
...
+    def __init__(self, *args, project=None, lazy=True, **kwargs):
         """
         A data storage container which can store itself to/retrieve itself from file at the project level.
 
         Args:
             project (pyiron_base.Project): The project instance the storage is attached to.
         """
         super().__init__(*args, **kwargs)
+        # Projectdata should be lazy by default.
+        self._lazy = lazy
```
pmrv:
```suggestion
        super().__init__(*args, lazy=lazy, **kwargs)
```

This is probably what @liamhuber meant.

yasser-sulaiman: ok then, thanks

yasser-sulaiman: so what happens now?
will the issue [#835](https://github.com/pyiron/pyiron_base/issues/835) be closed?

liamhuber (approved):

liamhuber:
> so what happens now?
> will the issue https://github.com/pyiron/pyiron_base/issues/835 be closed?

I flagged it to run the integration tests, I don't anticipate any problems but for the sake of completeness we'll let them run. Once they all pass I'll merge this and you can indeed close #835 😄

pmrv:
Sorry, I lost track of this.  Since the tests pass now, I think we can merge it?

yasser-sulaiman:
I am waiting for it to be merged actually @pmrv :)

pmrv:
> I am waiting for it to be merged actually @pmrv :)

Sorry to keep you waiting. Some things just drop off the radar for a bit sometimes.  Feel free to ping me next time. :)

PR #1027, opened 2023-02-06
Tell user to restart server when pyiron install ends

samwaseda:


pyiron_base/state/install.py
```diff
@@ -118,6 +118,7 @@ def install_dialog(silently=False):
                 giturl_for_zip_file="https://github.com/pyiron/pyiron-resources/releases/latest/download/resources.tar.gz",
                 git_folder_name="resources",
             )
+            print('pyiron resources installed - restart your server for the changes to be in effect')
```
jan-janssen, line 121: `restart` is not well defined in this context. It could be the computer, the jupyter server or as it is in this case the kernel of the jupyter notebook. As a rule of thumb in the pyiron context a restart of the jupyter server is only required when the `.bashrc` file is modified and a restart of the computer is never required. But I do not think this is clear for new users.

jan-janssen:
Please do not merge pull requests without at least one person reviewing them.

pyiron_base/state/install.py
samwaseda: Doesn't it update the the resource path?

jan-janssen:
> Please do not merge pull requests without at least one person reviewing them.

In particular when the tests fail !

pyiron_base/state/install.py
jan-janssen: restarting the computer updates the resource path. But it is not necessary, restarting the Jupyter kernel is sufficient.

Issue #1033, opened 2023-02-17
Why are jobs removed in two places when `delete_existing_job=True`?


samwaseda:
https://github.com/pyiron/pyiron_base/blob/f1334b623cb2bbdecf96aa928193439316e1ca29/pyiron_base/jobs/job/generic.py#L683-L692

In these lines the job is obviously removed. Then the same function, namely `def run`, calls [`run_job_with_status_finished`](https://github.com/pyiron/pyiron_base/blob/f1334b623cb2bbdecf96aa928193439316e1ca29/pyiron_base/jobs/job/runfunction.py#L226-L245), which does nearly the same things (I copy and paste the lines below)

```python
def run_job_with_status_finished(job, delete_existing_job=False, run_again=False):
    if run_again:
        delete_existing_job = True
    if delete_existing_job:
        parent_id = job.parent_id
        job.parent_id = None
        job.remove()
        job._job_id = None
        job.status.initialized = True
        job.parent_id = parent_id
        job.run()
```

Why?

samwaseda:
Side note: doing the same things twice is not quite relevant for the very first job of the database, because of `if self.job_id` (which is false for `job_id == 0`)

samwaseda:
https://github.com/pyiron/pyiron_base/blob/f1334b623cb2bbdecf96aa928193439316e1ca29/pyiron_base/jobs/job/generic.py#L683-L716

While reading the lines above, I realized that when `delete_existing_job` is True, it never executes `run_job_with_status_finished`, because `status` is changed to `True`, so we can also very well drop all lines under `if delete_existing_job:` in `run_job_with_status_finished`

PR #1034, opened 2023-02-17
Clean `generic.py` and `core.py`

samwaseda:
This PR is mostly about changing `if ID` to `if ID is not None`. So reviewing this PR should be relatively straightforward. There is just one function that I added: `remove_and_reset_id`, which should remove redundant lines in `kill`, `run` and `run_job_with_status_finished`

Edit: following [my comment](https://github.com/pyiron/pyiron_base/issues/1033), I removed the lines in `run_job_with_status_finished`

### DIFF
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -642,9 +642,7 @@ class GenericJob(JobCore):
         """
         Reset the job id sets the job_id to None in the GenericJob as well as all connected modules like JobStatus.
         """
-        if job_id is not None:
-            job_id = int(job_id)
-        self._job_id = job_id
+        super().reset_job_id(job_id=job_id)
         self._status = JobStatus(db=self.project.db, job_id=self._job_id)
 
     @deprecate(
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -117,7 +117,7 @@ class GenericJob(JobCore):
     __doc__ = (
         """
     Generic Job class extends the JobCore class with all the functionality to run the job object. From this class
-    all specific Hamiltonians are derived. Therefore it should contain the properties/routines common to all jobs.
+    all specific job types are derived. Therefore it should contain the properties/routines common to all jobs.
     The functions in this module should be as generic as possible.
 
     Sub classes that need to add special behavior after :method:`.copy_to()` can override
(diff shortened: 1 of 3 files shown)


pyiron_base/jobs/job/generic.py
```diff
@@ -990,10 +988,16 @@ def job_file_name(self, file_name, cwd=None):
         Returns:
             str: absolute path to the file in the current working directory
         """
-        if not cwd:
+        if cwd is None:
             cwd = self.project_hdf5.working_directory
         return posixpath.join(cwd, file_name)
 
+    def _set_hdf(self, hdf=None, group_name=None)
```
```suggestion
    def _set_hdf(self, hdf=None, group_name=None):
```

samwaseda:
https://github.com/pyiron/pyiron_base/blob/104460b5a9b5c55a9842275624b8d0c7d46775da/pyiron_base/jobs/job/core.py#L829-L839

It sounds like it should raise `NotImplementedError`, but in reality it is defined here. Can I move the current lines to `GenericJob` and make the one in `JobCore` raise `NotImplementedError`? @jan-janssen

jan-janssen:
> It sounds like it should raise `NotImplementedError`, but in reality it is defined here. Can I move the current lines to `GenericJob` and make the one in `JobCore` raise `NotImplementedError`? @jan-janssen

As the `job_id` is defined on the `JobCore` level and `reset_job_id()` is used in the `copy()` function and the `_internal_copy_to()` function, I guess it is not a good idea to move it right now. Still if you want to move it, I would suggest making a separate pull request, so we can analyze the dependencies in more detail.

jan-janssen:
@samwaseda To me this again is a pull request with too many different changes. It really simplifies the reviewing process if you make separate pull requests for changes that belong together. In this case the introduction of the `remove_and_reset_id(self, _protect_childs=True)` function is a separate topic which should be a separate pull request to make it easier for the reviewers to see which changes are related to this change.

samwaseda:
> @samwaseda To me this again is a pull request with too many different changes. It really simplifies the reviewing process if you make separate pull requests for changes that belong together. In this case the introduction of the `remove_and_reset_id(self, _protect_childs=True)` function is a separate topic which should be a separate pull request to make it easier for the reviewers to see which changes are related to this change.

Last time you specifically said 100 lines, and I was sticking to it in this PR + I did not create a new functionality; I only removed redundant lines and `if None` statements. I don't really understand your criterion of "too many changes", so I'm gonna close this one

jan-janssen:
> Last time you specifically said 100 lines, and I was sticking to it in this PR + I did not create a new functionality; I only removed redundant lines and `if None` statements. I don't really understand your criterion of "too many changes", so I'm gonna close this one

I cannot really follow this reaction. Just by separating this pull request into to https://github.com/pyiron/pyiron_base/pull/1035 and https://github.com/pyiron/pyiron_base/pull/1036 it becomes much more obvious which changes belong together.

PR #1036, opened 2023-02-18
Clean `generic.py` and `core.py` - part 2

### LINKED ISSUE #1033: Why are jobs removed in two places when `delete_existing_job=True`?

jan-janssen:


pyiron_base/jobs/job/generic.py
```diff
@@ -642,9 +648,7 @@ def reset_job_id(self, job_id=None):
         """
         Reset the job id sets the job_id to None in the GenericJob as well as all connected modules like JobStatus.
         """
-        if job_id is not None:
-            job_id = int(job_id)
-        self._job_id = job_id
+        super().reset_job_id(job_id=job_id)
```
```suggestion
        self.reset_job_id(job_id=job_id)
```

samwaseda: No it must be `super`, because otherwise it's gonna call itself.

PR #1041, opened 2023-02-20
Raise warning when the job_name is changed

jan-janssen:


samwaseda:
To be honest I don't want a warning at this point, because it would just make it annoying when you have a large number of jobs. What's much more important is that the user can run and load the job consistently, i.e.

```python
job = pr.create.job.MyJob('job name')
...
job.run()
job = pr.load('job name')
```

This must work, and how pyiron internally stores the data shouldn't matter to the user.

PR #1044, opened 2023-02-23
Implement @to_hdf_decorator decorator

jan-janssen:
Currently, the decorator is primarily used to check if the HDF5 file should be written or not. But the decorator could also be extended to other functionalities.

### DIFF
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -45,7 +45,7 @@ from pyiron_base.jobs.job.util import (
     _kill_child,
     _job_store_before_copy,
     _job_reload_after_copy,
-    to_hdf_decorator
+    to_hdf_decorator,
 )
 from pyiron_base.utils.instance import static_isinstance
 from pyiron_base.utils.deprecate import deprecate
@@ -173,7 +173,9 @@ class GenericJob(JobCore):
         if self._data_storage_disabled_implemented:
             self._data_storage_enabled = enabled
         else:
-            raise NotImplementedError("This JobType does not support disabling the data storage.")
+            raise NotImplementedError(
+                "This JobType does not support disabling the data storage."
+            )
 
     @property
     def version(self):
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -434,7 +434,7 @@ class GenericJob(JobCore):
         """
         Refresh job status by updating the job status with the status from the database if a job ID is available.
         """
-        if self.data_storage_enabled and self.job_id:
+        if self.data_storage_enabled and self.job_id is not None:
             self._status = JobStatus(
                 initial_status=self.project.db.get_job_status(self.job_id),
                 db=self.project.db,
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -45,6 +45,7 @@ from pyiron_base.jobs.job.util import (
     _kill_child,
     _job_store_before_copy,
     _job_reload_after_copy,
+    to_hdf_decorator
 )
 from pyiron_base.utils.instance import static_isinstance
 from pyiron_base.utils.deprecate import deprecate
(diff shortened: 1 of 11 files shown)


pmrv:
I only had a quick look: Does this mean every job class has to redecorate their `to_hdf` with the decorator?  I'm also not sure why we need this here in base, I thought we agreed we keep the new develop in parallel only in contrib?

jan-janssen:
> I only had a quick look: Does this mean every job class has to redecorate their `to_hdf` with the decorator? 

In the long / mid-term yes, but for now it is sufficient to implement it for the jobs with should work without data storage. 

> I'm also not sure why we need this here in base, I thought we agreed we keep the new develop in parallel only in contrib?

Yes, that was the precious agreement. Still seeing how disabling the file system simplifies key features like the MPI based parallelization, I guess we should move ahead with this migration more quickly. As some of us are participating at the IPAM   workshop on "New Mathematics for the Exascale: Applications to Materials Science"[1] I see that the impact pyiron can have on the community can drastically increase when we support parallel execution. A more detailed explanation can be found in the discussions[2]. Besides the parallel execution another advantage is that we no longer need job_names, which makes using pyiron much more pythonic. 

The way I envision the next version of pyiron is that file system storage and database become completely optional. Those who want to use these features can enable them, but for the default use of pyiron they are no longer required. There could be three different versions: 
* **pure python** - no database and no files are written 
* **external executables** - HDF5 based storage in combination with the file table, which enables the submission to the queuing system and the use of external executable. 
* **database based** - SQL database accelerates the indexing of the HDF5 files, this particularly accelerates high throughput DFT studies. 

[1]: http://www.ipam.ucla.edu/programs/long-programs/new-mathematics-for-the-exascale-applications-to-materials-science/
[2]: https://github.com/orgs/pyiron/teams/pyiron/discussions/160

pyiron_base/jobs/job/generic.py
```diff
@@ -418,13 +434,13 @@ def refresh_job_status(self):
         """
         Refresh job status by updating the job status with the status from the database if a job ID is available.
         """
-        if self.job_id:
```
samwaseda, line 421: If you are anyway changing this line, can you change it to `if self.job_id is not None`?

samwaseda: sorry I wanted to comment on the changed line, but you know what I mean.

samwaseda:
I have the feeling that `to_hdf` is anyway a function that should disappear from everywhere, so I might be a bit reluctant to add functionalities. At least I don't really have the feeling that we've talked it over yet.

pyiron_base/jobs/job/generic.py
jan-janssen: I updated it.

jan-janssen:
> I have the feeling that `to_hdf` is anyway a function that should disappear from everywhere, so I might be a bit reluctant to add functionalities. At least I don't really have the feeling that we've talked it over yet.

I agree, still disabling the functionality now until we have a more general functionality available seem to be a way to move forward and gain more experience.

liamhuber:
> The way I envision the next version of pyiron is that file system storage and database become completely optional. 

Do we have a clear picture where Joerg sits on this? Personally, I actually like moving in this direction, but this is a _really_ fundamental shift in philosophy; building a full consensus would be good, but at the very least he needs to be 100% sold on such a change.

> At least I don't really have the feeling that we've talked it over yet.

From this I guess the consensus piece is at least not there!

EDIT: I don't mean this comment to stop the merging of this particular PR, but we seem to have a lot of key players in the thread already, so I wanted to raise the concern of moving too far too fast before we get others on board.

jan-janssen:
> From this I guess the consensus piece is at least not there!

I agree. Still we also have to consider the scope of this pull request. The only thing that changes for existing users is that we have one more if statement. At the same time for the users who are interested in using pyiron without file system access this drastically reduces the number of lines to implement new classes https://github.com/pyiron/pyiron_contrib/pull/600 .

pmrv:
So the more maintainable way would be to derive `GenericJob` from [HasHDF](https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/interfaces/has_hdf.py) and then add this new `data_enabled` flag to that interface.  Then the decorator is not needed anymore and its functionality can be implemented in `HasHDF.to_hdf`.

jan-janssen:
> So the more maintainable way would be to derive `GenericJob` from [HasHDF](https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/interfaces/has_hdf.py) and then add this new `data_enabled` flag to that interface. Then the decorator is not needed anymore and its functionality can be implemented in `HasHDF.to_hdf`.

I agree but that to me sounds like a larger / separate change.

jan-janssen:
Based on today's discussion, I created the https://github.com/pyiron/pyiron_lammps . It is the first draft of a **pure python** pyiron version. Currently, it only supports `Lammps` but I plan to add the `GenericMasters` next, to have the same functionality as `nofiles` in `pyiron_contrib`. I have to admit the more I work on the idea of a standalone module, the more I like it, as it allowed me to drastically reduce the dependencies. In the mid-term I see `pyiron_atomistics` depending on both `pyiron_base` and `pyiron_lammps`.

PR #1054, opened 2023-03-07
HasHDF: Use generic list_groups instead of list_dirs

pmrv:
For ProjectHDFio they are the same anyway.


jan-janssen (approved):
Looks good to me

PR #1057, opened 2023-03-16
Use pprint to represent DataContainers, not forcing HDFStubs

### LINKED ISSUE #1031: Rendering DataContainers on an ipython shell
I just noticed that only jupyter lab renders the DataContainer class as foldable structure, while the same rendering does not work in jupyter notebooks or the ipython shell. This makes it difficult to work with hierarchical input files, for example the Sphinx DFT code. Example Code: 
```
d = DataContainer({"a": 1, "b": 2})
d
```
For the `GenericParameters` class this was not an issue: 
```
from pyiron_base import GenericParameters
d = GenericParameters()
d.set_dict({"a": 1, "b": 2})
d
```
It renders fine in jupyter lab, jupyter notebooks and on an ipython shell. 

@samwaseda, @pmrv Any suggestions? Are you using jupyter lab exclusively? Or is there something I am missing, which enables the rendering of the Datacontainer objects?

niklassiemer:
to_builtin got a new argument load_stubs=True such that loading HDFStubs can be avoided.

Might close #1031

### DIFF
--- a/pyiron_base/storage/datacontainer.py
+++ b/pyiron_base/storage/datacontainer.py
@@ -218,7 +218,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
     >>> g.my_fancy_field
     42
     >>> e
-    ExtendedContainer({'foo': 1, 'bar': 5, 'sub': ExtendedContainer({'fnord': 23})})
+    ExtendedContainer({'foo': 1, 'bar': 5, 'sub': {'fnord': 23}})
 
     For that reason most of time you'll actually want a class that uses a DataContainer for storage, but doesn't derive
     from it.
@@ -415,9 +415,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
 
     def __repr__(self):
         name = self.__class__.__name__
-        return (
-            f"{name}({pformat(self.to_builtin(load_stubs=False), indent=2, depth=2)})"
-        )
+        return f"{name}({pformat(self.to_builtin(load_stubs=False, _no_str_keys=True), indent=2, depth=2, sort_dicts=False)})"
 
     @property
     def read_only(self):
@@ -441,7 +439,7 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
             "finished.".format(cls.__name__)
         )
 
-    def to_builtin(self, stringify=False, load_stubs=True):
+    def to_builtin(self, stringify=False, load_stubs=True, _no_str_keys=False):
         """
         Convert the container back to builtin dict's and list's recursively.
 
@@ -462,7 +460,8 @@ class DataContainer(MutableMapping, HasGroups, HasHDF):
                 # requires all string keys when storing as json), since
                 # _normalize calls int() on all digit string keys this is
                 # transparent for the rest of the module
-                k = str(k)
+                if not _no_str_keys:
+                    k = str(k)
                 if isinstance(v, DataContainer):
                     dd[k] = v.to_builtin(stringify=stringify, load_stubs=load_stubs)
                 else:
--- a/pyiron_base/storage/datacontainer.py
+++ b/pyiron_base/storage/datacontainer.py
@@ -12,6 +12,7 @@ from collections.abc import Sequence, Set, Mapping, MutableMapping
 
 import numpy as np
 import pandas
+from pprint import pformat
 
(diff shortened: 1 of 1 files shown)


niklassiemer:
With these changes, the `DataContainer` representation is better to read on the command line and in notebooks - at least in my opinion. However, it also slightly changes for subgroups, which are not shown as `DataContainer` any more:
For the `DataContainer` 
```python
dc = DataContainer({"a": 1, "b": 2, 'c':{'A':1}})
```
we get with the current main 
```python
DataContainer({'a': 1, 'b': 2, 'c': DataContainer({'A': 1})})
```
but with these changes we would get
```python
DataContainer({"a": 1, "b": 2, 'c':{'A':1}})
```
Furthermore, it (currently) omits everything nested deeper than the first two levels by '...'.

pyiron_base/storage/datacontainer.py
```diff
@@ -451,50 +439,53 @@ def _read_only_error(cls):
             "finished.".format(cls.__name__)
         )
 
-    def to_builtin(self, stringify=False):
+    def to_builtin(self, stringify=False, load_stubs=True, _no_str_keys=False):
```
pmrv, line 442: `load_stubs` should have a short documentation even though it's fairly obvious.  If I see correctly `_no_str_keys` is not used.

pmrv:
> With these changes, the `DataContainer` representation is better to read on the command line and in notebooks - at least in my opinion. However, it also slightly changes for subgroups, which are not shown as `DataContainer` any more: For the `DataContainer`
> 
> ```python
> dc = DataContainer({"a": 1, "b": 2, 'c':{'A':1}})
> ```
> 
> we get with the current main
> 
> ```python
> DataContainer({'a': 1, 'b': 2, 'c': DataContainer({'A': 1})})
> ```
> 
> but with these changes we would get
> 
> ```python
> DataContainer({"a": 1, "b": 2, 'c':{'A':1}})
> ```
> 
> Furthermore, it (currently) omits everything nested deeper than the first two levels by '...'.

I have a slight preference to keep the class name in the output to differentiate true dict members from `DataContainer` sub groups or their subclasses (e.g. `Group` from Sphinx). Omitting after a certain depth is ok with me.

niklassiemer:
I agree, however, the pprint does not naively work on the `DataContainer` with depth truncation. Therefore, I use the to_builtin under the hood, which of course converts to plain dict... I'll take another shot at it. Maybe I find a way to tell pprint to use the `DataContainer` directly.

pyiron_base/storage/datacontainer.py
niklassiemer: Yes, documentation is missing and I forgot to pass the `_no_str_keys` to the sub groups. However, it is used and switches off the key conversion such that the keys stay e.g. int if they are.

pyiron_base/storage/datacontainer.py
```diff
@@ -451,50 +439,53 @@ def _read_only_error(cls):
...
     # allows "nice" displays in jupyter lab
     def _repr_json_(self):
-        return self.to_builtin(stringify=True)
-
-    # allows 'nice' display in notebooks
-    def _repr_html_(self):
-        name = self.__class__.__name__
-        plain = f"{name}({json.dumps(self.to_builtin(stringify=True), indent=2, default=str)})"
-        return "<pre>" + plain + "</pre>"
+        return self.to_builtin(stringify=True, load_stubs=False)
```
niklassiemer, line 488: Also a bit the question: Do we want to not load the stubs in this case? Also the nice jupyter lab interface would not show the full container from the beginning but the stubs. To have the full view one would need to call to_builtin beforehand.

pmrv:
I found [this](https://bugs.python.org/issue1373762) and [this](https://stackoverflow.com/questions/3258072/best-way-to-implement-custom-pretty-printers), which suggests to subclass the `PrettyPrinter`.  Your call if that's worth it. :') 

Not forcing stubs is prefered, I think, since it might incur significant I/O.

pmrv:
See also the second answer from the SO link and [this POTW](https://pymotw.com/3/pprint/).  Those manage without subclassing, by doing `__repr__` in certain way, but haven't looked at the details.

pmrv:
[Last link I promise.](https://ipython.readthedocs.io/en/stable/config/integrating.html#pretty-printing).

PR #1064, opened 2023-04-04
Allow unknown credentials

niklassiemer:


pyiron_base/state/settings.py
```diff
@@ -372,10 +379,48 @@ def _get_config_from_environment(self) -> Union[Dict, None]:
             elif k in self.environment_credential_map:
                 config[self.environment_credential_map[k]] = v
         config = self._fix_boolean_var_in_config(config=config)
-        config = self._add_credentials_from_file(config)
         return config if len(config) > 0 else None
 
-    def _add_credentials_from_file(self, config: dict) -> Dict:
+    def _add_credentials_from_file(self) -> Dict:
+        if (
```
liamhuber, line 405: It might be nice to have an `else` clause that adds something to the log about no credentials file being found.

pyiron_base/state/settings.py
```diff
@@ -372,10 +379,48 @@ def _get_config_from_environment(self) -> Union[Dict, None]:
...
+            parser.read(credential_file)
+            credentials = {}
+            for sec_name, section in parser.items():
+                credentials_w = {}
+
+                for k, v in section.items():
+                    if k.upper() in self.file_credential_map:
+                        credentials_w[self.file_configuration_map[k.upper()]] = v
+                    else:
+                        credentials_w[k.lower()] = v
```
liamhuber: This is a little weird looking, but I guess it's something like this: there is a str:str map with all upper-case keys and lower-case values, and if you can find the credential key in there, you store it at the remapped key; else you force it to be lower and store it regardless.

Could we extract this into its own function to make the current scope more readable? Something like this maybe:

```
credentials_w[self._get_remapped_credential_key(k)] = v
```

and 

```
def _get_remapped_credential_key(self, k):
    """
    A docstring like my comment if my comment was correct, maybe with an explanation of 
    why we would want to keep a (k,v) even if k isn't in our nice map
    """
    if k.upper() in self.file_credential_map:
        return self.file_configuration_map[k.upper()]
    else:
        return k.lower()
```

Actually, now I'm a little confused why there are two separate maps being used here! I didn't even notice that until I sat down to re-write the method 😝 So it is maybe even more helpful than I thought to extract it as a method and include a little docstring helping devs understand why we're doing what we're doing here.

pyiron_base/state/settings.py
```diff
@@ -372,10 +379,48 @@ def _get_config_from_environment(self) -> Union[Dict, None]:
...
+                if len(credentials_w) > 0:
+                    credentials[sec_name.upper()] = credentials_w
+            return credentials
+
+    def _update_credentials_from_std_pyiron_config(self):
+        pyiron_dict_name = "PYIRON"
```
liamhuber: I'm not diving deep into the surrounding scope here, but it smells a little bit like this should be a (module-level) global `PYIRON_DICT_NAME = "PYIRON"`. You will have a much better sense whether or not that's the case, I just want to nudge the consideration.

tests/state/test_settings.py
```diff
@@ -166,6 +168,12 @@ def test_validate_no_database_configuration(self):
                 }
             )
 
+    def _test_config_and_credential_synchronization(self):
+        pyiron_key = "PYIRON"
```
liamhuber: Looping back to my comment about `PYIRON_DICT_NAME` as a global, now it _really_ smells like that's the case 😉 

Ah, I see now this is actually in a different file. But the comment stands; Right now if we wanted to update its value to "PYIRON_CONFIG_STUFF", we'd need to also update this literal in the test file, so let's define the literal at the class or module level and use it self-consistently here.

liamhuber (commented):
Nice. I didn't review for actual performance -- I'm not deeply into this corner of the code, and your tests look complete, so on this front I return a value of "trust" with high confidence. Rather, I tried to review from the perspective of what I'd want to see if this broke sometime in the future and I had to come in, learn it, and fix it.

pyiron_base/state/settings.py
niklassiemer: I will extract it :) and two maps are not required... Right now the credential map is a strict subset of the configuration map and it slipped my notice. I would rather warn if I find a non-credential but configuration key in the _credentials_ file since this will not be used to configure pyiron in this case.

pyiron_base/state/settings.py
niklassiemer: This would only be something for a debug log imoh. This is just like that option has not been used in the main credentials. In that case we first should start to warn for unrecognized keys in the config such that users may ceck the spelling.

tests/state/test_settings.py
liamhuber: I declare unresolved! 😉 `pyiron_key` can now be imported instead of redefined

niklassiemer: Oh, well done :) It already is imported and used at various other places where `'PYIRON'` was before. However I forgot this one for some reason...

tests/state/test_settings.py
```diff
@@ -198,10 +207,12 @@ def test_get_config_from_environment(self):
             )
 
         local_loc = Path(self.cwd + "/.pyiron_credentials")
-        local_loc.write_text(f"[DEFAULT]\nPASSWD = something_else\n")
+        local_loc.write_text(
+            f"[DEFAULT]\nPASSWD = something_else\n[OTHER]\nNoPyironKey = token"
```
niklassiemer:
```suggestion
            "[DEFAULT]\nPASSWD = something_else\n[OTHER]\nNoPyironKey = token"
```

PR #1067, opened 2023-04-20
Fix run_time_to_db() for jobs submitted to remote computers

jan-janssen:
@ligerzero-ai

### DIFF
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -1336,7 +1336,7 @@ class GenericJob(JobCore):
         """
         Internal helper function to store the run_time in the database
         """
-        if self.job_id is not None:
+        if not state.database.database_is_disabled and self.job_id is not None:
             self.project.db.item_update(self._runtime(), self.job_id)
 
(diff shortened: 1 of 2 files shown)


niklassiemer (commented):
That solution would be ok for me. However, couldn't we shift the 'is a db available check' into the database interface itself?

ligerzero-ai:
wip - we are trying to hunt down a phantom error

ligerzero-ai:
It should work:

image

Fetching with `wait_for_jobs` works.

Remote system with PBSPro queueing system. Still needs to be tested for slurm

jan-janssen:
> That solution would be ok for me. However, couldn't we shift the 'is a db available check' into the database interface itself?

While I agree that cleaning up the database interface to handle the specific case of a local installation with database and a remote installation without an database. This pull request list primarily focused on solving the issues that @Leimeroth reported quite some time ago.

jan-janssen:
> It should work:

To clarify - it works if the calculation finishes successful, currently there is still an issue for failed or not converged calculation. This still needs to be addressed.

PR #1074, opened 2023-04-25
Reindex file table

### LINKED ISSUE #1073: FileTable not updating its path

MWE:
```python

pr = Project('tmp')
f = FileTable(pr.path)
print(f._project)
>>> ".../tmp"

pr2 = Project('tmp2')
f2 = FileTable(pr2.path)
print(f._project)
>>> ".../tmp"
print(f2._project)
>>> ".../tmp" 
```

liamhuber:
I introduce a new singleton class that re-runs `__init__` on each instantiation, and use it to make `FileTable` re-index itself iff it is being instantiated with a new location.

Closes #1073 

Necessary for #951 to get passing tests

### DIFF
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -7,6 +7,7 @@ File based database interface
 
 import datetime
 from collections.abc import Iterable
+from typing import Optional
 import numpy as np
 import os
 import pandas
@@ -48,7 +49,7 @@ table_columns = {
 
 
 class FileTable(IsDatabase, metaclass=UpdatableSingleton):
-    def __init__(self, path=None):
+    def __init__(self, path: Optional[str] = None):
         try:
             if path is not None and os.path.abspath(path) != self._project:
                 # Only re-index if you receive a new location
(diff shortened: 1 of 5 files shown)


pyiron_base/database/filetable.py
```diff
@@ -47,13 +48,26 @@
...
-    def __init__(self, project):
-        self._fileindex = None
-        self._job_table = None
-        self._project = os.path.abspath(project)
-        self._columns = list(table_columns.keys())
-        self.force_reset()
+class FileTable(IsDatabase, metaclass=UpdatableSingleton):
+    def __init__(self, path: Optional[str] = None):
+        try:
+            if path is not None and os.path.abspath(path) != self._project:
```
jan-janssen, line 54: I feel this is dangerous. The typical example that comes to mind is a project with path `a/b/c` followed by a subproject with path `a/b/c/d` .

liamhuber: Isn't it just "dangerous" in the sense that you may pay the computational cost of re-indexing when you don't need to?

IMO the fact this thing is still a singleton _at all_ is what's dangerous.

I opened another PR (#1075) which follows your other advice of just making new instances.

I'd like to continue discussion of both on the issue #1073

liamhuber:
Not as good as #1076

PR #1075, opened 2023-04-25
File table not singleton

### LINKED ISSUE #1073: FileTable not updating its path

liamhuber:
Just don't make the darned thing a singleton.

Projects get an if-clause so they only instantiate (and thus index) one instance per instance.

Closes #1073 

In competition with #1074.

Something like this is necessary for #951.

### DIFF
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -47,10 +47,10 @@ table_columns = {
 
 
 class FileTable(IsDatabase):
-    def __init__(self, project):
+    def __init__(self, path):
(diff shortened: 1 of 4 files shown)


liamhuber:
Not as good as #1076

PR #1076, opened 2023-04-28
Efficient and stable file table

liamhuber:
Addresses #1073, where we showed that running without the database makes it so you can't even instantiate a second project (and perhaps other problems).

Safer than #1074, and faster than #1075 (assuming that indexing is actually expensive, which sounds very true to me.)

Here we use a path-based-singleton setup suggested by @niklassiemer so each new path gets its own file table instance -- this fixes the problem while instantiating (and thus indexing) as few file paths as possible.
I guess memory scaling would be bad if you instantiate a billion of these at a billion different paths, but that seems like strange behaviour and a much less severe problem than not even being able to instantiate a second project successfully...

I am still not 100% sure how this (or the current HEAD!) interact with `GenericJob.transfer_from_remote`, which, in the case that you _are_ using the database, creates a file table _for the job alone_ as an intermediate step and then updates the database with that.
In the case that you're not using the database while you call the `update` method of the `FileTable`, and this change should have no effect.


liamhuber:
```
FAIL: test_re_initialization (database.test_filetable.TestFileTable.test_re_initialization)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/database/test_filetable.py", line 38, in test_re_initialization
    self.assertTrue(
AssertionError: False is not true : We promise re-initialization to be much faster, but got times 6.44e-06 for the first and 4.29e-06 for the second
```

This surprises me. Locally I get 8.2e-03 and 8.8e-06 seconds for first (with indexing) and second initialization, respectively. These both look like re-initializations at this path...

liamhuber:
Riiiight, because when I'm running it locally I'm not running the other tests and there's another test in the database folder that initializes a file table _in this location_. So, indeed, we really are seeing the re-initialization times for both runs. 

This is easily fixed.

pyiron_base/database/filetable.py
```diff
@@ -47,11 +47,38 @@
...
+    singleton system that does this once for each path instead.
+    """
+
+    _instances = {}
+
+    def __call__(cls, index_from):
+        _path = os.path.abspath(index_from)
+        if _path not in cls._instances:
+            cls._instances[_path] = super(FileTableSingleton, cls).__call__(index_from)
+        return cls._instances[_path]
```
jan-janssen, line 62: I like this implementation, there is just one case where it is inefficient: 
```
pr_a = Project("a")
pr_b = Project("a/b")
```
We can imagine the sub project `a/b` containing all the data, while `a` contains just three groups `a/b`, `a/c` and `a/d`. In this case we waste resources as the whole index has to be calculated twice. But maybe that is something that can be more easily fixed from the `pyfileindex` level.

liamhuber: I think we will need at least some edits here, right, because the `PyFileIndex` objects instantiated by `FileTable` don't know about each other? I can imagine that if the `_path` is a subset of and `_instances.keys()` path, that we extract the `PyFileIndex` attribute from the super-path instance, slice it (or otherwise reduce it down, I'm not exactly sure the details of `PyFileIndex`) and use _that_ (or a copy of it) when instantiating the new `FileTable` instead of creating the new `PyFileIndex` object from scratch.

jan-janssen: > I think we will need at least some edits here, right, because the `PyFileIndex` objects instantiated by `FileTable` don't know about each other? 

Yes. 

> I can imagine that if the `_path` is a subset of and `_instances.keys()` path, that we extract the `PyFileIndex` attribute from the super-path instance, slice it (or otherwise reduce it down, I'm not exactly sure the details of `PyFileIndex`) and use _that_ (or a copy of it) when instantiating the new `FileTable` instead of creating the new `PyFileIndex` object from scratch.

I started working on this in https://github.com/pyiron/pyfileindex/pull/74/

jan-janssen: For the case that we have multiple overlapping paths we could use: 
```
max([p for p in cls._instances.keys() if os.path.commonpath([_path, p]) == p], key=len)
```

Still this needs some additional modifications for the FileTable to accept an existing `PyFileIndex`.

liamhuber: Yes. I would suggest to merge this as-is, so that we can instantiate mutliple projects without failure, then open a second PR with changes to `FileTable` to allow instantiating from an existing `PyFileIndex` object.

PR #1080, opened 2023-05-01
add function container

jan-janssen:
This was a demonstration, primarily for the IPAM workshop on workflows, how we could combine the concepts of ScriptJob and GenericMaster. It still requires the integration with `cloudpickle` and as we discussed last time it most likely makes more sense to move these developments first to `pyiron_contrib`.


pyiron_base/jobs/functioncontainer.py
```diff
@@ -0,0 +1,14 @@
+from pyiron_base.jobs.job.generic import GenericJob
+
+
+class FunctionContainer(GenericJob):
```
liamhuber, line 4: Do you mean `PythonTemplateJob`?

pyiron_base/jobs/functioncontainer.py
```diff
@@ -0,0 +1,14 @@
...
+    def __init__(self, project, job_name):
+        super().__init__(project, job_name)
+        self._python_only_job = True
+
+    def run_static(self):
+        for kwargs in self.input:
+            self.function(**kwargs)
```
liamhuber, line 11: `GenericJob` doesn't have an `input` class, so it's not totally clear what `in` is doing here, but naively I would expect this to call `self.function` multiple times. Do you mean something like this?

```suggestion
        self.function(**self.input.to_dict())
```

pyiron_base/jobs/functioncontainer.py
jan-janssen: The issue is that `DataContainer` cannot store projects, that is why I cannot use the `PythonTemplateJob`.

pmrv: What is missing to store projects in `DataContainer` and how does this sidestep the problem?  I don't think `Project` comes with a `to_hdf`, but that also seems easy to add (just save the path).

PR #1082, opened 2023-05-03
Add support for additional arguments in the queue scripts

### LINKED ISSUE #1083: Queue submission interactive additions
Hi All,

In interacting with clusters with mutliple users on different accounts and potentially unreliable nodes, adding options for 1. runtime assignment of nodes to exclude and 2. slurm accounts (for example) to utilize for the jobs - having integrations with something like setting job.server.exclude and job.server.account would be ideal. As far as I can tell this change would involve pysqa and the pyiron_base libraries.

-Michael.

jan-janssen:
Example: 
```
job.server.additional_arguments["account"] = "my_account"
job.server.additional_arguments["exclude"] = "node-85,node-86,node-87"
```

And in your queue template: 
```
#!/bin/bash
#SBATCH --output=time.out
#SBATCH --job-name={{job_name}}
#SBATCH --chdir={{working_directory}}
#SBATCH --get-user-env=L
#SBATCH --partition=slurm
{%- if account %}
#SBATCH --account={{account}}
{%- endif %}
{%- if exclude %}
#SBATCH --exclude={{exclude}}
{%- endif %}
{%- if run_time_max %}
#SBATCH --time={{ [1, run_time_max // 60]|max }}
{%- endif %}
{%- if memory_max %}
#SBATCH --mem={{memory_max}}G
{%- endif %}
#SBATCH --cpus-per-task={{cores}}

{{command}}
```

These additional arguments should be propagated.

### DIFF
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -94,7 +94,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         self._send_to_db = False
         self._structure_id = None
         self._accept_crash = False
-        self._additional_arguments = {}
+        self.additional_arguments = {}
 
     @property
     def send_to_db(self):
@@ -398,10 +398,6 @@ class Server:  # add the option to return the job id and the hold id to the serv
         """
         return self.view_queues()
 
-    @property
-    def additional_arguments(self):
-        return self._additional_arguments
-
     @staticmethod
     def list_queues():
         """
@@ -449,7 +445,8 @@ class Server:  # add the option to return the job id and the hold id to the serv
         hdf_dict["run_time"] = self.run_time
         hdf_dict["memory_limit"] = self.memory_limit
         hdf_dict["accept_crash"] = self.accept_crash
-        hdf_dict.update(self._additional_arguments)
+        if len(self._additional_arguments) > 0:
+            hdf_dict["additional_arguments"] = self.additional_arguments
 
         if group_name is not None:
             with hdf.open(group_name) as hdf_group:
@@ -491,24 +488,9 @@ class Server:  # add the option to return the job id and the hold id to the serv
             self._accept_crash = hdf_dict["accept_crash"] == 1
         if "threads" in hdf_dict.keys():
             self._threads = hdf_dict["threads"]
+        if "additional_arguments" in hdf_dict.keys():
+            self.additional_arguments = hdf_dict["additional_arguments"]
         self._new_hdf = hdf_dict["new_h5"] == 1
-        options_used = [
-            "user",
-            "host",
-            "run_mode",
-            "queue",
-            "qid",
-            "structure_id",
-            "cores",
-            "run_time",
-            "memory_limit",
-            "accept_crash",
-            "threads",
-            "new_h5",
-        ]
-        for k, v in hdf_dict.items():
-            if k not in options_used:
-                self._additional_arguments[k] = v
 
     def db_entry(self):
         """
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -445,7 +445,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         hdf_dict["run_time"] = self.run_time
         hdf_dict["memory_limit"] = self.memory_limit
         hdf_dict["accept_crash"] = self.accept_crash
-        if len(self._additional_arguments) > 0:
+        if len(self.additional_arguments) > 0:
             hdf_dict["additional_arguments"] = self.additional_arguments
 
         if group_name is not None:
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -449,8 +449,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         hdf_dict["run_time"] = self.run_time
         hdf_dict["memory_limit"] = self.memory_limit
         hdf_dict["accept_crash"] = self.accept_crash
-        for k, v in self._additional_arguments.items():
-            hdf_dict[k] = v
+        hdf_dict.update(self._additional_arguments)
 
         if group_name is not None:
             with hdf.open(group_name) as hdf_group:
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -94,6 +94,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         self._send_to_db = False
         self._structure_id = None
         self._accept_crash = False
+        self._additional_arguments = {}
 
     @property
     def send_to_db(self):
@@ -397,6 +398,10 @@ class Server:  # add the option to return the job id and the hold id to the serv
         """
         return self.view_queues()
 
+    @property
+    def additional_arguments(self):
+        return self._additional_arguments
+
     @staticmethod
     def list_queues():
         """
@@ -444,6 +449,8 @@ class Server:  # add the option to return the job id and the hold id to the serv
         hdf_dict["run_time"] = self.run_time
         hdf_dict["memory_limit"] = self.memory_limit
         hdf_dict["accept_crash"] = self.accept_crash
+        for k, v in self._additional_arguments.items():
+            hdf_dict[k] = v
 
         if group_name is not None:
             with hdf.open(group_name) as hdf_group:
@@ -486,6 +493,13 @@ class Server:  # add the option to return the job id and the hold id to the serv
         if "threads" in hdf_dict.keys():
             self._threads = hdf_dict["threads"]
         self._new_hdf = hdf_dict["new_h5"] == 1
+        options_used = [
+            "user", "host", "run_mode", "queue", "qid", "structure_id", "cores",
+            "run_time", "memory_limit", "accept_crash", "threads", "new_h5"
+        ]
+        for k, v in hdf_dict.items():
+            if k not in options_used:
+                self._additional_arguments[k] = v
 
     def db_entry(self):
         """
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -494,8 +494,18 @@ class Server:  # add the option to return the job id and the hold id to the serv
             self._threads = hdf_dict["threads"]
         self._new_hdf = hdf_dict["new_h5"] == 1
         options_used = [
-            "user", "host", "run_mode", "queue", "qid", "structure_id", "cores",
-            "run_time", "memory_limit", "accept_crash", "threads", "new_h5"
+            "user",
+            "host",
+            "run_mode",
+            "queue",
+            "qid",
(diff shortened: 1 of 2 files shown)


jan-janssen:
@mgt16-LANL Can you test if this works for you?

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -397,6 +398,10 @@ def queue_view(self):
         """
         return self.view_queues()
 
+    @property
+    def additional_arguments(self):
+        return self._additional_arguments
+
```
liamhuber: Is it absolutely necessary to wrap this as a private variable with read-only access, or is it sufficient to define `self.additional_arguments = {}` in line 97 and just let users suffer the consequences if they decide to overwrite this field?

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -444,6 +449,8 @@ def to_hdf(self, hdf, group_name=None):
         hdf_dict["run_time"] = self.run_time
         hdf_dict["memory_limit"] = self.memory_limit
         hdf_dict["accept_crash"] = self.accept_crash
+        for k, v in self._additional_arguments.items():
+            hdf_dict[k] = v
```
liamhuber:
```suggestion
        hdf_dict.update(self._additional_arguments)
```

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -486,6 +493,23 @@ def from_hdf(self, hdf, group_name=None):
...
+            "cores",
+            "run_time",
+            "memory_limit",
+            "accept_crash",
+            "threads",
+            "new_h5",
+        ]
+        for k, v in hdf_dict.items():
+            if k not in options_used:
+                self._additional_arguments[k] = v
```
liamhuber: I'm really uncomfortable with big hard-coded exemption lists like this -- they are hard to grok and a pain to maintain. If we absolutely need to build such a list, let's instead look at doing it programatically. Here, it just looks like you're making sure that arg names are not duplicated. [Here](https://github.com/pyiron/pyiron_contrib/blob/b44cf1ebfc966cf48c1895dda9654c39eb7becdf/pyiron_contrib/workflow/node.py#L381) is an example of pulling out signatures for later use.

I'm also confused why this list contains the values it does. Some of these are clearly being disallowed because they are already kwargs or attributes of `Server` (`host`, `cores`, etc). Others kind of overlap? Like you prohibit `user` and `Server._user` exists (I double checked, there is no public `Server.user` wrapping this) -- but this is not actually used by the invocation over in runfunction.py. Are there other places these `additional_arguments` are getting used? Because right now you prohibit a bunch of things that wouldn't have any impact?

Also, it looks like you _allow_ a bunch of things that will cause a downstream crash. Like, when I go look at `state.queue_adapter` I see that the default is a `PySQAAdpter`. And if I look at `PySQAAdpter.submit_job` -- aha, nevermind. My local copy of pysqa is out of date. I see now that [pysqa takes and tosses a bunch of `**kwargs`](https://github.com/pyiron/pysqa/pull/169). _**NO**_, double aha! [`pysqa.QueueAdapter.submit_job`](https://github.com/pyiron/pysqa/blob/832e4c4559257f0a6e7251acac7f19e7d1fb54ee/pysqa/queueadapter.py#L134) that pyiron invokes does _not_ take `**kwargs` yet! So shouldn't this whole thing throw an unrecognized kwarg error if you actually try to use it with any `additional_arguments`? The tests here are passing though, so either I am somehow wrong about this or the tests need to be expanded.

Long story short, I find this messy and hard to understand/follow, and have concerns about maintainability and (lack of) encapsulation/integration with other tools we use (i.e. pysqa).

liamhuber (commented):
~Please provide some high-level description when opening the PR. It was not a big deal for
#1080 which was a draft, but for a green, open PR, context would be extremely helpful.~ Got added after the PR was created, but not long after so that's fine.

I have some concerns here about structure/maintainability, and as far as I can tell, this should just throw an error if you try to actually use a populated `additional_kwargs` because the main upstream class (`pysqa.QueueAdapter.submit_job`) can't handle it?

jan-janssen:
> I have some concerns here about structure/maintainability, and as far as I can tell, this should just throw an error if you try to actually use a populated `additional_kwargs` because the main upstream class (`pysqa.QueueAdapter.submit_job`) can't handle it?

Support was just added in `0.0.21`: https://github.com/pyiron/pyiron_base/pull/1081

pyiron_base/jobs/job/extension/server/generic.py
jan-janssen: I agree that is it not required right now, but we might want to add additional checks here in future. Still I am also fine with updating it and making it a global variable of the class.

pyiron_base/jobs/job/extension/server/generic.py
jan-janssen: The `pysqa` version was just updated to `0.0.21` which now supports the `kwargs` so all I want the user to add is the arguments which are not yet handled by `pyiron_base`. As shown above examples could be `account` or `exclude`. 

Still I agree that the maintainability of this part is tricky. A more clean solution would be to move the whole storage of the server object to one central dictionary or ideally a `DataContainer` but that to me sounds like a larger change.

liamhuber:
> > I have some concerns here about structure/maintainability, and as far as I can tell, this should just throw an error if you try to actually use a populated `additional_kwargs` because the main upstream class (`pysqa.QueueAdapter.submit_job`) can't handle it?
> 
> Support was just added in `0.0.21`: #1081

I don't think it was? Go look at the signature for `pysqa.queueadapter.QueueAdapter.submit_job`: https://github.com/pyiron/pysqa/blob/832e4c4559257f0a6e7251acac7f19e7d1fb54ee/pysqa/queueadapter.py#L134

jan-janssen:
> I don't think it was? Go look at the signature for `pysqa.queueadapter.QueueAdapter.submit_job`: https://github.com/pyiron/pysqa/blob/832e4c4559257f0a6e7251acac7f19e7d1fb54ee/pysqa/queueadapter.py#L134

You are right - I added a pull request to fix this https://github.com/pyiron/pysqa/pull/171

pyiron_base/jobs/job/extension/server/generic.py
liamhuber: Yeah, I'm not dead set against making it private either. But if you're comfortable then let's go with the more concise definition for now, and add the private variable+wrapper if and when we add those additional checks

pyiron_base/jobs/job/extension/server/generic.py
liamhuber: I think I'm in favour of a broader but cleaner change? I understand it would take more time now, but with the implementation here, I worry about lost time in the future both in terms of stuff breaking/misbehaving and no one understanding how to fix it, and how much study will be required by future devs before they can make changes here.

At an absolute minimum we should break this list up into components with good variable names or even explicit comments about where these strings are coming from and why they should be forbidden.

liamhuber:
> > I don't think it was? Go look at the signature for `pysqa.queueadapter.QueueAdapter.submit_job`: https://github.com/pyiron/pysqa/blob/832e4c4559257f0a6e7251acac7f19e7d1fb54ee/pysqa/queueadapter.py#L134
> 
> 
> 
> You are right - I added a pull request to fix this https://github.com/pyiron/pysqa/pull/171

I think it's also important to add a test here, now, while it's at the front of mind. It's not good that the changes were made here with no warnings to us that we had broken interface demands made by pysqa.

jan-janssen:
> I think it's also important to add a test here, now, while it's at the front of mind. It's not good that the changes were made here with no warnings to us that we had broken interface demands made by pysqa.

The challenge is that we do not have a queuing system in our test environment. I am currently doing quite a bit of refactoring on the `pysqa` side to increase the test coverage by moving the function which calls the queuing system on the command line to the outside https://github.com/pyiron/pysqa/pull/170 . But this does not solve the general challenge of mock-up tests.

liamhuber:
> The challenge is that we do not have a queuing system in our test environment. I am currently doing quite a bit of refactoring on the `pysqa` side to increase the test coverage by moving the function which calls the queuing system on the command line to the outside https://github.com/pyiron/pysqa/pull/170 . But this does not solve the general challenge of mock-up tests. 

I would support getting a mock-up running so we can at least test the signature interface. But I'm not volunteering to do it, so I'm not going to insist on it here.

If we now have certain interface demands, maybe a middle ground is to document them better, either in Server or probably better in QueueAdapters. In practice we only use a pysqa adapter (if any), but in ~practice~ principle we allow other adapters. Can we specify, perhaps in QueueAdapters.adapter what our expectations are in terms of methods, signatures, and behaviour?

pyiron_base/jobs/job/extension/server/generic.py
pmrv: > Still I agree that the maintainability of this part is tricky. A more clean solution would be to move the whole storage of the server object to one central dictionary or ideally a `DataContainer` but that to me sounds like a larger change.

Not sure, how big of a change it actually would be, in #728 I did it for the `Executable` class which should be very similar.  Anyways, even without converting to `DataContainer` based storage, can't we just set a dictionary with the extra keys (which h5io should handle), i.e. have the new keys in an extra hierarchy level instead of a flat version?

jan-janssen: I am also fine with moving the additional arguments to a different dictionary.

mgt16-LANL:
Hi all - I was able to build and test this successfully with a slurm cluster pyiron installation. Worked like a charm!

jan-janssen:
@pmrv and @liamhuber anymore feedback? otherwise I would merge it like this.

pmrv (approved):
lgtm, but as @liamhuber said, probably good to think about how to document/codify the interface midterm.

liamhuber (approved):
Yes, this looks fine now.

It's not _part_ of this PR, but the _process_ of this PR really highlighted for me the fragility of our interface with the queue adapter. I'll make an issue for that though.

niklassiemer:
I like the flexibility.

Wouldn't this also allow for 

```
job.server.additional_arguments["env_command"] = 'conda activate pyiron-stable' 
```
With the  queue template: 
```
#!/bin/bash
#SBATCH --output=time.out
#SBATCH --job-name={{job_name}}
#SBATCH --chdir={{working_directory}}
#SBATCH --get-user-env=L
#SBATCH --partition=slurm
{%- if account %}
#SBATCH --account={{account}}
{%- endif %}
{%- if exclude %}
#SBATCH --exclude={{exclude}}
{%- endif %}
{%- if run_time_max %}
#SBATCH --time={{ [1, run_time_max // 60]|max }}
{%- endif %}
{%- if memory_max %}
#SBATCH --mem={{memory_max}}G
{%- endif %}
#SBATCH --cpus-per-task={{cores}}

{{env_command}}

{{command}}
```

mgt16-LANL:
That's pretty much what I had tested when building this on our local cluster.

jan-janssen:
@niklassiemer Yes, exactly you can basically do anything with the additional commands. The only limitation I see is that the `additional_arguments` do not provide any standardization. So you might call it `env_command` and somebody else might call it `conda_environment_path` and so on.

PR #1093, opened 2023-05-13
Provide PyFileIndex as additional input 

jan-janssen:


liamhuber:
👍👍 will review on Monday after the meeting.

pyiron_base/database/filetable.py
```diff
@@ -56,8 +56,22 @@ class FileTableSingleton(ABCMeta):
...
+        _path = os.path.abspath(os.path.expanduser(index_from))
         if _path not in cls._instances:
+            path_match_lst = [
+                p for p in cls._instances.keys() if os.path.commonpath([_path, p]) == p
+            ]
+            if len(path_match_lst) > 0:
+                cls._instances[_path] = super(FileTableSingleton, cls).__call__(
+                    index_from=index_from,
+                    fileindex=cls._instances[
+                        max(path_match_lst, key=len)
```
liamhuber: Cool, I didn't know you could `key` `max` like this!

pyiron_base/database/filetable.py
```diff
@@ -56,8 +56,22 @@ class FileTableSingleton(ABCMeta):
     _instances = {}
 
     def __call__(cls, index_from):
-        _path = os.path.abspath(index_from)
+        _path = os.path.abspath(os.path.expanduser(index_from))
         if _path not in cls._instances:
+            path_match_lst = [
```
liamhuber: Let's refactor the path matching out as a static method, then we can have a unit test for the behaviour.

Method draft:
```python
@staticmethod
def get_most_precise_common_path(path, reference_paths):
    try:
        return max(
            [
                p for p in reference_paths 
                if os.path.commonpath([path, p]) == p
            ],
            key=len
        )
    except ValueError:
        # When no common path is found
        return None
```

Test sketch:
```python
reference_paths = [
    os.path.abspath(os.path.expanduser("~/work")),
    os.path.abspath(os.path.expanduser("~/work/pyiron"))
]
test_sibling = os.path.abspath(os.path.expanduser("~/personal"))
test_parent = os.path.abspath(os.path.expanduser("~"))
test_identity = os.path.abspath(os.path.expanduser("~/work/pyiron"))
test_child = os.path.abspath(os.path.expanduser("~/work/events"))

for p in [
    test_sibling,
    test_parent,
    test_identity,
    test_child
]:
    print(p, foo(p, reference_paths))
>/Users/huber/personal None
>/Users/huber None
>/Users/huber/work/pyiron /Users/huber/work/pyiron
>/Users/huber/work/events /Users/huber/work
```

pyiron_base/database/filetable.py
```diff
@@ -56,8 +56,22 @@ class FileTableSingleton(ABCMeta):
...
+                p for p in cls._instances.keys() if os.path.commonpath([_path, p]) == p
+            ]
+            if len(path_match_lst) > 0:
+                cls._instances[_path] = super(FileTableSingleton, cls).__call__(
```
liamhuber, line 65: In fact, I think we can go even further and add a class method to avoid nested if-else in the main call function, so it looks like this:

```python
    def __call__(cls, index_from):
        path = os.path.abspath(os.path.expanduser(index_from))
        if path not in cls._instances:
            cls._instances[_path] = super(FileTableSingleton, cls).__call__(
                index_from=index_from,
                fileindex=cls._get_common_fileindex_if_available(path)
            )
        return cls._instances[path]
```

where 

```python
@classmethod
_get_common_fileindex_if_available(cls, path) -> PyFileIndex | None:
    common_path = cls._get_most_precise_common_path(
        path,
        cls._instances.keys() 
    )
    if common_path is None:
        return None
    else:
        return cls._instances[common_path]._fileindex.open(path)
```

liamhuber (commented):
Yeah, looks good!

I've suggested a refactor of the `__call__` to break it apart into a nested class and static method; the end result is that `__call__` is super readable (no nested if-else, just calls a clearly named class method) and the raw manipulations of file paths can be subjected to unit testing (since it's now just a static method).

pyiron_base/database/filetable.py
```diff
@@ -56,9 +56,20 @@ class FileTableSingleton(ABCMeta):
...
 
     def __call__(cls, index_from):
-        _path = os.path.abspath(index_from)
+        _path = os.path.abspath(os.path.expanduser(index_from))
         if _path not in cls._instances:
-            cls._instances[_path] = super(FileTableSingleton, cls).__call__(index_from)
+            common_path = get_most_common_path(
+                path=_path, reference_paths=cls._instances.keys()
+            )
+            if common_path is not None:
```
liamhuber, line 64: I see that this was extracted to a method then that change was reverted. I can live with it as is -- although I still prefer to avoid the nested if-else by pulling it out -- but I'm curious why it got reverted? Does it not work?

pyiron_base/database/filetable.py
```diff
@@ -636,3 +657,11 @@ def get_job_status_from_file(hdf5_file, job_name):
         return read_hdf5(hdf5_file, job_name + "/status")
     else:
         return None
+
+
+def get_most_common_path(path, reference_paths):
```
liamhuber: Unless we explicitly expect this to get used somewhere else, I'd just make it private for now. If someone finds it and wants it, we can always make it public later.

liamhuber (approved):
One question and one nit, but otherwise lgtm.

pyiron_base/database/filetable.py
jan-janssen: I got a strange error calling `__call__()`: 
```
  File "/home/runner/work/pyiron_base/pyiron_base/pyiron_base/database/filetable.py", line 69, in _get_common_fileindex
    return super(FileTableSingleton, cls).__call__(index_from=path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: descriptor '__call__' of 'type' object needs an argument
```

pyiron_base/database/filetable.py
jan-janssen: Agreed - I made it a private function. In the mid-term I would like to move it to a more general place as we do something similar when matching the current directory to a project path. Still that can be done in a separate pull request.

pyiron_base/database/filetable.py
liamhuber: Cool, no worries then. Thanks for the trace too, I'll experiment with a MWE to see if I can figure out the python of what's going on here, just for educational purposes.

PR #1095, opened 2023-05-13
Updates to FileTable 

### LINKED ISSUE #1073: FileTable not updating its path

jan-janssen:
Fixes: https://github.com/pyiron/pyiron_base/issues/1073

Combines: https://github.com/pyiron/pyiron_base/pull/1076 and https://github.com/pyiron/pyiron_base/pull/1093 

Creating a separate pull request on the `main` branch triggers the unit tests while a pull request on from one branch to the next does not.


liamhuber:
> Creating a separate pull request on the `main` branch triggers the unit tests while a pull request on from one branch to the next does not.

In principle that's fine if what you're trying to do is a stacked PR. If you target this one to another branch, then merge that branch to main, it will automatically rebase onto main, and at that point I think the tests will get run

jan-janssen:
@liamhuber Can you take another look? I implemented your suggestions and would like to merge this pull request ideally before the end of the week, as we have a hackathon coming up next week.

liamhuber:
> @liamhuber Can you take another look? I implemented your suggestions and would like to merge this pull request ideally before the end of the week, as we have a hackathon coming up next week. 

For sure. Will get back to you by the end of the day.

liamhuber:
@jan-janssen, I left two more comments on #1093, but I would accept it being merged as-is.

What then is the purpose for this PR? Was it just to force CI on the whole shebang? From my perspective #1076 fixes the bug, then #1093 makes things more efficient. There's no fundamental problem to leaving the code in the #1076 state, so we can just merge those two in relatively quick succession and close this.

PR #1101, opened 2023-05-18
Polish filetable

liamhuber:
So this PR is 100% just nit resolutions.

@pmrv had some weird trouble with GitHub but let me know by e-mail his good insight into the trouble with [`TypeError: descriptor '__call__' of 'type' object needs an argument`](https://github.com/pyiron/pyiron_base/pull/1093#discussion_r1197082285) -- we have a nasty interaction of `classmethod` and `super` where we're passing types instead of instances. I played around with a MWE in a notebook for just a couple minutes and didn't find a good solution, but I realized we don't need to instantiate the new object in the class method anyhow, we just need to get a file index out of it (or not) and that's easy. So here I refactor away the if-else clause, which reeeeaaally is just a nit, but I still like it.

The other thing is that while I was doing this I ran into the situation that the filetable tests failed, and on re-running them they failed again because some folders they created already existed. So now we just pull out the creation/deletion to make sure that cleanup happens regardless of the test state. No more manually deleting local test folders for me!

### DIFF
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -56,13 +56,13 @@ class FileTableSingleton(ABCMeta):
     _instances = {}
 
     def __call__(cls, index_from):
-        _path = os.path.abspath(os.path.expanduser(index_from))
-        if _path not in cls._instances:
-            cls._instances[_path] = super(FileTableSingleton, cls).__call__(
-                    index_from=_path,
-                    fileindex=cls._get_fileindex_if_theres_a_common_path(_path),
+        path = os.path.abspath(os.path.expanduser(index_from))
+        if path not in cls._instances:
+            cls._instances[path] = super(FileTableSingleton, cls).__call__(
+                    index_from=path,
+                    fileindex=cls._get_fileindex_if_theres_a_common_path(path),
                 )
-        return cls._instances[_path]
+        return cls._instances[path]
 
     def _get_fileindex_if_theres_a_common_path(cls, path):
         common_path = _get_most_common_path(
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -59,9 +59,9 @@ class FileTableSingleton(ABCMeta):
         path = os.path.abspath(os.path.expanduser(index_from))
         if path not in cls._instances:
             cls._instances[path] = super(FileTableSingleton, cls).__call__(
-                    index_from=path,
-                    fileindex=cls._get_fileindex_if_theres_a_common_path(path),
-                )
+                index_from=path,
+                fileindex=cls._get_fileindex_if_theres_a_common_path(path),
+            )
         return cls._instances[path]
 
     def _get_fileindex_if_theres_a_common_path(cls, path):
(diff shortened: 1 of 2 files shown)


jan-janssen (approved):
Looks good to me

pyiron_base/database/filetable.py
```diff
@@ -56,21 +56,22 @@ class FileTableSingleton(ABCMeta):
     _instances = {}
 
     def __call__(cls, index_from):
-        _path = os.path.abspath(os.path.expanduser(index_from))
-        if _path not in cls._instances:
-            common_path = _get_most_common_path(
-                path=_path, reference_paths=cls._instances.keys()
+        path = os.path.abspath(os.path.expanduser(index_from))
+        if path not in cls._instances:
+            cls._instances[path] = super(FileTableSingleton, cls).__call__(
```
pmrv: I'm a bit confused still, because I don't know the details of meta classes.  Is `__call__` called on an instance of the meta class?  If no then the `super` call here is the same as in the closed PR and should make the same problems.  If not then we should rename `cls` to `self`.

liamhuber: @pmrv that is because it is confusing 😆 

Indeed, the `cls` variable means different things in the metaclass `__call__` signature and in the metaclass class method signatures. In `FileTableSingleton.__call__`, `cls` is the derived child class using that metaclass (in our case, `FileTable`); if you stick a class method on the metaclass, then `cls` is just the metaclass. This is why the `super().__call__(cls)` call didn't work inside the metaclass class method -- the super it was expecting  the derived class but got the metaclass.

Example:
```python
from abc import ABCMeta

class MyMeta(ABCMeta):
    def __call__(cls):
        print(f"In metaclass __call__, cls = {cls}({type(cls)})")
        cls.meta_class_method()
        cls.meta_method()
        return super().__call__()
    
    @classmethod
    def meta_class_method(meta_cls):
        print(f"In metaclass class method, meta_cls = {meta_cls}({type(meta_cls)})")
        
    def meta_method(cls):
        print(f"In metaclass method, cls = {cls}({type(cls)})")
        
        
class MyClass(metaclass=MyMeta):
    def __init__(self):
        print(f"In class __init__, self = {self}({type(self)})")
        
MyClass() 
>>>In metaclass __call__, cls = <class '__main__.MyClass'>(<class '__main__.MyMeta'>)
>>>In metaclass class method, meta_cls = <class '__main__.MyMeta'>(<class 'type'>)
>>>In metaclass method, cls = <class '__main__.MyClass'>(<class '__main__.MyMeta'>)
>>>In class __init__, self = <__main__.MyClass object at 0x14edec410>(<class '__main__.MyClass'>)
```

IMO, `cls` is the right name for the first argument, but it is the _derived_ `cls` not the metaclass `cls`! If we ever needed an actual class method on a metaclass, I'd probably name this argument `meta_cls` (I couldn't quickly find on google if there's already some other standard for this).

Thus, the current implementation with the method on the metaclass using `cls` is consistent and, IMO, good. I believe the old implementation would have worked if it had been a regular method instead of class method, but I actually like this refactoring more anyhow.

I'll clean up the super calls as a plain `super()` should work.

PR #1104, opened 2023-05-23
Allow for project_path to be relative

### LINKED ISSUE #1103: :bug: `FileTable` does not play well with non-`None` project paths
This has caused me an embarassing amount of trouble over on #951. The crux of the problem is that the `project` column in the `FileTable` database-like object is using abspaths, while the `FileTable` logic itself is looking at the `Project.project_path` which is only relative to the `project_paths` from the settings.

# Demonstration

This works great:

```python
from pyiron_base import Project, state

pr = Project("test")
print(pr.root_path, pr.project_path)
>>> None /Users/huber/work/pyiron/notebooks/test/
# Note that I don't have TOP_LEVEL_DIRS specified in my config

job = pr.create.job.ScriptJob(job_name="foo")
job.save()

state.update({"disable_database": True})

print(len(pr.job_table(recursive=False)))
>>> 1
      
pr.remove(enable=True)
```

While this fails to find the job:

```python
from pyiron_base import Project, state

from pathlib import Path
state.update({"project_paths": str(Path(".").absolute())})

pr = Project("test")
print(pr.root_path, pr.project_path)
>>> /Users/huber/work/pyiron/notebooks/ test/

job = pr.create.job.ScriptJob(job_name="foo")
job.save()

state.update({"disable_database": True})

print(len(pr.job_table(recursive=False)))
>>> 0

print(pr.job_table().project[0])
>>> /Users/huber/work/pyiron/notebooks/test/

pr.remove(enable=True)
```

# Why

The offending code is [here in `FileTable`](https://github.com/pyiron/pyiron_base/blob/0fe1191f8c8a655a5ffc2a19dad33a54089b7bf1/pyiron_base/database/filetable.py#L602), where our `if recursive` clause uses `else: return self._job_table[self._job_table.project == project_path]`. `project_path` comes from `Project.project_path`, which is relative to the `Project.root_path`, but the jobs use the absolute path (I can't find the code responsible for this, but you see it in the snippet above).

# What to do?

IMO, the quickest and easiest solution is to change `self._job_table.project == project_path` to `self._job_table.project.str.endswith(project_path)`. I'll go make a PR that does that right after I post this. But, this isn't a corner I really play with -- I don't think I've ever specified `TOP_LEVEL_DIRS` -- so I wanted to get some feedback to see if anyone else sees a danger of side effects.

liamhuber:
Closes #1103.

This allows `project_path` to be a relative path and the `project` column of the `job_table` to be absolute paths at the same time.

### DIFF
--- a/pyiron_base/database/filetable.py
+++ b/pyiron_base/database/filetable.py
@@ -604,7 +604,9 @@ class FileTable(IsDatabase, metaclass=FileTableSingleton):
                     self._job_table.project.str.contains(project_path)
                 ]
             else:
-                return self._job_table[self._job_table.project == project_path]
+                return self._job_table[
+                    self._job_table.project.str.endswith(project_path)
+                ]
         else:
             return self._job_table
 


jan-janssen:
But should not the path be always absolute? I thought the project already resolves the path to be absolute.

liamhuber:
> But should not the path be always absolute? I thought the project already resolves the path to be absolute. 

I would be perfectly happy if project path were always absolute, but it is not now. Go look at the linked issue and you can see evidence of this.

pyiron_base/database/filetable.py
```diff
@@ -604,7 +604,9 @@ def _get_job_table(
                     self._job_table.project.str.contains(project_path)
                 ]
             else:
-                return self._job_table[self._job_table.project == project_path]
+                return self._job_table[
+                    self._job_table.project.str.endswith(project_path)
+                ]
```
jan-janssen, line 609: This is the same as enforcing `recursive=True` so it reduces the functionality of pyiron and in particular for the file table this can be dangerous.

jan-janssen (changes requested):
This is the same as enforcing `recursive=True` so it reduces the functionality of pyiron and in particular for the file table this can be dangerous.

pyiron_base/database/filetable.py
liamhuber: This is incorrect. If a job has the project "/foo/bar/baz/" and the `project_path` is "/foo/bar/", then `.contains` (for `recursive=True`) will find the project and `.endswith` (for `recursive=False`) will not.

PR #1105, opened 2023-05-24
Remove working directory warnings

jan-janssen:
At least for pure python jobs as those do not write input and output files, but rather just call `run_static()`. In addition the developer of a traditional job which writes input and output files can be removed by setting `job._write_work_dir_warinings = False` .

### DIFF
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -158,6 +158,7 @@ class GenericJob(JobCore):
         self._process = None
         self._compress_by_default = False
         self._python_only_job = False
+        self._write_work_dir_warnings = True
         self.interactive_cache = None
         self.error = GenericError(job=self)
 
@@ -378,7 +379,7 @@ class GenericJob(JobCore):
         Write the input files for the external executable. This method has to be implemented in the individual
         hamiltonians.
         """
-        if state.settings.configuration["write_work_dir_warnings"]:
+        if state.settings.configuration["write_work_dir_warnings"] and self._write_work_dir_warnings and not self._python_only_job:
             with open(
                 os.path.join(self.working_directory, "WARNING_pyiron_modified_content"),
                 "w",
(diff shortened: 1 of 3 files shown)


niklassiemer:
I am fine with adding an active switch for jobs which do not modify the possibly known input formats of the external executable.
However, if we introduce the active switch `_write_work_dir_warnings` I am against another implicit switch with the python only jobs. With the active switch being true, I would always expect the warning and not some other side effect suppressing this.

jan-janssen:
> However, if we introduce the active switch `_write_work_dir_warnings` I am against another implicit switch with the python only jobs. With the active switch being true, I would always expect the warning and not some other side effect suppressing this.

But a python only job does not even call `write_input()` so there is never any input file. I do not understand why we still need the warning file.

niklassiemer:
I agree, I just do not want two variables to control the same thing. For the python only job the explicit control is absolutely meaningless. I would be open to just use the python only job switch and developers could be expected to set the config correctly?

jan-janssen:
> I would be open to just use the python only job switch and developers could be expected to set the config correctly?

The config is user controlled, the internal property is developer controlled.

niklassiemer:
> > I would be open to just use the python only job switch and developers could be expected to set the config correctly?
> 
> The config is user controlled, the internal property is developer controlled. 

Sure, however, the intent of the warning is to warn the user that the content in the working directory is pyiron property and might be different then expected. Once this is understood by the user, they can turn this warning file off? 
Is the problem about non-python jobs which handle input and output not file based and, thus, have empty working directories? 

The proposed changes are nothing I am really concerned with. Just from the code design I do not like the scenario with two variables controlling one thing, where the one with an unrelated name has the higher priority for switching this off.

jan-janssen:
>  Is the problem about non-python jobs which handle input and output not file based and, thus, have empty working directories?

I guess there are some jobs where pyiron just writes the input, naming the `randspg` job as one example, while others like VASP we have this inconsistency that we discussed. Currently the warning is written everywhere. I would prefer a warning that developers can turn on for their code. But then obviously every developer has to remember to add it. Now the compromise is to at least give the developer the option to disable the warning file, if they feel it is annoying for them.

PR #1112, opened 2023-06-02
Only try to lock Traits objects

pmrv:
Previously we tried to duck-type this, but it doesn't play well with some exotic `__getattr__` overloads that I have.  I left the previous check in, though, because I wasn't sure if `lock()`ing can fail in other cases.

### DIFF
--- a/pyiron_base/storage/has_stored_traits.py
+++ b/pyiron_base/storage/has_stored_traits.py
@@ -309,20 +309,15 @@ class HasStoredTraits(HasTraits, HasStorage, ABC, metaclass=ABCTraitsMeta):
         """Recursively make all traits read-only."""
         self._read_only = True
         for sub in self.trait_values().values():
-            if not isinstance(sub, HasStoredTraits): continue
-            try:
+            if isinstance(sub, HasStoredTraits):
                 sub.lock()
-            except AttributeError:
-                pass
 
     def unlock(self):
         """Recursively make all traits both readable and writeable"""
         self._read_only = False
         for sub in self.trait_values().values():
-            try:
+            if isinstance(sub, HasStoredTraits):
                 sub.unlock()
-            except AttributeError:
-                pass
 
     def __setattr__(self, key, value):
         if key == "_read_only":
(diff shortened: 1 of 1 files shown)


liamhuber (commented):
I'm on mobile and can't check deeply atm -- isn't lock a method of `trait let's.HasTraits` s.t. Everything returned here in `trait_values()` also has a valid `lock` method? I don't object and I believe you that it causes trouble, I would benefit from a better hint on how it causes trouble 😂

pmrv:
If I understand it correctly `trait_values` gives the values behind the traitlet objects and then the current lock method then tries to see if any of those are also `HasTraits` by duck-typing on the `lock` method. This works for normal objects and for `HasTraits` objects, but when you define a class that always returns something on attribute access the check naturally fails.

liamhuber:
Ah, thanks for the refresher and I get it and to the best of my current understanding you're right. In this case wouldn't it be better to check the typing for `HasTraits` directly instead of `HasStoredTraits`? It would still work in the later case, and would still catch cases where someone has directly used trailets downstream and stuck a different `HasTraits` child in there. Also since we've got the if, can't we replace the try/except with an else? (And probably then reverse the if case so if positive: do, else: pass, for readability)

pmrv:
`lock` is from us, `HasTraits`doesn't define it, if didn't miss it.  A-ok on the other points though.

pyiron_base/storage/has_stored_traits.py
```diff
@@ -309,6 +309,7 @@ def lock(self):
         """Recursively make all traits read-only."""
         self._read_only = True
         for sub in self.trait_values().values():
+            if not isinstance(sub, HasStoredTraits): continue
             try:
                 sub.lock()
             except AttributeError:
```
liamhuber: I thought we no longer need the duck typing at all? (Also the suggestion is incomplete because GitHub won't let me touch the `pass` line at the end...)

```suggestion
            if isinstance(sub, HasStoredTraits):
                sub.lock()
```

pmrv: Added a commit to the same effect.

liamhuber (approved):
Right, good remembering `unlock` too!

PR #1113, opened 2023-06-02
Do not try to double write objects that define to_hdf

pmrv:
`ProjecHDFio` checks whether values to be written define a `to_hdf` method and call it correctly, but then the conditional falls through and `h5py` is then called to write the same value again.  Which obviously fails.  This just adds an early return to avoid this.


liamhuber (approved):
Good catch!

PR #1115, opened 2023-06-04
HasStoredTraits: Use setitem instead of setattr

pmrv:
Previously the class puts its trait values into the underlying storage by doing

setattr(self.storage, key, value)

This creates problems when `key` clashes with a predefined method name, because DataContainer first resolves attribute names to predefined methods and attributes.  So if you define a trait with a conflicting name HasStoredTraits will then overwrite the method or attribute instead of writing to the container, which means this clashing value is not properly stored.

The solution is simply to use item access, because that never resolves to predefined methods and attributes like

self.storage[key] = value


liamhuber (approved):
lgtm

PR #1120, opened 2023-06-06
Implement run_mode flux based on the flux python interface

jan-janssen:
Example: 
```
>>> import flux.job
>>> from pyiron_atomistics import Project
>>> pr = Project('test')
>>> job = pr.create.job.Lammps("lmp")
>>> job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
>>> job.flux_executor = flux.job.FluxExecutor()
>>> job.server.cores = 2
>>> fs = job.run()
>>> print(fs.done())
>>> print(fs.result())
>>> print(fs.done())
False
0
True
```

The `fs` object here is a python `concurrent.futures` object, so you can use `fs.done()` to check if the calculation is finished or call `fs.result()` to wait until the calculation is finished.

### DIFF
--- a/pyiron_base/jobs/job/extension/server/runmode.py
+++ b/pyiron_base/jobs/job/extension/server/runmode.py
@@ -26,6 +26,7 @@ run_mode_lst = [
     "thread",
     "worker",
     "srun",
+    "flux",
     "interactive",
     "interactive_non_modal",
 ]
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -693,7 +693,7 @@ class GenericJob(JobCore):
                 elif status == "initialized":
                     self._run_if_new(debug=debug)
                 elif status == "created":
-                    self._run_if_created()
+                    return self._run_if_created()
                 elif status == "submitted":
                     run_job_with_status_submitted(job=self)
                 elif status == "running":
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -38,6 +38,7 @@ from pyiron_base.jobs.job.runfunction import (
     run_job_with_runmode_interactive_non_modal,
     run_job_with_runmode_queue,
     run_job_with_runmode_srun,
+    run_job_with_runmode_flux,
     execute_job_with_external_executable,
 )
 from pyiron_base.jobs.job.util import (
@@ -840,6 +841,13 @@ class GenericJob(JobCore):
         calculation to separate nodes in a SLURM based HPC cluster.
         """
         run_job_with_runmode_srun(job=self)
+
+    def run_if_flux(self):
+        """
+        The run if flux function is called by run to execute the simulation using flux, this allows distributing
+        calculation to separate nodes in a flux based HPC cluster.
+        """
+        return run_job_with_runmode_flux(job=self)
 
     def run_if_manually(self, _manually_print=True):
         """
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -701,7 +701,7 @@ class GenericJob(JobCore):
                 if repair and self.job_id and not self.status.finished:
                     self._run_if_repair()
                 elif status == "initialized":
-                    self._run_if_new(debug=debug)
+                    return self._run_if_new(debug=debug)
                 elif status == "created":
                     return self._run_if_created()
                 elif status == "submitted":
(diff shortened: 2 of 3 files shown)


liamhuber:
> Example:
> 
> ```
> >>> import flux.job
> >>> from pyiron_atomistics import Project
> >>> pr = Project('test')
> >>> job = pr.create.job.Lammps("lmp")
> >>> job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
> >>> job.flux_executor = flux.job.FluxExecutor()
> >>> job.server.cores = 2
> >>> fs = job.run()
> >>> print(fs.done())
> >>> print(fs.result())
> >>> print(fs.done())
> False
> 0
> True
> ```
> 
> The `fs` object here is a python `concurrent.futures` object, so you can use `fs.done()` to check if the calculation is finished or call `fs.result()` to wait until the calculation is finished.

I don't see this information anywhere in the codebase. How are users supposed to know this is how to do things?

jan-janssen:
> I don't see this information anywhere in the codebase. How are users supposed to know this is how to do things?

I agree and this is something I plan to add. We currently have some internal documentation for the Exascale Workshop and the users at Los Alamos National Lab, once the feature is more stabilized and maybe extended to other run_modes like non_modal, I am going to add the corresponding examples.

liamhuber:
> > I don't see this information anywhere in the codebase. How are users supposed to know this is how to do things?
> 
> I agree and this is something I plan to add. We currently have some internal documentation for the Exascale Workshop and the users at Los Alamos National Lab, once the feature is more stabilized and maybe extended to other run_modes like non_modal, I am going to add the corresponding examples.

I will be blunt, I find this entirely backwards. If you have an intended use and a user-facing interface in mind, start from there and make the code match it. Documenting existing stuff is a pain in the butt, but documenting new stuff is the easiest thing in the world and gives serious gains to how you code the new stuff. Getting it working and having secret documentation for key users opens up dangerous temptation of just never going back and documenting what becomes "old".

Further, if it isn't stabilized why isn't it in `contrib`? I can see how that might require explicit invocation of `run_job_with_runmode_flux` instead of just `job.run()`, but if this is flat-out unstable, undocumented, and untested, should it really be on `pyiron_base:HEAD`?

This just all feels extremely at odds with our discussions about making pyiron stable and maintainable.

jan-janssen:
> This just all feels extremely at odds with our discussions about making pyiron stable and maintainable.

Both tests and documentation for this new functionality are planned. As explained in https://github.com/pyiron/pyiron_base/pull/1126 the workshop on Exascale computing ends next week and I want to demonstrate that pyiron is "ready" for the Exascale. Obviously "ready" is a bit of an exaggeration as you noticed many features were added in a rush only one week before the workshop. Still obviously for the success of the workshop the demonstration of a fresh installation from conda on a new cluster and the ability to sub divide existing allocations with GPUs is quite impressive, so that is what I am aiming for. So forgive me for pushing this forward into a bit of an unstable regime at the moment. I guess I have a sufficient track record of stabilizing the pyiron functionality that this is a reasonable trade off.

pmrv:
I was on vacation so I couldn't directly comment before it was merged.  I share @liamhuber concerns and have a few on my own.

Interface-wise I'm not so happy with `run()` returning a future, we don't do this for other background run modes and there waiting on the job can be generally done with `pr.wait_for_job`.  The problem is not with using futures as such, but that there's now two different scenarios of what can happen and what is returned from the function.  Imo a valid approach would also be to always return our own futures, though naturally a lot more work.

A bigger worry is the `flux_executor` living directly on the job class.  Abstractly I think, this couples too strongly to flux and it would be better kept in the server object or even behind `pysqa` (which I thought was what we had originally discussed in a meeting).  More concretely, I think this opens a lot of potential problems with children/master or derived jobs.   Is this property always copied or otherwise transferred between reference and master jobs?  Are the semantics the same as our normal queueing mode (e.g. it seems flux will only ever ask for one node viz. L495)?  Is it honored and correctly used in all subclasses?  These may or may not be issues, but from the architecture and the lack of tests we cannot know.

My point here is not that PRs have to be perfect or all encompassing, but that for more than a year now we all agree that maintainability is the primary problem and concern that we have with pyiron.  Over this time we discussed multiple times and exhaustively that working more slowly and thinking harder about separating and cleanly interfacing the different aspects of pyiron would be desirable to make it easier to maintain pyiron.  Yet the next workshop or conference is always on the horizon and all the nice resolutions are thrown to the wind.

More generally, I have to ask myself why we are spending months over in contrib to prototype and think about how functionality like this could work in generic and extensible way, why we keep even simple and orthogonal new functionality changes (e.g. [#1008](https://github.com/pyiron/pyiron_atomistics/pull/1008),  but you somehow don't hold yourself to the same standard.  A workshop or a presentation is not a valid reason.  Am I to understand that a three month workshop would have failed because you couldn't have presented a <100 line patch and ten line example in the last week? 

> > This just all feels extremely at odds with our discussions about making pyiron stable and maintainable.
> 
> I guess I have a sufficient track record of stabilizing the pyiron functionality that this is a reasonable trade off.

I guess stability lies in the eye of the beholder.

PR #1125, opened 2023-06-07
Drop future support

jan-janssen:


.ci_support/environment-docs.yml
```diff
@@ -3,7 +3,6 @@ channels:
 dependencies:
 - nbsphinx
 - dill
-- future
```
liamhuber, line 6: This is fine and afaik unrelated to the imports down the line. Since we dropped python 2(?) `__future__` has been a built-in module.

pyiron_base/storage/has_stored_traits.py
```diff
@@ -6,17 +6,14 @@
 A parent class for managing input to pyiron jobs.
 """
 
-from __future__ import annotations
```
liamhuber: From what I can see in this module, there's no problem removing this here. Be careful though, as there are places this maneuver is [doing real work](https://stackoverflow.com/questions/61544854/from-future-import-annotations).

pyiron_base/storage/has_stored_traits.py
```diff
@@ -6,17 +6,14 @@
...
 from abc import ABC, ABCMeta
-from typing import Optional, TYPE_CHECKING
+from typing import Optional
 
 from traitlets import HasTraits, MetaHasTraits
 
 from pyiron_base.interfaces.object import HasStorage
 
-if TYPE_CHECKING:
-    from pyiron_base.storage.hdfio import ProjectHDFio
```
liamhuber: So the `TYPE_CHECKING` is not strictly necessary filter is not strictly necessary here -- there are no circular imports -- but I like to use it anyhow to signal intent: this import is used _only_ for type hinting in this module, it's not actually called. (There is potentially some speedup too, since `TYPE_CHECKING` is just `False` and this way we avoid importing all the stuff inside the clause at run-time -- but that's not my motivation.) 

My opinion here is not strong but is completely unambiguous; i.e. it's not a hill I want to die on, but I'd rather keep this.

liamhuber (commented):
Dependency update looks great. The type checking stuff I have mixed feelings about.

pyiron_base/storage/has_stored_traits.py
jan-janssen: If I keep the `TYPE_CHECKING` the tests fail - see first commit.

liamhuber: I guess `annotations` is doing some work here after all. Then my preference would be to keep both `from __future__ import annotations` and `if TYPE_CHECKING`, which should be the previous-and-passing state, yes?

liamhuber: `from __future__ import annotations` is [scheduled to become mandatory (automatic) behaviour eventually](https://docs.python.org/3/library/__future__.html#id1) anyhow

jan-janssen:
@liamhuber I reset the pull request to only remove future from the dependencies.

liamhuber (approved):

PR #1126, opened 2023-06-08
Add GPU support

jan-janssen:


jan-janssen:
Waiting for https://github.com/conda-forge/status/issues/144

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -490,6 +508,8 @@ def from_hdf(self, hdf, group_name=None):
             self._threads = hdf_dict["threads"]
         if "additional_arguments" in hdf_dict.keys():
             self.additional_arguments = hdf_dict["additional_arguments"]
+        if "gpus" in hdf_dict.keys():
+            self._gpus = hdf_dict["accept_crash"]
```
pmrv, line 512: Is there a reason why the names don't match and clash with an already existing entry?

liamhuber: Looking at the PRs merged over the last two days, I'm also quite concerned that we now have both Flux and GPU "support", but the tests and notebooks directories have not been touched at all.

jan-janssen: Good catch - I fixed it in https://github.com/pyiron/pyiron_base/pull/1127 

@liamhuber the challenging part for both is the integration in the test environment. But I agree I am going to work on adding tests once the presentation on Monday was successful.

liamhuber: For flux it seems super straightforward: add `flux` to the notebook dependencies, take your example from #1120 and change it so it uses a python template job instead of atomistics, and add it to an example notebook. That would give us both a bare-minimum of testing _and_ at least somewhere in the codebase that the intended use is shown.

For GPUs I agree, we cannot directly test the execution on GitHub CI, but we could at least have `AssertRaises` tests to make sure that setting the `gpus` flag actually gets us to the right part of the code.

liamhuber: > But I agree I am going to work on adding tests once the presentation on Monday was successful.

@jan-janssen I am super not comfortable with merges to base being done under "I have a presentation on Monday" pressure. Is it not possible to leave these in a branch and do the example from there, or is your example really getting the audience to `conda install pyiron` such that you need these changes publicly available?

jan-janssen: >  is your example really getting the audience to `conda install pyiron` such that you need these changes publicly available?

Yes, that is the goal - tell people during the three month we developed an Exascale ready version of pyiron and they can now install it directly from conda, with flux and all required dependencies.

jan-janssen: > For flux it seems super straightforward: add `flux` to the notebook dependencies, take your example from #1120 and change it so it uses a python template job instead of atomistics, and add it to an example notebook. That would give us both a bare-minimum of testing _and_ at least somewhere in the codebase that the intended use is shown.

Tests and documentation are in progress - I add them soon.

liamhuber: > Yes, that is the goal - tell people during the three month we developed an Exascale ready version of pyiron and they can now install it directly from conda, with flux and all required dependencies.

Well, for me this only strengthens the case that the tests should be present at the time the functionality is merged, but in any case I wish you a smooth presentation.

PR #1131, opened 2023-06-13
Check pyiron installation

jan-janssen:
Current checks implemented:
- executable bit is set for shell scripts
- conda package is installed

These can be executed using:
```
import pyiron_base
pyiron_base.check_executables_status()
```
Which returns a pandas dataframe with all executables pyiron found.

### DIFF
--- a/pyiron_base/state/check.py
+++ b/pyiron_base/state/check.py
@@ -5,6 +5,7 @@ import pandas
 
 try:
     from conda.cli import python_api
+
     conda_imported_successful = True
 except ImportError:
     conda_imported_successful = False
@@ -18,7 +19,7 @@ conda_package_dict = {
     "runner": "runner",
     "randspg": "randspg",
     "sphinx": "sphinxdft",
-    "lammps": "lammps"
+    "lammps": "lammps",
 }
 
 
@@ -33,7 +34,11 @@ def check_for_conda_package(name):
 def check_executable_bit(resource_paths):
     def check_bit(script_path):
         filemode = os.stat(script_path).st_mode
-        return bool(filemode & stat.S_IXUSR or filemode & stat.S_IXGRP or filemode & stat.S_IXOTH)
+        return bool(
+            filemode & stat.S_IXUSR
+            or filemode & stat.S_IXGRP
+            or filemode & stat.S_IXOTH
+        )
 
     path_lst = []
     for res_path in resource_paths:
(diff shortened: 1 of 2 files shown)


pmrv:
Would be nice to have this also available via `python -m pyiron_base.cli`, but lgtm otherwise.

pyiron_base/state/check.py
```diff
@@ -0,0 +1,80 @@
...
+        filemode = os.stat(script_path).st_mode
+        return bool(
+            filemode & stat.S_IXUSR
+            or filemode & stat.S_IXGRP
+            or filemode & stat.S_IXOTH
+        )
+
+    path_lst = []
+    for res_path in resource_paths:
+        path_lst += glob.glob(res_path + "/*/*/*.sh")
```
liamhuber, line 47: This feels very fragile. It looks to me like this requires all executables to be at exactly the third level of depth. This may be the case right now, but so we specify or document this anywhere? Do we anticipate other sh files appearing here, or is it safe to run an endswith command recursively inside the resource path instead?

liamhuber (commented):
This still needs a test. We can explicitly add some dummy scripts to the static directory. Whenever there is path based stuff, it's much safe to have it appear in the CI so we don't accidentally specify something in a windows non-compliant way.

pyiron_base/state/check.py
pmrv: Right now it needs to be exactly at the third level because the expected structure is `/$resources/$module_name/bin/$exe.sh`.  The documentation for this is [here](https://pyiron.readthedocs.io/en/latest/source/installation.html#use-your-own-executable-for-lammps-s-phi-nx-or-gpaw), though it could be more explicit.

I have a little class that codifies this in [tinybase](https://github.com/pyiron/pyiron_contrib/blob/tinybase/pyiron_contrib/tinybase/shell.py#L24) which could be moved to base.  Replacing this everywhere might be a bit more effort, so imo it's ok to do it in a separate PR.

liamhuber: I'm ok with codefying it as long as we're sufficiently explicit. One option I see is to search for scripts at exactly three levels deep and also searching for them anywhere in the resource tree, then giving the user a warning if the lists aren't the same.

Actually, I think I really like that, and it should be straightforward to make a test setup in static to make sure both aspects work.

Similarly we should also be checking if non-supported codes have directories. I'm on mobile so I can't easily bail on this comment to check if that's already done, but this would be another nice thing to give a warning about.

PR #1132, opened 2023-06-14
Fix maintenance and add tests

niklassiemer:


tests/project/test_maintenance.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _assert_hdf_rewrite(self):
+        job_hdf = self.project['toy_1'].project_hdf5
+        array = self.project['toy_1/user/some']
+        self.assertEqual(array, _test_array())
+        self.assertLess(job_hdf.file_size(job_hdf), self.initial_toy_1_hdf_file_size)
+
+    def test_repository_status(self):
+        df = self.project.maintenance.get_repository_status()
+        self.assertIn('pyiron_base', df, 'Environment dependent, but pyiron_base should be in there!')
```
niklassiemer:
```suggestion
        self.assertIn('pyiron_base', df.Module.values, 'Environment dependent, but pyiron_base should be in there!')
```

tests/project/test_maintenance.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def assertArrayEqual(self, first, second, msg=None):
+        self.assertIsInstance(first, np.ndarray, 'First argument is not a numpy array')
+        self.assertIsInstance(second, np.ndarray, 'Second argument is not a numpy array')
+
+        if not np.array_equal(first, second):
+            if msg is None:
+                self.fail(f'Arrays differ in shape or value {first} and {second}.')
+            else:
+                self.fail(msg)
```
pmrv: There's also [this nice utility](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html) from numpy itself.  I think we also came up with something like this a few times now,  I'll have a look at the other code quickly.

PR #1133, opened 2023-06-15
Add numpy equality check for unit tests

pmrv:
With this one can do `self.assertEqual()` on arrays directly.  Downside side that really both values have to be arrays, so I had to add some conversions. 

Because this relies on `setUp` I had to add a bunch of `super` calls.

### DIFF
--- a/tests/project/test_data.py
+++ b/tests/project/test_data.py
@@ -13,6 +13,7 @@ from pyiron_base._tests import PyironTestCase
 class TestProjectData(PyironTestCase):
     @classmethod
     def setUpClass(cls):
+        super().setUpClass()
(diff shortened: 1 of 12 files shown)


tests/project/test_data.py
```diff
@@ -13,6 +13,7 @@
 class TestProjectData(PyironTestCase):
     @classmethod
     def setUpClass(cls):
+        super().setUpClass()
```
niklassiemer, line 16: I am happy to include this here, however, this would is not required for the functionality, is it?

niklassiemer (approved):
LGTM!

tests/project/test_data.py
pmrv: Yeah, that was just because I was at it already anyway.

PR #1138, opened 2023-06-16
Check if project path exists before creating Project

### LINKED ISSUE #1137: table job of packed Project fails to load data
I just stubbled across an unintended behavior for table jobs which have been `pack`ed and are now tried to be loaded in completely different environment (e.g. binder) without access to the original analysis project (not even the path can be build..., which causes the error in my case). I suggest to check if the project paths exists before trying to make a `Project` there...

https://github.com/pyiron/pyiron_base/blob/0b6399ddad6ab4a734282b08ac8e024bebd973f4/pyiron_base/jobs/datamining.py#L684-L693

niklassiemer:
closes #1137

### DIFF
--- a/pyiron_base/jobs/datamining.py
+++ b/pyiron_base/jobs/datamining.py
@@ -683,7 +683,7 @@ class TableJob(GenericJob):
         with self.project_hdf5.open("input") as hdf5_input:
             if "project" in hdf5_input.list_nodes():
                 project_dict = hdf5_input["project"]
-                if os.path.exists(project_dict['path']):
+                if os.path.exists(project_dict["path"]):
                     project = self.project.__class__(
                         path=project_dict["path"],
                         user=project_dict["user"],
--- a/pyiron_base/jobs/datamining.py
+++ b/pyiron_base/jobs/datamining.py
@@ -683,14 +683,15 @@ class TableJob(GenericJob):
         with self.project_hdf5.open("input") as hdf5_input:
             if "project" in hdf5_input.list_nodes():
                 project_dict = hdf5_input["project"]
-                project = self.project.__class__(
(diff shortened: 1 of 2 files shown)


pyiron_base/jobs/datamining.py
```diff
@@ -683,14 +683,15 @@ def from_hdf(self, hdf=None, group_name=None):
...
-                self.analysis_project = project
+                if os.path.exists(project_dict["path"]):
+                    project = self.project.__class__(
+                        path=project_dict["path"],
+                        user=project_dict["user"],
+                        sql_query=project_dict["sql_query"],
+                    )
+                    project._filter = project_dict["filter"]
+                    project._inspect_mode = project_dict["inspect_mode"]
+                    self.analysis_project = project
```
pmrv, line 694: What about adding a warning here, so that users can know they've instantiated a "dangling" table?

tests/table/test_datamining.py
```diff
@@ -30,6 +32,17 @@ def setUp(self):
...
+        self.assertIs(self.project, self.table.analysis_project)
+        self.assertEqual(self.project.path, self.project.load(self.table.name).analysis_project.path)
+
+    def test_analysis_project_no_valid_path(self):
+        self.table.project_hdf5['input/project']['path'] = '/some/wrong/path'
+        self.table.to_hdf()
+
+        table_loaded: pyiron_base.TableJob = self.project.load(self.table.name)
+        self.assertIs(table_loaded.analysis_project, None)
+
```
niklassiemer:
```suggestion
```
I do not have time right now to correctly get an unfitting pyiron table right now to test the warning... Thus, I take it out for now.

PR #1142, opened 2023-06-19
Re-introduce flux

liamhuber:
Contains commit 5b92123 except for a bit of docstring stuff that never got removed.

Tests, documentation, and possibly (probably?) a re-design of the interface can be merged into here until the feature is ready for `main:HEAD`.


liamhuber:
Ok, there we go. I merged main when I should have re-based to it. The correct changes are all showing up in the diff now.

Issue #1143, opened 2023-06-21
Numpy update fails tests...what was intended?


liamhuber:
There's a number of failures in #1140, but this one caught my eye:

```
======================================================================
ERROR: test_array_type_conversion (generic.test_fileHDFio.TestFileHDFio) [object_array_with_lists]
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/pyiron_base/pyiron_base/tests/generic/test_fileHDFio.py", line 185, in test_array_type_conversion
TestWithProject: Setting up test project
TestWithProject: Setting up test project
    np.array_equal(array, object_array_with_lists)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/numpy/core/numeric.py", line 2439, in array_equal
    return bool(asarray(a1 == a2).all())
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

The relevant portion of the tests looks like this:

```python
    def test_array_type_conversion(self):
        object_array_with_lists = np.array(
            [[[1, 2, 3], [2, 3, 4]], [[4, 5, 6]]], dtype=object
        )
        int_array_as_objects_array = np.array([[1, 2, 3], [3, 4, 5]], dtype=object)
        float_array_as_objects_array = np.array(
            [[1.1, 1.3, 1.5], [2, 2.1, 2.2]], dtype=object
        )

        hdf = self.i_o_hdf5.open("arrays")

        hdf["object_array_with_lists"] = object_array_with_lists
        hdf["int_array_as_objects_array"] = int_array_as_objects_array
        hdf["float_array_as_objects_array"] = float_array_as_objects_array

        warn_msg_start = "WARNING:pyiron_log:Deprecated data structure! Returned array was converted "

        with self.subTest("object_array_with_lists"):
            array = hdf["object_array_with_lists"]
            np.array_equal(array, object_array_with_lists)
            self.assertIsInstance(array, np.ndarray)
            self.assertEqual(
                array.dtype,
                np.dtype(object),
                msg="dtype=object array falsely converted.",
            )
```

[...]

PR #1147, opened 2023-06-26
Turn np.array_equal into proper asserts

pmrv:
Based on our discussion [here](https://github.com/pyiron/pyiron_base/issues/1143) I turned the bare `array_equals` into asserts.

I'm working under the pragmatic assumption that we are fine with the implicit conversion of 

```python
np.array([[1,2,3],[4,5,6]], dtype=object)
```

to 

```python
np.array([np.array([1,2,3]),np.array([4,5,6])], dtype=object)
```

as currently done by `ProjectHDFio`.


tests/generic/test_fileHDFio.py
```diff
@@ -182,7 +182,9 @@ def test_array_type_conversion(self):
 
         with self.subTest("object_array_with_lists"):
             array = hdf["object_array_with_lists"]
-            np.array_equal(array, object_array_with_lists)
+            # HDF layer implicitely converts sub lists to array, so enforce
+            # that here as well
+            self.assertEqual(array, np.array([np.asarray(a) for a in object_array_with_lists]))
```
pmrv: Remove the conversion here and fix `ProjectHDFio` accordingly.

pmrv:
While I have removed the array conversion *reading from HDF*, I've kept the conversion *writting to HDF*, because a numpy array with lists inside cannot actually be written by h5io.

PR #1148, opened 2023-06-26
Drop pathlib2

pmrv:
If I see this correctly `pathlib2` is purely a backport of `pathlib` to python2.  We were using it only in one place and all usage there seems to be supported by standard lib `pathlib` since python3.4.  Hence I kick it out here and see if anything breaks.

### DIFF
--- a/.ci_support/environment-docs.yml
+++ b/.ci_support/environment-docs.yml
@@ -8,7 +8,6 @@ dependencies:
 - h5py
 - numpy
 - pandas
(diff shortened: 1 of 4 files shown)


liamhuber (approved):
Yep, looks that way to me too.

jan-janssen:
I guess you should also remove it in https://github.com/pyiron/pyiron_base/blob/main/.ci_support/environment-docs.yml

PR #1154, opened 2023-07-03
Reimplement the flux interface

jan-janssen:
Example code:
```
>>> import flux.job
>>> from pyiron_atomistics import Project
>>> pr = Project('test')
>>> job = pr.create.job.Lammps("lmp")
>>> job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
>>> job.server.executor = flux.job.FluxExecutor()
>>> job.server.cores = 2
>>> fs = job.run()
>>> print(fs.done())
>>> print(fs.result())
>>> print(fs.done())
False
0
True
```

This pull request is based on: 
https://github.com/pyiron/pyiron_base/pull/1151
https://github.com/pyiron/pyiron_base/pull/1152
https://github.com/pyiron/pyiron_base/pull/1153

### DIFF
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -8,6 +8,8 @@ import os
 import posixpath
 import subprocess
 
+from jinja2 import Template
+
 from pyiron_base.utils.deprecate import deprecate
 from pyiron_base.jobs.job.wrapper import JobWrapper
 from pyiron_base.state import state
@@ -495,7 +497,7 @@ def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
         return run_job_with_runmode_executor_futures(job=job, executor=executor)
     else:
         raise NotImplementedError(
-            "Currently only concurrent.futures.ProcessPoolExecutor is supported."
+            "Currently only flux.job.FluxExecutor and concurrent.futures.ProcessPoolExecutor are supported."
         )
 
 
@@ -551,7 +553,12 @@ def run_job_with_runmode_executor_flux(job, executor, gpus_per_slot=None):
 
     >>> from flux.job import FluxExecutor
     >>> job.server.executor = FluxExecutor()
-    >>> fs = job.run()
+    >>> job.run()
+    >>> job.server.future.done()
+    False
+    >>> job.server.future.result()
+    >>> job.server.future.done()
+    True
 
     A word of caution - flux is currently only available on Linux, for all other operation systems the ProcessPoolExecutor
     from the python standard library concurrent.futures is recommended. The advantage of flux over the ProcessPoolExecutor
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -16,6 +16,13 @@ from pyiron_base.state import state
 from pyiron_base.utils.instance import static_isinstance
 
 
+try:
+    import flux.job
+
+    flux_available = True
+except ImportError:
+    flux_available = False
+
 """
 The function job.run() inside pyiron is executed differently depending on the status of the job object. This module 
 introduces the most general run functions and how they are selected. 
@@ -475,7 +482,11 @@ def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
         raise NotImplementedError(
             "Currently job.server.run_mode.executor does not support GenericMaster jobs."
         )
-    if isinstance(executor, ProcessPoolExecutor):
+    if flux_available and isinstance(executor, flux.job.FluxExecutor):
+        return run_job_with_runmode_executor_flux(
+            job=job, executor=executor, gpus_per_slot=gpus_per_slot
+        )
+    elif isinstance(executor, ProcessPoolExecutor):
         return run_job_with_runmode_executor_futures(job=job, executor=executor)
     else:
         raise NotImplementedError(
@@ -529,6 +540,75 @@ def run_job_with_runmode_executor_futures(job, executor):
             debug=False,
             connection_string=None,
         )
+
+
+def run_job_with_runmode_executor_flux(job, executor, gpus_per_slot=None):
+    """
+    Interface for the flux.job.FluxExecutor executor. Flux is a hierarchical resource management. It can either be used to
+    replace queuing systems like SLURM or be used as a user specific queuing system within an existing allocation.
+    pyiron provides two interfaces to flux, this executor interface as well as a traditional queuing system interface
+    via pysqa. This executor interface is designed for the development of asynchronous simulation protocols, while the
+    traditional queuing system interface simplifies the transition from other queuing systems like SLURM. The usuage
+    is analog to the concurrent.futures.Executor interface:
+
+    >>> from flux.job import FluxExecutor
+    >>> job.server.executor = FluxExecutor()
+    >>> fs = job.run()
+
+    A word of caution - flux is currently only available on Linux, for all other operation systems the ProcessPoolExecutor
+    from the python standard library concurrent.futures is recommended. The advantage of flux over the ProcessPoolExecutor
+    is that flux takes over the resource management, like monitoring how many cores are available while with the
+    ProcessPoolExecutor this is left to the user.
+
+    Args:
+        job (GenericJob): pyiron job object
+        executor (flux.job.FluxExecutor): flux executor class which implements the executor interface defined in the
+                                      python concurrent.futures.Executor class.
+        gpus_per_slot (int): number of GPUs per MPI rank, typically 1
+
+     Returns:
+         concurrent.futures.Future: future object to develop asynchronous simulation protocols
+    """
+    if not flux_available:
+        raise ModuleNotFoundError(
+            "No module named 'flux'. Running in flux mode is only available on Linux;"
+            "For CPU jobs, please use `conda install -c conda-forge flux-core`; for "
+            "GPU support you will additionally need "
+            "`conda install -c conda-forge flux-sched libhwloc=*=cuda*`"
+        )
+    if not state.database.database_is_disabled:
+        executable_template = Template(
+            "#!/bin/bash\n",
+            "python -m pyiron_base.cli wrapper -p {{working_directory}} -j {{job_id}}",
+        )
+        exeuctable_str = executable_template.render(
+            working_directory=job.working_directory,
+            job_id=str(job.job_id),
+        )
+        job_name = "pi_" + str(job.job_id)
+    else:
+        executable_template = Template(
+            "#!/bin/bash\n",
+            "python -m pyiron_base.cli wrapper -p {{working_directory}} -f {{file_name}}{{h5_path}}",
+        )
+        exeuctable_str = executable_template.render(
+            working_directory=job.working_directory,
+            file_name=job.project_hdf5.file_name,
+            h5_path=job.project_hdf5.h5_path,
+        )
+        job_name = "pi_" + job.job_name
+
+    jobspec = flux.job.JobspecV1.from_batch_command(
+        jobname=job_name,
+        script=exeuctable_str,
+        num_nodes=1,
+        cores_per_slot=1,
+        gpus_per_slot=gpus_per_slot,
+        num_slots=job.server.cores,
+    )
+    jobspec.cwd = job.project_hdf5.working_directory
+    jobspec.environment = dict(os.environ)
+    return executor.submit(jobspec)
 
 
 def run_time_decorator(func):
--- a/.ci_support/environment-docs.yml
+++ b/.ci_support/environment-docs.yml
@@ -6,6 +6,7 @@ dependencies:
 - gitpython
 - h5io
 - h5py
+- jinja2
 - numpy
 - pandas
 - pint
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -8,6 +8,7 @@ dependencies:
 - gitpython =3.1.31
 - h5io =0.1.7
 - h5py =3.9.0
+- jinja2 =3.1.2
 - numpy =1.24.3
 - pandas =2.0.3
 - pint =0.22
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ setup(
         'gitpython==3.1.31',
         'h5io==0.1.7',
         'h5py==3.9.0',
+        'jinja2==3.1.2',
         'numpy==1.24.3',
         'pandas==2.0.3',
         'pint==0.22',


jan-janssen:
The demo is also available at https://github.com/pyiron-dev/pyiron-flux-demo

jan-janssen:
Missing parts: 
* Disable `run_mode` flux on `GenericMasters`.
* Test `copy()` function.

samwaseda (changes requested):
DocStrings

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +451,53 @@ def run_job_with_runmode_srun(job):
     )
 
 
+def run_job_with_runmode_flux(job, executor, gpus_per_slot=None):
```
samwaseda: Docstring either here or in `runmode.py` because currently the idea is unclear.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +451,53 @@ def run_job_with_runmode_srun(job):
...
+            "GPU support you will additionally need "
+            "`conda install -c conda-forge flux-sched libhwloc=*=cuda*`"
+        )
+    if not state.database.database_is_disabled:
+        executable_template = Template(
+            """\
+#!/bin/bash
+python -m pyiron_base.cli wrapper -p {{working_directory}} -j {{job_id}}
+"""
+        )
```
```suggestion
        executable_template = Template(
            "#!/bin/bash\n"
            "python -m pyiron_base.cli wrapper -p {{working_directory}} -j {{job_id}}\n"
        )
```

Just a suggestion, but it makes it a lot easier for folding. Same for the lines below.

jan-janssen: Done

jan-janssen:
Python `concurrent.futures.Executor` example: 
```

pr = Project('test')

job = pr.create.job.Lammps("lmp")
job.server.executor = ProcessPoolExecutor()
job.server.cores = 2
fs = job.run()

print(fs.done())
print(fs.result())
print(fs.done())
```
With the `ProcessPoolExecutor` it is possible to develop modern workflows locally before submitting them to an HPC with flux.

jan-janssen:
> Missing parts:
> 
> * Disable `run_mode` flux on `GenericMasters`.
> * Test `copy()` function.

Done

jan-janssen:
> DocStrings

I added Docstrings, can you take another look?

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +453,150 @@ def run_job_with_runmode_srun(job):
     )
 
 
+def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
+    """
+    Introduced in Python 3.2 the concurrent.futures interface enables the development of asynchronous python programs.
```
```suggestion
    Introduced in Python 3.2 the concurrent.futures interface enables the asynchronous execution of python programs.
```

I presume, no?

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +453,150 @@ def run_job_with_runmode_srun(job):
...
+            "Currently job.server.run_mode.executor does not support GenericMaster jobs."
+        )
+    if flux_available and isinstance(executor, flux.job.FluxExecutor):
+        return run_job_with_runmode_executor_flux(
+            job=job, executor=executor, gpus_per_slot=gpus_per_slot
+        )
+    elif isinstance(executor, ProcessPoolExecutor):
+        return run_job_with_runmode_executor_futures(job=job, executor=executor)
+    else:
+        raise TypeError(
```
```suggestion
        raise NotImplementedError(
```

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +453,150 @@ def run_job_with_runmode_srun(job):
...
+        gpus_per_slot (int): number of GPUs per MPI rank, typically 1
+
+     Returns:
+         concurrent.futures.Future: future object to develop asynchronous simulation protocols
+    """
+
+    if static_isinstance(
+        obj=job, obj_type="pyiron_base.jobs.master.generic.GenericMaster"
+    ):
+        raise TypeError(
```
```suggestion
        raise NotImplementedError(
```

samwaseda (commented):
I really like the introductory text in the function. Is there a possibility to extend the example a bit more so that the code becomes full? Especially it's not very clear right now where `Executor()` comes from. I would really start from `import flux.job` and put `job.server.executor = flux.job.FluxExecutor()`.

Apart from Doc, is there actually a practical example which shows the capability of `flux`? I'm asking this because I'm not 100% convinced that we should introduce a new way of accessing job data - so far in pyiron we always worked with only `job`, and I'm wondering whether returning another object in `job.run()` is really necessary.

samwaseda:
How can I actually unblock merging?

jan-janssen:
> How can I actually unblock merging?

I guess you can just create a new review and approve the merge request, that should unblock the merging.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +453,150 @@ def run_job_with_runmode_srun(job):
...
 
 
+def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
+    """
+    Introduced in Python 3.2 the concurrent.futures interface enables the asynchronous execution of python programs.
+    A function is submitted to the executor and a future object is returned. The future object is updated in the
+    background once the executor finished executing the function. The job.server.run_mode.executor implements the same
+    functionality for pyiron jobs. An executor is set as an attribute to the server object:
+
+    >>> job.server.executor = Executor()
```
jan-janssen:
```suggestion
    >>> job.server.executor = concurrent.futures.Executor()
```

jan-janssen: @samwaseda Does this clarify where the executor comes from?

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +453,150 @@ def run_job_with_runmode_srun(job):
...
+        )
+
+
+def run_job_with_runmode_executor_flux(job, executor, gpus_per_slot=None):
+    """
+    Interface for the flux.job.Executor executor. Flux is a hierarchical resource management. It can either be used to
+    replace queuing systems like SLURM or be used as a user specific queuing system within an existing allocation.
+    pyiron provides two interfaces to flux, this executor interface as well as a traditional queuing system interface
+    via pysqa. This executor interface is designed for the development of asynchronous simulation protocols, while the
+    traditional queuing system interface simplifies the transition from other queuing systems like SLURM.
```
jan-janssen:
```suggestion
    traditional queuing system interface simplifies the transition from other queuing systems like SLURM. The usuage
    is analog to the concurrent.futures.Executor interface: 
    
    >>> from flux.job import FluxExecutor 
    >>> job.server.executor = FluxExecutor()
    >>> fs = job.run()
```

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +453,150 @@ def run_job_with_runmode_srun(job):
...
+            job_id=None,
+            file_path=job.project_hdf5.file_name + job.project_hdf5.h5_path,
+            debug=False,
+            connection_string=None,
+        )
+
+
+def run_job_with_runmode_executor_flux(job, executor, gpus_per_slot=None):
+    """
+    Interface for the flux.job.Executor executor. Flux is a hierarchical resource management. It can either be used to
```
jan-janssen:
```suggestion
    Interface for the flux.job.FluxExecutor executor. Flux is a hierarchical resource management. It can either be used to
```

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +453,150 @@ def run_job_with_runmode_srun(job):
...
+    replace queuing systems like SLURM or be used as a user specific queuing system within an existing allocation.
+    pyiron provides two interfaces to flux, this executor interface as well as a traditional queuing system interface
+    via pysqa. This executor interface is designed for the development of asynchronous simulation protocols, while the
+    traditional queuing system interface simplifies the transition from other queuing systems like SLURM.
+
+    Args:
+        job (GenericJob): pyiron job object
+        executor (flux.job.Executor): flux executor class which implements the executor interface defined in the
```
jan-janssen:
```suggestion
        executor (flux.job.FluxExecutor): flux executor class which implements the executor interface defined in the
```

jan-janssen:
> I really like the introductory text in the function. Is there a possibility to extend the example a bit more so that the code becomes full? Especially it's not very clear right now where `Executor()` comes from. I would really start from `import flux.job` and put `job.server.executor = flux.job.FluxExecutor()`.

I added the import statement to clarify the usage.

> Apart from Doc, is there actually a practical example which shows the capability of `flux`? I'm asking this because I'm not 100% convinced that we should introduce a new way of accessing job data - so far in pyiron we always worked with only `job`, and I'm wondering whether returning another object in `job.run()` is really necessary.

To develop Exascale Workflows with pyiron - which is an essential part of my PostDoc here in Los Alamos - we need asynchronous workflows. Fitting a machine learning potential is one example we are currently working on. 

Basically, we start thousands of DFT calculation, one per GPU and once 100 DFT calculation are finished we start fitting the first potential. We then use the potential for active learning and generating new structures we want to include in the training set. During this initial fitting 90% of the DFT calculation are still running. Once more DFT calculation are completed we fit another potential and update the active learning to use the new potential. During the whole time the cyclic workflow of 1) DFT calculation, 2) Fitting and 3) Active Learning is executed asynchronously and in parallel. So at any point in time there are some DFT calculation running, a new potential is fitted to the already finished DFT calculation and a previous potential is used in an active learning procedure to identify new structures. 

The standard interface in python for developing such asynchronous procedures is the `concurrent.futures` module. So this pull request aims to make pyiron compatible to the standard interface. In the future I want to connect this to the workflow developments of you and @liamhuber but that is a next step. For the moment I want to leave it to the user to manage the concurrent future objects, as I am not yet sure what is the best strategy. But to enable users to test this interface we need the integration inside `pyiron_base` as the goal is to use the current interface in production runs in Los Alamos within the next month.

samwaseda:
> Example code:
> 
> ```python
> >>> import flux.job
> >>> from pyiron_atomistics import Project
> >>> pr = Project('test')
> >>> job = pr.create.job.Lammps("lmp")
> >>> job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
> >>> job.server.executor = flux.job.FluxExecutor()
> >>> job.server.cores = 2
> >>> fs = job.run()
> >>> print(fs.done())
> >>> print(fs.result())
> >>> print(fs.done())
> False
> 0
> True
> ```

Actually I cannot find `flux`. Is it on conda?

jan-janssen:
> Actually I cannot find `flux`. Is it on conda?

Yes, it is `flux-core` on conda-forge.

samwaseda:
I got this error message: `FileNotFoundError: [Errno 2] Unable to connect to Flux: broker socket /run/flux/local was not found`

jan-janssen:
> I got this error message: `FileNotFoundError: [Errno 2] Unable to connect to Flux: broker socket /run/flux/local was not found`

As illustrated in https://github.com/pyiron-dev/pyiron-flux-demo the first step is to call `flux start` this returns a shell and then in this shell the python process has to be executed. For a jupyter notebook you can use `flux start jupyter notebook`. In addition just like for any other queuing system, it is necessary to modify the `run_*.sh` scripts for the individual codes to use `flux run` just like they would use `srun` for SLURM.

samwaseda:
> To develop Exascale Workflows with pyiron - which is an essential part of my PostDoc here in Los Alamos - we need asynchronous workflows. Fitting a machine learning potential is one example we are currently working on.

I know *what* it should do, but the question is *how* it does it. So far I haven't been able to test it myself, but while in pyiron there were different types of run modes, there has never been anything more than just `run`. Now there's a returned object just for `flux` which follows a very different syntax after that. I just don't really understand: 1, how the user is supposed to know how it works; 2, why it should deviate so strongly from the unified pyiron syntax.

This being said, since I haven't been able to run it myself so far there might have been a fundamental misunderstanding from my side, so feel free to correct me if you think that's the case.

jan-janssen:
> Now there's a returned object just for `flux` which follows a very different syntax after that. I just don't really understand: 1, how the user is supposed to know how it works; 2, why it should deviate so strongly from the unified pyiron syntax.

Yes, asynchronous workflows are different from the synchronous workflows we developed before. Before each step in a workflow had a clear start and a clear end and only after one step ended a new step started. In asynchronous workflows the individual steps overlap. To address your concerns: 

1) The implementation tries to follow the python standard library as closely as possible. https://docs.python.org/3/library/concurrent.futures.html This does not say it is easy to develop asynchronous workflows, but if you already develop asynchronous workflows with python, then pyiron feels natural - at least this is the intention. The second example I posted above as well as the unit tests use the `concurrent.futures.ProcessPoolExecutor` which is part of the standard library and should be easy to test on any system. 
2) Yes, the syntax is different but only in a minimal way, as the run function returns a pointer which links to the execution of the current job. This pointer is only used to check if the calculation is finished. Maybe in future this is again going to be interested in the job.status. But at the moment to be able to use the standard python libraries for asynchronous programming we need direct access to the future object. 

An alternative suggestion would be to attach the future object to `job.server.future`, then the run function does not return any output, if that is more desirable.

samwaseda:
> As illustrated in https://github.com/pyiron-dev/pyiron-flux-demo the first step is to call `flux start` this returns a shell and then in this shell the python process has to be executed. For a jupyter notebook you can use `flux start jupyter notebook`. In addition just like for any other queuing system, it is necessary to modify the `run_*.sh` scripts for the individual codes to use `flux run` just like they would use `srun` for SLURM.

I have serious concern about the utility of `flux` based on this example, because I don't see straightforwardly how to relate other pyiron jobs to this example. I'm sure that this is the right way, but I just don't think it's ripe for `pyiron_base`. Would it be imaginable to move it to `pyiron_contrib`? If not, I think this demo has to be extended to include steps to make other programs available using `flux` in a *pyironic* way (i.e. it shouldn't be a string inside a notebook that's exported and executed).

Maybe @pmrv or @liamhuber has an opinion?

niklassiemer:
Indeed, the returned Future upon run is 'only' for status checking, isnt it? One concern is also that a 'standard' pyiron user forgets to assign the run() since we never did this... in that case it starts running but one could not get the Future object any more? Thus, I would like to store the future somewhere...

jan-janssen:
>  it shouldn't be a string inside a notebook that's exported and executed.

If you start your notebook using `flus start jupyter notebook` then you can use all the code in the string directly in the jupyter notebook. The issue here is just on the mybinder side, because in mybinder I cannot manipulate the way how the jupyter notebook is started.

jan-janssen:
> Indeed, the returned Future upon run is 'only' for status checking, isnt it? One concern is also that a 'standard' pyiron user forgets to assign the run() since we never did this... in that case it starts running but one could not get the Future object any more? Thus, I would like to store the future somewhere...

As suggested above: 
> An alternative suggestion would be to attach the future object to job.server.future, then the run function does not return any output, if that is more desirable.

jan-janssen:
> Would it be imaginable to move it to `pyiron_contrib`? 

I thought about this for a long time, but on the one hand this would result in a lot of duplicate code to have LAMMPS and VASP run with the new interface, as we would have to have modified job classes and so on. We had similar discussions about pyiron interface without files in https://github.com/pyiron/pyiron_contrib/tree/main/pyiron_contrib/nofiles which ultimately resulted in the release of https://github.com/pyiron/pyiron_lammps . While in this case having a standalone code really helped us to accelerate the development, in particular for large number of small LAMMPS simulations where pyiron struggles. The flux integration is primarily designed for the up-scaling of DFT calculation and orchestrating large numbers of DFT calculations in one allocation. So this directly addresses the primary application of `pyiron_atomistics`, so from my perspective the flux code has to be integrated in `pyiron_base`.

liamhuber:
@samwaseda:
>  @liamhuber has an opinion?

I'm certain I do -- I'm just not sure what it is yet 😉 I've been dealing with `.workflow` issues so far this morning, but will dig into the flux stuff presently.

@samwaseda 
> This being said, since I haven't been able to run it myself 

@jan-janssen 
> The issue here is just on the mybinder side, because in mybinder I cannot manipulate the way how the jupyter notebook is started.

This is potentially a big issue for me, and not a great sign that Sam so far faces the same issue. 

I don't currently have a linux OS installed on any of my machines (I have pyiron dev setups for windows and OSX though) and I'm relying on binder to be able to run this at all. Binder _does_ have `postBuild` tools, so the first thing I'm going to look into after getting a light lunch is whether it has `preBuild` tools so that we can get this running on binder.

I haven't had a chance to read through all the comments yet, but does this limitation also mean that it's impossible to do direct flux tests on the github CI? I see the 

> the first step is to call `flux start`

comment from Jan, so I guess I'm cautiously optimistic this is a non-issue??

I'll read up and have something more productive to say later today.

jan-janssen:
Once more for users who do not have access to Linux, we can use the `ProcessPoolExecutor` as illustrated above. It uses the same executor interface as flux, but the user has to keep track how many resources are used and in addition the `ProcessPoolExecutor` is limited to a single node while flux can manage thousands of compute nodes.

jan-janssen:
To simplify the discussion, I created a separate pull request for the `ProcessPoolExecutor` interface: 
https://github.com/pyiron/pyiron_base/pull/1155

This pull request is based on `ProcessPoolExecutor` interface and adds the flux interface.

samwaseda:
> If you start your notebook using `flus start jupyter notebook` then you can use all the code in the string directly in the jupyter notebook. The issue here is just on the mybinder side, because in mybinder I cannot manipulate the way how the jupyter notebook is started.

This sounds like "this is a demo that people shouldn't follow", which is contraction in itself... For me it doesn't have to be quite as interactive as binder, but we need something more enlightening that people can straightforwardly follow.

samwaseda:
> I thought about this for a long time, but on the one hand this would result in a lot of duplicate code to have LAMMPS and VASP run with the new interface, as we would have to have modified job classes and so on. We had similar discussions about pyiron interface without files in https://github.com/pyiron/pyiron_contrib/tree/main/pyiron_contrib/nofiles which ultimately resulted in the release of https://github.com/pyiron/pyiron_lammps . While in this case having a standalone code really helped us to accelerate the development, in particular for large number of small LAMMPS simulations where pyiron struggles.

If you are saying it should be included in `pyiron_base` because it's otherwise not efficient, I have to say it's not really a reason... Btw. I guess the last phrase requires an auxiliary phrase that might have been crucial.

> The flux integration is primarily designed for the up-scaling of DFT calculation and orchestrating large numbers of DFT calculations in one allocation. So this directly addresses the primary application of pyiron_atomistics, so from my perspective the flux code has to be integrated in pyiron_base.

Again, I understand what it was made for, and I agree that it's the future (in the literal sense), but the question is how it should be implemented.

> As suggested above:
> 
> > An alternative suggestion would be to attach the future object to job.server.future, then the run function does not return any output, if that is more desirable.

This statement shows for me that the discussion has not converged yet.

Just to make sure, this PR has two major problems for me:
- It breaks the generality of `run`
- It is not clear how to make other jobs `flux`-compatible within the pyiron framework

For both points it's difficult for me to understand why it's happening this way and estimate what would be the steps to resolve the problems, but I have the feeling that we need a discussion on the more fundamental level. At least I'm sure we cannot maintain the code reliably in the current format.

jan-janssen:
> * It breaks the generality of `run`

As suggested already above, I am fine with attaching the futures object to `job.server.future` if that resolves your concerns. 

> * It is not clear how to make other jobs `flux`-compatible within the pyiron framework

I guess this is a general misunderstanding. Each job is flux-compatible, as flux can be used just like all the other queuing systems via pysqa. https://github.com/pyiron/pyiron_base/pull/1155 implements a new asynchronous job interface, based on the `concurrent.futures.ProcessPoolExecutor` class. This pull request extends https://github.com/pyiron/pyiron_base/pull/1155 by adding an asynchronous flux interface. So flux is basically the first queuing system which supports the asynchronous job interface as well as the traditional interface based on `pysqa`.

jan-janssen:
I updated the flux mybinder example https://github.com/pyiron-dev/pyiron-flux-demo/tree/main - it now starts jupyter with `flux start jupyter notebook`. This is done by manipulating the `kernel.json` file.

liamhuber:
I'm not going to be able to get to reviewing this PR today. Overall I thought #1155 was pretty good, although I may want to play around with some different combinations (like interactively opening a job with a flux executor set) to see how it goes. The baseline behaviour is ok by me though

> > * It breaks the generality of `run`
> 
> As suggested already above, I am fine with attaching the futures object to `job.server.future` if that resolves your concerns.

This also comes up in #1155. I am ok with this solution; it may not be the final resting place of the futures object, but it will do for now and doesn't clutter up the main job namespace.

> > * It is not clear how to make other jobs `flux`-compatible within the pyiron framework
> 
> I guess this is a general misunderstanding. Each job is flux-compatible, as flux can be used just like all the other queuing systems via pysqa. #1155 implements a new asynchronous job interface, based on the `concurrent.futures.ProcessPoolExecutor` class. This pull request extends #1155 by adding an asynchronous flux interface. So flux is basically the first queuing system which supports the asynchronous job interface as well as the traditional interface based on `pysqa`.

I think @samwaseda means master jobs? In which case #1155 certainly does not provide a universal interface and if this PR is similar his concern is totally valid!

I do think that at this stage it's fair enough to fail cleanly rather than support everything. Although obviously I would be much _more_ comfortable with a "fail clean instead of working" attack in contrib than here. (I saw you're earlier comment on the topic of moving to contrib, and I'm not currently of the opinion that we should force the flux stuff over there, but that still doesn't mean I'm _happy_ about this rough-around-the-edges stuff being in base...)


~~@jan-janssen, take a look at my comments in #1155. I'll come review this one tomorrow, but if it's similar to that then I guess I will be able to get behind the gist of it. I do share Sam's other concern:~~

> ~~This sounds like "this is a demo that people shouldn't follow", which is contraction in itself... For me it doesn't have to be quite as interactive as binder, but we need something more enlightening that people can straightforwardly follow.~~

~~The binder on https://github.com/pyiron-dev/pyiron-flux-demo  full on fails, so what is even the point? But anyhow, while a somehow-functional demo is an important thing to have before merging this, this seems like a simple technical problem that I trust we'll find a way around --~~(EDIT: Super, binder issue got resolved while I was writing this. New version of the binder link is working fine for me!) more core is coming to consensus about the architecture of where to put the futures and how to handle master jobs.

samwaseda:
> I guess this is a general misunderstanding. Each job is flux-compatible, as flux can be used just like all the other queuing systems via pysqa. #1155 implements a new asynchronous job interface, based on the `concurrent.futures.ProcessPoolExecutor` class. This pull request extends #1155 by adding an asynchronous flux interface. So flux is basically the first queuing system which supports the asynchronous job interface as well as the traditional interface based on `pysqa`.

ok then I'm now a lot more comfortable, but with the demo before you updated it it was not very clear how to make that happen. I finish for today but I promise that I'll take a look as soon as I come back tomorrow morning.

liamhuber:
This is built on top of #1155 and keeps merging those changes into it, but targets main; let's make it a draft until #1155 is merged.

samwaseda:
One thing I don't understand is the coexistence of `fs.done()` and job status. Is there a reason that they cannot be merged except for historical reasons?

jan-janssen:
> One thing I don't understand is the coexistence of `fs.done()` and job status. Is there a reason that they cannot be merged except for historical reasons?

I agree, if I would start a new version of `pyiron_base` I would only use the `concurrent.futures` and remove the job status completely, but this change would be much more drastic, so for the moment we are going to have both concepts until we have move experience with the `concurrent.futures`.

samwaseda:
> 
> I agree, if I would start a new version of `pyiron_base` I would only use the `concurrent.futures` and remove the job status completely, but this change would be much more drastic, so for the moment we are going to have both concepts until we have move experience with the `concurrent.futures`.

but isn’t this then the reason why it should be in contrib and not base? We are breaking the concept with this change

jan-janssen:
> but isn’t this then the reason why it should be in contrib and not base? We are breaking the concept with this change

I agree that this implementation adds an alternative to the previous implementation and I am convinced that if I would rewrite `pyiron_base` from scratch I would only use the `concurrent.futures`. Still I also think we both agree that is not what we want to do right now, as we have so many other developments in the pipeline regarding the workflows/ graphs and so on. 

So the current pull requests introduce the functionality in a way that allows existing users who leverage pyiron for DFT calculation to already benefit from the new functionality, without interfering with existing users. As discussed in https://github.com/pyiron/pyiron_base/pull/1155 I want to add the integration with `wait_for_jobs` and `remove_jobs` so the only new command a user has to learn is setting `job.server.executor` while all the rest behaves exactly the same. A user can access the internal `job.server.future` interface but they do not have to.

niklassiemer:
> > but isn’t this then the reason why it should be in contrib and not base? We are breaking the concept with this change
> 
> I agree that this implementation adds an alternative to the previous implementation and I am convinced that if I would rewrite `pyiron_base` from scratch I would only use the `concurrent.futures`. Still I also think we both agree that is not what we want to do right now, as we have so many other developments in the pipeline regarding the workflows/ graphs and so on.
> 
> So the current pull requests introduce the functionality in a way that allows existing users who leverage pyiron for DFT calculation to already benefit from the new functionality, without interfering with existing users. As discussed in #1155 I want to add the integration with `wait_for_jobs` and `remove_jobs` so the only new command a user has to learn is setting `job.server.executor` while all the rest behaves exactly the same. A user can access the internal `job.server.future` interface but they do not have to.

I like the general idea and I am fine with having this in `pyiron_base` if we agree on the interface, if it does not affect current users, if it will be integrated in the pyiron syntax as good as possible and if it raises clear error massages if something is not supported. I deem we are on a good way to achieve these goals :) Maybe, the job status should be able to react on a present `Future` object and as this instead of the file or the database if there is a `Future`? This would streamline the user experience.

@jan-janssen (in future perspective) why does the user need to set the executor? Could this not be handled by pyiron upon using a flux based queue type? Right now, it is of course fine that users have to do this to enable the new run functionality!

samwaseda:
Now I thought about it again and think that this should also change `status`. Essentially at least for the time being it should work in the conventional way, meaning status=suspended should be there, and `refresh_blablabla` should update the statuses.

liamhuber:
@jan-janssen it's a bit hard to say until the diff gets updated, but it looks like once #1155 gets merged the changes here will be extremely minimal and good-to-go. 

It would be cool to have the infrastructure in our CI to have OS-dependent envs -- I can imagine an implementation of this and it should be possible -- then we could write linux-only tests for the flux stuff. But that's for the future and IMO now that #1155 has a bunch of tests we can get away without ~~extensive~~ any (for the foreseeable future at least) testing for flux here.

jan-janssen:
> @jan-janssen it's a bit hard to say until the diff gets updated, but it looks like once #1155 gets merged the changes here will be extremely minimal and good-to-go.

I just merged the changes into main and main into this pull request. It is basically just a single function that is added plus `jinja2` as additional dependency, but we already use `jinja2` in `pysqa` so it was already an implicit dependency before.

liamhuber (approved):
lgtm! 

I guess some of the stuff about how to boot your jupyter notebook such that it fluxes might be nice to document, but also that's basically just flux documentation and not pyiron-flux documentation so I'm also ok with leaving it out. Now that this is on top of #1155 it's nice clean changes, so in addition to the working binder demo I'm totally happy with it.

jan-janssen:
@samwaseda Do you want to take another look? I would be very happy to merge this before Monday.

niklassiemer (approved):
I think with change to store the `Future` in `job.server.future` the example code needs an edit. The code changes LGTM.

samwaseda (approved):
Yeah looks good to me now.

PR #1155, opened 2023-07-04
Implement concurrent.futures.ProcessPoolExecutor interface

jan-janssen:
Example code:
```

pr = Project('test')

job = pr.create.job.Lammps("lmp")
job.server.executor = ProcessPoolExecutor()
job.server.cores = 2
fs = job.run()

print(fs.done())
print(fs.result())
print(fs.done())
```

### DIFF
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -456,7 +456,11 @@ class Server:  # add the option to return the job id and the hold id to the serv
 
     @property
     def executor(self) -> Union[Executor, None]:
-        return self._executor
+        if self.run_mode.executor:
+            return self._executor
+        else:
+            self._executor = None
+            return self._executor
 
     @executor.setter
     def executor(self, exe: Union[Executor, None]):
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -5,7 +5,7 @@
 Server object class which is connected to each job containing the technical details how the job is executed.
 """
 
-from concurrent.futures import Executor
+from concurrent.futures import Executor, Future
 from collections import OrderedDict
 import numbers
 import socket
@@ -428,6 +428,49 @@ class Server:  # add the option to return the job id and the hold id to the serv
         """
         return self.view_queues()
 
+    @property
+    def executor(self) -> Union[Executor, None]:
+        if not self.run_mode.executor and self._executor is not None:
+            self._executor = None
+        return self._executor
+
+    @executor.setter
+    def executor(self, exe: Union[Executor, None]):
+        if isinstance(exe, Executor):
+            self._executor = exe
+            self.run_mode.executor = True
+        elif exe is None:
+            self._executor = None
+            self.run_mode.modal = True
+        else:
+            raise TypeError(
+                "The executor has to be derived from the concurrent.futures.Executor class."
+            )
+
+    @property
+    def future(self):
+        """
+        Python concurrent.futures.Future object to track the status of the execution of the job this server object is
+        attached to. This is an internal pyiron feature and most users never have to interact with the future object
+        directly.
+
+        Returns:
+             concurrent.futures.Future: future object to track the status of the execution
+        """
+        return self._future
+
+    @future.setter
+    def future(self, future_obj: Future):
+        """
+        Set a python concurrent.futures.Future object to track the status of the execution of the job this server object
+        is attached to. This is an internal pyiron feature and most users never have to interact with the future object
+        directly.
+
+        Args:
+            future_obj (concurrent.futures.Future): future object to track the status of the execution
+        """
+        self._future = future_obj
+
     @staticmethod
     def list_queues():
         """
@@ -453,33 +496,6 @@ class Server:  # add the option to return the job id and the hold id to the serv
             return state.queue_adapter.queue_view
         else:
             return None
-
-    @property
-    def executor(self) -> Union[Executor, None]:
-        if not self.run_mode.executor and self._executor is not None:
-            self._executor = None
-        return self._executor
-
-    @executor.setter
-    def executor(self, exe: Union[Executor, None]):
-        if isinstance(exe, Executor):
-            self._executor = exe
-            self.run_mode.executor = True
-        elif exe is None:
-            self._executor = None
-            self.run_mode.modal = True
-        else:
-            raise TypeError(
-                "The executor has to be derived from the concurrent.futures.Executor class."
-            )
-
-    @property
-    def future(self):
-        return self._future
-
-    @future.setter
-    def future(self, future_obj):
-        self._future = future_obj
 
     def to_hdf(self, hdf, group_name=None):
         """
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -8,9 +8,11 @@ Server object class which is connected to each job containing the technical deta
 from concurrent.futures import Executor
 from collections import OrderedDict
 import numbers
+import socket
+from typing import Union
+
 from pyiron_base.state import state
 from pyiron_base.jobs.job.extension.server.runmode import Runmode
-import socket
 
 __author__ = "Jan Janssen"
 __copyright__ = (
@@ -76,6 +78,10 @@ class Server:  # add the option to return the job id and the hold id to the serv
         .. attribute:: executor
 
             the executor can be used to execute the job object
+
+        .. attribute:: future
+
+            the concurrent.futures.Future object for monitoring the execution of the job object
     """
 
     def __init__(
@@ -108,6 +114,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         self._new_hdf = new_hdf
         self._send_to_db = False
         self._structure_id = None
+        self._future = None
         self._accept_crash = False
         self.additional_arguments = {}
 
@@ -448,18 +455,29 @@ class Server:  # add the option to return the job id and the hold id to the serv
             return None
 
     @property
-    def executor(self):
+    def executor(self) -> Union[Executor, None]:
         return self._executor
 
     @executor.setter
-    def executor(self, exe):
+    def executor(self, exe: Union[Executor, None]):
         if isinstance(exe, Executor):
             self._executor = exe
             self.run_mode.executor = True
+        elif exe is None:
+            self._executor = None
+            self.run_mode.executor = False
         else:
             raise TypeError(
                 "The executor has to be derived from the concurrent.futures.Executor class."
             )
+
+    @property
+    def future(self):
+        return self._future
+
+    @future.setter
+    def future(self, future_obj):
+        self._future = future_obj
 
     def to_hdf(self, hdf, group_name=None):
         """
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -254,10 +254,22 @@ class Server:  # add the option to return the job id and the hold id to the serv
 
     @property
     def gpus(self):
+        """
+        Total number of GPUs to use for this calculation.
+
+        Returns:
+            int: Total number of GPUs to use for this calculation.
+        """
         return self._gpus
 
     @gpus.setter
     def gpus(self, number_of_gpus):
+        """
+        Total number of GPUs to use for this calculation.
+
+        Args:
+            number_of_gpus (int): Total number of GPUs to use for this calculation.
+        """
         self._gpus = number_of_gpus
 
     @property
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -463,15 +463,14 @@ class Server:  # add the option to return the job id and the hold id to the serv
     @executor.setter
     def executor(self, exe: Union[Executor, None]):
         if isinstance(exe, Executor):
-            self._executor = exe
             self.run_mode.executor = True
         elif exe is None:
-            self._executor = None
             self.run_mode.modal = True
         else:
             raise TypeError(
                 "The executor has to be derived from the concurrent.futures.Executor class."
             )
+        self._executor = exe
 
     @property
     def future(self):
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -114,7 +114,6 @@ class Server:  # add the option to return the job id and the hold id to the serv
         self._new_hdf = new_hdf
         self._send_to_db = False
         self._structure_id = None
-        self._future = None
         self._accept_crash = False
         self.additional_arguments = {}
 
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -5,6 +5,7 @@
 Server object class which is connected to each job containing the technical details how the job is executed.
 """
 
+from concurrent.futures import Executor
 from collections import OrderedDict
 import numbers
 from pyiron_base.state import state
@@ -71,6 +72,10 @@ class Server:  # add the option to return the job id and the hold id to the serv
         .. attribute:: new_hdf
 
             defines whether a subjob should be stored in the same HDF5 file or in a new one.
+
+        .. attribute:: executor
+
+            the executor can be used to execute the job object
     """
 
     def __init__(
@@ -91,6 +96,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         self._memory_limit = None
         self._host = self._init_host(host=host)
         self._run_mode = Runmode()
+        self._executor = None
 
         self.queue = queue
 
@@ -440,6 +446,20 @@ class Server:  # add the option to return the job id and the hold id to the serv
             return state.queue_adapter.queue_view
         else:
             return None
+
+    @property
+    def executor(self):
+        return self._executor
+
+    @executor.setter
+    def executor(self, exe):
+        if isinstance(exe, Executor):
+            self._executor = exe
+            self.run_mode.executor = True
+        else:
+            raise TypeError(
+                "The executor has to be derived from the concurrent.futures.Executor class."
+            )
 
     def to_hdf(self, hdf, group_name=None):
         """
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -102,7 +102,8 @@ class Server:  # add the option to return the job id and the hold id to the serv
         self._memory_limit = None
         self._host = self._init_host(host=host)
         self._run_mode = Runmode()
-        self._executor = None
+        self._executor: Union[Executor, None] = None
+        self._future: Union[Future, None] = None
 
         self.queue = queue
 
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -465,7 +465,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
             self.run_mode.executor = True
         elif exe is None:
             self._executor = None
-            self.run_mode.executor = False
+            self.run_mode.modal = True
         else:
             raise TypeError(
                 "The executor has to be derived from the concurrent.futures.Executor class."
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -485,7 +485,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         self._executor = exe
 
     @property
-    def future(self):
+    def future(self) -> Union[Future, None]:
         """
         Python concurrent.futures.Future object to track the status of the execution of the job this server object is
         attached to. This is an internal pyiron feature and most users never have to interact with the future object
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -246,10 +246,22 @@ class Server:  # add the option to return the job id and the hold id to the serv
 
     @property
     def threads(self):
+        """
+        The number of threads selected for the current simulation
+
+        Returns:
+            (int): number of threads
+        """
         return self._threads
 
     @threads.setter
     def threads(self, number_of_threads):
+        """
+        The number of threads selected for the current simulation
+
+        Args:
+            number_of_threads (int): number of threads
+        """
         self._threads = number_of_threads
 
     @property
@@ -375,7 +387,8 @@ class Server:  # add the option to return the job id and the hold id to the serv
         Get the run mode of the job
 
         Returns:
-            (str/pyiron_base.server.runmode.Runmode): ['modal', 'non_modal', 'queue', 'manual']
+            (str/pyiron_base.server.runmode.Runmode): ['modal', 'non_modal', 'queue', 'manual', 'thread', 'worker',
+                        'interactive', 'interactive_non_modal', 'srun', 'executor']
         """
         return self._run_mode
 
@@ -385,7 +398,8 @@ class Server:  # add the option to return the job id and the hold id to the serv
         Set the run mode of the job
 
         Args:
-            new_mode (str): ['modal', 'non_modal', 'queue', 'manual']
+            new_mode (str): ['modal', 'non_modal', 'queue', 'manual', 'thread', 'worker', 'interactive',
+                        'interactive_non_modal', 'srun', 'executor']
         """
         self._run_mode.mode = new_mode
         if new_mode == "queue":
@@ -442,12 +456,24 @@ class Server:  # add the option to return the job id and the hold id to the serv
 
     @property
     def executor(self) -> Union[Executor, None]:
+        """
+        Executor to execute the job object this server object is attached to in the background.
+
+        Returns:
+            concurrent.futures.Executor:
+        """
         if not self.run_mode.executor and self._executor is not None:
             self._executor = None
         return self._executor
 
     @executor.setter
     def executor(self, exe: Union[Executor, None]):
+        """
+        Executor to execute the job object this server object is attached to in the background.
+
+        Args:
+            exe (concurrent.futures.Executor):
+        """
         if isinstance(exe, Executor):
             self.run_mode.executor = True
         elif exe is None:
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -476,7 +476,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
         """
         if isinstance(exe, Executor):
             self.run_mode.executor = True
-        elif exe is None:
+        elif exe is None and self.run_mode.executor:
             self.run_mode.modal = True
         else:
             raise TypeError(
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -478,7 +478,7 @@ class Server:  # add the option to return the job id and the hold id to the serv
             self.run_mode.executor = True
         elif exe is None and self.run_mode.executor:
             self.run_mode.modal = True
-        else:
+        elif exe is not None:
             raise TypeError(
                 "The executor has to be derived from the concurrent.futures.Executor class."
             )
--- a/pyiron_base/jobs/job/extension/server/generic.py
+++ b/pyiron_base/jobs/job/extension/server/generic.py
@@ -456,11 +456,9 @@ class Server:  # add the option to return the job id and the hold id to the serv
 
     @property
     def executor(self) -> Union[Executor, None]:
-        if self.run_mode.executor:
-            return self._executor
-        else:
+        if not self.run_mode.executor and self._executor is not None:
             self._executor = None
-            return self._executor
+        return self._executor
 
     @executor.setter
     def executor(self, exe: Union[Executor, None]):
--- a/pyiron_base/jobs/job/extension/server/queuestatus.py
+++ b/pyiron_base/jobs/job/extension/server/queuestatus.py
@@ -228,8 +228,8 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                     finished = True
                     break
                 elif isinstance(job.server.future, Future):
-                    finished = True
                     job.server.future.result(timeout=interval_in_s)
+                    finished = job.server.future.done()
                     break
                 else:
                     time.sleep(interval_in_s)
--- a/pyiron_base/jobs/job/extension/server/queuestatus.py
+++ b/pyiron_base/jobs/job/extension/server/queuestatus.py
@@ -228,6 +228,7 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                     finished = True
                     break
                 elif isinstance(job.server.future, Future):
+                    finished = True
                     job.server.future.result()
                     break
                 else:
--- a/pyiron_base/jobs/job/extension/server/queuestatus.py
+++ b/pyiron_base/jobs/job/extension/server/queuestatus.py
@@ -5,6 +5,7 @@
 Set of functions to interact with the queuing system directly from within pyiron - optimized for the Sun grid engine.
 """
 
+from concurrent.futures import Future
 import pandas
 import time
 import numpy as np
@@ -226,7 +227,11 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                 if job.status.string in job_status_finished_lst:
                     finished = True
                     break
-                time.sleep(interval_in_s)
+                elif isinstance(job.server.future, Future):
+                    job.server.future.result()
+                    break
+                else:
+                    time.sleep(interval_in_s)
             if not finished:
                 raise ValueError(
                     "Maximum iterations reached, but the job was not finished."
--- a/pyiron_base/jobs/job/extension/server/queuestatus.py
+++ b/pyiron_base/jobs/job/extension/server/queuestatus.py
@@ -229,7 +229,7 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                     break
                 elif isinstance(job.server.future, Future):
                     finished = True
-                    job.server.future.result()
+                    job.server.future.result(timeout=interval_in_s)
                     break
                 else:
                     time.sleep(interval_in_s)
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -480,7 +480,7 @@ class GenericJob(JobCore):
         # Copy executor - it cannot be copied and is just linked instead
         if self.server.executor is not None:
             copied_self.server.executor = self.server.executor
-        if self.server.future is not None and not self.sever.future.done():
+        if self.server.future is not None and not self.server.future.done():
             raise RuntimeError(
                 "Jobs whose server has executor and future attributes cannot be copied unless the future is `done()`"
             )
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -5,6 +5,7 @@
 Generic Job class extends the JobCore class with all the functionality to run the job object.
 """
 
+from concurrent.futures import Future
 from datetime import datetime
 import os
 import posixpath
@@ -430,6 +431,15 @@ class GenericJob(JobCore):
                     self.project_hdf5.file_name, self.job_name + "/status"
                 )
             )
+        if (
+            isinstance(self.server.future, Future)
+            and not self._status.finished
+            and self.server.future.done()
+        ):
+            if self.server.future.cancelled():
+                self._status.aborted = True
+            else:
+                self._status.finished = True
 
     def clear_job(self):
         """
@@ -604,6 +614,18 @@ class GenericJob(JobCore):
             input_only=True,
             new_database_entry=False,
         )
+
+    def remove(self, _protect_childs=True):
+        """
+        Remove the job - this removes the HDF5 file, all data stored in the HDF5 file an the corresponding database entry.
+
+        Args:
+            _protect_childs (bool): [True/False] by default child jobs can not be deleted, to maintain the consistency
+                                    - default=True
+        """
+        if isinstance(self.server.future, Future) and not self.server.future.done():
+            self.server.future.cancel()
+        super().remove(_protect_childs=_protect_childs)
 
     def remove_child(self):
         """
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -694,9 +694,9 @@ class GenericJob(JobCore):
                 if repair and self.job_id and not self.status.finished:
                     self._run_if_repair()
                 elif status == "initialized":
-                    return self._run_if_new(debug=debug)
+                    self._run_if_new(debug=debug)
                 elif status == "created":
-                    return self._run_if_created()
+                    self._run_if_created()
                 elif status == "submitted":
                     run_job_with_status_submitted(job=self)
                 elif status == "running":
@@ -1195,7 +1195,7 @@ class GenericJob(JobCore):
         Args:
             debug (bool): Debug Mode
         """
-        return run_job_with_status_initialized(job=self, debug=debug)
+        run_job_with_status_initialized(job=self, debug=debug)
 
     def _run_if_created(self):
         """
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -481,7 +481,9 @@ class GenericJob(JobCore):
         if self.server.executor is not None:
             copied_self.server.executor = self.server.executor
         if self.server.future is not None and not self.sever.future.done():
-            raise RuntimeError("Jobs whose server has executor and future attributes cannot be copied unless the future is `done()`")
+            raise RuntimeError(
+                "Jobs whose server has executor and future attributes cannot be copied unless the future is `done()`"
+            )
         return copied_self
 
     def _internal_copy_to(
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -439,7 +439,7 @@ class GenericJob(JobCore):
             if self.server.future.cancelled():
                 self.status.aborted = True
             else:
-                self._status.finished = True
+                self.status.finished = True
 
     def clear_job(self):
         """
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -433,7 +433,7 @@ class GenericJob(JobCore):
             )
         if (
             isinstance(self.server.future, Future)
-            and not self._status.finished
+            and not self.status.finished
             and self.server.future.done()
         ):
             if self.server.future.cancelled():
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -466,6 +466,10 @@ class GenericJob(JobCore):
         _job_reload_after_copy(
             job=copied_self, delete_file_after_copy=delete_file_after_copy
         )
+
+        # Copy executor - it cannot be copied and is just linked instead
+        if self.server.executor is not None:
+            copied_self.server.executor = self.server.executor
         return copied_self
 
     def _internal_copy_to(
@@ -690,9 +694,9 @@ class GenericJob(JobCore):
                 if repair and self.job_id and not self.status.finished:
                     self._run_if_repair()
                 elif status == "initialized":
-                    self._run_if_new(debug=debug)
+                    return self._run_if_new(debug=debug)
                 elif status == "created":
-                    self._run_if_created()
+                    return self._run_if_created()
                 elif status == "submitted":
                     run_job_with_status_submitted(job=self)
                 elif status == "running":
@@ -1191,7 +1195,7 @@ class GenericJob(JobCore):
         Args:
             debug (bool): Debug Mode
         """
-        run_job_with_status_initialized(job=self, debug=debug)
+        return run_job_with_status_initialized(job=self, debug=debug)
 
     def _run_if_created(self):
         """
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -437,7 +437,7 @@ class GenericJob(JobCore):
             and self.server.future.done()
         ):
             if self.server.future.cancelled():
-                self._status.aborted = True
+                self.status.aborted = True
             else:
                 self._status.finished = True
 
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -480,6 +480,8 @@ class GenericJob(JobCore):
         # Copy executor - it cannot be copied and is just linked instead
         if self.server.executor is not None:
             copied_self.server.executor = self.server.executor
+        if self.server.future is not None and not self.sever.future.done():
+            raise RuntimeError("Jobs whose server has executor and future attributes cannot be copied unless the future is `done()`")
         return copied_self
 
     def _internal_copy_to(
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -16,7 +16,6 @@ from pyiron_base.state import state
 from pyiron_base.utils.instance import static_isinstance
 
 
-
 """
 The function job.run() inside pyiron is executed differently depending on the status of the job object. This module 
 introduces the most general run functions and how they are selected. 
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -110,6 +110,8 @@ def run_job_with_status_created(job):
             gpus_per_slot = int(job.server.gpus / job.server.cores)
         else:
             gpus_per_slot = None
+        if gpus_per_slot < 0:
+            raise ValueError("Both job.server.gpus and job.server.cores have to be greater than zero.")
         run_job_with_runmode_executor(
             job=job,
             executor=job.server.executor,
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -8,8 +8,6 @@ import os
 import posixpath
 import subprocess
 
-from jinja2 import Template
-
 from pyiron_base.utils.deprecate import deprecate
 from pyiron_base.jobs.job.wrapper import JobWrapper
 from pyiron_base.state import state
@@ -81,7 +79,7 @@ def run_job_with_status_initialized(job, debug=False):
         print("job exists already and therefore was not created!")
     else:
         job.save()
-        return job.run()
+        job.run()
 
 
 def run_job_with_status_created(job):
@@ -112,7 +110,7 @@ def run_job_with_status_created(job):
             gpus_per_slot = int(job.server.gpus / job.server.cores)
         else:
             gpus_per_slot = None
-        return run_job_with_runmode_executor(
+        run_job_with_runmode_executor(
             job=job,
             executor=job.server.executor,
             gpus_per_slot=gpus_per_slot,
@@ -454,7 +452,12 @@ def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
     functionality for pyiron jobs. An executor is set as an attribute to the server object:
 
     >>> job.server.executor = concurrent.futures.Executor()
-    >>> fs = job.run()
+    >>> job.run()
+    >>> job.server.future.done()
+    False
+    >>> job.server.future.result()
+    >>> job.server.future.done()
+    True
 
     When the job is executed by calling the run() function a future object is returned. The job is then executed in the
     background and the user can use the future object to check the status of the job.
@@ -464,13 +467,15 @@ def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
         executor (concurrent.futures.Executor): executor class which implements the executor interface defined in the
                                                 python concurrent.futures.Executor class.
         gpus_per_slot (int): number of GPUs per MPI rank, typically 1
-
-     Returns:
-         concurrent.futures.Future: future object to develop asynchronous simulation protocols
     """
 
     if static_isinstance(
-        obj=job, obj_type="pyiron_base.jobs.master.generic.GenericMaster"
+        obj=job,
+        obj_type="pyiron_base.jobs.master.generic.GenericMaster"
+        # The static check is used to avoid a circular import:
+        # runfunction -> GenericJob -> GenericMaster -> runfunction
+        # This smells a bit, so if a better architecture is found in the future, use it
+        # to avoid string-based specifications
     ):
         raise NotImplementedError(
             "Currently job.server.run_mode.executor does not support GenericMaster jobs."
@@ -479,7 +484,7 @@ def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
         return run_job_with_runmode_executor_futures(job=job, executor=executor)
     else:
         raise NotImplementedError(
-            "Currently only flux.job.FluxExecutor and concurrent.futures.ProcessPoolExecutor are supported."
+            "Currently only concurrent.futures.ProcessPoolExecutor is supported."
         )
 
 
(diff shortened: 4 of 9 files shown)


jan-janssen:
An alternative suggestion would be to attach the future object to job.server.future, then the run function does not return any output, if that is more desirable.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -441,6 +447,20 @@ def view_queues():
         else:
             return None
 
+    @property
+    def executor(self):
```
liamhuber:
```suggestion
    def executor(self) -> Executor | None:
```

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -441,6 +447,20 @@ def view_queues():
...
+        return self._executor
+
+    @executor.setter
+    def executor(self, exe):
```
liamhuber:
```suggestion
    def executor(self, exe: Executor | None):
```

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -441,6 +447,20 @@ def view_queues():
...
+        if isinstance(exe, Executor):
+            self._executor = exe
+            self.run_mode.executor = True
+        else:
```
liamhuber:
```suggestion
        elif exe is None:
            self._executor = None
            self.run_mode.executor = False
        else:
```

Let's add a way to turn the executor back off if the user changes their mind.

pyiron_base/jobs/job/generic.py
```diff
@@ -690,9 +694,9 @@ def run(
                 if repair and self.job_id and not self.status.finished:
                     self._run_if_repair()
                 elif status == "initialized":
-                    self._run_if_new(debug=debug)
+                    return self._run_if_new(debug=debug)
```
liamhuber: Per the other discussion and your comment, I think I lean towards your alternative idea of holding the resulting `Future` object somewhere else (e.g. the `Server`) rather than returning it. Just so that this run mode looks more like the others.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -1,15 +1,19 @@
...
 # Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
 # Distributed under the terms of "New BSD License", see the LICENSE file.
+from concurrent.futures import ProcessPoolExecutor
 from datetime import datetime
 import multiprocessing
 import os
 import posixpath
 import subprocess
 
+from jinja2 import Template
```
liamhuber: This needs to be added to `setup.py` and `.ci_support/environment.yml`, or? I have it on my machine, but it's not in [the standard library](https://docs.python.org/3/library/), so I think we should specify it.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +446,91 @@ def run_job_with_runmode_srun(job):
...
+    """
+    Introduced in Python 3.2 the concurrent.futures interface enables the asynchronous execution of python programs.
+    A function is submitted to the executor and a future object is returned. The future object is updated in the
+    background once the executor finished executing the function. The job.server.run_mode.executor implements the same
+    functionality for pyiron jobs. An executor is set as an attribute to the server object:
+
+    >>> job.server.executor = concurrent.futures.Executor()
+    >>> fs = job.run()
+
+    When the job is executed by calling the run() function a future object is returned. The job is then executed in the
```
liamhuber, line 466: To update after implementation change

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +446,91 @@ def run_job_with_runmode_srun(job):
...
+        executor (concurrent.futures.Executor): executor class which implements the executor interface defined in the
+                                                python concurrent.futures.Executor class.
+        gpus_per_slot (int): number of GPUs per MPI rank, typically 1
+
+     Returns:
+         concurrent.futures.Future: future object to develop asynchronous simulation protocols
+    """
+
+    if static_isinstance(
+        obj=job, obj_type="pyiron_base.jobs.master.generic.GenericMaster"
```
liamhuber:
```suggestion
        obj=job, obj_type="pyiron_base.jobs.master.generic.GenericMaster"
        # The static check is used to avoid a circular import:
        # runfunction -> GenericJob -> GenericMaster -> runfunction
        # This smells a bit, so if a better architecture is found in the future, use it 
        # to avoid string-based specifications
```

I don't see an easy workaround right now and we have the tool to do this so let's just do it, I just want a quick note _why_ we do it in the hopes that a future dev (or future us) will understand quicker and/or see a better solution.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +446,91 @@ def run_job_with_runmode_srun(job):
...
+    ):
+        raise NotImplementedError(
+            "Currently job.server.run_mode.executor does not support GenericMaster jobs."
+        )
+    if isinstance(executor, ProcessPoolExecutor):
+        return run_job_with_runmode_executor_futures(job=job, executor=executor)
+    else:
+        raise NotImplementedError(
+            "Currently only flux.job.FluxExecutor and concurrent.futures.ProcessPoolExecutor are supported."
```
liamhuber:
```suggestion
            "Currently only concurrent.futures.ProcessPoolExecutor is supported."
```

I guess this is a holdover from some copypasta. This string will get updated in #1154, but let's keep it true for the PR it's in.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +446,91 @@ def run_job_with_runmode_srun(job):
...
+
+def run_job_with_runmode_executor_futures(job, executor):
+    """
+    Interface for the ProcessPoolExecutor implemented in the python standard library as part of the concurrent.futures
+    module. The ProcessPoolExecutor does not provide any resource management, so the user is responsible to keep track of
+    the number of compute cores in use, as over-subscription can lead to low performance.
+
+    >>> from concurrent.futures import ProcessPoolExecutor
+    >>> job.server.executor = ProcessPoolExecutor()
+    >>> fs = job.run()
```
liamhuber: Will need to be updated with the implementation

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +446,91 @@ def run_job_with_runmode_srun(job):
...
+
+    Args:
+        job (GenericJob): pyiron job object
+        executor (concurrent.futures.Executor): executor class which implements the executor interface defined in the
+                                                python concurrent.futures.Executor class.
+
+     Returns:
+         concurrent.futures.Future: future object to develop asynchronous simulation protocols
```
liamhuber: Update with implementation

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +446,91 @@ def run_job_with_runmode_srun(job):
...
+            )
+    else:
+        return executor.submit(
+            multiprocess_wrapper,
+            working_directory=job.project_hdf5.working_directory,
+            job_id=None,
+            file_path=job.project_hdf5.file_name + job.project_hdf5.h5_path,
+            debug=False,
+            connection_string=None,
+        )
```
liamhuber:
```suggestion
    if state.database.database_is_disabled:
        file_path = job.project_hdf5.file_name + job.project_hdf5.h5_path
        connection_string = None
    else:
        file_path = None
        if state.database.using_local_database:
            connection_string = str(job.project.db.conn.engine.url)
        else:
            connection_string = None
    
    return executor.submit(
                multiprocess_wrapper,
                working_directory=job.project_hdf5.working_directory,
                job_id=job.job_id,
                file_path=file_path,
                debug=False,
                connection_string=connection_string,
            )
```

Improve readability by factoring out the common code and avoiding logical negations

liamhuber: I find the above to be the most readable, but here is a line-minimized version that I don't find too bad:

```python
    connection_string = None
    file_path = None

    if not state.database.database_is_disabled and state.database.using_local_database: 
        connection_string=str(job.project.db.conn.engine.url)
    elif state.database.database_is_disabled:
        file_path=job.project_hdf5.file_name + job.project_hdf5.h5_path
    return executor.submit(
                multiprocess_wrapper,
                working_directory=job.project_hdf5.working_directory,
                job_id=job.job_id,
                file_path=file_path,
                debug=False,
                connection_string=connection_string,
            )
```

tests/job/test_genericJob.py
```diff
@@ -502,5 +503,25 @@ def test_return_codes(self):
...
+        fs = j.run()
+        fs.result()
+        self.assertTrue(fs.done())
+
+    def test_job_executor_copy(self):
+        j1 = self.project.create_job(ReturnCodeJob, "job_with_executor_copy")
+        j1.input["accepted_codes"] = [1]
+        j1.server.executor = ProcessPoolExecutor()
+        j2 = j1.copy()
+        self.assertTrue(j2.server.run_mode.executor)
```
liamhuber, line 545:
```suggestion
        self.assertIs(j2.server.executor, j.server.executor)
        self.assertTrue(j2.server.run_mode.executor)
```

I believe we also want them to literally have the same executor, so let's check that too.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +446,91 @@ def run_job_with_runmode_srun(job):
...
+            "Currently only flux.job.FluxExecutor and concurrent.futures.ProcessPoolExecutor are supported."
+        )
+
+
+def run_job_with_runmode_executor_futures(job, executor):
+    """
+    Interface for the ProcessPoolExecutor implemented in the python standard library as part of the concurrent.futures
+    module. The ProcessPoolExecutor does not provide any resource management, so the user is responsible to keep track of
+    the number of compute cores in use, as over-subscription can lead to low performance.
+
```
liamhuber, line 500:
```suggestion

    The [ProcessPoolExecutor docs](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) state: "The __main__ module must be importable by worker subprocesses. This means that ProcessPoolExecutor will not work in the interactive interpreter." (i.e. Jupyter notebooks). For standard usage this is a non-issue, but for the edge case of job classes defined in-notebook (e.g. children of `PythonTemplateJob`), the using the ProcessPoolExecutor will result in errors. To resolve this, relocate such classes to an importable .py file.

```

This is definitely an edge case, but I don't think the average user is going to be poking around in the docstring of `run_job_with_runmode_executor_futures` just for fun, so I'd rather take a couple lines to raise the issue.

liamhuber (changes requested):
Yeah, looks good. I do think the future should be put somewhere else, probably on the server as you suggest. But that's the only non-local/big change; the rest is just a bunch of very specific requests.

EDIT: Ok actually, I have one more UX concern: the error raised when you mix `executor` children with masters is not very clear. E.g.

```python

pr = Project('tmp')

job = pr.create.job.Lammps("lmp")
job.server.executor = ProcessPoolExecutor(max_workers=4)

master = pr.create.job.Murnaghan("murn")
master.ref_job = job
master.run()
```

Runs fine without the executor line, but with it gives a stack trace ending in

```
File ~/work/pyiron/pyiron_base/pyiron_base/jobs/master/parallel.py:556, in ParallelMaster.run_static(self)
    554             self._run_if_master_modal_child_non_modal(job)
    555         else:
--> 556             raise TypeError()
    557 else:
    558     self.status.collect = True

TypeError: 
```

It would be great to have something more informative. I also checked `SerialMasterBase` and it's got the same safety `else: raise TypeError()` clause that should keep the behaviour reasonable, but will need its own informative error message.

pyiron_base/jobs/job/runfunction.py
jan-janssen: `jinja2` is required for `flux` so I move it to the other pull request.

jan-janssen:
```suggestion
```

pyiron_base/jobs/job/runfunction.py
jan-janssen: Done

pyiron_base/jobs/job/runfunction.py
jan-janssen: Done

pyiron_base/jobs/job/generic.py
jan-janssen: Done

pyiron_base/jobs/job/runfunction.py
jan-janssen: Done

jan-janssen:
> Yeah, looks good. I do think the future should be put somewhere else, probably on the server as you suggest. But that's the only non-local/big change; the rest is just a bunch of very specific requests.

The future is now moved to `job.server.future`. 

> EDIT: Ok actually, I have one more UX concern: the error raised when you mix `executor` children with masters is not very clear. E.g.

Done.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -441,6 +454,31 @@ def view_queues():
...
+        return self._executor
+
+    @executor.setter
+    def executor(self, exe: Union[Executor, None]):
+        if isinstance(exe, Executor):
+            self._executor = exe
+            self.run_mode.executor = True
+        elif exe is None:
+            self._executor = None
+            self.run_mode.executor = False
```
liamhuber:
```suggestion
            self.run_mode.modal = True
```

There is a critical test missing where we turn the executor on and off again -- my original suggestion raises an error! The suggestion here is also not going to cut it, because `executor` is not playing well with the other run modes, but it is at least a step in the right direction.

Problem with my original suggestion:

```python

pr = Project('tmp')

job = pr.create.job.Lammps("lmp")
job.server.executor = ProcessPoolExecutor()
job.server.executor = None
```

gives 

```
Cell In[7], line 10
      8 job.server.executor = ProcessPoolExecutor()
      9 print(job.server.run_mode)
---> 10 job.server.executor = None
     12 # master = pr.create.job.Murnaghan("murn")
     13 # master.ref_job = job
     14 # master.run()

File ~/work/pyiron/pyiron_base/pyiron_base/jobs/job/extension/server/generic.py:468, in Server.executor(self, exe)
    466 elif exe is None:
    467     self._executor = None
--> 468     self.run_mode.executor = False
    469 else:
    470     raise TypeError(
    471         "The executor has to be derived from the concurrent.futures.Executor class."
    472     )

File ~/work/pyiron/pyiron_base/pyiron_base/jobs/job/extension/server/runmode.py:103, in Runmode.__setattr__(self, name, value)
    101         self.mode = name
    102     else:
--> 103         raise ValueError("A run mode can only be activated using [True].")
    104 else:
    105     super(Runmode, self).__setattr__(name, value)

ValueError: A run mode can only be activated using [True].
```

liamhuber (changes requested):
I like the modifications and they addressed all my original concerns, but per [my comment on 1154](https://github.com/pyiron/pyiron_base/pull/1154#issuecomment-1620742898) I played around a bit combining it with other elements of the codebase and ran into trouble.

The issue I see is that this is not yet deeply integrated enough with the rest of the `Server.run_mode` options. What is needed is to think through the different combinations of user actions, decide what the outcome of these actions should be, and then enforce these outcomes in the tests -- and the test part really is crticial, e.g. "turning it on and turning it off again" was never actually tested and my original suggestion resulted in an exception!

I'm not familiar off-hand with all the ways we have to modify the `run_mode`, but here is one example of what I'm thinking of:

```python

pr = Project('tmp')

job = pr.create.job.Lammps("lmp")
job.structure = pr.atomistics.structure.bulk("Al")
job.server.executor = ProcessPoolExecutor()
print(job.server.run_mode)
# >>> executor
job.interactive_open()
print(job.server.run_mode, job.server.executor is not None)
# >>> interactive True
job.run()
print(job.output.energy_pot, job.server.future)
# >>> [-3.36] None

job2 = job.copy()
print(job.server.run_mode, job2.server.run_mode)
# >>> interactive executor
```

I think it's fine that turning on `interactive_open()` supplanted the `executor` and left the `future` empty, but then It's weird that the `server.executor` stays populated. Even worse, if we _copy_ the interactively-opened job, the copy winds up with a different run mode! In this particular case, I guess setting the run mode to interactive needs to clear the executor to `None`?

I don't have super strong opinions right now about what the behaviour should be, but it shouldn't be left up to chance/side effects and decisions about the intended interface need to be enforced by the tests.

@samwaseda, I think that after this PR is resolved most of the remaining changes for flux in particular over in #1154 will be pretty straightforward; it's probably more worthwhile for you to review here before reading there.

jan-janssen:
> I think it's fine that turning on `interactive_open()` supplanted the `executor` and left the `future` empty, but then It's weird that the `server.executor` stays populated. Even worse, if we _copy_ the interactively-opened job, the copy winds up with a different run mode! In this particular case, I guess setting the run mode to interactive needs to clear the executor to `None`?

I fixed the issue and reset the executor when the run_mode is changed afterwards. Currently, the interactive mode is not yet compatible to the executor, these are two separate developments and we have to think more how an integration could look like.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -441,6 +454,35 @@ def view_queues():
         else:
             return None
 
+    @property
+    def executor(self) -> Union[Executor, None]:
+        if self.run_mode.executor:
+            return self._executor
+        else:
+            self._executor = None
+            return self._executor
```
liamhuber:
```suggestion
        if not self.run_mode.executor and self._executor is not None:
            self._executor = None
        return self._executor
```

Cleaner with a single return loop, IMO

jan-janssen:
@liamhuber , @niklassiemer  and @samwaseda Any other feedback? Otherwise I would like to merge this pull request, so we can merge the flux pull request on Friday and have everything ready next Monday.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -431,6 +444,91 @@ def run_job_with_runmode_srun(job):
     )
 
 
+def run_job_with_runmode_executor(job, executor, gpus_per_slot=None):
```
pmrv, line 451: Should this or `run_job_with_runmode_executor_futures` be decorated with `run_time_decorator` or is this somehow taken care of the `multiprocess_wrapper`?

pmrv:
Am generally happy with this new interface, it feels more natural and powerful.  If this works with other concurrent futures executors (and I don't see why not) it might also give us a why replace remote jobs with specific executors, like parsl or dask (notwithstanding  their other shortcomings). 

A small additional feature I would like is that `Project.wait_for_job` is aware of executor jobs and waits on the future objects.

One can of worms I think we need to open is canceling futures.  Say I submit a bunch of things and make a mistake or my workflow converges and I don't need additional calculations.  I suppose `job.server.executor.shutdown()` is always possible, but users will likely have one executor for a notebook, so that's a bit harsh.  `job.server.future.cancel()` I guess will work to stop the calculation, but probably won't update the database entry of the job.  I'm not sure if it is possible to attach something to the future that will take care of this, but a minimal fix would be to make `Project.refresh_job_status` aware of executors, so that it can reset the job status if it finds one with a canceled future.  Whatever the solution `GenericJob.kill` should also cancel the underlying future if there is one.

Other than that, I think it's good stuff.

jan-janssen:
> A small additional feature I would like is that `Project.wait_for_job` is aware of executor jobs and waits on the future objects.

I like the idea and I guess we can extend this concept in a way that if you remove the jobs using `Project.remove_jobs` the futures are also cancelled. 

The other way of cancelling the futures resulting the job status to change sounds more difficult to me and would require more of a rewrite of the integration of the `job.status` with the `concurrent.futures`. I agree that this would be great, but for me this is out of the scope of the current pull request.

samwaseda:
Do I understand correctly that with the [last commit](https://github.com/pyiron/pyiron_base/pull/1155/commits/c9ef6ee4cf3ade4827d5b827d45dfad51552abbb) we can simply use `job.status` and access data from `job` if the job is finished? Or do I still need to call `job.server.future.result()`?

jan-janssen:
> Do I understand correctly that with the [last commit](https://github.com/pyiron/pyiron_base/pull/1155/commits/c9ef6ee4cf3ade4827d5b827d45dfad51552abbb) we can simply use `job.status` and access data from `job` if the job is finished? Or do I still need to call `job.server.future.result()`?

You can always use `job.status` to check the status of the job, this worked the whole time. The last commit just implements to security checks: 
1) it cancels the future object when you remove the job before it is executed, this helpful if you create ten jobs, submit them all to an executor and then decide to remove job five to eight before they are executed. Without this fix pyiron would start to run the job just to realize that it no longer exists. With this fix the Executor is informed when you delete the job. 
2) if the user decides to cancel the job, using `job.server.future.cancel()` rather than the traditional `job.status.aborted = True`, then this is now correctly transferred to the job status, so the job status is set to `aborted` when the future object is cancelled.

samwaseda:
ok that was not quite my question. I wanted to ask whether `job.server.future.result()` has to be called before the final results can be accessed.

jan-janssen:
> ok that was not quite my question. I wanted to ask whether `job.server.future.result()` has to be called before the final results can be accessed.

no, and this was never the case. You can use `job.server.future.result()` as alternative to `pr.wait_for_job(job)` but it is not mandatory.

jan-janssen:
Furthermore, with the last commit when the user calls `pr.wait_for_job(job)` it now internally uses `job.server.future.result()`.

samwaseda:
ok, what is now the difference between `non_modal` and `executor`? I'm not talking about the technical detail but as a concept.

jan-janssen:
> ok, what is now the difference between `non_modal` and `executor`? I'm not talking about the technical detail but as a concept.

`executor` is the better `non_modal`, I am happy to remove `non_modal` once we have the `executor` working for `GenericMasters`.

jan-janssen:
> One can of worms I think we need to open is canceling futures. Say I submit a bunch of things and make a mistake or my workflow converges and I don't need additional calculations. I suppose `job.server.executor.shutdown()` is always possible, but users will likely have one executor for a notebook, so that's a bit harsh. `job.server.future.cancel()` I guess will work to stop the calculation, but probably won't update the database entry of the job. I'm not sure if it is possible to attach something to the future that will take care of this, but a minimal fix would be to make `Project.refresh_job_status` aware of executors, so that it can reset the job status if it finds one with a canceled future. Whatever the solution `GenericJob.kill` should also cancel the underlying future if there is one.

This is now implemented as well.

jan-janssen:
@liamhuber , @niklassiemer , @pmrv and @samwaseda : Any other comments? Otherwise, can you approve the PR, so I know who agrees with merging it? I would really like to get the flux integration done tomorrow.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -441,6 +454,33 @@ def view_queues():
...
+        if isinstance(exe, Executor):
+            self._executor = exe
+            self.run_mode.executor = True
+        elif exe is None:
+            self._executor = None
+            self.run_mode.modal = True
+        else:
+            raise TypeError(
+                "The executor has to be derived from the concurrent.futures.Executor class."
+            )
```
samwaseda:
```suggestion
        if isinstance(exe, Executor):
            self.run_mode.executor = True
        elif exe is None:
            self.run_mode.modal = True
        else:
            raise TypeError(
                "The executor has to be derived from the concurrent.futures.Executor class."
            )
        self._executor = exe
```

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -441,6 +454,33 @@ def view_queues():
...
+
+    @property
+    def future(self):
```
samwaseda: I think this one merits a Doc

liamhuber:
> @liamhuber , @niklassiemer , @pmrv and @samwaseda : Any other comments? Otherwise, can you approve the PR, so I know who agrees with merging it? I would really like to get the flux integration done tomorrow. 

I will be able to review again today.

pyiron_base/jobs/job/generic.py
```diff
@@ -430,6 +431,15 @@ def refresh_job_status(self):
                     self.project_hdf5.file_name, self.job_name + "/status"
                 )
             )
+        if (
+            isinstance(self.server.future, Future)
+            and not self._status.finished
```
samwaseda:
```suggestion
            and not self.status.finished
```

pyiron_base/jobs/job/generic.py
```diff
@@ -430,6 +431,15 @@ def refresh_job_status(self):
...
+            and self.server.future.done()
+        ):
+            if self.server.future.cancelled():
+                self._status.aborted = True
```
samwaseda:
```suggestion
                self.status.aborted = True
```

pyiron_base/jobs/job/generic.py
```diff
@@ -430,6 +431,15 @@ def refresh_job_status(self):
...
+            else:
+                self._status.finished = True
```
samwaseda:
```suggestion
                self.status.finished = True
```

pyiron_base/jobs/job/runfunction.py
```diff
@@ -102,6 +105,16 @@ def run_job_with_status_created(job):
         job.run_static()
     elif job.server.run_mode.srun:
         run_job_with_runmode_srun(job=job)
+    elif job.server.run_mode.executor:
+        if job.server.gpus is not None:
+            gpus_per_slot = int(job.server.gpus / job.server.cores)
```
samwaseda, line 110: I'm a bit uncomfortable with the fact that there's no doc in `job.server.gpus` and also no error check there, so I might want to see a simple check like `assert gpus_per_slot > 0` at this point.

samwaseda (approved):
I made a few comments, but now I'm generally happy with this implementation.

liamhuber:
> > @liamhuber , @niklassiemer , @pmrv and @samwaseda : Any other comments? Otherwise, can you approve the PR, so I know who agrees with merging it? I would really like to get the flux integration done tomorrow.
> 
> I will be able to review again today.

@jan-janssen I have a meeting with Joerg before I can get to a real review, but I can tell you already this is not going to pass. There have been _zero_ changes to `test/*` since my initial review, not even my very initial request of testing "turning it on and turning it off again". 

The whole point of having the dev releases was to avoid merging stuff into `pyiron_base:main:HEAD` that is not ready, and to avoid insane time crunches. Your statement about wanting to use flux in production by the end of the month seems ambitious but reasonable -- getting this merged by the end of the week is really rushing things.

I would strongly suggest that we slow down and try to do a better job here before merging this to main. If you want people using this by next week, let's take the functionality on this branch, make a new "in progress" branch off of it (well, off the flux branch, which sits on top of all the changes here), and make a new dev release.

pyiron_base/jobs/job/extension/server/generic.py
jan-janssen: Done

pyiron_base/jobs/job/runfunction.py
jan-janssen: The `run_time_decorator` is only applied when the job is actually executed. Here the `multiprocess_wrapper` creates a new `JobWrapper` object which then calls `run_static()`. So yes, it is taken care of inside the `multiprocess_wrapper`.

pyiron_base/jobs/job/runfunction.py
jan-janssen: I added a docstring to `job.server.gpus` as when as the recommended check if `gpus_per_slot > 0`.

jan-janssen:
> The general fear I have is that the pyiron development slows down more and more, approaching what computer scientists would call the `waterfall model`, while I personally prefer a more agile approach as we used to practice it before. 

The agile manifesto will be the first one to state the importance of tests.  Anyway, I personally feel that agile vs. waterfall is mostly a rhetorical debate at this point, so it's maybe not the most productive discussion to have.


I'm generally happy with the functionality here.  I'll try to give it a detailed look again after dinner, but will otherwise leave to @liamhuber .

pyiron_base/jobs/job/extension/server/queuestatus.py
```diff
@@ -226,7 +227,12 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                 if job.status.string in job_status_finished_lst:
                     finished = True
                     break
-                time.sleep(interval_in_s)
+                elif isinstance(job.server.future, Future):
+                    finished = True
+                    job.server.future.result()
```
pmrv:
```suggestion
                    job.server.future.result(timeout=interval_in_s)
```

pyiron_base/jobs/job/generic.py
```diff
@@ -430,6 +431,15 @@ def refresh_job_status(self):
...
             )
+        if (
+            isinstance(self.server.future, Future)
+            and not self.status.finished
+            and self.server.future.done()
+        ):
+            if self.server.future.cancelled():
+                self.status.aborted = True
+            else:
+                self.status.finished = True
```
pmrv, line 442: That works for me for now.

jan-janssen: I merged the changes you suggested, so if there are no more modification request, can you approve the pull request? Then we have everything clearly documented here.

liamhuber:
> > The general fear I have is that the pyiron development slows down more and more, approaching what computer scientists would call the `waterfall model`, while I personally prefer a more agile approach as we used to practice it before.
> 
> The agile manifesto will be the first one to state the importance of tests. Anyway, I personally feel that agile vs. waterfall is mostly a rhetorical debate at this point, so it's maybe not the most productive discussion to have.

I don't see the original waterfall comment, but please don't misunderstand my comments as suggesting we move in that direction; I absolutely do not think that we need to pre-engineer everything. What I want to see is this:

- Want a thing
- Write a thing
- Now it's good enough for `pyiron_contrib`! If you want to merge the thing to `pyiron_base`, keep going
- Write a test to make sure the thing will keep working like you envision
- Write documentation so other people know how to use your new thing
- Ask for review
- Author and reviewers should stop for a second and think: does this thing interact with other parts of the user-facing code? If yes...
  - Decide what that interaction should look like
  - Modify the code to either behave that way, or fail cleanly if such a modification is too much work
  - Add tests so that this user-facing behaviour stays robust
  - Stacked PRs may be super useful here to keep each set of changes well-scoped
- Now it's good enough for `base`

We absolutely will not see _all_ the interactions and side effects a priori. At some point we merge things, use them, learn, and adapt. _But_ if we _can_ see hiccups and interactions and side effects a priori, ignoring them is not "agile", it's irresponsible.

So we don't need to solve all future problems in this PR! But something like "A user has set an executor. Now they want to remove that executor -- what should happen?" is absolutely within the scope of this PR. The functionality has been added, but IMO we also need to see it in the tests so that this element of the _user interface_ behaves and stays behaving as expected.

> I'm generally happy with the functionality here. I'll try to give it a detailed look again after dinner, but will otherwise leave to @liamhuber .

Absolutely! I'm excited to have executor functionality and agree it really supplants "non-modal" as something superior! I still need to do a detailed re-review, but it also looks like everyone else thought of some important interactions and that these have been handled.

"after dinner" sounds like you and I are in substantially different time zones again. Since this is time sensitive, after I review I'll open a stacked PR to start introducing the tests that are still needed. It's possible that I won't be able to figure out from the discussion here what the intended interface is for certain things and that we'll need a round of feedback on my tests in the CET-Friday AM, but fingers crossed they'll look alright and be comprehensive enough to just merge in here.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -102,6 +114,7 @@ def __init__(
         self._new_hdf = new_hdf
         self._send_to_db = False
         self._structure_id = None
+        self._future = None
```
liamhuber:
```suggestion
```

Readability nit: let's slide this up to be adjacent to `_executor`

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -91,6 +102,7 @@ def __init__(
         self._memory_limit = None
         self._host = self._init_host(host=host)
         self._run_mode = Runmode()
+        self._executor = None
```
liamhuber:
```suggestion
        self._executor: Union[Executor, None] = None
        self._future: Union[Future, None] = None
```

Readability nit: slid `_future` up to be close to the thing it's related to and added type hint.

jan-janssen:
> > The general fear I have is that the pyiron development slows down more and more, approaching what computer scientists would call the `waterfall model`, while I personally prefer a more agile approach as we used to practice it before.
> 
> The agile manifesto will be the first one to state the importance of tests. Anyway, I personally feel that agile vs. waterfall is mostly a rhetorical debate at this point, so it's maybe not the most productive discussion to have.
> 
> I'm generally happy with the functionality here. I'll try to give it a detailed look again after dinner, but will otherwise leave to @liamhuber .

@pmrv It seems like you modified my comment above, that made the whole discussion rather confusing: 
Screenshot 2023-07-06 at 11 17 47 AM

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -415,6 +454,60 @@ def queue_view(self):
         """
         return self.view_queues()
 
+    @property
+    def executor(self) -> Union[Executor, None]:
```
liamhuber, line 458: Nothing wrong with this PR, but I'm looking forward to dropping <3.10 so that we can just write `Executor | None` 😝

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -415,6 +454,60 @@ def queue_view(self):
...
+        """
+        Executor to execute the job object this server object is attached to in the background.
+
+        Args:
+            exe (concurrent.futures.Executor):
+        """
+        if isinstance(exe, Executor):
+            self.run_mode.executor = True
+        elif exe is None:
+            self.run_mode.modal = True
```
liamhuber:
```suggestion
        elif exe is None:
            if self.executor is not None:
                self.run_mode.modal = True
            # else pass, it's already None and you're setting it to None
```

I know I wrote this, but I expressed some concern then and I am still pretty sure it's bad. E.g. consider that a user thinks "oh shoot, I forgot to disable the executor" and tries to do it manually after some other run-mode-modifier:

```python
job.interactive_open()
job.server.executor = None
```

Now they've interactive opened but the run mode is modal!

I _think_ this suggestion now avoids that by first going through the `executor` property, which will force the value over to `None` if the run mode has changed, but I'll need to write up some tests to be sure it's doing what I imagine.

liamhuber: It's also probably that the whole setter can now be refactored to have a more readable logic flow, but that's hard for me to see in review+suggestion mode

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -415,6 +454,60 @@ def queue_view(self):
...
+        else:
+            raise TypeError(
+                "The executor has to be derived from the concurrent.futures.Executor class."
+            )
+        self._executor = exe
+
+    @property
+    def future(self):
```
liamhuber:
```suggestion
    def future(self) -> Union[Future, None]:
```

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -415,6 +454,60 @@ def queue_view(self):
...
+        attached to. This is an internal pyiron feature and most users never have to interact with the future object
+        directly.
+
+        Returns:
+             concurrent.futures.Future: future object to track the status of the execution
+        """
+        return self._future
+
+    @future.setter
+    def future(self, future_obj: Future):
```
liamhuber, line 500: This type hint raises a good question. What do we want to do to to `job.server.future` when the job has run with an executor and now the executor is being removed/the run mode is being changed?

I'm a little surprised we allow run mode changes after running the job tbh, but we do. I also don't really care what we do with the future -- I don't see any reason leaving it there orphaned from the executor would be a problem. But I would like it to be a choice we're making.

pyiron_base/jobs/job/extension/server/queuestatus.py
```diff
@@ -226,7 +227,12 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
                 if job.status.string in job_status_finished_lst:
                     finished = True
                     break
-                time.sleep(interval_in_s)
+                elif isinstance(job.server.future, Future):
```
liamhuber, line 230: Aha, so my previous comment about what to do with `server.future` might actually be more relevant than I thought. So just from reading the code here it's hard for me to tell if this is actually a problem, but I'm concerned about a situation like this:
- A user runs a job with an executor, populating the `future` attribute
- They do something to kind of clean up/restart the job with a _different_ run mode, but the `future` sticks around
- They run the job and then try to `wait`, but the wait terminates prematurely because the `future` is old and already has a result

I'm just not sure if that middle bullet point is actually doable without playing a bit -- do we have any restarts etc. that would reset the status without cleaning the future? I'll try play with it in the tests stacked PR but maybe someone else already knows what happens/where to look.

liamhuber: I had a quick go at this in the notebook and I don't see any way to awkwardly complete step 2, but maybe I am just not imaginative enough. I'm optimistic we're safe here.

pyiron_base/jobs/job/generic.py
```diff
@@ -466,6 +476,10 @@ def copy(self):
         _job_reload_after_copy(
             job=copied_self, delete_file_after_copy=delete_file_after_copy
         )
+
+        # Copy executor - it cannot be copied and is just linked instead
+        if self.server.executor is not None:
```
liamhuber, line 481: The `future` doesn't get copied. Should it?

pyiron_base/jobs/job/generic.py
```diff
@@ -601,6 +615,18 @@ def copy_template(self, project=None, new_job_name=None):
...
 
+    def remove(self, _protect_childs=True):
+        """
+        Remove the job - this removes the HDF5 file, all data stored in the HDF5 file an the corresponding database entry.
+
+        Args:
+            _protect_childs (bool): [True/False] by default child jobs can not be deleted, to maintain the consistency
+                                    - default=True
+        """
+        if isinstance(self.server.future, Future) and not self.server.future.done():
```
liamhuber, line 630: This code holds spec -- namely, that if a job running on an executor is `remove()`d, the server should first cancel the future. I'm not sure where is the best place to document this promise though -- right here, or over on the attribute docs for `Server.future`? I kind of lean towards the latter; it will make the attribute doc blow up a bit, but my initial feeling is that it's better to have the future/executor documentation centralized somewhere; if people want to use future/executor, they will want/need to look at the server docs anyhow.

tests/job/test_genericJob.py
```diff
@@ -502,5 +504,56 @@ def test_return_codes(self):
             pass
         self.assertTrue(j.status.aborted, "Job did not abort even though return code is 2!")
 
+    def test_job_executor_set(self):
```
liamhuber: I guess this test can be moved over to the `Server` tests since the fact it lives on a job is an uneccesary complication.

tests/job/test_genericJob.py
```diff
@@ -502,5 +504,56 @@ def test_return_codes(self):
...
+        j = self.project.create_job(ReturnCodeJob, "job_with_executor_set")
+        j.server.executor = ProcessPoolExecutor()
+        self.assertTrue(j.server.run_mode.executor)
+        j.server.run_mode.modal = True
+        self.assertFalse(j.server.run_mode.executor)
```
liamhuber:
```suggestion
        self.assertFalse(j.server.run_mode.executor)
        self.assertIsNone(
            j.server.executor,
            msg="The run mode executor and the presence of an executor object should be 1:1"
        )
```

liamhuber: This part should also include tests to verify that whatever behaviour we want for `Server.future` when the run mode is changed away from `executor` are actually happening

tests/job/test_genericJob.py
```diff
@@ -502,5 +504,56 @@ def test_return_codes(self):
...
+        j1 = self.project.create_job(ReturnCodeJob, "job_with_executor_copy")
+        j1.input["accepted_codes"] = [1]
+        j1.server.executor = ProcessPoolExecutor()
+        j2 = j1.copy()
+        self.assertIs(j2.server.executor, j1.server.executor)
+        self.assertTrue(j2.server.run_mode.executor)
+        j2.run()
+        j2.server.future.result()
+        self.assertTrue(j2.server.future.done())
+
```
liamhuber, line 554:
```suggestion
        j2.server.future = Future()
        # Manually override the future with one that isn't done() to test copy spec:
        # No copying jobs with futures that aren't done
        with self.assertRaises(RuntimeError):
            j2.copy()
```

Assuming this is the behaviour we want when copying finished executor jobs. I'm open to the opposite behaviour if there's a case for it that I'm not seeing.

liamhuber (changes requested):
Super! A handful of tests got added between my last comment and reviewing, and most cases I can think of have some coverage now. The outstanding issue I see is that we need to decide what should happen to `Server.future` under a few conditions and then test that those design choices are being actualized. Here is a summary of my review comments:

Statements:
- The `server.future` should get copied along with the `server.executor` when we copy jobs. (In a suggestion)
- Setting the `server.executor` to `None` when the run mode is not `executor` should not change the run mode. (In a suggestion)
- One test should be moved over to the `Server` tests since it doesn't care about jobs

Questions:
- What should happen to `server.future` when the run mode/executor are changed?
  - We use `job.server.future` throughout for a bunch of checks; does it make sense to protect this attribute with regards to changes in run mode the same way we protect `server.executor`? I couldn't quickly find an edge case where a user changes the mode of a _completed_ executor-executed job and runs into trouble, but I still have a vague sense of unease here, so I lean somewhat towards clearing the future like we clear the executor.
- Does it make sense to summarize the overall spec for executors+futures somewhere? I imagine something like a list of promises we are making for user-facing behaviour as it pertains to executors "execution can be done using a ...Executor; this populates the ...Future object on ...future; copying a job(server?) will copy all of its executor/future data; removing a job with an executor will send a cancel message to that executor; etc..."

Tests still needed:
- Copying a completed job copies the future too. (In a suggestion)
- Setting `server.executor` to `None` when it is already `None` should have no impact under any condition (namely, the condition that the `run_mode` is something other than `executor` or `modal`)
- Test that `server.future` is whatever it should be if the `server.run_mode` of a job that started executor execution subsequently gets changed
  - Probably this means testing `self.AssertIsNone(job.server.future)` after the change, but I'm open to alternative design choices.

liamhuber:
@jan-janssen 
> I added the tests you suggested

I like them! With the comment edit I didn't realize they had been added after my comment until I had scrolled down to the bottom of the review where they were a pleasant surprise! I think they nicely cover most of the weird cases we could see a priori that a user might run into. Cf. my review about making similar choices and tests for `Server.future`, but I think this PR is now extremely close to giving us a clearly defined interface to `Executor`.

I'll work on other stuff for a while, but I'll check back here -- if you ask or if all reasonable working hours are finished in your time zone, I'm happy to stack a PR that resolves my outstanding concerns.

pmrv:
> > > The general fear I have is that the pyiron development slows down more and more, approaching what computer scientists would call the `waterfall model`, while I personally prefer a more agile approach as we used to practice it before.
> > 
> > 
> > The agile manifesto will be the first one to state the importance of tests. Anyway, I personally feel that agile vs. waterfall is mostly a rhetorical debate at this point, so it's maybe not the most productive discussion to have.
> > I'm generally happy with the functionality here. I'll try to give it a detailed look again after dinner, but will otherwise leave to @liamhuber .
> 
> @pmrv It seems like you modified my comment above, that made the whole discussion rather confusing: Screenshot 2023-07-06 at 11 17 47 AM

Oh, I'm really (**EDIT**) *not* sure how that happened.  Sorry.  Here's @jan-janssen original comment as per my email notifications.

>> There have been zero changes to test/* since my initial review, not even my very initial request of testing "turning it on and turning it off again".
> 
>I added the tests you suggested.
> 
>> I would strongly suggest that we slow down and try to do a better job here before merging this to main. If you want people using this by next week, let's take the functionality on this branch, make a new "in progress" branch off of it (well, off the flux branch, which sits on top of all the changes here), and make a new dev release.
>
> I agree that the development releases are the right way to move forward, but this functionality is not yet available, so for now I am to get his merged in the main branch.
>
> The general fear I have is that the pyiron development slows down more and more, approaching what computer scientists would call the waterfall model, while I personally prefer a more agile approach as we used to practice it before.

Again, sorry, this is on me.

EDIT:  First the edit and then a relevant typo, I think I better stay away for the computer for the rest of the day. >>

pyiron_base/jobs/job/extension/server/generic.py
jan-janssen: The official end of life for 3.8 is 2024-10 and the official end of life for 3.9 is 2025-10. So even if we move quickly and deprecate versions one year ahead of the end of life, we are going to stick with Python 3.9 for approximately another year.

jan-janssen: https://devguide.python.org/versions/

pyiron_base/jobs/job/generic.py
pmrv: Probably doesn't make sense to copy it.  I guess we are mostly copying jobs to use them as reference for some master or move them to other projects.  In neither case we to "duplicate" a result so to speak.

jan-janssen: While it is technically no problem to copy the future object, we have to ask ourselves what would this mean. The future is only used while the job is executed, copying a job during execution sounds very dangerous to me. If we want to be extra careful we can check if a future exists and then raise a `ValueError`, unless the future is `done()`.

tests/job/test_genericJob.py
jan-janssen: Again, changing the `run_mode` during or after the execution, when a future object exists sounds very dangerous to me.

tests/job/test_genericJob.py
jan-janssen: From my perspective the future object is only reasonable during the execution, after the execution, when the status is set to finished, then the future is no longer required and it is recommended to just check the job status.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -415,6 +454,60 @@ def queue_view(self):
...
+    def executor(self, exe: Union[Executor, None]):
+        """
+        Executor to execute the job object this server object is attached to in the background.
+
+        Args:
+            exe (concurrent.futures.Executor):
+        """
+        if isinstance(exe, Executor):
+            self.run_mode.executor = True
+        elif exe is None:
```
jan-janssen:
```suggestion
        elif exe is None and self.run_mode.executor:
```

pyiron_base/jobs/job/extension/server/generic.py
pmrv: Leaving it orphaned might make the difficult for the code in `Project.{remove_job,wait_for_job}` to reliably check what it should it do.  Should be set to `None` therefor I think.  

Mid-term I agree there's no point in changing the run mode on a finished/running job.

pyiron_base/jobs/job/extension/server/queuestatus.py
pmrv: I guess one way to trigger it would be `job.run(run_again=True)`, which we have deprecated.  So agree: probably no immediate danger.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -415,6 +454,60 @@ def queue_view(self):
...
+            self.run_mode.modal = True
+        else:
```
jan-janssen:
```suggestion
        elif exe is not None:
```

pyiron_base/jobs/job/extension/server/generic.py
jan-janssen: I added an alternative suggestion above - only switch the `run_mode` to `modal` when the `run_mode` is still set to `executor` otherwise the `run_mode` is not modified.

pyiron_base/jobs/job/extension/server/queuestatus.py
```diff
@@ -226,7 +227,12 @@ def wait_for_job(job, interval_in_s=5, max_iterations=100):
...
+                    finished = True
+                    job.server.future.result(timeout=interval_in_s)
```
pmrv:
```suggestion
                    job.server.future.result(timeout=interval_in_s)
                    finished = job.server.future.done()
```
Something to that effect has to there, otherwise it will break the loop early, no?

Sorry, I wasn't more explicit earlier.  I was in a rush and you see where it got me.  It is the last thing that I see for now.

pyiron_base/jobs/job/generic.py
liamhuber: > If we want to be extra careful we can check if a future exists and then raise a `ValueError`, unless the future is `done()`.

👍 I like this

liamhuber: > The future is only used while the job is executed,

In this case, should we register a callback or otherwise clean up `server.future` when it's finished successfully?

tests/job/test_genericJob.py
liamhuber: If it's not supposed to be looked at anymore, then we should just get rid of it

pyiron_base/jobs/job/extension/server/generic.py
liamhuber: > I added an alternative suggestion above 

I don't see this, but that sounds fine too. It _should_ be formally equivalent, as any time `server.executor is not None` I expect `server.run_mode.executor`. Anything that stops a None-executor being set to more None from updating the run mode will satisfy me.

pyiron_base/jobs/job/extension/server/generic.py
liamhuber: > Should be set to None therefor I think.

This is what I lean towards

tests/job/test_genericJob.py
liamhuber: > Again, changing the `run_mode` during or after the execution, when a future object exists sounds very dangerous to me.

I totally agree, I wouldn't do it. But we let users do it. My point is that I would like a 1:1 relationship between `server.executor is not None` and `server.run_mode.executor` under all operations that the user can permissibly execute. The behaviour in `Server.executor` setter and getter do this now, so I just want to test that it stays that way.

pyiron_base/jobs/job/extension/server/generic.py
liamhuber: Ah, I see it. It was just "below" for me and not "above". Hard to read with all the layered "suggestion" bits, but what you describe here sounds fine and I trust that's what the suggestions result in.

pmrv (approved):

liamhuber:
> I'm happy to stack a PR that resolves my outstanding concerns.

Aaand one of my kids needs to be picked up from daycare sick. Probably won't be able to do this.

pyiron_base/jobs/job/generic.py
jan-janssen: The issue with the call back is we do not want the future to be removed when the job is finished but rather we want it to be called when the user recognized the job is finished, so I currently do not see a way how to achieve this with a call back and I would just keep the Future object active. We could even go one step further and if the job is finished, always set a future which is `done`.

niklassiemer (approved):
I am fine with this state of the new interface.

liamhuber:
> I agree that the development releases are the right way to move forward, but this functionality is not yet available, so for now I am to get his merged in the main branch.

Just for clarity, the development releases can target _any branch_ (correct me if I'm wrong about this, @srmnitc). So while it looks like we'll merge this by the end of the week anyhow, there is no actual need to merge it in order to dev release the functionality -- the dev release can target #1154 directly. This is the point of the dev-release workflow: to allow releasing functionality without feeling time-pressured to merge things into main.

jan-janssen:
> Just for clarity, the development releases can target _any branch_ (correct me if I'm wrong about this, @srmnitc). So while it looks like we'll merge this by the end of the week anyhow, there is no actual need to merge it in order to dev release the functionality -- the dev release can target #1154 directly. This is the point of the dev-release workflow: to allow releasing functionality without feeling time-pressured to merge things into main.

As far as I can see the dev release is not yet pushed to conda.

srmnitc:
> > I agree that the development releases are the right way to move forward, but this functionality is not yet available, so for now I am to get his merged in the main branch.
> 
> Just for clarity, the development releases can target _any branch_ (correct me if I'm wrong about this, @srmnitc). So while it looks like we'll merge this by the end of the week anyhow, there is no actual need to merge it in order to dev release the functionality -- the dev release can target #1154 directly. This is the point of the dev-release workflow: to allow releasing functionality without feeling time-pressured to merge things into main.

Yes correct!

srmnitc:
> > Just for clarity, the development releases can target _any branch_ (correct me if I'm wrong about this, @srmnitc). So while it looks like we'll merge this by the end of the week anyhow, there is no actual need to merge it in order to dev release the functionality -- the dev release can target #1154 directly. This is the point of the dev-release workflow: to allow releasing functionality without feeling time-pressured to merge things into main.
> 
> As far as I can see the dev release is not yet pushed to conda.

I was hoping some of you would add a review, since the tests pass, I will go ahead and merge.

srmnitc:
@jan-janssen @liamhuber Both base and atomistics have the dev release functionality now.

pyiron_base/jobs/job/generic.py
liamhuber: I see your point about the callback. Also even in the example the future is used post-facto. I'll add a suggestion for the check on copying.

pyiron_base/jobs/job/extension/server/generic.py
liamhuber: Even the example looks a the future post-facto, so I guess we can't kill it as soon as it's done. I have a vague sense of unease about this variable hanging around and causing trouble other places where assumptions are made about its existence/status, but since there are no concrete counter examples of how to break things (barring doing something really stupid like manually assigning some other random future to the attribute, but w/e, we don't need to police that), and since I don't have any alternative suggestion, I'm happy to just make a mental note of the potential pain point and move on.

pyiron_base/jobs/job/generic.py
liamhuber: There is some time pressure and there seems to be no appetite for this. It would be useful but I don't think it's strictly necessary.

pyiron_base/jobs/job/generic.py
```diff
@@ -466,6 +476,10 @@ def copy(self):
         _job_reload_after_copy(
             job=copied_self, delete_file_after_copy=delete_file_after_copy
         )
+
+        # Copy executor - it cannot be copied and is just linked instead
+        if self.server.executor is not None:
+            copied_self.server.executor = self.server.executor
```
liamhuber, line 482:
```suggestion
            copied_self.server.executor = self.server.executor
        if self.server.future is not None and not self.sever.future.done():
            raise RuntimeError("Jobs whose server has executor and future attributes cannot be copied unless the future is `done()`")
```

liamhuber: Probably breaks line length and will need re-blacking

tests/job/test_genericJob.py
```diff
@@ -4,6 +4,8 @@
 
 import unittest
 import os
+from time import sleep
+from concurrent.futures import ProcessPoolExecutor
```
liamhuber:
```suggestion
from concurrent.futures import Future, ProcessPoolExecutor
```

Necessary for copy test suggestion below

liamhuber:
I'm re-reviewing. Almost done. Please hold changes for a minute

tests/job/test_genericJob.py
liamhuber: Ok, we agree elsewhere it still needs to be around, but shouldn't copy if it's not done. That behaviour is implemented in another suggestion and tested here.

liamhuber (changes requested):
- I implemented your suggestion to not copy jobs with futures that aren't done and tested it
- I still want the job-independent tests moved out of the job tests and into the server tests

liamhuber:
> I'm re-reviewing. Almost done. Please hold changes for a minute

K super, thanks! EDIT: thanks for waiting. For future readers I'm not just thanking myself -- there was a timely thumbs up 😝 

All done. A couple more minor requests re testing. I still get a little nervous about potential weird states that `server.future` can get into given how it's used elsewhere, but since no one can see any explicit issues or an alternative suggestion, let's just go for it. Definitely I see your point that we don't want to just delete it immediately. Disallowing copying jobs with a non-`done()` future was a good suggestion though, so that's in and tested.

pyiron_base/jobs/job/generic.py
```diff
@@ -466,6 +476,14 @@ def copy(self):
...
+        if self.server.future is not None and not self.sever.future.done():
```
```suggestion
        if self.server.future is not None and not self.server.future.done():
```

jan-janssen:
@liamhuber I merged your changes plus fixes - so from my perspective it is ready to be merged now. What do you think?

liamhuber (approved):
I would still like the server-only tests moved to the lowest place they can go, but that's not approval-prohibiting.

🚀

jan-janssen:
> I would still like the server-only tests moved to the lowest place they can go, but that's not approval-prohibiting.

Done

Issue #1157, opened 2023-07-07
Forbid changing the job.server.run_mode after job.run


pmrv:
Over in #1155 we discussed that it's quite weird that we allow to change the run mode of a job after it has or while it is running.  Refreshed and full of energy I went this morning to fix this---only to discover why we don't disallow this yet. `Server` doesn't have a back reference to the job and can therefor not check its status in the `run_mode` setter. But actually you don't really want to change anything about the server object after the job is run, so alternatively we could add a `freeze` method to the server, that the job calls on `run`, after which no attribute modifications are allowed any more.  Special considerations would have to be taken for the `future` object of #1155, because that one we do want to change after the fact, but that's ok imo.  Thoughts?

jan-janssen:
I like the idea, in principle every object attached to the job should have this option, as well as an unfreeze option when somebody copies only the input of a given job to run it again.

pmrv:
I've started working on this in #1162.

PR #1162, opened 2023-07-10
Add an interface to turn objects read-only

pmrv:
As we discussed in #1157 it be nice to have some way of blocking users from changing values or calling methods after a certain point in time.  This is implemented here with some doc test examples.  For now it locks this recursively, but I have a vague fear that this might be not be super performant.  I'll have to do some tests.  I'll also move over at least `DataContainer` and `Server` in this PR.

@liamhuber Do you remember why `HasStoredTraits` implements its own `read_only` logic instead of reusing the existing one from `DataContainer`?


pmrv:
Current failure seems to be related to https://github.com/pyiron/pyiron_base/pull/1200

pmrv:
Functionality and documentation is done from my side.  There's some tests as this is now used by `DataContainer` and `Server`, but I can add additional specific tests after we've discussed whether we want this in a pyiron meeting.

liamhuber:
@pmrv, this hasn't escaped my notice, I just didn't manage to get this far today!

pyiron_base/interfaces/lockable.py
```diff
@@ -0,0 +1,347 @@
...
+        Raises:
+            ValueError: if `method` is not an allowed value
+        """
+        if method not in ["error", "warning", None]:
+            raise ValueError(f"Unrecognized lock method {method}!")
+        if method is not None:
+            object.__setattr__(self, "_lock_method", method)
+        self.read_only = True
+
+    def unlocked(self) -> _UnlockContext:
```
liamhuber, line 341: nit: I think this should be `unlock` to match the verb tense of `lock`

pyiron_base/interfaces/lockable.py
```diff
@@ -0,0 +1,347 @@
...
+        if changed:
+            self._read_only = value
+            if value:
+                self._on_lock()
+            else:
+                self._on_unlock()
+
+    def _on_lock(self):
+        for it in _iterate_lockable_subs(self):
+            it.lock()
```
liamhuber, line 315:
```suggestion
        pass
```

maybe (see main review comment)

pyiron_base/interfaces/lockable.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def _on_unlock(self):
+        for it in _iterate_lockable_subs(self):
+            it.read_only = False
```
liamhuber, line 319:
```suggestion
        pass
```

maybe (see main review comment)

pyiron_base/interfaces/lockable.py
```diff
@@ -0,0 +1,347 @@
...
+    subs = lockable_groups.list_all()
+    for n in subs["nodes"]:
+        node = lockable_groups[n]
+        if isinstance(node, Lockable):
+            yield node
+
+    for g in subs["groups"]:
+        group = lockable_groups[g]
+        yield group
+        yield from _iterate_lockable_subs(g)
```
liamhuber, line 122: pyiron_base-specific

liamhuber (commented):
Really nice, Marvin. For specific application here, I only have the one nit to refactor:rename `unlocked()` to `unlock()`. Otherwise lgtm, tests can be written, and I think this is useful in base.

Zooming out a bit though, I think this is an elegant solution to the very generic problem of transforming python attributes to being read-only. I think this would be an excellent candidate for the standard-library-only toolbox discussed in #1208. Here it would just be a matter of purging reference to `HasGroups`, which could then be re-introduced here in `pyiron_base` by defining `_iterate_lockable_subs` and a new class inheriting from the generic tool that overrides `_on_lock` and `_on_unlock`. Unlike `Singleton`, which is a breeze to re-implement as needed, what you have here is a bit more complex and might really be useful to the community as a standalone tool (or at least stand-with-other-lightweight-tools-in-a-nice-toolbox)

pyiron_base/interfaces/lockable.py
pmrv: I named it asymmetrical on purpose, because `lock` and `unlocked` are not symmetric either.

pmrv: What I wanted to avoid is that some code unlocks an object and then crashes (or forgets it) before it can relock it.  So `obj.unlocked()` is a context manager that does the relocking.

liamhuber: Clever. I approve. But I was too dense to pick up on it on the first read, so maybe a comment or docstring to clarify intent is good.

pyiron_base/interfaces/lockable.py
```diff
@@ -0,0 +1,347 @@
...
+    Traceback (most recent call last):
+        ...
+    lockable.Locked: Object is currently locked!  Use unlocked() if you know what you are doing.
+    >>> d['b']['c']
+    1
+
+    but we can unlock it as usual
+
+    >>> with d.unlocked() as dopen:
+    ...   dopen['b']['d'] = 23
```
pmrv:
```suggestion
    >>> with d.unlocked():
    ...   d['b']['d'] = 23
```

pyiron_base/interfaces/lockable.py
```diff
@@ -0,0 +1,347 @@
...
+            raise ValueError(f"Unrecognized lock method {method}!")
+        if method is not None:
+            object.__setattr__(self, "_lock_method", method)
+        self.read_only = True
+
+    def unlocked(self) -> _UnlockContext:
+        """
+        Unlock the object temporarily.
+
+        Context manager returns this object again.
```
pmrv:
```suggestion
        Context manager returns this object again and relocks it after the `with` statement finished.
        
        .. note:: `lock()` vs. `unlocked()`
        
            There is a small asymmetry between these two methods.  :meth:`.lock` can only be done once (meaningfully), while :meth:`.unlocked` is a context manager and can be called multiple times.
```

pmrv:
Integration tests are currently not passing, because base wants pint==0.23, but atomistics still requires <=0.22.  I assume this will solve itself when atomistics is updated, but will go ahead merge this here anyway.

jan-janssen:
> Integration tests are currently not passing, because base wants pint==0.23, but atomistics still requires <=0.22. I assume this will solve itself when atomistics is updated, but will go ahead merge this here anyway.

It seems like there was more wrong than just the `pint` dependency as now the tests fail on `pyiron_atomistics` - https://github.com/pyiron/pyiron_atomistics/pull/1265 - another indication that we should not merge broken code with `main`.

PR #1165, opened 2023-07-12
Bug fixes for flux

jan-janssen:


pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -476,6 +477,8 @@ def executor(self, exe: Union[Executor, None]):
         """
         if isinstance(exe, Executor):
             self.run_mode.executor = True
+        elif static_isinstance(exe, "flux.job.executor.FluxExecutor"):
+            self.run_mode.executor = True
```
jan-janssen, line 481: This part is no longer necessary once https://github.com/flux-framework/flux-core/pull/5323 is included in the next flux version.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -617,7 +617,7 @@ def run_job_with_runmode_executor_flux(job, executor, gpus_per_slot=None):
     )
     jobspec.cwd = job.project_hdf5.working_directory
     jobspec.environment = dict(os.environ)
-    return executor.submit(jobspec)
+    job.server.future = executor.submit(jobspec)
```
pmrv, line 620: I didn't follow everything in detail, but are we not doing the `fs = job.run()` anymore?  Is that taken of somewhere else or is it not needed anymore?

niklassiemer: Yes, we do not do it anymore. We decided to put the future next to the executor onto the server object. As such you cannot forget to assign the run call. Especially since we never did this before..

PR #1168, opened 2023-07-17
Fake out old tale.datamining for backwards compat

### LINKED ISSUE #1163: Pyiron tables: Backwards compatibility broken again by #796
By moving the modules around without providing backwards compatibility #796 broke the loading of pyiron tables created before that change.  That PR contained a [stop gap](https://github.com/pyiron/pyiron_base/blob/d2d9b1b49e1b2dde1305ce8b3a982e5244eb4ba5/pyiron_base/storage/hdfio.py#L1378) which remaps a few class paths, but the issue runs deeper for functions that are `dill`ed or `pickle`ed, because those libraries also write the import paths and we have no way of intercepting them.

The workaround is a finger of scotch and putting this bespoke line in your notebooks before you start loading the old tables.

```python
sys.modules['pyiron_base.table.datamining'] = sys.modules['pyiron_base.jobs.datamining']
```

pmrv:
Closes #1163.

### DIFF
--- a/pyiron_base/table/datamining.py
+++ b/pyiron_base/table/datamining.py
@@ -1,6 +1,8 @@
 from pyiron_base.jobs.datamining import *
 import warnings
+
 warnings.warning(
     "pyiron_base.table.datamining is deprecated, import from pyiron_base.job.datamining",
-    category=DeprecationWarning, stacklevel=1
+    category=DeprecationWarning,
+    stacklevel=1,
 )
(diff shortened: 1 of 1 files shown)


pmrv:
So it seems that if you define a class in one module, then re-export it in another and use it from there the `__module__` still points to the original module.  Which is good, because it means that the solution that we discussed in the meeting would work, i.e.

1. bump `HDF_VERSION` of `PyironTable`
2. in the `_from_hdf` implement the `sys.path` hack to be able to read the old function defs
3. write it back to save the new path

I'll keep this open until I get to do that.

PR #1180, opened 2023-08-22
Revert "Turn np.array_equal into proper asserts"

niklassiemer:
Reverts pyiron/pyiron_base#1147

since this broke the pyiron_atomistics compatibility tests:
```
ERROR: test_get_structure (atomic.job.test_atomistic.TestAtomisticGenericJob)
get_structure() should return structures with the exact values from the HDF files even if the size of
  File "/home/runner/work/pyiron_base/pyiron_atomistics/tests/atomic/job/test_atomistic.py", line 109, in test_get_structure
    self.assertTrue(np.allclose(job.output.cells[i], struct.cell.array))
  File "<__array_function__ internals>", line 200, in allclose
  File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/numpy/core/numeric.py", line 2270, in allclose
    res = all(isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan))
  File "<__array_function__ internals>", line 200, in isclose
  File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/numpy/core/numeric.py", line 2377, in isclose
    xfin = isfinite(x)
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

```

However, I am not sure on which side the actual problem is. With the (to be reverted) PR our hdfio should now put out what was put in, but for the `Atoms.cell` this seems to change the return type?!

### DIFF
--- a/pyiron_base/storage/hdfio.py
+++ b/pyiron_base/storage/hdfio.py
@@ -149,7 +149,10 @@ class FileHDFio(HasGroups, MutableMapping):
                 # underlying file once, this reduces the number of file opens in the most-likely case from 2 to 1 (1 to
                 # check whether the data is there and 1 to read it) and increases in the worst case from 1 to 2 (1 to
                 # try to read it here and one more time to verify it's not a group below).
-                return read_hdf5(self.file_name, title=self._get_h5_path(item))
+                obj = read_hdf5(self.file_name, title=self._get_h5_path(item))
+                if self._is_convertable_dtype_object_array(obj):
+                    obj = self._convert_dtype_obj_array(obj.copy())
+                return obj
             except (ValueError, OSError, RuntimeError, NotImplementedError):
                 # h5io couldn't find a dataset with name item, but there still might be a group with that name, which we
                 # check in the rest of the method
--- a/tests/generic/test_fileHDFio.py
+++ b/tests/generic/test_fileHDFio.py
@@ -182,14 +182,7 @@ class TestFileHDFio(PyironTestCase):
 
         with self.subTest("object_array_with_lists"):
             array = hdf["object_array_with_lists"]
-            self.assertEqual(len(array), len(object_array_with_lists),
-                             msg="object array read with incorrect length!")
-            for a_read, a_written in zip(array, object_array_with_lists):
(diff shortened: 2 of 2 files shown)


niklassiemer:
Impationed me already released the current main before the additional compatibility tests were performed... Sorry! Thus, we should fix this on either end relatively quickly and a) release a new version or provide a bug-fix path here if needed or b) release pyiron_atomistics with structure behaving as expected?

niklassiemer:
Just adding a few prints in the above test-function, there was a somewhat unexpected change with PR #1147
```
job.output.cells[0]
[[18.0 1.102182119232618e-15 1.102182119232618e-15]
 [0.0 18.0 1.102182119232618e-15]
 [0.0 0.0 18.0]]
struct.cell.array
[[1.80000000e+01 1.10218212e-15 1.10218212e-15]
 [0.00000000e+00 1.80000000e+01 1.10218212e-15]
 [0.00000000e+00 0.00000000e+00 1.80000000e+01]]
```

However before (or with this PR) we get
```
job.output.cells[0]
2023-08-23 14:54:43,854 - pyiron_log - WARNING - Deprecated data structure! Returned array was converted from dtype
='O' to dtype=float64 via `np.array(result.tolist())`.
Please run rewrite_hdf5() (from a job: job.project_hdf5.rewrite_hdf5() ) to update this data! To update all your da
ta run Project.maintenance.update.base_v0_3_to_v0_4('all').
[[1.80000000e+01 1.10218212e-15 1.10218212e-15]
 [0.00000000e+00 1.80000000e+01 1.10218212e-15]
 [0.00000000e+00 0.00000000e+00 1.80000000e+01]]
struct.cell.array
[[1.80000000e+01 1.10218212e-15 1.10218212e-15]
 [0.00000000e+00 1.80000000e+01 1.10218212e-15]
 [0.00000000e+00 0.00000000e+00 1.80000000e+01]]
```

niklassiemer:
Thus, it seems to me, that the problem is that we are testing on data with the `dtype=object` data structure we wrote accidentally some time back! I will skip the corresponding test for now and replace the testing data. Thus, I close this PR.

Issue #844, opened 2022-09-20
RFC: Bulk delete of jobs


pmrv:
We discussed a few times already that deleting a lot of jobs takes too much time.  Removing thousands of jobs can take a few hours.

Here's an idea of how to do it better:

1. get a list of jobs to be deleted
2. prun or enrich the list based on linked jobs (master, parent, etc.)
3. prun the list based on job status (e.g. running or submitted) 
4. make a backup of the database entries
5. delete database entries in bulk (doing in it sequentially already only takes a few tens of seconds for a ~1k jobs)
6. delete the working directories of each job, aborting once there's an error
7. if there's been an error restore the database entries of jobs that have been skipped.

Optionally we could stop after 2. & 3. to be cautious.

Some points why I picked this order
- 3. comes after 1. and 2. because jobs added in 2. might not be finished yet.
- 5. is done to avoid a potential race condition between multiple pyiron processes where a working directory is deleted already but the database is not updated yet.
- 4. & 7. are there to restore the database entries of jobs that have not been deleted on disk because of errors.

mgt16-LANL:
I'll comment here in support of speeding this up. From my end restarting from a set of bulk calculations, and using pyiron takes an absurd amount of time - it's taking ~3 minutes with delete_exisiting_jobs=True vs ~1 second to submit jobs if there is no existing job. I am finding it faster to prune out the jobs that didn't previously finish before restarting the pyiron workflow. Having a faster job removal can be key for having "resubmit-able" workflows.

jan-janssen:
I have a pull request for this open https://github.com/pyiron/pyiron_base/pull/1182  maybe you can check if it solves your issue, then we should move forward and include it in the next release.

pmrv:
I'm guessing #1182 and #1215 won't help much with the use case of something like

```python
for _ in ...:
  j = pr.create.job.MyJob(..., delete_existing_job=True)
  j.run()
```

since it will still touch every job sequentially.

pmrv:
@mgt16-LANL Can you run your workflow (re-)submission under `cProfile` and send the resulting profile here?

```python
import cProfile
prof = cProfile.Profile()
prof.enable()

... # your workflow

prof.disable()
prof.dump_stats('myprofile.pyprof')
```

mgt16-LANL:
> I have a pull request for this open #1182 maybe you can check if it solves your issue, then we should move forward and include it in the next release.

[...]

PR #1215, opened 2023-10-21
WIP: Implement #844 (bulk delete jobs)

pmrv:


pmrv:
This will need more tests for master jobs and such, but in the current state this already allows to delete ~5k jobs in ~45s (with a central database).  To address #844 fully, we'll need to think about changes to the file table, but generally I think the direction is right.

jan-janssen:
I converted this pull request to `draft` as the unit tests are currently failing.

PR #1222, opened 2023-11-07
Fix iterjobs() without database

### LINKED ISSUE #1221: `iter_jobs()` fails without database
```python

pr = Project('test')
pr.state.update({"disable_database": True})

pr_2 = pr.open("sub_1")
job_1 = pr_2.create.job.Lammps("lmp1")
job_1.structure = pr_2.create.structure.ase.bulk("Al")
job_1.run()

pr_3 = pr.open("sub_2")
job_1 = pr_3.create.job.Lammps("lmp2")
job_1.structure = pr_3.create.structure.ase.bulk("Al")
job_1.run()

try:
    for job in pr.iter_jobs(status="finished"):
        print(job.job_name)
finally:
    pr.remove(enable=True)
```
This fails with the following error: 
```
ValueError: The job with ID 2 is not listed in the database anymore.
```

Work around: 
```
for job in pr.iter_jobs(status="finished", convert_to_object=False):
    print(job.project)
    print(job.project.load(job.job_name))
```

EDIT @liamhuber: disable database and cleanup in example

jan-janssen:


pyiron_base/project/generic.py
```diff
@@ -581,14 +581,41 @@ def iter_jobs(
...
+            db_entry_lst = [row[1].to_dict() for row in self.job_table().iterrows()]
+            table_columns = [
+                "job",
+                "subjob",
+                "projectpath",
+                "project",
+                "status",
+                "hamilton",
+                "hamversion",
+            ]
```
liamhuber, line 600: I notice this is only a subset of the columns mentioned in [`pyiron_base.database.filetable.table_columns`](https://github.com/pyiron/pyiron_base/blob/59d0852ddba64490535de203fa3cf99a9fba13b6/pyiron_base/database/filetable.py#L30) and in the [full column list for the job table](https://github.com/pyiron/pyiron_base/blob/59d0852ddba64490535de203fa3cf99a9fba13b6/pyiron_base/database/generic.py#L178). Will this truncation possibly get us into trouble? If the truncation is intentional it might be helpful to future maintainers to add a short comment on the intent of the truncation.

jan-janssen: We do not want to use the `id`, `masterid` or `parentid` as these can lead to inconsistencies when using pyiron without a database. In `pyiron_base.jobs.job.path` we define the `from_db_entry()` function: 
```
    @classmethod
    def from_db_entry(cls, db_entry):
        """
        Load a job path from a database entry.

        Args:
            db_entry (dict): database entry {"job":, "subjob":, "projectpath":, "project":, "hamilton":, "hamversion":,
                                            "status":} and optional entries are {"id":, "masterid":, "parentid":}
        """
        hdf5_file = db_entry["subjob"].split("/")[1] + ".h5"
        job_path = db_entry["projectpath"]
        if job_path is None:
            job_path = ""
        job_path += db_entry["project"] + hdf5_file + db_entry["subjob"]
        job = cls(job_path=job_path)

        if "hamilton" in db_entry.keys():
            job.__name__ = db_entry["hamilton"]
        if "hamversion" in db_entry.keys():
            job.__version__ = db_entry["hamversion"]

        if "id" in db_entry.keys():
            job._job_id = db_entry["id"]
        if "status" in db_entry.keys():
            job._status = db_entry["status"]
        if "masterid" in db_entry.keys():
            job._master_id = db_entry["masterid"]
        if "parentid" in db_entry.keys():
            job._parent_id = db_entry["parentid"]
        return job
```
So I selected all columns which can be used by this function and are not related to the job id.

liamhuber: That sounds perfectly reasonable to me! In the code the list just appears out of thin air, and I really think it would help future developers to have a truncated version of that comment appear next to the list defined here.

jan-janssen: I added a short comment.

jan-janssen:
@liamhuber It would be great if you can take another look at this pull request. It took far too long on my side but I definitely want to merge it soon, to move on.

pyiron_base/project/generic.py
```diff
@@ -581,14 +581,43 @@ def iter_jobs(
             case, you may seriously wish to consider setting `convert_to_object=False` and access only the HDF5/JobCore
             representation of the jobs instead.
         """
-        job_id_lst = self.job_table(recursive=recursive, **kwargs)["id"]
+        job_table = self.job_table(recursive=recursive, **kwargs)
+        if not isinstance(self.db, FileTable):
+            job_lst = [[job_id, None] for job_id in job_table["id"]]
+        else:
```
liamhuber, line 587: Nit: it is a bit of a smell to `if not isinstance(); else` instead of `if isinstance(); else` as it increases cognitive load to mentally negate, but I also see the argument for putting the "default" case at the top so I want to point it out but am comfortable either way.

liamhuber (commented):
I appreciate the comment above the magic-value list, I think that clarifies it perfectly.

Taking a deeper dive, there are a few things I still don't like about this PR:

1) It is made to patch a bug, but there is no change to the test suite. I can't tell it's fixing the bug without manually running the example in #1221, and without a test we're not guaranteeing that this new and improved behaviour will stick around.

2) The high-level `inspect` and `load` calls get replaced with the low-level `load_from_jobpath` call. It didn't take much digging to see why -- `load` and `inspect` don't accept `db_entry`, but these high-level calls were created for a reason: to more clearly signify the intent of the action. Instead, let's empower them to also handle a `db_entry` instead of a job specifier.

3) By forcing `convert_to_object=True` in line 619, we are changing behaviour. This actually invalidates the signature of `iter_jobs` as `convert_to_object` becomes totally disused. I see the comment a couple lines above that we will deprecate `convert_to_object`...but deprecate it to what? Always false or always true? Now we're deprecating the variable, but the comment sticks around?? I don't know what the intended future behaviour is, but without clarifying that and updating the "backwards compatibility" comment, I'm not comfortable changing the behaviour.


I'm stacking a patching PR now that addresses all three of these concerns. EDIT: #1228

pyiron_base/project/generic.py
```diff
@@ -581,14 +581,43 @@ def iter_jobs(
...
+                    job_id=job_id,
+                    db_entry=db_entry,
+                    convert_to_object=False,
+                )[path]
             else:  # Backwards compatibility - in future the option convert_to_object should be removed
-                yield self.load(job_id, convert_to_object=convert_to_object)
+                yield self.load_from_jobpath(
+                    job_id=job_id,
+                    db_entry=db_entry,
+                    convert_to_object=True,
```
jan-janssen:
```suggestion
                    convert_to_object=convert_to_object,
```

jan-janssen:
> 1. It is made to patch a bug, but there is no change to the test suite. I can't tell it's fixing the bug without manually running the example in [`iter_jobs()` fails without database  #1221](https://github.com/pyiron/pyiron_base/issues/1221), and without a test we're not guaranteeing that this new and improved behaviour will stick around.

I agree - thanks for adding the test. 

> 2. The high-level `inspect` and `load` calls get replaced with the low-level `load_from_jobpath` call. It didn't take much digging to see why -- `load` and `inspect` don't accept `db_entry`, but these high-level calls were created for a reason: to more clearly signify the intent of the action. Instead, let's empower them to also handle a `db_entry` instead of a job specifier.

From my perspective there are two levels of APIs in pyiron, the user interface which we use interactively and the internal interface the different pyiron objects use to interact with each other. I do not expect a user to write the `db_entry` in a dictionary to load an object, but for an internal function this is usually easy. Consequently, I do not think the internal calls should be forced to use the same interface as the users. 

> 3. By forcing `convert_to_object=True` in line 619, we are changing behaviour. This actually invalidates the signature of `iter_jobs` as `convert_to_object` becomes totally disused. I see the comment a couple lines above that we will deprecate `convert_to_object`...but deprecate it to what? Always false or always true? Now we're deprecating the variable, but the comment sticks around?? I don't know what the intended future behaviour is, but without clarifying that and updating the "backwards compatibility" comment, I'm not comfortable changing the behaviour.

This was a bug - that is fixed now.

liamhuber (approved):

PR #1228, opened 2023-11-21
Use inspect and load

liamhuber:
Revert to using the high-level interface for inspecting/loading, which also brings us back to the original signature for `Project.iter_jobs` that includes `convert_to_object`. This involved extending the inspector/loader to handle `db_entry` input instead of only a job identifier.

Also added the example from #1221 as a test.

### DIFF
--- a/pyiron_base/project/jobloader.py
+++ b/pyiron_base/project/jobloader.py
@@ -53,11 +53,29 @@ class _JobByAttribute(ABC):
     def __getitem__(self, item):
         return self.__getattr__(item)
 
-    def __call__(self, job_specifier, convert_to_object=None):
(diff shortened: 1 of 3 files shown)


pyiron_base/project/jobloader.py
```diff
@@ -53,11 +53,29 @@ def __getattr__(self, item):
     def __getitem__(self, item):
         return self.__getattr__(item)
 
-    def __call__(self, job_specifier, convert_to_object=None):
+    def __call__(self, job_specifier=None, db_entry=None, convert_to_object=None):
```
jan-janssen, line 56: From my perspective the `jobloader` is a user facing module and therefore should not use the `db_entry`, because I do not expect a user to define the dictionary of database entries.

PR #1234, opened 2023-11-30
Add a function to remap module paths for stored objects

pmrv:
Previously HDF storage would break every time we renamed a module.  This also happens for pickled or dilled objects.  This change introduces a mapping between old and new module paths and monkey patches sys.modules on the fly so that both ProjectHDFio.to_object and pickle.load automatically find the new modules.  In both cases do the new objects point to the correct new module paths, so that writing or pickling them points to the correct thing.  ProjectHDFio.to_object also updates the TYPE field automatically after loading an object.

### DIFF
--- a/pyiron_base/storage/hdfio.py
+++ b/pyiron_base/storage/hdfio.py
@@ -36,6 +36,28 @@ __status__ = "production"
 __date__ = "Sep 1, 2017"
 
 
+# we sometimes move classes between modules; this would break HDF storage,
+# since objects save there the module path from which their classes can be
+# imported.  We can work around this by defining here an explicit map that
+# _to_object can use to find the new modules and update the HDF5 files
+_MODULE_CONVERSION_DICT = {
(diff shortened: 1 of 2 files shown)


pmrv:
Supersedes #1168, fixes #1163.

pyiron_base/storage/hdfio.py
```diff
@@ -119,22 +165,18 @@ def _to_object(hdf, class_name=None, **kwargs):
         raise ValueError(
             "Object type in hdf5-file must be identical to input parameter"
         )
+
+    # add moved modules to sys path so our import mechanism and pickle can find them
```
pmrv:
```suggestion
    # add moved modules to sys.modules so our import mechanism and pickle can find them
```

pmrv:
- move all of this to a module updater 
- also add a pyiron table updater for the re-pickling

pmrv:
Superseded by #1243

PR #1235, opened 2023-11-30
Use proper monotonic clock for benchmark

pmrv:


liamhuber (approved):
Looks good. 

Did you ctrl+f for time throughout? The only improvement to this pr that I can think of is just to make sure we make the same good change everywhere possible

pmrv:
> Looks good.
> 
> Did you ctrl+f for time throughout? The only improvement to this pr that I can think of is just to make sure we make the same good change everywhere possible

Ah, I had not done that.  I'll give it a check in the office tomorrow.

pmrv:
> > Looks good.
> > Did you ctrl+f for time throughout? The only improvement to this pr that I can think of is just to make sure we make the same good change everywhere possible
> 
> Ah, I had not done that. I'll give it a check in the office tomorrow.

Seems that this is the only occurrence.  The benchmark tests use `timeit`.

PR #1242, opened 2023-12-04
Remove blosc2 dependency as it is now included in pytables

jan-janssen:


jan-janssen:
Should be fixed with https://github.com/conda-forge/pytables-feedstock/pull/96 .

PR #1243, opened 2023-12-05
Add maintenance function to update TYPE in HDF5 files

pmrv:
Problem as discussed in #1234.

This adds `pyiron_base.project.maintenance.add_module_conversion` that takes to module paths and saves them for updates.  Upon calling the corresponding maintenance function HDF5 files (jobs and project.data) are scanned and the TYPE field is updated, substituting the old for the new module path.  When an object is loaded and the import fails while its module path is known to have been updated a readable error message is raised directing users to use the maintenance function.

In the future *whenever* classes are moved between packages and modules a corresponding call to `add_module_conversion` has to be inserted into the `__init__` of the package where those classes were previously defined.


pyiron_base/project/maintenance.py
```diff
@@ -7,9 +7,41 @@
...
+# _to_object can use to find the new modules and update the HDF5 files
+_MODULE_CONVERSION_DICT = {
+    "pyiron_base.generic.datacontainer": "pyiron_base.storage.datacontainer",
+    "pyiron_base.generic.inputlist": "pyiron_base.storage.inputlist",
+    "pyiron_base.generic.flattenedstorage": "pyiron_base.storage.flattenedstorage",
+    "pyiron_base.table.datamining": "pyiron_base.jobs.datamining",
+}
+
+
+def add_module_conversion_path(old: str, new: str):
```
pmrv:
```suggestion
def add_module_conversion(old: str, new: str):
```

pyiron_base/project/maintenance.py
```diff
@@ -7,9 +7,41 @@
...
+        old (str): path to module that previously defined objects in storage
+        new (str): path to module that should be imported instead
+    Raises:
+        ValueError: if an entry for `old` already exists and does not point to `new`.
+    """
+    if old not in _MODULE_CONVERSION_DICT:
+        _MODULE_CONVERSION_DICT[old] = new
+    elif _MODULE_CONVERSION_DICT[old] != new:
+        raise ValueError(
+            f"Module path '{old}' already found in conversion dict, pointing to '{new}'!"
```
pmrv:
```suggestion
            f"Module path '{old}' already found in conversion dict, pointing to '{_MODULE_CONVERSION_DICT[old]}'!"
```

pmrv:
Reminder to myself: forgot to add new test file.

pyiron_base/project/maintenance.py
```diff
@@ -2,14 +2,48 @@
...
+    "pyiron_base.generic.inputlist": "pyiron_base.storage.inputlist",
+    "pyiron_base.generic.flattenedstorage": "pyiron_base.storage.flattenedstorage",
+    "pyiron_base.table.datamining": "pyiron_base.jobs.datamining",
+}
+
+
+def add_module_conversion(old: str, new: str):
+    """
+    Add a new module conversion.
+    After this call any object that was previously defined in module `old` and is now defined in `new`, will be
```
liamhuber: This is not strictly true, or? It looks to me like you need to invoke `add_module_conversion` then still do a maintenance cycle.

pyiron_base/project/maintenance.py
```diff
@@ -102,6 +139,67 @@ def defragment_storage(
...
             hdf.rewrite_hdf5(job.name)
 
+    def update_hdf_types(
+        self,
+        recursive: bool = True,
+        progress: bool = True,
+        **kwargs: dict,
+    ):
+        """
+        Rewrite TYPE fields in hdf5 files for renamed modules.
```
liamhuber, line 150: Should `_MODULE_CONVERSION_DICT` be an attribute of the maintenance class? It feels like here we should direct the reader to where these "renamed modules" are, and that hint makes more sense to me if they also just live on this class.

pyiron_base/storage/hdfio.py
```diff
@@ -120,23 +136,16 @@ def _to_object(hdf, class_name=None, **kwargs):
...
-    }
-    if class_path in class_convert_dict.keys():
-        class_name_new = "<class '" + class_convert_dict[class_path] + "'>"
-        class_object = _import_class(class_name_new)
-    elif not class_path.startswith("abc."):
-        class_object = _import_class(class_name)
+    type_field = class_name or hdf.get("TYPE")
+    module_path, class_name = _extract_module_class_name(type_field)
+
+    if not module_path.startswith("abc."):
```
liamhuber, line 144: We talked about this briefly as I had concerns that this was about dynamically defined classes but the `abc` there can be somehow overloaded. However, I don't remember why it's _actually_ here. It's also not clear to me from reading what the intent is. Can you add a `#comment` about why we want to do this, i.e. what case does it cover?

liamhuber (approved):
Nice. I have some questions/suggestions, but they're "accept with minor revision not requiring re-review" 😜 I also assume you'll get the integration tests passing, but I see an open PR over on pyiron_atomistics on this topic, so I am confident you're already on top of this 👍 

One thing on my wishlist is that we actually have an old project that we test the updates on? Maybe we have this and I'm just not aware? If such a thing is indeed missing, I can accept that it's outside scope for this PR.

pyiron_base/project/maintenance.py
pmrv: Correct, I had just copied this over from #1234, my bad.

pyiron_base/project/maintenance.py
pmrv: I had it as global, because I didn't want someone to accidentally mess with it.  Definitely this docstring should reference `add_module_conversion`.  If `_MODULE_CONVERSION_DICT` becomes an attribute, `add_module_conversion` would have to become a method.  Feels a bit weird, because the other methods are all for actual fixes, whereas that one only sets some fixes up, but at least it would be all in one location.  Dunno, will mull over it on the train tomorrow.

pyiron_base/storage/hdfio.py
pmrv: Will add a comment to the following effect.  From my understanding, this is because [dynamic jobs](https://github.com/pyiron/pyiron_base/blob/e6cca773c11bfade35b445b76f83d8225552f0c7/pyiron_base/jobs/dynamic.py#L132) (which were patched in quickly about this time last year, if I remember correctly).  One can add scripts into a magic location in the resources and then the job factory creates job classes dynamically (one class for each script).  Apparently when writing them to HDF, nothing is handled and so TYPE ends up being the `abc.` path, that means when reading them we have to guess that all objects starting with `abc.` are dynamic jobs and look this up in the `JOB_DYN_DICT`.

Afaik, this is nowhere documented in more detail, or did I miss this somewhere @jan-janssen ?

tests/storage/test_module_path_maintenance.py
```diff
@@ -0,0 +1,62 @@
...
+        hdf = self.project.create_hdf(self.project.path, "project_data")["../data"]
+        # manipulate type to fake an old module
+        old_path_project_data = "project.data.module"
+        hdf["test_data__index_0/TYPE"] = f"<class '{old_path_project_data}.DataContainer'>"
+
+        job = self.project.create_job(ToyJob, "test_job")
+        job.run()
+
+        # manipulate type to fake an old module
+        old_path_job = "job.module"
```
liamhuber, line 33: Beautiful. Even my stretch goal for this PR is satisfied 😂 🚀

pyiron_base/project/maintenance.py
pmrv: I've left it as is for now, but added comments.  Having it as a global has the advantage that we only need to maintain one such dictionary, instead of carrying it around on each project, which might make it easy to loose it when we define the additional Project sub classes.

pmrv:
From my side this is done and I'll merge it tonight and bump the version if there's no other comments.

@jan-janssen Is it realistic for you update the PR on contrib until Monday? 

https://github.com/pyiron/pyiron_contrib/pull/887

pmrv:
Compatibility tests are failing because of some mamba issues.

PR #1244, opened 2023-12-07
add tutorial on user defined jobs

samwaseda:
In the PMD Hackathon, we decided to make an example for the user defined jobs, using `TemplateJob` and `PythonTemplateJob`, doing linear fitting via Gnuplot and numpy. I thought it might be a nice one for external users if they want to implement their jobs, so I upload it here.


samwaseda:
ok maybe in the end it wasn't such a nice example because we would have to include Gnuplot in the environment only for this

jan-janssen:
@samwaseda I guess we have to make a decision at some point, either merge this pull request or close it. I am fine either way, I just would prefer to have it off the list of open pull requests.

samwaseda:
ah right. I'm not really sure if we want to include gnuplot or not, because it takes forever to download it...

jan-janssen:
> ah right. I'm not really sure if we want to include gnuplot or not, because it takes forever to download it...

Ok, then we close the pull request. We can still restore the branch later, or go back to this pull request.

PR #1247, opened 2023-12-16
HDF5 interface: add write_dict_to_hdf() and read_dict_from_hdf()

jan-janssen:
This pull request refactors the helper functions in the `pyiron_base.storage` module and adds the `write_dict_to_hdf()` and `read_dict_from_hdf()` functions. The `write_dict_to_hdf()` function takes a dictionary where the keys represent the path in the HDF5 file and the values represent the objects which should be stored in these nodes. The corresponding HDF5 groups are created automatically. The `read_dict_from_hdf()` function reads the nodes from a given HDF5 group and subgroups can be included using the optional `group_paths` parameter. The function returns a hierarchical dictionary which can be directly converted to a `DataContainer`.

### DIFF
--- a/pyiron_base/storage/helper_functions.py
+++ b/pyiron_base/storage/helper_functions.py
@@ -205,7 +205,7 @@ def read_dict_from_hdf(file_name, h5_path, group_paths=[], slash="ignore"):
     Args:
        hdf (pyiron_base.storage.hdfio.FileHDFio): HDF5 file object
        file_name (str): Name of the file on disk
-       h5_path (str): Path to a group in the HDF5 file where the data is stored
+       h5_path (str): Path to a group in the HDF5 file from where the data is read
        group_paths (list): list of additional groups to be included in the dictionary, for example:
                            ["input", "output", "output/generic"]
                            These groups are defined relative to the h5_path.
--- a/pyiron_base/storage/helper_functions.py
+++ b/pyiron_base/storage/helper_functions.py
@@ -72,19 +72,12 @@ def write_hdf5_with_json_support(value, path, file_handle):
         ) from None
 
 
-def write_dict_to_hdf(hdf, data_dict, groups=[]):
+def write_dict_to_hdf(hdf, data_dict):
     with open_hdf5(hdf.file_name, mode="a") as store:
         for k, v in data_dict.items():
-            if k not in groups:
-                write_hdf5_with_json_support(
-                    file_handle=store, value=v, path=hdf.get_h5_path(k)
-                )
-        for group in groups:
-            hdf_group = hdf.create_group(group)
-            for k, v in data_dict[group].items():
-                write_hdf5_with_json_support(
-                    file_handle=store, value=v, path=hdf_group.get_h5_path(k)
-                )
+            write_hdf5_with_json_support(
+                file_handle=store, value=v, path=hdf.get_h5_path(k)
+            )
 
 
 def _check_json_conversion(value):
--- a/pyiron_base/storage/helper_functions.py
+++ b/pyiron_base/storage/helper_functions.py
@@ -39,8 +39,7 @@ def read_hdf5(fname, title, slash="ignore"):
     Args:
         fname (str): Name of the file on disk, or file-like object.  Note: for files created with the 'core' driver,
                      HDF5 still requires this be non-empty.
-        title (str): The top-level directory name to use. Typically it is useful to make this your package name,
-                     e.g. ``'mnepython'``.
+        title (str): the HDF5 internal dataset path from which should be read, slashes indicate sub groups
         slash (str): 'ignore' | 'replace' Whether to replace the string {FWDSLASH} with the value /. This does
                      not apply to the top level name (title). If 'ignore', nothing will be replaced.
 
--- a/pyiron_base/storage/helper_functions.py
+++ b/pyiron_base/storage/helper_functions.py
@@ -223,15 +223,11 @@ def read_dict_from_hdf(file_name, h5_path, group_paths=[], slash="ignore"):
         }
 
     def resolve_nested_dict(group_path, data_dict):
-        group_lst = group_path.split("/")
-        if len(group_lst) > 1:
-            return {
-                group_lst[0]: resolve_nested_dict(
-                    group_path="/".join(group_lst[1:]), data_dict=data_dict
-                )
-            }
-        else:
-            return {group_lst[0]: data_dict}
+        groups = group_path.split("/")
+        nested_dict = data_dict
+        for g in groups[::-1]:
+            nested_dict = {g: nested_dict}
+        return nested_dict
 
     with open_hdf5(file_name, mode="r") as store:
         output_dict = get_dict_from_nodes(store=store, h5_path=h5_path, slash=slash)
--- a/pyiron_base/storage/helper_functions.py
+++ b/pyiron_base/storage/helper_functions.py
@@ -266,7 +266,7 @@ def get_h5_path(h5_path, name):
 
 def _check_json_conversion(value):
     """
-    Check if the object can be converted to JSON to optimize the HDF5 performance. This can change the data tupe of the
+    Check if the object can be converted to JSON to optimize the HDF5 performance. This can change the data type of the
     object which is going to be stored in the HDF5 file.
 
     Args:
--- a/pyiron_base/storage/helper_functions.py
+++ b/pyiron_base/storage/helper_functions.py
@@ -198,7 +198,7 @@ def list_groups_and_nodes(hdf, h5_path):
     return list(groups), list(nodes)
 
 
-def read_dict_from_hdf5(file_name, h5_path, group_paths=[], slash="ignore"):
+def read_dict_from_hdf(file_name, h5_path, group_paths=[], slash="ignore"):
     """
     Read data from HDF5 file into a dictionary - by default only the nodes are converted to dictionaries, additional
     groups can be specified using the group_paths parameter.
--- a/pyiron_base/storage/helper_functions.py
+++ b/pyiron_base/storage/helper_functions.py
@@ -193,7 +193,6 @@ def list_groups_and_nodes(hdf, h5_path):
                 nodes.add(k)
     except KeyError:
         pass
-    print("list_groups_and_nodes():", h5_path, groups, nodes)
     return list(groups), list(nodes)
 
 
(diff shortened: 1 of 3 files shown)


jan-janssen:
@pmrv As discussed in the pyiron meeting, I plan to go ahead and merge this so we can focus on defining the interface in a separate pull request.

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +53,81 @@ def write_hdf5(
...
+                )
+        for group in groups:
+            hdf_group = hdf.create_group(group)
+            for k, v in data_dict[group].items():
+                write_hdf5_with_json_support(
+                    file_handle=store, value=v, path=hdf_group.get_h5_path(k)
+                )
+
+
+def _check_json_conversion(value):
```
pmrv, line 272: Docstring should mention that `value` is potentially transformed.

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +53,81 @@ def write_hdf5(
...
+            if k not in groups:
+                write_hdf5_with_json_support(
+                    file_handle=store, value=v, path=hdf.get_h5_path(k)
+                )
+        for group in groups:
+            hdf_group = hdf.create_group(group)
+            for k, v in data_dict[group].items():
+                write_hdf5_with_json_support(
+                    file_handle=store, value=v, path=hdf_group.get_h5_path(k)
+                )
```
pmrv: I don't understand the purpose of `groups` or how it works.  `groups` only seems to refer top level sub dicts of `data_dict`, what about deeper nested ones?  At least clarify expected structure of `data_dict` in docstring.

pmrv (changes requested):

pyiron_base/storage/helper_functions.py
jan-janssen: The issue is that the dictionary we get from the job is not stored in the HDF5 file as a single entry but rather can contain certain sub groups which we want to represent as individual groups in the HDF5 file. So far I introduced a single level of sub groups but you are right we can have an arbitrary level of hierarchy and the developers should be able to decide which entries they want to store individually so they can be reloaded more quickly. Or in terms of the HDF5 file which entries should be compressed as one JSON dictionary and which other entries are resolved as individual nodes on the HDF5 level.

jan-janssen: I thought about this a bit more and I think the best way to address this is to build a flat dictionary, where the keys represent the h5_path, something like this:
```
data_dict = {
    'job_name/input/control_inp/data_dict': {}, 
    'job_name/output/generic/energy_tot': [...],
    'job_name/output/generic/energy_pot': [...],
}
```

jan-janssen:
@pmrv I fixed the remarks and would like to merge this. So if you have time to take another look that would be great.

jan-janssen:
While I made a couple more changes to add the `read_dict_from_hdf()` function - still the pull request is now again ready for review.

pyiron_base/storage/helper_functions.py
```diff
@@ -1,21 +1,52 @@
...
+                    a        Read/write if exists, create otherwise
+        swmr (bool): Open the file in SWMR read mode. Only used when mode = 'r'.
+
+    Returns:
+        h5py.File: open HDF5 file object
+    """
     if swmr and mode != "r":
-        store = h5py.File(filename, mode=mode, libver="latest")
+        store = h5py.File(name=filename, mode=mode, libver="latest")
         store.swmr = True
```
pmrv:
```suggestion
```
According to docstring.  I checked the [HDF5 docs](https://docs.h5py.org/en/stable/swmr.html) and apparently the correct flags for a writer is `swmr_mode` instead of just `swmr`, so I'm guessing this statement never had an effect.  Anyway the docs say one cannot create new groups or datasets in SWMR writer mode, so we probably never want to enable it in the first place.

pyiron_base/storage/helper_functions.py
```diff
@@ -1,8 +1,52 @@
...
+    """
+    if swmr and mode != "r":
+        store = h5py.File(name=filename, mode=mode, libver="latest")
+        store.swmr = True
+        return store
+    else:
+        return h5py.File(name=filename, mode=mode, libver="latest", swmr=swmr)
 
 
 def read_hdf5(fname, title="h5io", slash="ignore"):
```
pmrv:
```suggestion
def read_hdf5(fname, title, slash="ignore"):
```
not originally from this PR, but I've just noticed this.

pyiron_base/storage/helper_functions.py
```diff
@@ -1,8 +1,52 @@
...
+    """
+    Read data from HDF5 file
+
+    Args:
+        fname (str): Name of the file on disk, or file-like object.  Note: for files created with the 'core' driver,
+                     HDF5 still requires this be non-empty.
+        title (str): The top-level directory name to use. Typically it is useful to make this your package name,
+                     e.g. ``'mnepython'``.
```
pmrv:
```suggestion
        title (str): the HDF5 internal dataset path from which should be read, slashes indicate sub groups
```
Original descriptions feels as if it came from h5io documentation.  For our functions we should document our typical usage of the arguments.

pmrv: Same applies for `write_hdf5` docstring.

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +104,218 @@ def write_hdf5(
...
+        ) from None
+
+
+def write_dict_to_hdf(file_name, h5_path, data_dict, compression=4, slash="error"):
+    """
+    Write dictionary to HDF5 file
+
+    Args:
+        file_name (str): Name of the file on disk
+        h5_path (str): Path to a group in the HDF5 file where the data_dict is going to be stored
```
pmrv:
```suggestion
        h5_path (str): Path to a group in the HDF5 file where the data_dict is going to be stored; all entries of `data_dict` will be stored beneath it.
```

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +104,218 @@ def write_hdf5(
...
+
+def read_dict_from_hdf(file_name, h5_path, group_paths=[], slash="ignore"):
+    """
+    Read data from HDF5 file into a dictionary - by default only the nodes are converted to dictionaries, additional
+    groups can be specified using the group_paths parameter.
+
+    Args:
+       hdf (pyiron_base.storage.hdfio.FileHDFio): HDF5 file object
+       file_name (str): Name of the file on disk
+       h5_path (str): Path to a group in the HDF5 file where the data is stored
```
pmrv:
```suggestion
       h5_path (str): Path to a group in the HDF5 file from where the data is read
```

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +104,218 @@ def write_hdf5(
...
+            )
+            for n in list_groups_and_nodes(hdf=store, h5_path=h5_path)[1]
+        }
+
+    def resolve_nested_dict(group_path, data_dict):
+        group_lst = group_path.split("/")
+        if len(group_lst) > 1:
+            return {
+                group_lst[0]: resolve_nested_dict(
+                    group_path="/".join(group_lst[1:]), data_dict=data_dict
```
pmrv:
```suggestion
        group_lst = group_path.split("/", maxsplit=1)
        if len(group_lst) > 1:
            return {
                group_lst[0]: resolve_nested_dict(
                    group_path=group_lst[1], data_dict=data_dict
```

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +104,218 @@ def write_hdf5(
...
+
+    def get_dict_from_nodes(store, h5_path, slash="ignore"):
+        return {
+            n: read_hdf5(
+                fname=store, title=get_h5_path(h5_path=h5_path, name=n), slash=slash
+            )
+            for n in list_groups_and_nodes(hdf=store, h5_path=h5_path)[1]
+        }
+
+    def resolve_nested_dict(group_path, data_dict):
```
pmrv, line 225:
```suggestion
    def resolve_nested_dict(group_path, data_dict):
        """Turns a dict with a key containing slashes into a nested dict.  {'/a/b/c': 1} -> {'a': {'b': {'c': 1}"""
```

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +104,218 @@ def write_hdf5(
...
+                nodes.add(k)
+    except KeyError:
+        pass
+    return list(groups), list(nodes)
+
+
+def read_dict_from_hdf(file_name, h5_path, group_paths=[], slash="ignore"):
+    """
+    Read data from HDF5 file into a dictionary - by default only the nodes are converted to dictionaries, additional
+    groups can be specified using the group_paths parameter.
```
pmrv:
```suggestion
    sub groups can be specified using the group_paths parameter.
```

pyiron_base/storage/helper_functions.py
pmrv: Actually maybe the whole body can be just
```python
groups = group_path.split("/")
nested_dict = data_dict
for g in groups[::-1]:
  nested_dict = {g: nested_dict}
```
avoid the recursion which I found a bit confusing.

pyiron_base/storage/helper_functions.py
```diff
@@ -40,3 +104,218 @@ def write_hdf5(
...
+
+    Returns:
+        str: combined path
+    """
+    return posixpath.join(h5_path, name)
+
+
+def _check_json_conversion(value):
+    """
+    Check if the object can be converted to JSON to optimize the HDF5 performance. This can change the data tupe of the
```
pmrv:
```suggestion
    Check if the object can be converted to JSON to optimize the HDF5 performance. This can change the data type of the
```

pmrv (approved):
I left a number of small remarks, but can be merged after those are considered without re-review.

The only bigger thing I've seen is the `slash` parameter.  While it was present before this PR introduces a lot of code now that offers it, but afaik we never use this, so it can be just removed completely imo

jan-janssen:
> The only bigger thing I've seen is the `slash` parameter. While it was present before this PR introduces a lot of code now that offers it, but afaik we never use this, so it can be just removed completely imo

I would like to make the `slash` parameter as well as the compression setting accessible to the outside once we release the stand-alone HDF5 interface, so I already added it to the central functions in the helper class.

PR #1249, opened 2023-12-16
GenericParameters: add to_dict() and from_dict()

jan-janssen:
In analogy to https://github.com/pyiron/pyiron_base/pull/1248 this pull request adds the `to_dict()` and `from_dict()` function to the `GenericParameters` class. This is required to enable more flexible storage backends.

### DIFF
--- a/pyiron_base/interfaces/has_dict.py
+++ b/pyiron_base/interfaces/has_dict.py
@@ -19,6 +19,7 @@ __date__ = "Dec 20, 2023"
 
 class HasDict(ABC):
     __dict_version__ = "0.1.0"
+
     @abstractmethod
     def from_dict(self, obj_dict: dict, version: str = None):
         pass
--- a/pyiron_base/interfaces/has_dict.py
+++ b/pyiron_base/interfaces/has_dict.py
@@ -1,0 +1,40 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+"""Interface for classes to serialize to dictionary."""
+
+from abc import ABC, abstractmethod
+
+__author__ = "Jan Janssen"
+__copyright__ = (
+    "Copyright 2023, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Jan Janssen"
+__email__ = "janssen@mpie.de"
+__status__ = "production"
+__date__ = "Dec 20, 2023"
+
+
+class HasDict(ABC):
+    __dict_version__ = "0.1.0"
+
+    @abstractmethod
+    def _from_dict(self, data_dict: dict, version: str = None):
+        pass
+
+    @abstractmethod
+    def _to_dict(self):
+        pass
+
+    def _store_type_to_dict(self):
+        type_dict = {
+            "NAME": self.__class__.__name__,
+            "TYPE": str(type(self)),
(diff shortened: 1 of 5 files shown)


pyiron_base/storage/parameters.py
```diff
@@ -936,17 +943,19 @@ def _lines_to_dict(self, lines):
                 lst["Comment"].append("")
         return lst
 
-    def _type_to_hdf(self, hdf):
+    def _type_to_dict(self):
```
pmrv: So this is one of the reasons I keep insisting on a common mixin class for to_dict/from_dict.  If you had this it could be predefined automatically, instead you had to add this now manually here, in #1260 and in #1250 (#1248 should have also had it).

pyiron_base/storage/parameters.py
```diff
@@ -499,6 +499,14 @@ def define_blocks(self, block_dict):
             raise AssertionError()
         self._block_dict = block_dict
 
+    def to_dict(self):
+        data_dict = self._type_to_dict()
+        data_dict["data_dict"] = self._dataset
+        return data_dict
+
+    def from_dict(self, generic_parameters_dict):
+        self._dataset = generic_parameters_dict
```
pmrv:
```suggestion
        self._dataset = generic_parameters_dict["data_dict"]
```
imo `to_dict` and `from_dict` should be symmetric, meaning I should be able to call `obj.from_dict(obj.to_dict()) == obj`.  Everything else outside should treat the dict as opaque.

pmrv (changes requested):

pyiron_base/storage/parameters.py
pmrv: (now that it's done, go ahead with this now anyway, just making my point.)

pyiron_base/storage/parameters.py
jan-janssen: I made the change in a separate commit, to also update the `from_hdf()` function accordingly.

pyiron_base/storage/parameters.py
jan-janssen: @pmrv I added an interface class, maybe you can have another look tomorrow.

pyiron_base/interfaces/has_dict.py
```diff
@@ -0,0 +1,38 @@
...
+__email__ = "janssen@mpie.de"
+__status__ = "production"
+__date__ = "Dec 20, 2023"
+
+
+class HasDict(ABC):
+    __dict_version__ = "0.1.0"
+
+    def from_dict(self, obj_dict: dict, version: str = None):
+        raise NotImplementedError
```
```suggestion
    @abstractmethod
    def from_dict(self, obj_dict: dict, version: str = None):
        pass
```

so that already subclassing `HasDict` without providing an implementation is an error.

pyiron_base/interfaces/has_dict.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def to_dict(self):
+        raise NotImplementedError
```
```suggestion
    @abstractmethod
    def to_dict(self):
        pass
```

pyiron_base/interfaces/has_hdf.py
```diff
@@ -39,7 +40,7 @@ def __exit__(self, *args):
             self._hdf.close()
 
 
-class HasHDF(ABC):
+class HasHDF(HasDict, ABC):
```
pmrv: This then actually also needs to implement the necessary methods for the interface.  Just leave the subclass out here, if you don't want to do this now.

pmrv (approved):
Just nits, otherwise lgtm.

pyiron_base/interfaces/has_hdf.py
jan-janssen:
```suggestion
class HasHDF(ABC):
```

PR #717, opened 2022-03-08
Extension of the dynamic job class to support external executables

jan-janssen:

PR #1253, opened 2023-12-17
Add create_new_job_type() function to dynamically initialize a job object

jan-janssen:


jan-janssen:
Extension of the dynamic jobs in `pyiron_base` which were initially implemented in https://github.com/pyiron/pyiron_base/pull/701 https://github.com/pyiron/pyiron_base/pull/705 https://github.com/pyiron/pyiron_base/pull/717 and https://github.com/pyiron/pyiron_base/pull/799

jan-janssen:
Based on this new functionality, Quantum Espresso can be implemented using the following lines:
```python
import os
from ase.calculators.espresso import Espresso
from ase.io import write
from pyiron_atomistics import ase_to_pyiron
from pyiron_base.jobs.dynamic import create_new_job_type
from pwtools import io


def write_input_calc_minimize(input_dict, working_directory="."):
    filename = os.path.join(working_directory, 'relax.pwi')
    os.makedirs(working_directory, exist_ok=True)
    input_data_relax = {
        'calculation': 'vc-relax',
        'cell_dofree': 'ibrav',
    }
    write(
        filename=filename, 
        images=input_dict["structure"], 
        Crystal=True, 
        kpts=input_dict["kpts"], 
        input_data=input_data_relax, 
        pseudopotentials=input_dict["pseudopotentials"],
        tstress=True, 
        tprnfor=True
    )

def collect_output_calc_minimize(working_directory="."):
    filename = os.path.join(working_directory, 'relax.pwo')
    # return {"structure": io.read_pw_md(filename)[-1].get_ase_atoms()}
    return {"structure": ase_to_pyiron(io.read_pw_md(filename)[-1].get_ase_atoms())}


QEMinimizeJob = create_new_job_type(
    class_name="QEMinimize", 
    input_dict={
        "structure": None, 
        "pseudopotentials": {"Al": "Al.pbe-n-kjpaw_psl.1.0.0.UPF"}, 
        "kpts": (3, 3, 3),
    },
    write_input_funct=write_input_calc_minimize, 
    collect_output_funct=collect_output_calc_minimize, 
    executable_str="mpirun -np 1 pw.x -in relax.pwi > relax.pwo",
)

pr = Project("test")
job_qe_mini = pr.create_job(job_type=QEMinimizeJob, job_name="qe_mini")
```

tests/job/test_dynamic.py
```diff
@@ -0,0 +1,32 @@
...
+    with open(os.path.join(working_directory, "input.json"), "w") as f:
+        json.dump(input_dict, f)
+
+
+def collect_output(working_directory="."):
+    with open(os.path.join(working_directory, "output.json"), "r") as f:
+        return json.load(f)
+
+
+class DynamicTest(unittest.TestCase):
```
pmrv: `TestWithProject` could be helpful here.

pmrv:
How does this interact with the TYPE field in HDF5 storage? @liamhuber and I [recently discussed](https://github.com/pyiron/pyiron_base/pull/1243/files#r1416278469) that currently dynamic classes are a bit weird from a storage perspective.  It might be nice to have dynamic classes somehow advertise themselves as such via the TYPE field, which would require providing an overload for `GenericJob._type_to_hdf`.  

I'm ok with classes defined in notebooks not being able to reload, but for dynamic classes in modules might be useful to add them to JOB_DICT directly in `create_new_job_type`.

tests/job/test_dynamic.py
jan-janssen: That's a good idea - I updated the test to use the `TestWithProject` class.

jan-janssen:
> I'm ok with classes defined in notebooks not being able to reload, but for dynamic classes in modules might be useful to add them to JOB_DICT directly in `create_new_job_type`.

I guess currently in a module I would use something like: 
```
from pyiron_base import JOB_CLASS_DICT

JOB_CLASS_DICT["MyClass"] = create_new_job_type(
    class_name="MyClass",
    ...
)
```

Obviously, it is a bit redundant, as we have to type the class name twice. But I guess this is something we can work with for now.

pmrv:
I also have this small wrapper that allows to write functions to HDF that might be useful for this or #1164 

```python

import codecs, dill

def _to_pickle(value):
    return codecs.encode(dill.dumps(value), "base64").decode()


def _from_pickle(value):
    return dill.loads(codecs.decode(value.encode(), "base64"))

class FunctionContainer(HasHDF):
    __slots__ = ('func',)

    def __init__(self, func=None):
        self.func = func
        self.__doc__ = func.__doc__

    def _to_hdf(self, hdf):
        hdf['dill'] = _to_pickle(self.func)

    def _from_hdf(self, hdf, version=None):
        self.func = _from_pickle(hdf['dill'])

    def __call__(self, *args, **kwargs):
        return self.func(*args, **kwargs)
```

If you had this you could also add `to_hdf/from_hdf` automatically to write the user provided functions to HDF5 storage so that even dynamically created classes in notebooks could be restored.

samwaseda:
I am very unhappy with the lack of doc string...

samwaseda:
I cannot really say what would be the appropriate name, but I think it makes sense to put this function in the `Creator` of `Project`. Otherwise it will be undermined just like `TemplateJob`.

pmrv (changes requested):
Agree with @samwaseda's points.  The absolute minimum would be docs that specify of what shape or format the provided functions should be.

pyiron_base/jobs/dynamic.py
```diff
@@ -132,4 +132,53 @@ def _get_template_classes():
...
+                working_directory=self.working_directory,
+                input_dict=self.input.to_builtin(),
+            )
+
+        return write_input
+
+    def _init_constructor_dynamic_func(class_name, input_dict, executable_str):
+        def __init__(self, project, job_name):
+            super(self.__class__, self).__init__(project, job_name)
+            self.__name__ = class_name
```
pmrv, line 152: Should be handled already by [this](https://github.com/pyiron/pyiron_base/blob/e359d95f17c17c7489120fe81989fcc35511182c/pyiron_base/jobs/job/generic.py#L130).

pyiron_base/jobs/dynamic.py
```diff
@@ -132,4 +132,53 @@ def _get_template_classes():
     }
 
 
+def create_new_job_type(
+    class_name, input_dict, write_input_funct, collect_output_funct, executable_str
+):
+    from pyiron_base.jobs.job.template import TemplateJob
+
+    def _get_write_input_func(write_input_funct):
```
pmrv, line 140: Is the double nesting necessary?  In my quick testing, just defining `write_input` immediately in the function body works as well.

jan-janssen:
This pull request is replaced by https://github.com/pyiron/pyiron_base/pull/1286

PR #1258, opened 2023-12-19
Workaround for SerialMaster

pmrv:
https://github.com/pyiron/pyiron_atomistics/pull/1265 flagged a failing test for SerialMasterBase, that tries to re-run a master job and fails because the server object is locked.  This is a hack to make it work again.


pmrv:
Closed in favor of removing `SerialMaster` which is the only thing that failed after bumping base to 0.6.14.

PR #1261, opened 2023-12-20
DataContainer: split _to_hdf() 

jan-janssen:
The `_to_hdf()` function is split into a `_get_dict_for_storage()` function and the `to_hdf()` function which internally calls the `_get_dict_for_storage()` function. While this is just a refactoring it highlights the different types of data stored in DataContainers and allows us to transition one type at the time to a more abstract `to_dict()` based storage scheme.


pmrv (approved):
lgtm, we not yet using the feature to re-use open files here, but the docstring clearly document the intent and what is necessary to do so, so this is perfect imo.

PR #1276, opened 2024-01-02
[feedback wanted] Use regex for job table sorting

samwaseda:
@sudarsan-surendralal and I introduced some time ago the sorting of job tables such as:

```python
pr.job_table(status="finished")
```

which returns the job table containing only entries where the status is `finished`. While doing this, we allowed also *some* flexibility with `*` (and `!` added some time later). This feature isn't so bad, but I would like to replace it with regex, because we've got the following problems:

- This functionality is super limited compared to regex
- Our current code: 16 lines vs. with regex: 1 line
- Ours is not a standardised method?
  - At least we have to write in the doc string what would be understood for each case, while with regex we can simply say "It's regex"

If I remember correctly, using regex was originally @niklassiemer's idea, and we sort of gave him the cold shoulder and now I would like to come back to it because I'm super frustrated by how limited our current functionality is.

I am not going to merge this one until the next pyiron meeting, and even if it's merged the current functionality will be maintained, because if `regex=True` not set, you will only get a warning saying it's deprecated. Anyway ahead of it I would like to get some feedback from you.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -85,7 +85,7 @@ class IsDatabase(ABC):
 
     @staticmethod
     def _get_filtered_job_table(
-        df: pandas.DataFrame, **kwargs: dict, regex=False,
+        df: pandas.DataFrame, regex=False, **kwargs: dict,
     ) -> pandas.DataFrame:
         """
         Get a job table in a project based on matching values from any column in the project database
@@ -113,7 +113,9 @@ class IsDatabase(ABC):
                     f"Column name {key} does not exist in the project database!"
                 )
         if len(kwargs) > 0 and not regex:
-            logger.warn("regex=False when using keyword sorting is deprecated")
+            logger.warning(
+                "regex=False when using keyword sorting is deprecated",
+            )
         for key, val in kwargs.items():
             if regex:
                 update = df[key].apply(lambda x: re.search(val, x)).astype(bool)
@@ -219,7 +221,7 @@ class IsDatabase(ABC):
                 "`job_name_contains` is deprecated - use `job='*term*'` instead"
             )
             kwargs["job"] = "*{}*".format(job_name_contains)
-        df = self._get_filtered_job_table(df, **kwargs, regex=regex)
+        df = self._get_filtered_job_table(df, regex=regex, **kwargs)
         if sort_by is not None:
             return df.sort_values(by=sort_by)
         return df
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -16,6 +16,7 @@ from datetime import datetime
 from pyiron_base.utils.deprecate import deprecate
 import pandas
 import typing
+import fnmatch
 from sqlalchemy import (
     create_engine,
     MetaData,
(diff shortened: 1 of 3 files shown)


pmrv:
Since we are anyway only filtering the table we get from the database client side, we should also investigate if `DataFrame.query`  doesn't already most of the things we want and if we can make that more accessible.

samwaseda:
> Since we are anyway only filtering the table we get from the database client side, we should also investigate if `DataFrame.query` doesn't already most of the things we want and if we can make that more accessible.

I just took a look but isn't it something totally different? I guess `query` can already be applied, but I don't for example understand how I can get only job names starting with something like `abc` and ending with `xyz` using `query`.

pmrv:
You can call some methods inside the query so for your example it would be

```python
df = pr.job_table()
df.query('job.str.startswith("abc") and job.str.endswith("xyz")')
```

regex are not so easy it seems, but I think most of the filtering that we support now could be done with query.  It's just annoying to have to do that separately.   Anyway I just wanted to bring it up because it seems related.

samwaseda:
> regex are not so easy it seems, but I think most of the filtering that we support now could be done with query. It's just annoying to have to do that separately. Anyway I just wanted to bring it up because it seems related.

I don't necessarily stick to regex, but it seems to me that it's the only one way to deal with strings. Your example shows how much you have to write as soon as it's string handling and not about numbers, so I guess we have to think about something beyond `query`.

samwaseda:
Conclusions:
- Replace current lines by `glob`
  - Fail silently if string starts with `!`
- `regex=False`: function argument
- Filtering applied **after** query

pmrv:
You mentioned the concern where searching for `finished` also matches `abcfinishedfobar`.  Maybe worthwhile to add '^' + regex + `$` or use `re.fullmatch` instead of search, but I don't actually have a strong opinion.

For large job tables we could `re.compile` the expression before, but probably not necessary.

niklassiemer:
> You mentioned the concern where searching for `finished` also matches `abcfinishedfobar`. Maybe worthwhile to add '^' + regex + `$` or use `re.fullmatch` instead of search, but I don't actually have a strong opinion.
> 
> For large job tables we could `re.compile` the expression before, but probably not necessary.

If it does not cause a huge overhead on small tables, `re.compile` sounds like it might be better and is unlikely to cause problems :+1:

I would think if `regex=False` is the default, a plain `re.search` is the way to go. Otherwise it might hamper some queries for which you want the regex in the first place a lot! (Does the structure contain Al and O? `structure="Al.*O|O.*Al"` would be restrained to be `AlO` or `OAl`)

liamhuber:
I think replacing `regex:bool = False` with `mode: typing.Literal["search", "regex", "glob"] = "search"` (Or whatever). Even if there are only two options, to me it still is more readable than a bool for `regex`. This is not a hill I'd die on though.

liamhuber:
Screen Shot 2024-01-18 at 09 24 38

😂 😂 ❤️

liamhuber (approved):
Lgtm.

I like there's even a helpful "did you mean" and everything.

samwaseda:
I haven’t got a reply from @pmrv and @niklassiemer but since the default behaviour doesn’t change (except if there’s `!` at the beginning), I’m gonna merge this one. Feel free to open another PR .

PR #1284, opened 2024-01-12
Enforce environment for subprocess

jan-janssen:


jan-janssen:
The notebooks work with these changes https://github.com/jan-janssen/debug-pyiron-integration-tests/tree/subprocess_env_bug

PR #1285, opened 2024-01-13
Simplify submitting functions to the queue

jan-janssen:
Now that the `DataContainer` is finally able to store arbitrary objects - at least with python 3.11 and the latest version of h5io https://github.com/h5io/h5io/releases/tag/h5io-0.2.1 - I finally come back to the FunctionContainer pull request https://github.com/pyiron/pyiron_base/pull/1080 .

Example: 

```python
def test_function(a, b=8):
    return a+b

from pyiron_base import Project
pr = Project("test")
job = pr.wrap_python_function(test_function)
job.input["a"] = 4
job.input["b"] = 5
job.run()
job.output["result"] 
```

Features: 

* Any kind of function can be wrapped. In the background I am using `cloudpickle` to pickle the functions. 
* Job names are automatically generated with the md5 hash function, so multiple calls of the same function with different parameters result in different jobs. Still when the same job is submitted with the same function and the same parameters, then the job is correctly reloaded. 
* Functions can be executed in the background using `job.server.run_mode.non_modal = True` or submitted to the queuing system. 
* The default arguments of the function are copied to `job.input`. Parameters which do not have a default, are set to `None`.

PR #1286, opened 2024-01-13
Implement ExecutableJobContainer

jan-janssen:
Following a similar reasoning as https://github.com/pyiron/pyiron_base/pull/1285 this pull request is a replacement for https://github.com/pyiron/pyiron_base/pull/1253 based on the recent developments in https://github.com/h5io/h5io/releases/tag/h5io-0.2.1 . 

Example:  
```python
def write_input(input_dict, working_directory="."):
    with open(os.path.join(working_directory, "input_file"), "w") as f:
        f.write(str(input_dict["energy"]))


def collect_output(working_directory="."):
    with open(os.path.join(working_directory, "output_file"), "r") as f:
        return {"energy": float(f.readline())}


pr = Project("test")
pr.create_job_class(
    class_name="CatJob",
    write_input_funct=write_input,
    collect_output_funct=collect_output,
    default_input_dict={"energy": 1.0},
    executable_str="cat input_file > output_file",
)
job = pr.create.job.CatJob(job_name="job_test")
job.input["energy"] = 2.0
job.run()
job.output
```

Features: 
* Any kind of external executable can be wrapped. 
* The write_input() and collect_output() function are stored in the HDF5 file using `cloudpickle`. 
* The newly defined classes can be submitted to the queuing system. 
* Classes are automatically registered in the `JOB_CLASS_DICT`, so users can easily create multiple jobs of the same type. Still while the jobs can still be reloaded, the class is removed from the `JOB_CLASS_DICT` once the current python session is closed.


pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, project, job_name):
+        super().__init__(project, job_name)
+        self._write_input_funct = None
+        self._collect_output_funct = None
+
+    def set_job_type(
+        self,
+        write_input_funct,
+        collect_output_funct,
+        default_input_dict,
```
samwaseda: We might need an optional argument `collect_output_dict`, because the output parsing sometimes depend on the input.

samwaseda (commented):
I have a suggestion to make it this way:

```python
pr = Project("test")
new_job_class = pr.create_job_class(
    class_name="CatJob",
    write_input_funct=write_input,
    collect_output_funct=collect_output,
    default_input_dict={"energy": 1.0},
    executable_str="cat input_file > output_file",
)
job = new_job_class(job_name="job_test")
```

Reasons:

1. It's not intuitively clear that the new class is to be accessed via `pr.create.job.NewClass`
2. `pr` is repeated, which might create the problem that some people might create a job in `pr_1` and try to use it in `pr_2`
3. Setting the new job class on the same level as other libraries (such as `Lammps`) might create the confusion that the new class is stored somewhere in pyiron, which I guess is not the case (right?). I think assigning just a variable makes a user understand better that it only lives in this active notebook and it's not stored anywhere else.

jan-janssen:
> Reasons:
> 
> 1. It's not intuitively clear that the new class is to be accessed via `pr.create.job.NewClass`

I agree and I am fine to support both, the creation via `pr.create.job.NewClass` and via the `new_job_class` function you create above. Still for consistency I would require both a `job_name` and a `project` as inputs. I explain this below.

> 2. `pr` is repeated, which might create the problem that some people might create a job in `pr_1` and try to use it in `pr_2`

I agree that having the registration of a new job class attached to a project is confusing. As it is anyway a special use case we could ask the user to import the function: 

`from pyiron_base import create_job_class`

I not a fan of implicitly hiding the project in the function that creates the job. For example when a user creates a sub project, then he can create all other jobs as before but for the custom classes he would have to initialize them again. 

> 3. Setting the new job class on the same level as other libraries (such as `Lammps`) might create the confusion that the new class is stored somewhere in pyiron, which I guess is not the case (right?). I think assigning just a variable makes a user understand better that it only lives in this active notebook and it's not stored anywhere else.

I am fine with creating a standalone function which returns a creator. My motivation to register the class like any other was that some developers might use the function in their module definition.

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: I agree that this case exists but I think those are special cases so it is not handled by the current pull request. In the same way atomistic functions like `calc_md()` are currently not available in this interface. So it is going to take some time until this simplified interface can replace jobs in `pyiron_atomistics` - also the interface for interactive execution is currently still missing. Nevertheless for many simple use cases even beyond atomistic simulation I could see this functionality accelerate the implementation of new job classes in pyiron.

jan-janssen:
@samwaseda I extended the functionality to cover the use case you suggested:

```python
from pyiron_base import create_job_factory


def write_input(input_dict, working_directory="."):
    with open(os.path.join(working_directory, "input_file"), "w") as f:
        f.write(str(input_dict["energy"]))


def collect_output(working_directory="."):
    with open(os.path.join(working_directory, "output_file"), "r") as f:
        return {"energy": float(f.readline())}


pr = Project("test")
create_job_cat = create_job_factory(
    write_input_funct=write_input,
    collect_output_funct=collect_output,
    default_input_dict={"energy": 1.0},
    executable_str="cat input_file > output_file",
)
job = create_job_cat(project=pr, job_name="job_test")
job.input["energy"] = 2.0
job.run()
job.output
```

This new functionality is introduced in addition to the functionality introduced above.

jan-janssen:
@samwaseda Can you take another look at this pull request?

PR #1289, opened 2024-01-14
Project: Add __getstate__() and __setstate__()

jan-janssen:
Projects can now be pickled. This allows us to store projects in data containers: 
```python
import pickle 

pr = Project("test")
pr_reload = pickle.loads(pickle.dumps(pr))
```
The reloaded Project `pr_reload` has the same properties as the initial project `pr`.

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -1548,6 +1548,32 @@ class Project(ProjectPath, HasGroups):
             {"groups": self.list_dirs(skip_hdf5=True), "nodes": self.list_nodes()}
         )
 
+    def __getstate__(self):
+        return {
+            "user": self.user,
+            "sql_query": self.sql_query,
+            "history": self._history,
+            "root_path": self._root_path,
+            "project_path": self._project_path,
+            "filter": self._filter,
+            "inspect_mode": self._inspect_mode,
+        }
+
+    def __setstate__(self, state):
+        self.user = state["user"]
+        self.sql_query = state["sql_query"]
+        self._history = state["history"]
+        self._root_path = state["root_path"]
+        self._project_path = state["project_path"]
+        self._filter = state["filter"]
+        self._inspect_mode = state["inspect_mode"]
+        self._data = None
+        self._creator = Creator(project=self)
+        self._loader = JobLoader(project=self)
+        self._inspector = JobInspector(project=self)
+        self.job_type = JobTypeChoice()
+        self._maintenance = None
+
     def _get_item_helper(self, item, convert_to_object=True):
         """
         Internal helper function to get item from project
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -1549,22 +1549,21 @@ class Project(ProjectPath, HasGroups):
         )
 
     def __getstate__(self):
-        return {
-            "user": self.user,
-            "sql_query": self.sql_query,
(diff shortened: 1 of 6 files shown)


jan-janssen:
This functionality requires Python 3.11.

jan-janssen:
@pmrv I switched to `isinstance()` as we discussed.

pyiron_base/project/generic.py
```diff
@@ -1548,6 +1548,32 @@ def __repr__(self):
...
+            "filter": self._filter,
+            "inspect_mode": self._inspect_mode,
+        }
+
+    def __setstate__(self, state):
+        self.user = state["user"]
+        self.sql_query = state["sql_query"]
+        self._history = state["history"]
+        self._root_path = state["root_path"]
+        self._project_path = state["project_path"]
```
pmrv: Ideally this would be set in `GenericPath/ProjectPath.__setstate__`, but I'm not sure why even have those as separate classes.

tests/generic/test_datacontainer.py
```diff
@@ -686,6 +687,16 @@ def test_overwrite_ordering(self):
         self.assertEqual(len(d), len(items),
                          "Number of items in HDF does not match length of container!")
 
+    @unittest.skipIf(sys.version_info < (3, 11), "__getstate__() and __setstate__() support in h5io requires Python 3.11")
+    def test_project_in_datacontainer(self):
```
pmrv, line 691:
```suggestion
    def test_project_in_datacontainer(self):
        """DataContainer should be able to save Project to HDF."""
```

tests/project/test_project.py
```diff
@@ -238,6 +239,11 @@ def test_maintenance_get_repository_status(self):
         df = self.project.maintenance.get_repository_status()
         self.assertIn('pyiron_base', df.Module.values)
 
+    def test_serialization(self):
```
pmrv:
```suggestion
    def test_pickle(self):
        """Project should be pickle-able."""
```

pyiron_base/project/generic.py
jan-janssen: I implemented `__getstate__()` and `__setstate__()` in `GenericPath` and `ProjectPath`. Still I agree that the interface should be refactored but not in this pull request.

pyiron_base/storage/datacontainer.py
```diff
@@ -837,10 +838,17 @@ def normalize_key(name):
...
+                    if isinstance(hdf_group, ProjectHDFio):
+                        return hdf_group.to_object()
+                    else:
+                        return hdf_group
+
                 items.append(
                     (
                         *normalize_key(g),
-                        hdf[g].to_object() if not self._lazy else HDFStub(hdf, g),
+                        to_object(hdf[g]) if not self._lazy else HDFStub(hdf, g),
```
pmrv, line 843: I've just noticed it, but this logic also needs to be applied in [HDFStub](https://github.com/pyiron/pyiron_base/blob/9dec6f9744b937548d8846bf4226520bae3b98e4/pyiron_base/storage/hdfstub.py#L101) otherwise lazy loading will not work.

pmrv (approved):

pyiron_base/storage/datacontainer.py
pmrv: Everything else lgtm, though.

PR #1299, opened 2024-01-30
Split DataContainer to a HDF and non HDF part

pmrv:
Basically DataContainer is renamed to DataContainerBase and the methods used for HasHDF/HDFStub are moved to a new sub class DataContainer. Docstrings are replicated where necessary.  No API changes for users.

In preparation of https://github.com/orgs/pyiron/projects/7/views/1

@liamhuber Pulling `DataContainer` out should be just a matter of copy/paste now.

### DIFF
--- a/pyiron_base/storage/datacontainer.py
+++ b/pyiron_base/storage/datacontainer.py
@@ -447,7 +447,11 @@ class DataContainerBase(MutableMapping, Lockable, HasGroups):
             )
         else:
             return list(
-                v.to_builtin(stringify=stringify) if isinstance(v, DataContainerBase) else v
+                (
+                    v.to_builtin(stringify=stringify)
+                    if isinstance(v, DataContainerBase)
+                    else v
+                )
                 for v in self.values()
             )
 
--- a/pyiron_base/storage/datacontainer.py
+++ b/pyiron_base/storage/datacontainer.py
@@ -47,16 +47,16 @@ def _normalize(key):
     return key
 
 
-class DataContainer(MutableMapping, Lockable, HasGroups, HasHDF):
+class DataContainerBase(MutableMapping, Lockable, HasGroups):
     """
     Mutable sequence with optional keys.
 
-    If no argument is given, the constructor creates a new empty DataContainer.  If
+    If no argument is given, the constructor creates a new empty DataContainerBase.  If
     specified init maybe a Sequence, Set or Mapping and all recursive
-    occurrences of these are also wrapped by DataContainer.
+    occurrences of these are also wrapped by DataContainerBase.
 
-    >>> pl = DataContainer([3, 2, 1, 0])
-    >>> pm = DataContainer({"foo": 24, "bar": 42})
+    >>> pl = DataContainerBase([3, 2, 1, 0])
+    >>> pm = DataContainerBase({"foo": 24, "bar": 42})
 
     Access can be like a normal list with integers or optionally with strings
     as keys.
@@ -70,23 +70,23 @@ class DataContainer(MutableMapping, Lockable, HasGroups, HasHDF):
 
     Keys do not have to be present for all elements.
 
-    >>> pl2 = DataContainer([1,2])
+    >>> pl2 = DataContainerBase([1,2])
     >>> pl2["end"] = 3
     >>> pl2
-    DataContainer({0: 1, 1: 2, 'end': 3})
+    DataContainerBase({0: 1, 1: 2, 'end': 3})
 
-    It is also allowed to set an item one past the length of the DataContainer,
+    It is also allowed to set an item one past the length of the DataContainerBase,
     this is then equivalent to appending that element.  This allows to use the
-    update method also with other DataContainers
+    update method also with other DataContainerBases
 
     >>> pl[len(pl)] = -1
     >>> pl
-    DataContainer([3, 2, 1, 0, -1])
+    DataContainerBase([3, 2, 1, 0, -1])
     >>> pl.pop(-1)
     -1
 
     Where strings are used they may also be used as attributes.  Getting keys
-    which clash with methods of DataContainer must be done with item access, but
+    which clash with methods of DataContainerBase must be done with item access, but
     setting them works without overwriting the instance methods, but is not
     recommended for readability.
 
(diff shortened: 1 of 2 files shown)


liamhuber:
> @liamhuber Pulling DataContainer out should be just a matter of copy/paste now.

I think this may be of even more imminent utility to @samwaseda?

Will review shortly, but I peeked at the diff and it was what I intuitively expected so that should go pretty smoothly.

pyiron_base/storage/datacontainer.py
```diff
@@ -47,16 +47,16 @@ def _normalize(key):
     return key
 
 
-class DataContainer(MutableMapping, Lockable, HasGroups, HasHDF):
+class DataContainerBase(MutableMapping, Lockable, HasGroups):
```
liamhuber, line 50: If we move this part off to its own repo (e.g. `pyiron/snippets`) and leave the child class here, then in the other location I would simply call this `DataContainer` and here do like `from pyiron.snippets import DataContainer as DataContainerBase`. In the context of making the break-apart go as smoothly as possible, I am totally behind the naming scheme used in this PR, I just want to double check that we have the same middle-term vision on naming here.

pyiron_base/storage/datacontainer.py
```diff
@@ -451,7 +430,7 @@ def to_builtin(self, stringify=False):
                 # _normalize calls int() on all digit string keys this is
                 # transparent for the rest of the module
                 k = str(k)
-                if isinstance(v, DataContainer):
+                if isinstance(v, type(self)):
```
liamhuber: Is this (and the other nearby `type(self)` appearances) really correct? I naively would have thought this should be `isinstance(v, DataContainerBase))`, since that is already what guarantees us to have `.to_builtin`. 

What _should_ the behaviour be if a `DataContainer` instance has a `DataContainerBase` instance as an item? I would want it to "just work", but I know that is vague. Here I think it certainly wouldn't as the type check would evaluate `False` and we'd get kicked over to the `repr` branch.

I don't know what the exact right behaviour is, but I think this needs a test added to solidify whatever choice you make.

pyiron_base/storage/datacontainer.py
```diff
@@ -566,11 +545,11 @@ def _search_parent(self, key, stop_on_first_hit=True):
         # descend into subgroups
         for it in self.groups():
             hit = self[it]._search_parent(key, stop_on_first_hit)
-            if isinstance(hit, DataContainer):
+            if isinstance(hit, type(self)):
```
liamhuber: Similar... if `DataClass` hits a `DataClassBase` instance, do we want to stop or not? Because right now it's "not". I could be convinced that's the right behaviour, but intuitively it feels wrong.

pyiron_base/storage/datacontainer.py
```diff
@@ -920,13 +962,27 @@ def _force_load(self, recursive=True):
 
         # values are loaded from HDF once they are accessed via __getitem__, which is implicitly called by values()
         for v in self.values():
-            if recursive and isinstance(v, DataContainer):
+            if recursive and isinstance(v, type(self)):
```
liamhuber: Similarly, should this be `(v, DataContainer)`? If my `MyDataContainer(DataContainer)` has `DataContainer` instances as items, don't those instances contain enough power to recurse like this?

liamhuber (changes requested):
99% as expected and no nits or anything.

The only thing I'm not sure of is the consistent use of `type(self)` for a bunch of recursion checks. Most of these seem to me like they should instead hard-code the class they're written in, so that if derived classes hold an instance of their parent class the recursion can continue uninterrupted. I'm not totally sure that's the right behaviour choice though.

At any rate, now that we explicitly have a base and child class, I think it's important to extend the tests to cover the case of child-holds-parent-instance to encode whatever behaviour decision you make.

pyiron_base/storage/datacontainer.py
pmrv: Yeah, that's exactly how I thought it would go down.

pyiron_base/storage/datacontainer.py
```diff
@@ -851,62 +941,14 @@ def normalize_key(name):
...
-        """
-        write(self.to_builtin(), file_name)
+    # HDFStub compat
+    def __getitem__(self, key):
+        value = super().__getitem__(key)
+        if not isinstance(value, HDFStub):
+            return value
+        else:
+            value = self[key] = value.load()
+            return value
```
samwaseda, line 955: Does it have a particular reason that it assigns `value` here?

pmrv: Once we have loaded a stub, we don't want to load it again a second time, so this overwrites the stub with the actual value.  I guess

```python
return self[key] := value.load()
```

would work as well, but is only available in newer python version (not sure which atm).

samwaseda: ok it's a small thing, but my problem is that by returning `value` it might give the impression that it's the copy of the data, while in reality if you change it, it changes the content in `self[key]` as well. So I would simply set `self[key] = value.load()` and `return self[key]`.

pmrv: Won't work because we are inside `__getitem__`, so using `self[key]` will recurse without stopping.

samwaseda: ah ok. Beautiful

pmrv, line 955:
```suggestion
            return self[key] := value.load()
```

the "walrus" is already in since 3.8 it turns out, so let's just use it.

pmrv: Ah, bummer it's not allowed in this place.  Will revert.

niklassiemer (commented):
LGTM once @liamhuber s concerns are matched

pyiron_base/storage/datacontainer.py
pmrv: Agree with your reasoning, it seems a more useful choice, I just hadn't considered it.  I've changed it back, but tests will have to come later.

pmrv:
```
  File "/home/runner/work/pyiron_base/pyiron_base/tests/flex/test_pythonfunctioncontainer.py", line 63, in test_with_executor_wait
```

@jan-janssen Sounds related to your recent developments.  Would that be fixed by #1300 or #1301?

jan-janssen:
> @jan-janssen Sounds related to your recent developments. Would that be fixed by #1300 or #1301?

The issue was solved by restarting the CI. We have to take a look at this and maybe increase the time out in case the issue continues to come up.

jan-janssen:
Once this is merged, we should also release a new `pyiron_base` version.

pmrv:
> > @jan-janssen Sounds related to your recent developments. Would that be fixed by #1300 or #1301?
> 
> The issue was solved by restarting the CI. We have to take a look at this and maybe increase the time out in case the issue continues to come up.

I think it makes more sense to just use the timeout directly in the tests, then this can never happen.

jan-janssen:
@pmrv As more tests were requested, I changed the status of this pull request to `draft` until the tests are added.

liamhuber (approved):
Nice, type checking and testing changes lgtm.

PR #1300, opened 2024-01-31
Fix reloading of executables for flexible jobs

jan-janssen:
The executable of a flexible job created using the `create_job_class()` function was empty when reloading the job from the HDF5 file. This issue is fixed in this pull request and the corresponding test is added.

### DIFF
--- a/pyiron_base/jobs/job/extension/executable.py
+++ b/pyiron_base/jobs/job/extension/executable.py
@@ -299,7 +299,10 @@ class Executable(HasStorage):
         try:
             return self.executable_lst[self.version]
         except KeyError:
-            return ""
+            if isinstance(self.version, str):
+                return self.version
+            else:
+                return ""
 
     def _get_hdf_group_name(self):
         return "executable"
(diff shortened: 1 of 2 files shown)


pmrv (approved):

pyiron_base/jobs/job/extension/executable.py
```diff
@@ -299,7 +299,10 @@ def _executable_select(self):
         try:
             return self.executable_lst[self.version]
         except KeyError:
-            return ""
+            if isinstance(self.version, str):
+                return self.version
+            else:
+                return ""
```
niklassiemer, line 305: I am slightly confused - why does the `version` hold the `str: absolute executable path` in this case?

niklassiemer (commented):
Test and behavior LGTM! One question, though.

pyiron_base/jobs/job/extension/executable.py
jan-janssen: In the current implementation when the `executable_path` is set this internally is stored in the `version` property: 
https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/jobs/job/extension/executable.py#L205

I added the comparison of the storage dictionary to demonstrate this. I am also open to changing the executable interface but from my perspective this goes beyond the scope of this pull request.

PR #1301, opened 2024-01-31
Fix wait_for_job() when for jobs with executors

jan-janssen:
There was a bug in the `wait_for_job()` function on the project level as a result the function failed with a `TimeoutError` after the first iteration rather than iterating up to `max_iterations`. This issue is fixed now and two additional tests are added to validate this functionality is consistent.


ligerzero-ai:
LGTM

PR #1303, opened 2024-01-31
Job class creator can be static

liamhuber:
~I want it so for `pyiron_workflow`, but~ PyCharm even nits me about it.

EDIT: Actually it doesn't look like I need it imminently, but I still agree with PyCharm.


jan-janssen (approved):
Looks good to me

PR #1311, opened 2024-02-04
Add output files class

jan-janssen:
In comparison to the previous work in https://github.com/pyiron/pyiron_base/pull/969 this pull request only lists the file paths for the files in the working directory.

Alternatively, we could also consider combining both pull requests.


jan-janssen:
@pmrv I merged your previous branch into this one and modified it to fulfil your requirements (based on the tests) as well as mine. From my perspective the pull request is now ready to be reviewed.

pyiron_base/jobs/job/core.py
```diff
@@ -982,6 +994,20 @@ def __delitem__(self, key):
...
+
+        Args:
+            file_name (str): the file to print
+            lines (int): how many lines to print
+
+        Raises:
+            FileNotFoundError: if the given file name does not exist in the job folder
+        """
+        print(*_job_read_file(self, file_name, tail=lines), sep="")
+
```
pmrv:
```suggestion
```
I had only introduced this method in this PR, so let's just not expose it at all.

pyiron_base/jobs/job/extension/files.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __dir__(self):
+        return list(self._get_file_dict().keys()) + [
+            "list",
+            "tail",
+            "__dir__",
+            "__getitem__",
+            "__getattr__",
+            "_ipython_display_",
+        ]
```
pmrv:
```suggestion
        return list(self._get_file_dict().keys()) + super().__dir__()
```

pyiron_base/jobs/job/extension/files.py
```diff
@@ -0,0 +1,125 @@
...
+        return File(os.path.join(self._working_directory, item))
+
+    def __getattr__(self, item):
+        try:
+            return self[self._get_file_dict()[item]]
+        except KeyError:
+            raise AttributeError(item) from None
+
+
+class File(str):
```
pmrv, line 109:
```suggestion
class File:
    __slots__ = ("_path",)
    def __init__(self, path: str):
        """
        Wrap a file.
        
        Args:
            path (str): path to file
        """
        self._path = path
```

pmrv: We not actually using any string functionality here and I'd rather not let people accidentally perform string operations on this object.

pyiron_base/jobs/job/extension/files.py
```diff
@@ -0,0 +1,125 @@
...
+    def tail(self, lines: int = 100):
+        print(
+            *_working_directory_read_file(
+                working_directory=os.path.dirname(self),
+                file_name=os.path.basename(self),
```
pmrv, line 114:
```suggestion
                working_directory=os.path.dirname(self._path),
                file_name=os.path.basename(self._path),
```
See above.

pyiron_base/jobs/job/extension/files.py
```diff
@@ -0,0 +1,125 @@
...
+    The :meth:`.tail` method prints the last lines of a file to stdout
+
+    >>> job.files.tail('error.out', lines=1)
+    Something went wrong!
+
+    For files that have valid python variable names can also be accessed by
+    attribute notation
+
+    >>> job.files.INCAR
+    ["SYSTEM=pyiron\n", "ENCUT=270\n", ...]
```
pmrv:
```suggestion
    >>> job.files.INCAR # doctest: +SKIP
    File('INCAR')
```

getattr/getitem return `File` object, if I understand correctly.

pmrv (approved):
Left some small nits, but otherwise lgtm.

pyiron_base/jobs/job/extension/files.py
jan-janssen: The reason why I would like to keep the string part in, is that then I can use `os.path.join()` on the object, this is what I need to access the path.

PR #1314, opened 2024-02-04
Do not import pint when it is not needed

jan-janssen:


niklassiemer (commented):
I am ok with the changes in general, especially for such a rare used case the recreation of the function is fine with me. However, I wonder if we could not simply do something like

```python
def _size_conversion(size: "pint.Quantity"):
    import pint
    ...
```
and leave the function where it was?

jan-janssen:
> I am ok with the changes in general, especially for such a rare used case the recreation of the function is fine with me. However, I wonder if we could not simply do something like
> 
> ```python
> def _size_conversion(size: "pint.Quantity"):
>     import pint
>     ...
> ```
> 
> and leave the function where it was?

I do not think that python typing allows defining types as strings.

jan-janssen:
@niklassiemer I moved the function to a separate module, this way it can still be imported when somebody wants to use it somewhere else but it is not imported by default.

niklassiemer (approved):
LGTM

PR #1317, opened 2024-02-07
Remove sql_query functionality to prepare unified db backends

niklassiemer:


jan-janssen:
I converted this pull request to `draft` as the unit tests are currently failing.

niklassiemer:
I am not sure why the macOS times out after ~5 minutes?

niklassiemer:
https://github.com/pyiron/pyiron_base/actions/runs/7972721340/usage?pr=1317 shows the total billable time for the entire workflow just passes the 30 minutes (At least in the first attempt) - I wonder if this is related to the 30 min max of the test workflow, while I expected this to be 30 min per OS/python-version combination....

jan-janssen:
This is an interesting case - the timeout on `main` was changed to 5 min to fail more quickly. And while this change was not yet applied to your branch, it seems like your runs are already limited to 5min. https://github.com/pyiron/pyiron_base/blob/main/.github/workflows/unittests.yml#L61 

I would suggest to merge main into this branch, as we had quite a few changes since this pull request was started.

niklassiemer:
Sure, and you are absolutely right-the github workflows are always taken from main! I forgot... This protects us from possible security issues.

PR #1323, opened 2024-02-12
Include files outside the archive in job.files.list()

jan-janssen:
Previously `job.files.list()` only lists the files in the job archive, this does not include the restart files. This pull request combines both, now `job.files.list()` lists the file in the job archive as well as the other files in the working directory.


pmrv (approved):
lgtm

PR #1324, opened 2024-02-12
remove unnecessary database queries

XzzX:
Improves performance of `remove_jobs`.

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -1185,27 +1185,27 @@ class Project(ProjectPath, HasGroups):
         if isinstance(job_specifier, (list, np.ndarray)):
             for job_id in job_specifier:
                 self.remove_job(job_specifier=job_id, _unprotect=_unprotect)
-        else:
-            if not self.db.view_mode:
-                try:
-                    job = self.inspect(job_specifier=job_specifier)
-                    if job is None:
-                        state.logger.warning(
-                            "Job '%s' does not exist and could not be removed",
-                            str(job_specifier),
-                        )
-                    elif _unprotect:
-                        job.remove_child()
-                    else:
-                        job.remove()
-                except IOError as _:
-                    state.logger.debug(
-                        "hdf file does not exist. Removal from database will be attempted."
-                    )
-                    job_id = self.get_job_id(job_specifier)
-                    self.db.delete_item(job_id)
+            return
+        if self.db.view_mode:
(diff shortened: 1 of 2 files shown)


pyiron_base/project/generic.py
```diff
@@ -1700,23 +1700,19 @@ def _remove_jobs_helper(self, recursive=False, progress=True):
         """
         if not isinstance(recursive, bool):
             raise ValueError("recursive must be a boolean")
-        if not self.db.view_mode:
-            job_id_lst = self.get_job_ids(recursive=recursive)
-            if progress and len(job_id_lst) > 0:
-                job_id_lst = tqdm(job_id_lst)
-            for job_id in job_id_lst:
-                if job_id not in self.get_job_ids(recursive=recursive):
-                    continue
```
jan-janssen: If I understand this pull request correctly, then the primary change is removing this database query, correct? Maybe we should split this pull request in two parts. One containing just the removal of this part and the other the refactoring. Having both mixed makes it a bit hard to read.

jan-janssen: To clarify these changes a bit more I created two separate pull requests https://github.com/pyiron/pyiron_base/pull/1326 and https://github.com/pyiron/pyiron_base/pull/1325 - these contain the changes to the database queries but not the refactoring done in this pull request.

pyiron_base/project/generic.py
```diff
@@ -1185,27 +1185,27 @@ def remove_job(self, job_specifier, _unprotect=False):
...
+            if _unprotect:
+                job.remove_child()
             else:
-                raise EnvironmentError("copy_to: is not available in Viewermode !")
+                job.remove()
+        except IOError as _:
+            state.logger.debug(
+                "hdf file does not exist. Removal from database will be attempted."
+            )
+            self.db.delete_item(job.id)
```
XzzX, line 1207: @jan-janssen You missed this line. It might not be called often but it is another redundant database query.

jan-janssen: But this is dangerous. For example when you create a job object and remove the HDF5 file, then the loading of the job object with `inspect()` fails and returns an `IOError` still in the except case the job object is not available, so we cannot use `job.id`.

jan-janssen: Ok, I realised that the `inspect()` function also works when the file is corrupted https://github.com/pyiron/pyiron_base/pull/1327

PR #1325, opened 2024-02-12
get_jobs() filter for columns directly

jan-janssen:


XzzX:
Since the columns argument is handled at the pandas dataframe stage and not at the sql stage this does not improve database performance at the moment but perhaps in the future.

jan-janssen:
> Since the columns argument is handled at the pandas dataframe stage and not at the sql stage this does not improve database performance at the moment but perhaps in the future.

In this case I guess we can leave this pull request open and I just set it to draft.

PR #1326, opened 2024-02-12
Accelerate removing of jobs by removing consistency check

jan-janssen:


niklassiemer (commented):
I am ok with these changes - it is a huge performance bottleneck!
Do we plan to handle cases where a sub-job of a Master is deleted in some way in the database? or do we just say it is not that important, especially with the workflow based pyiron coming up? (Or did I understand something wrong?)

jan-janssen:
> Do we plan to handle cases where a sub-job of a Master is deleted in some way in the database? or do we just say it is not that important, especially with the workflow based pyiron coming up? (Or did I understand something wrong?)

That part is still handled in the `remove_job()` function.

niklassiemer:
Ok, it is handled in `JobCore.remove`. Then I do not see any issue with this!

XzzX:
There is a lot of conversion between `job` and `jobid` always involving a database query. Allowing `job` as an argument to `remove_job` can reduce the number of database queries even more.

PR #1333, opened 2024-02-14
Clean traceback when trying to import a moved module

pmrv:
If it's not from None we get two nested tracebacks, which hides the error


jan-janssen (approved):
Looks good to me

PR #1296, opened 2024-01-24
Enable parallelism inside the PythonFunctionContainerJob

jan-janssen:

PR #1050, opened 2023-02-27
Parallel pyiron table

jan-janssen:
Currently the initialisation for the individual subprocesses is too expensive, so it does not make a lot of sense to focus on a parallel pyiron table implementation. Still it is maybe a good point to start. Example code:
```
from pyiron_atomistics import Project
pr = Project("test")
pr.remove_jobs(silently=True, recursive=True)

job = pr.create.job.Lammps(job_name="lmp")
job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
murn = job.create_job(pr.job_type.Murnaghan, "murn")
murn.run()

def filter_function(job):
    return job.__name__ == "Lammps"

table = pr.create_table()
table.filter_function = filter_function
table.add.get_energy_tot
table.add.get_volume
table.server.cores = 4
table.run(delete_existing_job=True)

table.get_dataframe()
```

PR #1334, opened 2024-02-14
Use the same executor for pyiron table and function container

jan-janssen:
Combine the work in https://github.com/pyiron/pyiron_base/pull/1050 and https://github.com/pyiron/pyiron_base/pull/1296

Example code - for parallel pyiron table:
```python
pr = Project("test")
pr.remove_jobs(silently=True, recursive=True)

job = pr.create.job.Lammps(job_name="lmp")
murn = job.create_job(pr.job_type.Murnaghan, "murn")
murn.run()

def filter_function(job):
    return job.__name__ == "Lammps"

table = pr.create_table()
table.filter_function = filter_function
table.add.get_energy_tot
table.add.get_volume
table.server.cores = 4
table.run(delete_existing_job=True)

table.get_dataframe()
```

EDIT: syntax highlight flag


pyiron_base/jobs/datamining.py
```diff
@@ -340,44 +345,47 @@ def _get_filtered_job_ids_from_project(self, recursive=True):
...
+        executor: concurrent.futures.Executor = None,
+    ) -> List[dict]:
         """
         Apply functions to job.
 
         Any functions that raise an error are set to `None` in the final list.
 
         Args:
-            job_lst (list of JobPath): all jobs to analyze
+            job_id_lst (list of JobPath): all jobs to analyze
```
liamhuber: `type(my_job.id)` is `int` for pyiron jobs `my_job`; I find it confusing naming that `type(job_id_lst)` is `list[JobPath]`

pyiron_base/jobs/job/executor.py
```diff
@@ -0,0 +1,64 @@
+from pyiron_base.jobs.job.generic import GenericJob
+
+
+class JobWithExecutor(GenericJob):
```
liamhuber: Needs a docstring justifying its existence over `GenericJob`. Additionally the ability to specify the executor to be used by string needs to be documented somewhere, either here or at least on the property.

pyiron_base/jobs/job/template.py
```diff
@@ -49,3 +50,17 @@ def __init__(self, project, job_name):
         super().__init__(project, job_name)
         self._python_only_job = True
         self._write_work_dir_warnings = False
+
+
+class PythonTemplateJobWithExecutor(PythonTemplateJob, JobWithExecutor):
```
liamhuber: Do we really need a new class here, or can `PythonTemplateJob` just be expanded to _always_ inherit from `JobWithExecutor`? As far as I can tell the storage and computational overhead of `JobWithExecutor` is extremely minimal, so I'd just beef up the main class instead of proliferating new user-facing tools.

tests/flex/test_pythonfunctioncontainer.py
```diff
@@ -61,3 +66,13 @@ def test_with_executor_wait(self):
             self.assertFalse(job.server.future.done())
             self.project.wait_for_job(job=job, interval_in_s=0.01, max_iterations=1000)
             self.assertTrue(job.server.future.done())
+
+    def test_with_internal_executor(self):
+        job = self.project.wrap_python_function(my_function_exe)
+        job.input["a_lst"] = [1, 2, 3, 4]
+        job.input["b_lst"] = [5, 6, 7, 8]
+        job.server.cores = 2
+        job.executor_type = "ProcessPoolExecutor"
```
liamhuber: Tests don't handle failure case; what happens if I assign something weird here? What should happen/when should it happen? IMO behaviour should be to get an exception at assignment time, and that should be codified in the tests, but I think right now you don't hit problems until you call `run` and I don't know whether the error message is informative or not.

pyiron_base/jobs/job/executor.py
```diff
@@ -0,0 +1,64 @@
...
+        super().__init__(project, job_name)
+        self._executor_type = None
+
+    @property
+    def executor_type(self):
+        return self._executor_type
+
+    @executor_type.setter
+    def executor_type(self, exe):
+        self._executor_type = exe
```
liamhuber: At a minimum, the setter should be used as an opportunity to fail early by comparing to a set of allowed values and raising an exception if it's not there.

At best, we can actually relax the allowable executors by just making sure it's available and that it's likely to follow the right UI (could use `inspect` to really verify the signature, but I won't do that here). 

Add the following imports:

```python
from concurrent.futures import Executor as StdLibExecutor
from inspect import isclass

from h5io import _import_class 
```
(Or redefine it somewhere more convenient is probably better, I just want to minimize my typing in the github review browser interface)

Then

```suggestion
    @staticmethod
    def _is_a_subclass_of_executor(obj):
        return isclass(exe) and issubclass(exe, StdLibExecutor):

    @executor_type.setter
    def executor_type(self, exe):
        if isinstance(exe, str):
            try:
                exe_class = _import_class(exe)  # Make sure it's available
                if not self._is_a_subclass_of_executor(exe_class):  # And what we want
                    raise TypeError(
                        f"{exe} imported OK, but {exe_class} is not a subclass of {StdLibExecutor}"
                    )
            except Exception as e:
                raise ImportError("Something went wrong trying to import {exe}") from e
        elif isclass(exe) and issubclass(exe, StdLibExecutor):
            self._executor_type = f"{exe.__module__}.{exe.__name__}"
        elif isinstance(exe, StdLibExecutor):
            raise NotImplementedError(
                "We don't want to let you pass an entire executor, because you might think its state comes "
                "with it. Try passing `.__class__` on this object instead."
            )
        else:
            raise TypeError(
                f"Expected an executor class or string representing one, but got {exe}"
            )
```

pyiron_base/jobs/job/executor.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def _get_executor(self, max_workers=None):
```
liamhuber: This is rather clunky, to the point that the `elif isinstance(self._executor_type, str):` case error message even misses one of the available options.

Ideally, safety can have already be done in the setter, and this can just be the `None` check and an import and instantiation, super short. If for some reason we don't want to allow arbitrary executors but only a whitelisted set, this set should be better abstracted, e.g. into some class like `_AllowedExecutors` with properties (to delay import, as is done here), where we can exploit the `__dir__()` or whatever to dynamically generate the error message so it includes all the choices.

pyiron_base/jobs/job/executor.py
liamhuber: Note that this is also a bit of a UI change, since we insist on `"concurrent.futures."` even ahead of the standard thread and process pool guys. If you still want to accept just the class name for these two special and common cases, it's easy enough to add the extra logic here.

liamhuber (changes requested):
Functionality seems good, I have requested changes but I see no fundamental barriers. Non-nit requests:

- I see no need for a new python job class, just power up the main one
- `JobWithExecutor.executor_type` needs to be more robustly handled for early failure, and I propose a way to make it both more elegant in implementation and flexible
- Test needs to cover the failure case of a bad executor assignment

jan-janssen:
Just for awareness I continue to benchmark and optimise the performance of the `pympipool` executor in https://github.com/pyiron/pympipool/pull/264

pyiron_base/jobs/datamining.py
jan-janssen: I updated the DocStrings.

pyiron_base/jobs/job/executor.py
jan-janssen: I removed this class and implemented the functionality in the `GenericJob` class.

pyiron_base/jobs/job/template.py
jan-janssen: I removed this class and implemented the functionality in the `GenericJob` class.

pyiron_base/jobs/job/executor.py
jan-janssen: I created `pyiron_base.jobs.job.extension.executor.EXECUTORDICT` as a central place to manage the available Executors.

pyiron_base/jobs/job/executor.py
jan-janssen: I currently tend to prefer white listing selected executors, rather than supporting all of them by default. This allows us to test the executors to check if they are working with `pyiron_base`. For example adding support for `dask` is technically possible, but I would prefer somebody test it first before we add it to the white list.

tests/flex/test_pythonfunctioncontainer.py
jan-janssen: I added additional tests for the failure case.

jan-janssen:
> Functionality seems good, I have requested changes but I see no fundamental barriers. Non-nit requests:
> 
> * I see no need for a new python job class, just power up the main one
> * `JobWithExecutor.executor_type` needs to be more robustly handled for early failure, and I propose a way to make it both more elegant in implementation and flexible
> * Test needs to cover the failure case of a bad executor assignment

@liamhuber Thanks for your review, I tried to address the individual points in the separate comments. I just do not feel so easy to allow all possible executors, I prefer to whitelist existing executors and when somebody wants to use specific executors with `pyiron_base` and adds the corresponding tests I am happy to add those to the white list. Apart from this I have an intrinsic motivation to promote the use of `pympipool` in particular as I am currently working on improving the performance of `pympipool` in comparison to the built-in Process Pool executor https://github.com/pyiron/pympipool/pull/264 .

tests/flex/test_pythonfunctioncontainer.py
```diff
@@ -61,3 +66,19 @@ def test_with_executor_wait(self):
...
+        job = self.project.wrap_python_function(my_function_exe)
+        job.input["a_lst"] = [1, 2, 3, 4]
+        job.input["b_lst"] = [5, 6, 7, 8]
+        job.server.cores = 2
+        with self.assertRaises(TypeError):
+            job.executor_type = "Executor"
+        with self.assertRaises(TypeError):
+            job.executor_type = ProcessPoolExecutor
+        with self.assertRaises(TypeError):
+            job.executor_type = None
```
liamhuber: # (Blocking) Be able to turn it off

Seeing this test, it actually becomes clear to me that this is the opposite behaviour we want. `job._executor_type = None` is the default state, which is to not use an executor, so I believe we explicitly _do_ want to allow this value to be set in order to allow users to turn off executor functionality.

pyiron_base/jobs/job/extension/executor.py
```diff
@@ -0,0 +1,37 @@
...
+
+    return PyFluxExecutor(max_workers=max_workers)
+
+
+EXECUTORDICT = {
+    "ThreadPoolExecutor": get_thread_executor,
+    "ProcessPoolExecutor": get_process_executor,
+    "PyMPIPoolExecutor": get_pympipool_mpi_executor,
+    "PyMPIPoolSlurmExecutor": get_pympipool_slurm_executor,
+    "PyMPIPoolFluxExecutor": get_pympipool_flux_executor,
```
liamhuber: # (Minor, non-blocking) UX consistency

It is a bit weird to me that some of these incorporate the module path and some don't. Not too bad for users, but the pattern/intent is unclear. What convention should future devs follow for extending the whitelist? 

Using a dict of functions demands that a choice be made for this at all, and in this particular case the difference in treatment and module capitalization makes it unclear exactly what that choice was.

liamhuber (changes requested):
One blocking comment re: turning the functionality back off and one minor comment re: UX.

Otherwise, I won't insist on these but I would encourage you to carefully consider this feedback, which is two closely related ideas:

> I just do not feel so easy to allow all possible executors, I prefer to whitelist existing executors and when somebody wants to use specific executors with `pyiron_base` and adds the corresponding tests I am happy to add those to the white list. 

# (Non-blocking) Walled-garden

By adhering to a strict whitelist paradigm, you force new users to adopt the pyiron ecosystem entirely, or avoid pyiron. If a user comes along with some new executor they want to use, they need to explicitly modify `pyiron_base` _just to check if it works_.

In contrast, if we place some reasonable constraint on the value (i.e. inherits from the base `concurrent.futures.Executor` to guarantee basic interface), and give guidance that certain only executors are _guaranteed_ to work, then users can experiment with new things and pull pyiron into their preferred toolset. Then, if it works, they are free to suggest that their executor be added to the test suite, etc.

IMO this is the much more pythonic approach, and avoids locking-out users with particular executor demands.

> Apart from this I have an intrinsic motivation to promote the use of `pympipool` in particular...

# (Non-blocking) Nudge don't force

This is an extremely fair perspective. I would encourage nudging users instead of forcing them to adopt a particular behaviour though. In this case, the `pympipool` executors should be "promoted" and "supported", so we mention them by name, have tests to guarantee they work, and lower the barrier to using them relative other executors.

This is exactly what I do over in `pyiron_workflow`, where `Node.executor` _must_ inherit from the base executor to guarantee interface, warnings are made that executors need to support serializing dynamic stuff with docs directing users to `pympipool`, and then the `pympipool` executors are made as accessible as possible by storing them in `Workflow.create.executor`.

I would suggest a similar architecture here, so that `pympipool` holds a place of extreme privilege without being exclusive.

pyiron_base/jobs/job/executor.py
liamhuber: >but I would prefer somebody test it first before we add it to the white list.

Per my main comment in the re-review, by having a strict whitelist we force that person to clone and modify `pyiron_base` to do this test. A more relaxed approach lets that testing happen without modifications to base itself.

tests/flex/test_pythonfunctioncontainer.py
jan-janssen: I added the option to reset the executor.

pyiron_base/jobs/job/extension/executor.py
jan-janssen: The `pyiron_base/jobs/job/extension/executor.py` is now removed.

jan-janssen:
@liamhuber I followed your suggestion and added the option to define executors as both strings and python classes. From my perspective this pull request is now ready to be merged.

tests/flex/test_pythonfunctioncontainer.py
```diff
@@ -61,3 +66,17 @@ def test_with_executor_wait(self):
...
+    def test_with_internal_executor(self):
+        job = self.project.wrap_python_function(my_function_exe)
+        job.input["a_lst"] = [1, 2, 3, 4]
+        job.input["b_lst"] = [5, 6, 7, 8]
+        job.server.cores = 2
+        with self.assertRaises(ImportError):
+            job.executor_type = "Executor"
+        job.executor_type = ProcessPoolExecutor
+        job.executor_type = None
+        job.executor_type = "concurrent.futures.ProcessPoolExecutor"
```
liamhuber: Better to extend these so to guarantee `._get_executor` then has the expected behaviour for each case

liamhuber (approved):
LGTM.

I recommend extending the tests to be a bit more thorough for the three executor cases.

tests/flex/test_pythonfunctioncontainer.py
jan-janssen: Done

PR #1335, opened 2024-02-14
Fix multi dimensional arrays with FlattenedStorage.to_pandas

pmrv:


jan-janssen (changes requested):
To guarantee the stability of these kinds of conversions it would be great to add tests.

jan-janssen:
@pmrv Can you merge `main` into your branch? Especially as the directory structure for the tests changed?

pmrv:
> @pmrv Can you merge `main` into your branch? Especially as the directory structure for the tests changed?

Should be fixed now.

pyiron_base/_tests.py
```diff
@@ -43,9 +43,9 @@ class PyironTestCase(unittest.TestCase, ABC):
     def setUp(self):
         self.addTypeEqualityFunc(np.ndarray, self._assert_equal_numpy)
 
-    def _assert_equal_numpy(self, a, b, msg=None):
+    def _assert_equal_numpy(self, a, b, msg=""):
```
jan-janssen: Is this necessary? I thought with the changes in line 48 now `None` is supported and maybe the better choice.

tests/unit/storage/test_flattenedstorage.py
```diff
@@ -945,8 +963,21 @@ def test_to_pandas(self):
...
                     )
 
-        dfe = store.to_pandas(explode=True)
+        try:
+            dfe = store.to_pandas(explode=True)
+        except:
+            # Regression test
+            self.fail(
+                "to_pandas must not fail when multidimensional arrays are present!"
+            )
```
jan-janssen: I am a bit confused about the `try except` part. I agree that it gives a nicer error message but at the same time it is not as readable.

pyiron_base/_tests.py
pmrv:
```suggestion
    def _assert_equal_numpy(self, a, b, msg=None):
```

jan-janssen (approved):
Looks good to me

PR #1336, opened 2024-02-14
Fix regression on #1188

pmrv:
A refactor in between must have missed renaming the variable that the check is using, essentially making it always False.

#1188


jan-janssen (approved):
Looks good to me

PR #1341, opened 2024-02-16
Catch signal of aborted job in job wrapper

jan-janssen:
We discussed this issue in the Friday FAQ meeting and @ahmedabdelkawy agreed to test it once in production, by reducing the `run_time` of a given calculation so the `SLURM` queuing system kills the job and to check if `pyiron_base` correctly identifies the job as killed and the status is set to aborted.


pmrv (approved):

jan-janssen:
I added a couple of tests, the pul request is now ready to review again.

PR #1343, opened 2024-02-16
Try to access files after HDF

pmrv:
Currently job[x] first checks whether x is a file in the directory of job.  Crucially it also does this when the files are compressed and when job was obtained in inspect mode.  Together this conspires to make pyiron tables much slower, because every single table entry will search the compress tar archive for (non-existing) files first.

This commit reorders the access so that first the HDF files is searched for x, before looking for local files, based on two assumptions:
1. accessing HDF is the common case and should be faster
2. users that want to read files and avoid searching HDF can simply use job.files

### DIFF
--- a/pyiron_base/jobs/job/core.py
+++ b/pyiron_base/jobs/job/core.py
@@ -907,13 +907,6 @@ class JobCore(HasGroups):
             dict, list, float, int, :class:`.DataContainer`, None: data or data object; if nothing is found None is returned
         """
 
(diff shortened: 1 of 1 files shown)


jan-janssen (approved):
looks good to me

niklassiemer (approved):
With almost certain no overlap (which would cause a behavior change) between filenames and hdf keys and the deprecation of this sort if file access, I am in favor of this change!

PR #1352, opened 2024-02-23
Fix the project type when the project is reloaded

### LINKED ISSUE #1351: Project type is not correctly set when reloading jobs
```python
from pyiron_base import Project as Project_base
from pyiron_atomistics import Project as Project_atomistics


pr_atomistics = Project_atomistics("test")
job = pr_atomistics.create.job.Lammps("lmp")
job.structure = pr_atomistics.create.structure.ase.bulk("Al")
job.run()

pr_base = Project_base("test")
job_reload = pr_base.load(job.job_name)

type(job.project), type(job_reload.project)
```
The output is: 
```
(pyiron_atomistics.project.Project, pyiron_base.project.generic.Project)
```

jan-janssen:
@srmnitc Can you check if this fixes the issue with `self.project.create` ?


srmnitc:
This fixes the issue, I can confirm.

PR #1349, opened 2024-02-22
Remove project.sql_query

pmrv:
In lieu of Niklas.

PR #1356, opened 2024-02-23
[patch] Introduce a flag for _not_ mangling the name of wrapped function jobs

liamhuber:
Uses a private interface and defaults false, so this is totally backwards compatible.

```python
def my_function(x):
    return x

job = self.project.wrap_python_function(my_function)
job.input["x"] = 42
job._mangle_name_on_save = False
job.save()
assert(job.job_name, my_function.__name__)
```


liamhuber:
The osx 12 tests failed, but it was a timeout and I have trouble seeing in the log but it looks like a completely unrelated test. I'm trying to just re-run, but the button doesn't do anything on my browser...


```python
/Users/runner/work/pyiron_base/pyiron_base/pyiron_base/jobs/datamining.py:651: PerformanceWarning: 
your performance may suffer as PyTables will pickle object types that it cannot
map directly to c-types [inferred_type->mixed,key->block1_values] [items->Index(['name', 'array'], dtype='object')]

  self.pyiron_table._df.to_hdf(
.
  0%|          | 0/4 [00:00<?, ?it/s]
100%|██████████| 4/4 [00:00<00:00, 41.70it/s]
TestWithProject: Setting up test project
.
The job test_a was saved and received the ID: 6
Loading and filtering jobs:   0%|          | 0/5 [00:00<?, ?it/s]
The job test_b was saved and received the ID: 7
Loading and filtering jobs: 100%|██████████| 5/5 [00:00<00:00, 645.32it/s]
The job test_c was saved and received the ID: 8

The job test_d was saved and received the ID: 9
The job test_table was saved and received the ID: 10
```

liamhuber:
> The osx 12 tests failed, but it was a timeout and I have trouble seeing in the log but it looks like a completely unrelated test. I'm trying to just re-run, but the button doesn't do anything on my browser...

Indeed, the same timeout error appears on #1349. It went through OK on #1352 though, so I'll probably try just re-running a few times...

pyiron_base/jobs/flex/pythonfunctioncontainer.py
```diff
@@ -64,21 +66,28 @@ def __call__(self, *args, **kwargs):
     def to_hdf(self, hdf=None, group_name=None):
         super().to_hdf(hdf=hdf, group_name=group_name)
         self.project_hdf5["function"] = np.void(cloudpickle.dumps(self._function))
+        self.project_hdf5["_mangle_name_on_save"] = self._mangle_name_on_save
```
jan-janssen: Is it necessary to store this part in the HDF5 file? Basically the `to_hdf()` function is called as part of the save function and when the save function is called the action - in this case renaming the job - is already executed. Or to ask it the other way around when is it required that the property is stored in the HDF5 file?

pyiron_base/jobs/flex/pythonfunctioncontainer.py
```diff
@@ -64,21 +66,28 @@ def __call__(self, *args, **kwargs):
...
+
+            self.job_name = job_name
+
+            if self.job_name in self.project.list_nodes():
+                self.from_hdf()
+                self.status.finished = True
+            else:
+                super().save()
         else:
             super().save()
```
jan-janssen:
```suggestion
        super().save()
```

pyiron_base/jobs/flex/pythonfunctioncontainer.py
```diff
@@ -44,6 +44,8 @@ def __init__(self, project, job_name):
         super().__init__(project, job_name)
         self._function = None
         self._executor_type = None
+        self._mangle_name_on_save = True  # Automatically rename job using function and
```
jan-janssen:
```suggestion
        self._overwrite_job_name_by_hash = True
```
 I feel when we have to explain the variable name in a comment, then maybe the variable name is not intuitive.

jan-janssen (changes requested):
In principle I agree with the changes, I just find the variable name `_mangle_name_on_save` hard to understand and I do not understand why this information is stored in the HDF5 file.

pyiron_base/jobs/flex/pythonfunctioncontainer.py
liamhuber: Yes, it is strictly necessary. Without saving this, the data is lost on re-load and you wind up with weird job renaming. In the following, _without_ saving the flag, the final job table does not contain `"my_job"` (but rather `"plus_one{hash}"` and `reloaded` is None (instead of a job):

```python
from pyiron_base import Project, JOB_CLASS_DICT

from pyiron_base.jobs.flex.pythonfunctioncontainer import PythonFunctionContainerJob

JOB_CLASS_DICT[PythonFunctionContainerJob.__name__] = PythonFunctionContainerJob.__module__

pr = Project("tmp")
pr.remove_jobs(silently=True, recursive=True)

assert(len(pr.job_table()) == 0)

def plus_one(x):
    return x + 1

# job = PythonFunctionContainerJob(pr, "my_job")  # Gave me an attribute error?!!?
job = pr.create.job.PythonFunctionContainerJob("my_job")
job.python_function = plus_one
job.input["x"] = 2
job._mangle_name_on_save = False

job.run()

loaded = pr.load("my_job")
loaded.run(delete_existing_job=True)

print(pr.job_table()["job"])
reloaded = pr.load("my_job")
type(reloaded)
```

pyiron_base/jobs/flex/pythonfunctioncontainer.py
liamhuber: This breaks the behaviour as we no longer avoid saving when the job is already found, but I agree it's more readable so I've committed it and will add an explicit `return` to recover the correct behaviour.

pyiron_base/jobs/flex/pythonfunctioncontainer.py
liamhuber: Yeah, I was aware of this deficiency -- I fixed it in [the stacked PR](https://github.com/pyiron/pyiron_base/pull/1358/), but it was still unprofessional of me to just leave it knowingly-bad here.

I find your suggestion loses information on when/why we would rename this, and focuses on the implementation (hashing) instead of the functionality (leveraging input values), so I opted to go with `_automatically_rename_on_save_using_input`.

This is definitely an improvement, although in the stacked PR I'm still inclined to leave to comment to give even more depth to why we're doing this.

pyiron_base/jobs/flex/pythonfunctioncontainer.py
jan-janssen: Ok, that makes sense.

jan-janssen (approved):
Looks good to me

PR #1364, opened 2024-03-02
Add two classes for HasHDF/HasDict compat

pmrv:
and outline transition path in docstrings there.

### DIFF
--- a/pyiron_base/interfaces/has_dict.py
+++ b/pyiron_base/interfaces/has_dict.py
@@ -112,6 +112,7 @@ class HasDict(ABC):
             obj_dict (dict): data previously returned from :meth:`.to_dict`
             version (str): version tag written together with the data
         """
+
         def load(inner_dict):
             if not isinstance(inner_dict, dict):
                 return inner_dict
--- a/pyiron_base/interfaces/has_dict.py
+++ b/pyiron_base/interfaces/has_dict.py
@@ -1,9 +1,25 @@
 # coding: utf-8
 # Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
 # Distributed under the terms of "New BSD License", see the LICENSE file.
-"""Interface for classes to serialize to dictionary."""
+"""Interface for classes to serialize to dictionary.
+
+This also contains classes to ease the transition from HDF based storage to
+dict based serialization.  Roughly we want to proceed as follows:
+    1. Any new object directly implements :class:`.HasDict` and derives from
+    :class:`.HasHDFfromDict` to be compatible to older code that still relies
+    on the HDF interface.
+    2. Any old object that doesn't yet directly implements :class:`.HasDict`
+    can trivially derive from :class:`.HasDictfromHDF` to be compatible to
+    newer code that uses the dict based serialization.
+    3. Step by step we can transition old objects to directly implement
+    :class:`.HasDict`.
+"""
 
 from abc import ABC, abstractmethod
+from typing import Any
+
+from pyiron_base.interfaces.has_hdf import HasHDF
+from pyiron_base.storage.hdfio import DummyHDFio
 
 __author__ = "Jan Janssen"
 __copyright__ = (
@@ -25,10 +41,10 @@ class HasDict(ABC):
         pass
 
     @abstractmethod
-    def to_dict(self):
+    def _to_dict(self):
         pass
 
-    def _type_to_dict(self):
+    def to_dict(self):
         type_dict = {
             "NAME": self.__class__.__name__,
             "TYPE": str(type(self)),
@@ -37,4 +53,60 @@ class HasDict(ABC):
         }
         if hasattr(self, "__version__"):
             type_dict["VERSION"] = self.__version__
-        return type_dict
+        return self._to_dict() | type_dict
+
+    @staticmethod
+    def _join_children_dict(
+            children: dict[str, dict[str, Any]]
+    ) -> dict[str, Any]:
+        """
+        Given a nested dictionary, flatten the first level.
+
+        >>> d = {'a': {'a1': 3}, 'b': {'b1': 4, 'b2': {'c': 42}}}
+        >>> _join_children_dict(d)
+        {'a/a1': 3, 'b/b1': 4, 'b/b2': {'c': 42}}
+
+        This is intended as a utility function for nested HasDict objects, that
+        to_dict their children and then want to give a flattened dict for
+        writing to ProjectHDFio.write_dict_to_hdf
+        """
+        return {'/'.join((k1, k2)): v2 for k1, v1 in children.items()
+                                        for k2, v2 in v2.items()}
+
+
+class HasHDFfromDict(HasHDF, HasDict):
+    """
+    Implements HasHDF in terms of HasDict.
+
+    This class is intended for "new-style" objects that are used in a context
+    that only assumes that they implements HasHDF.
+
+    Implementors may still override :meth:`.HasHDF._get_hdf_group_name`.
+    """
+
+    def _from_hdf(self, hdf, version=None):
+        self.from_dict(hdf.read_dict_from_hdf(recursive=True))
+
+    def _to_hdf(self, hdf):
+        hdf.write_dict_to_hdf(self.to_dict())
+
+class HasDictfromHDF(HasDict, HasHDF):
+    """
+    Implements HasDict in terms of HasHDF.
+
+    This class is intended for "old-style" objects that should be able to be
+    used as children for objects that already implement HasDict and expect
+    their children to implmement it.
+    """
+
+    def from_dict(self, obj_dict: dict, version: str = None):
+        # DummyHDFio(project=None) looks a bit weird, but it was added there
+        # only to support saving/loading jobs which already use the HasDict
+        # interface
+        hdf = DummyHDFio(None, '/', obj_dict)
+        self.from_hdf(hdf)
+
+    def _to_dict(self):
+        hdf = DummyHDFio(None, '/')
+        self.to_hdf(hdf)
+        return hdf.to_dict()
--- a/pyiron_base/interfaces/has_dict.py
+++ b/pyiron_base/interfaces/has_dict.py
@@ -44,7 +44,7 @@ class HasDict(ABC):
     def _to_dict(self):
         pass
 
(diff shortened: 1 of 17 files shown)


pmrv:
@jan-janssen and I discussed something like this last week.    The problem statement is essentially: We want everything as to_dict, but don't want to convert all to_hdf at once.  Some code also still expects to be able to call to_hdf and again we don't want to change those all at one as well.  So this introduces two mixins that can translate implementers of HasHDF and HasDict between each other. If we use them we can then slowly fade out to_hdf implementation by manually converting them to HasHDF over time.  Once all objects are converted, we deprecate HasHDFfromDict.to_hdf to show us what we need to update in the surrounding code, and when that is done we remove both classes again.

Tests will need to follow and it's not yet used by any objects, but I want to discuss this plan and the design first.

pyiron_base/interfaces/has_dict.py
```diff
@@ -37,4 +53,61 @@ def _type_to_dict(self):
...
+    Implements HasHDF in terms of HasDict.
+
+    This class is intended for "new-style" objects that are used in a context
+    that only assumes that they implements HasHDF.
+
+    Implementors may still override :meth:`.HasHDF._get_hdf_group_name`.
+    """
+
+    def _from_hdf(self, hdf, version=None):
+        entries = self._get_hdf_entries()
```
```suggestion
```

pyiron_base/interfaces/has_dict.py
```diff
@@ -37,4 +53,60 @@ def _type_to_dict(self):
...
+
+    def _to_hdf(self, hdf):
+        hdf.write_dict_to_hdf(self.to_dict())
+
+class HasDictfromHDF(HasDict, HasHDF):
+    """
+    Implements HasDict in terms of HasHDF.
+
+    This class is intended for "old-style" objects that should be able to be
+    used as children for objects that already implements HasDict and expect
```
```suggestion
    used as children for objects that already implement HasDict and expect
```

pyiron_base/storage/parameters.py
```diff
@@ -500,14 +500,13 @@ def define_blocks(self, block_dict):
...
+    def _to_dict(self):
         """
         Convert the GenericParameters object to a dictionary for storage
 
         Returns:
             dict: GenericParameters object as a dictionary
         """
-        data_dict = self._type_to_dict()
         data_dict["data_dict"] = self._dataset
         return data_dict
```
jan-janssen:
```suggestion
        return {"data_dict": self._dataset}
```

pyiron_base/storage/parameters.py
```diff
@@ -500,14 +500,13 @@ def define_blocks(self, block_dict):
...
 
-    def to_dict(self):
+    def _to_dict(self):
         """
         Convert the GenericParameters object to a dictionary for storage
 
         Returns:
             dict: GenericParameters object as a dictionary
         """
-        data_dict = self._type_to_dict()
```
jan-janssen, line 512: Removing this line causes the tests to fail

pmrv:
```
Error: The action 'Test' has timed out after 5 minutes.
```

pyiron_base/interfaces/has_dict.py
```diff
@@ -25,10 +41,10 @@ def from_dict(self, obj_dict: dict, version: str = None):
         pass
 
     @abstractmethod
-    def to_dict(self):
+    def _to_dict(self):
         pass
 
-    def _type_to_dict(self):
+    def to_dict(self):
```
```suggestion
    @HasDict.to_dict
    def to_dict(self):
```
What about this so people don't have to put `_...?`

pmrv: Do literature search what is common in the community?  Some reference to OOP best practices?

liamhuber: I poked around for this, but I found nothing conclusive. The best I got was [this SO thread](https://stackoverflow.com/questions/30599316/using-abstract-private-methods-as-an-interface) from eight years ago; they lay out the same choices we're already aware of (OP's default was to do it the way you have here, if that's anything), but the question is left unanswered as "well, that's opinion". Personally, I find the `@abstractmethod` flag on the private method of the same name the clearest...but I agree it's opinion.

pmrv: I would like to stick to the current convention then.  Since this is a private class used internally anyway, it should be fairly easy to change in the future anyway, in case we decide differently.

pmrv:
I've added a test in form of the `DataContainer` implementation of `HasDictFromHDF` and adapted `DummyHDFio` to work with this.

@jan-janssen I'd prefer to merge this in before proceeding with your `to_dict` series of PRs.   Changes needed there to use this would hopefully be minimal.

I've noticed as well when running the tests locally that there is one location where we create a job with `run_mode=="manual"` that hangs for a long time before the tests proceed.  This may be the issue of the timeout problems we are seeing with the CI, but I couldn't locate the test case yet.  I have a feeling this may be related to the `files` changes with the decompression that we did recently.  I have also seen a bunch unrelated test failures locally.

jan-janssen:
> @jan-janssen I'd prefer to merge this in before proceeding with your `to_dict` series of PRs. Changes needed there to use this would hopefully be minimal.

The only pull requests which use `HasDict` are https://github.com/pyiron/pyiron_base/pull/1380 and https://github.com/pyiron/pyiron_base/pull/1377 . As those are required for https://github.com/pyiron/pyiron_atomistics/pull/1356 I would prefer to merge those first before we change the interface.

pmrv:
The reason for the unit tests hanging is this line

https://github.com/pyiron/pyiron_base/blob/0bdde9b73eb32f85688808898b208f716b606689/tests/job/test_worker.py#L26

pmrv:
> The reason for the unit tests hanging is this line
> 
> https://github.com/pyiron/pyiron_base/blob/0bdde9b73eb32f85688808898b208f716b606689/tests/job/test_worker.py#L26

What I don't understand is why this is suddenly a problem as this  PR doesn't touch worker code at all.

pmrv:
Seems this only breaks, once I touched `GenericParameters`.

pmrv:
@jan-janssen I can't seem to fix this quickly today, so go ahead with the other PRs.

pyiron_base/jobs/job/extension/executable.py
```diff
@@ -203,13 +203,10 @@ def executable_path(self, new_path):
...
-        executable_dict = self._type_to_dict()
-        executable_storage_dict = self.storage._type_to_dict()
-        executable_storage_dict["READ_ONLY"] = self.storage._read_only
-        executable_storage_dict.update(self.storage.to_builtin())
-        executable_dict["executable"] = executable_storage_dict
-        return executable_dict
+    def _to_dict(self):
+        full_dict = self.storage.to_dict()
+        data = full_dict.pop("data")
+        return {"executable": {**full_dict, **data}}
```
jan-janssen: In principle, I like the idea with the `DummyHDFio`. But for classes like the `Executable`, I am not even sure if they need a `DataContainer`. I think the goal is to get to `pickle`-able classes in the future, with the `to_dict()` and `from_dict()` method being part of the transition. So I feel switching to `_to_dict()` here goes in the wrong direction.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -602,24 +602,22 @@ def view_queues():
...
+            "queue": self.queue,
+            "qid": self._queue_id,
+            "cores": self.cores,
+            "threads": self.threads,
+            "new_h5": self.new_hdf,
+            "structure_id": self.structure_id,
+            "run_time": self.run_time,
+            "memory_limit": self.memory_limit,
+            "accept_crash": self.accept_crash,
+        }
```
jan-janssen: In analogy to the `Executable` class, does the `Server` class need the transition from `to_dict()` to `_to_dict()` or can we modify the class in a way that the `__getstate__()` returns the same dictionary as `to_dict()`. Again my focus would be on simplifying the class hierarchy.

pmrv: Both Server and Executable derived from `HasDict` before this PR.  I'm ok with changing this in the future, but here I just did the minimal changes to keep existing sub classes working with the changes I did to `HasDict` (which is just adding the type info automatically to the results of to_dict).

jan-janssen (approved):
Looks good to me

pyiron_base/jobs/job/extension/server/generic.py
jan-janssen: I add this to the bucket list. I would like to replace both with data classes sooner rather than later then we can potentially also reuse them in the functional approach.

pyiron_base/jobs/job/extension/executable.py
pmrv: The DummyHDFio is not used for the executable or the data container, so I'm not sure what you mean.  I agree that in the future `Executable` could be a simple dataclass or so, but again here I just modified it, so it conforms to the new _to_dict.

jan-janssen:
@pmrv As https://github.com/pyiron/pyiron_base/pull/1578 and  https://github.com/pyiron/pyiron_base/pull/1580 are now merged - can you update this pull request, so we can include it in the next release? I think the three pull requests together streamline the `to_dict()` and `from_dict()` interface, so I would like to release them as part of `pyiron_base=0.10.0`.

pmrv:
contrib tests fail because its dependencies cannot currently be installed concurrently with the ones from atomistics.

pmrv:
This is in a good state now, but I'm still trying to add recursive to_dict/from_dict (in a separate PR).

jan-janssen:
@pmrv Can this be merged? Or should we wait for ta new `h5io_browser` version to include https://github.com/h5io/h5io_browser/pull/61 ?

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -561,10 +561,9 @@ def view_queues():
         else:
             return None
 
-    def to_dict(self):
-        server_dict = self._type_to_dict()
+    def _to_dict(self):
         self._data.run_mode = self._run_mode.mode
-        server_dict.update(asdict(self._data))
+        return asdict(self._data)
         return server_dict
```
jan-janssen, line 567:
```suggestion
        return asdict(self._data)
```

pmrv:
Both this and #1602 dont need the h5io fix, thats just something I found while trying to get the jobs read their dicts from HDF recursively.

PR #1365, opened 2024-03-02
Add Lockable to FlattenedStorage

pmrv:
I tend to have it on `HasStoredTraits` objects, which I can lock after a workflow is done, but since flattened storages are mutable, I could still mess with them by accident.


jan-janssen:
The tests are currently broken, so I mark this pull request as draft.

PR #1369, opened 2024-03-12
pyiron table: Change default executor to "pympipool.mpi.executor.PyMPIExecutor"

### LINKED ISSUE #1362: Tables broken with default executor
```python
tab = pr.create.table(...)
tab.add[...] = ...
tab.server.cores = 4
tab.run()
```

currently breaks, because by default the plain process pool is used.

https://github.com/pyiron/pyiron_base/blob/dec8f0931be9b289ba87ed1bf34f1caaa9b6a9c4/pyiron_base/jobs/datamining.py#L787

Since all pyiron tables use lambdas to wrap the user defined functions, this gives a pickle error.

I suppose using an executor from `pympipool` would fix this, but I couldn't check yet.

jan-janssen:
Change the default executor from `concurrent.futures.ProcessPoolExecutor` to `pympipool.mpi.executor.PyMPIExecutor` as the `ProcessPoolExecutor` does not support `lambda` functions.

### DIFF
--- a/pyiron_base/jobs/datamining.py
+++ b/pyiron_base/jobs/datamining.py
@@ -784,7 +784,7 @@ class TableJob(GenericJob):
             self.project.db.item_update({"timestart": datetime.now()}, self.job_id)
         with self.project_hdf5.open("input") as hdf5_input:
             if self._executor_type is None and self.server.cores > 1:
-                self._executor_type = "concurrent.futures.ProcessPoolExecutor"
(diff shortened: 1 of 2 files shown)


pmrv:
I've updated the `create_table` docstring to highlight the problem and point others to `pympipool`, but otherwise lgtm.  Thanks for the fix.

pmrv (approved):
lgtm

PR #1370, opened 2024-03-12
Switch to use conda as default

jan-janssen:
We require the `conda` package to be available, so we can expect that the `conda` executable is available but that does not mean that the `mamba` executable is available.

### DIFF
--- a/pyiron_base/project/condaenv.py
+++ b/pyiron_base/project/condaenv.py
@@ -21,7 +21,7 @@ class CondaEnvironment:
             )
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


pmrv (approved):
lgtm.

Is it actually possible to add mamba to our (optional) deps and set this to True again in the future?

jan-janssen:
> Is it actually possible to add mamba to our (optional) deps and set this to True again in the future?

Technically yes, still as conda is now using the mamba solver I do not see an advantage of using mamba anymore:
https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community

PR #1372, opened 2024-03-12
Extend conda interface: store conda environments in project

### LINKED ISSUE #1371: More flexible conda environments
pyiron at the moment uses the central conda environment path to store the environments. This is rather error-prone, so it would be better to store the conda environments also in the project path.

jan-janssen:
This pull request changes the location of the conda environment and stores it inside the project in a `conda` folder. In addition, it checks if an environment already exists and in that case just raises a warning rather than recreating the environment.


srmnitc:
How would this play out when there is limited number of files possible in the file system? For example, on a HPC cluster.  One might want to keep the project on a backed up space, but the conda environment there would be an overkill.

jan-janssen:
> How would this play out when there is limited number of files possible in the file system? For example, on a HPC cluster. One might want to keep the project on a backed up space, but the conda environment there would be an overkill.

I would exclude all directories which have `conda` in the name from the backup. It is similar to the way snakemake does it. The advantage is that it is more clear for the user to see which environment belongs to which project.

srmnitc:
> > How would this play out when there is limited number of files possible in the file system? For example, on a HPC cluster. One might want to keep the project on a backed up space, but the conda environment there would be an overkill.
> 
> I would exclude all directories which have `conda` in the name from the backup. It is similar to the way snakemake does it. The advantage is that it is more clear for the user to see which environment belongs to which project.

Is this possible, for example on cmti. I feel a lot of users might simply not know it. I do see the advantages. I suggest we discuss this in one of the pyiron meetings.

jan-janssen:
I would like to include this functionality in the `0.8` release of `pyiron_base` and as we have no pyiron meeting next week - given the easter holidays - it would be great to discuss this asynchronously.

srmnitc:
I still have my reservations about storing the whole environment in the project. Mainly about the number of files created and backup as I mentioned. Additionally for example wont the `pr.pack` function now include the environment files itself? But maybe @pmrv or others can also take a look.

jan-janssen:
> Mainly about the number of files created and backup as I mentioned. 

You can still create a central project and store all your environments there, it is just the default that is changing. 

> Additionally for example wont the `pr.pack` function now include the environment files itself? 

This is correct. Still I guess in most cases people are anyway copying their jobs to an archive project as discussed in https://github.com/pyiron/FAQs/issues/46

pyiron_base/project/generic.py
```diff
@@ -222,8 +222,9 @@ def conda_environment(self):
             raise ImportError(
                 "You need to have the conda python package installed to access conda environments."
             ) from None
-
-        return CondaEnvironment()
+        conda_dir = os.path.join(self.path, "conda")
+        os.makedirs(conda_dir, exist_ok=True)
+        return CondaEnvironment(env_path=conda_dir)
```
pmrv: Would it not be possible to make the conda path configurable here, that would side step @srmnitc concerns, which I think are valid.

jan-janssen:
@pmrv and @srmnitc based on your suggestion, I kept the default option to create a new central conda environment. Still now the users can disable the global installation and create a conda environment only in a given project using the `global_installation=False` option:
```python
pr.conda_environment.create(env_name="test", env_file="environment.yml", global_installation=False)
```

srmnitc (approved):
lgtm

PR #1373, opened 2024-03-13
ExecutableContainerJob: store command line output in job.output

jan-janssen:
Previously, the output of the command line was only stored in the `error.out` file, now it is also directly stored in the datacontainer of the job output: `job.output.command_line`

### DIFF
--- a/pyiron_base/jobs/flex/executablecontainer.py
+++ b/pyiron_base/jobs/flex/executablecontainer.py
@@ -79,6 +79,9 @@ class ExecutableContainerJob(TemplateJob):
                 working_directory=self.working_directory,
             )
 
+    def run_static(self):
+        self.storage.output.command_line = super().run_static()
+
     def collect_output(self):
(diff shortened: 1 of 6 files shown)


pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -79,6 +79,9 @@ def write_input(self):
                 working_directory=self.working_directory,
             )
 
+    def run_static(self):
+        self.storage.output.command_line = super().run_static()
```
pmrv: I would call this `stdout` or something like this, because when I read `command_line` I'd expect the command that was run, otherwise looks good to me.

pmrv (approved):

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: I like the suggestion and renamed `command_line` to `stdout`.

PR #1375, opened 2024-03-15
raise exception on failed delete from database

XzzX:


XzzX:
The exception is also raised if the job does not exist. Do we want to handle this separately?

jan-janssen:
@XzzX Can you add a unit test for this functionality? We already have a test for deleting jobs:
https://github.com/pyiron/pyiron_base/blob/main/tests/database/test_database_access.py#L184
So it would be great to add a test to delete a job that does not exist.

XzzX:
> @XzzX Can you add a unit test for this functionality? We already have a test for deleting jobs: https://github.com/pyiron/pyiron_base/blob/main/tests/database/test_database_access.py#L184 So it would be great to add a test to delete a job that does not exist.

You are already testing this, aren't you? There is a normal delete and a raise exception delete... I am confused. In my notebook there is no exception fired.

jan-janssen:
@XzzX the current test only covers the regular delete and trying to submit the job ID as list which is not supported. There is currently no test to delete a job which no longer exists:
```python
def test_delete_item(self):
        """
        Tests delete_item function
        Returns:
        """
        par_dict = self.add_items("BO")
        key = par_dict["id"]
        self.database.delete_item(key)
        self.assertRaises(
            Exception, self.database.delete_item, [key]
        )  # use only str or int
        # self.assertRaises(Exception, self.database.get_item_by_id, key)  # ensure item does not exist anymore
```

XzzX:
> @XzzX the current test only covers the regular delete and trying to submit the job ID as list which is not supported. There is currently no test to delete a job which no longer exists:
> 
> ```python
> def test_delete_item(self):
>         """
>         Tests delete_item function
>         Returns:
>         """
>         par_dict = self.add_items("BO")
>         key = par_dict["id"]
>         self.database.delete_item(key)
>         self.assertRaises(
>             Exception, self.database.delete_item, [key]
>         )  # use only str or int
>         # self.assertRaises(Exception, self.database.get_item_by_id, key)  # ensure item does not exist anymore
> ```

Aaaah, you are checking for a type error. Now also the comment makes sense. Ok, got it.

jan-janssen (approved):
Looks good to me

PR #1376, opened 2024-03-15
warn if job could not be deleted

XzzX:
Since debug messages are not printed by default I think this should be a warning. In particular since it resolves exceptions.


jan-janssen (approved):
Looks good to me

PR #1377, opened 2024-03-16
Executable: Switch from HasHDF to HasDict

jan-janssen:
This pull request reverts parts of https://github.com/pyiron/pyiron_base/pull/728 to simplify the `to_dict()` and `from_dict()` interface of the `GenericJob` class. In the future the interface should use `__getstate__()` and `__setstate__()`.

### DIFF
--- a/pyiron_base/jobs/job/extension/executable.py
+++ b/pyiron_base/jobs/job/extension/executable.py
@@ -3,7 +3,8 @@
 # Distributed under the terms of "New BSD License", see the LICENSE file.
 
 import os
-from pyiron_base.interfaces.object import HasStorage
+from pyiron_base.interfaces.has_dict import HasDict
+from pyiron_base.storage.datacontainer import DataContainer
 from pyiron_base.state import state
 
 """
@@ -22,7 +23,7 @@ __status__ = "production"
 __date__ = "Sep 1, 2017"
 
 
-class Executable(HasStorage):
+class Executable(HasDict):
     __hdf_version__ = "0.3.0"
 
     def __init__(
@@ -42,6 +43,7 @@ class Executable(HasStorage):
             overwrite_nt_flag (bool):
         """
         super().__init__()
+        self.storage = DataContainer()
         self.storage.table_name = "executable"
 
         if path_binary_codes is None:
--- a/pyiron_base/jobs/job/extension/executable.py
+++ b/pyiron_base/jobs/job/extension/executable.py
@@ -238,13 +238,9 @@ class Executable(HasDict):
             "mpi",
             "accepted_return_codes",
         ]
-        self.storage.update(
-            {
-                key: executable_dict["executable"][key]
-                for key in data_container_keys
-                if key in executable_dict["executable"].keys()
-            }
-        )
+        for key in data_container_keys:
+            if key in executable_dict["executable"]:
+                self.storage[key] = executable_dict["executable"][key]
         if executable_dict["executable"]["READ_ONLY"]:
             self.storage.read_only = True
 
--- a/pyiron_base/jobs/job/extension/executable.py
+++ b/pyiron_base/jobs/job/extension/executable.py
@@ -237,7 +237,7 @@ class Executable(HasStorage):
             'accepted_return_codes'
         ]
         self.storage.update({
-            executable_dict["executable"][key]
+            key: executable_dict["executable"][key]
             for key in data_container_keys
             if key in executable_dict["executable"].keys()
         })
--- a/pyiron_base/jobs/job/extension/executable.py
+++ b/pyiron_base/jobs/job/extension/executable.py
@@ -23,6 +23,8 @@ __date__ = "Sep 1, 2017"
 
 
 class Executable(HasStorage):
+    __hdf_version__ = "0.3.0"
+
     def __init__(
         self,
         path_binary_codes=None,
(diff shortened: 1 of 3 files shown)


pyiron_base/jobs/job/generic.py
```diff
@@ -50,6 +50,7 @@
 from pyiron_base.jobs.job.extension.server.generic import Server
 from pyiron_base.database.filetable import FileTable
 from pyiron_base.interfaces.has_dict import HasDict
+from pyiron_base.storage.datacontainer import DataContainer
```
pmrv, line 53:
```suggestion
```
Not used.

pyiron_base/jobs/job/extension/executable.py
```diff
@@ -217,6 +221,33 @@ def executable_path(self, new_path):
...
+            "mpi",
+            "accepted_return_codes",
+        ]
+        self.storage.update(
+            {
+                key: executable_dict["executable"][key]
+                for key in data_container_keys
+                if key in executable_dict["executable"].keys()
+            }
+        )
```
pmrv:
```suggestion
        for key in data_container_keys:
            if key in executable_dict["executable"]:
                self.storage[key] = executable_dict["executable"][key]
```
More readable imo.

pyiron_base/jobs/job/generic.py
```diff
@@ -1087,17 +1093,9 @@ def to_hdf(self, hdf=None, group_name=None):
             hdf (ProjectHDFio): HDF5 group object - optional
             group_name (str): HDF5 subgroup name - optional
         """
-
         self._set_hdf(hdf=hdf, group_name=group_name)
-        self._executable_activate_mpi()
```
pmrv, line 1092: Seem unrelated to the other changes, is that generally not necessary anymore?

pyiron_base/jobs/job/generic.py
```diff
@@ -1124,10 +1122,12 @@ def from_hdf(self, hdf=None, group_name=None):
...
         with self._hdf5.open("input") as hdf5_input:
             job_dict["input"] = hdf5_input.read_dict_from_hdf(recursive=True)
-        self.from_dict(job_dict=job_dict)
-
+        # Backwards compatibility to the previous HasHDF based interface
         if "executable" in self._hdf5.list_groups():
-            self.executable.from_hdf(self._hdf5)
+            exe_dict = self._hdf5["executable/executable"].to_object().to_builtin()
+            exe_dict["READ_ONLY"] = self._hdf5["executable/executable/READ_ONLY"]
+            job_dict["executable"] = {"executable": exe_dict}
```
pmrv, line 1129: This is triggered in both cases when on "old" executable or a "new" one is present, right?  Does `to_object().to_builtin()` really work in both cases?  Imo the preferable solution would be to keep the `Executable.from_hdf` around and have there logic that checks `__hdf_version__` and either does the old loading logic or `from_dict`s.

pmrv:
Actually wouldn't it be much easier to just give `HasStorage` a `HasDict` implementation?  That would be quite trivial, since it's just 
```python
def to_dict(self):
  return self.storage.to_builtin()
def from_dict(self, data_dict):
  self.storage.clear()
  self.storage.update(data_dict)
```

pyiron_base/jobs/job/extension/executable.py
```diff
@@ -40,6 +43,7 @@ def __init__(
             overwrite_nt_flag (bool):
         """
         super().__init__()
+        self.storage = DataContainer()
```
jan-janssen, line 46: @pmrv We still use the `DataContainer` here

pyiron_base/jobs/job/generic.py
jan-janssen: The `DataContainer` is still used in line 46.

pyiron_base/jobs/job/generic.py
```diff
@@ -1049,6 +1050,9 @@ def to_dict(self):
             "exclude_groups_hdf": self._exclude_groups_hdf,
         }
         data_dict["server"] = self._server.to_dict()
+        self._executable_activate_mpi()
```
jan-janssen, line 1053: @pmrv The `self._executable_activate_mpi()` was just moved up the to the point where the `to_dict()` method is called on the `executable` object.

pyiron_base/jobs/job/generic.py
jan-janssen: The function was just moved to line 1053.

pyiron_base/jobs/job/generic.py
jan-janssen: No the important part is the `list_groups()` call. The previous implementation created an HDF group, the new one just a single node with the content of the node being a JSON dictionary.

jan-janssen:
> Actually wouldn't it be much easier to just give `HasStorage` a `HasDict` implementation? That would be quite trivial, since it's just
> 
> ```python
> def to_dict(self):
>   return self.storage.to_builtin()
> def from_dict(self, data_dict):
>   self.storage.clear()
>   self.storage.update(data_dict)
> ```

As I commented in https://github.com/pyiron/pyiron_base/pull/1398 I do not think it is that easy but I am happy if you convince me otherwise.

pyiron_base/jobs/job/generic.py
pmrv: Ah, makes sense.  I still find it not so nice, that now `GenericJob` needs to know about backwards compat things from `Executable` instead of executable handling everything on its own, but I suppose it works.

PR #1378, opened 2024-03-16
PythonFunctionContainerJob and ExecutableContainerJob: Use to_dict() and from_dict()

jan-janssen:
Replace to_hdf() and from_hdf() with to_dict() and from_dict()

### DIFF
--- a/pyiron_base/jobs/flex/executablecontainer.py
+++ b/pyiron_base/jobs/flex/executablecontainer.py
@@ -86,24 +86,25 @@ class ExecutableContainerJob(TemplateJob):
             )
             self.to_hdf()
 
-    def to_hdf(self, hdf=None, group_name=None):
-        super().to_hdf(hdf=hdf, group_name=group_name)
+    def to_dict(self):
+        job_dict = self.to_dict()
         if self._write_input_funct is not None:
-            self.project_hdf5["write_input_function"] = np.void(
+            job_dict["write_input_function"] = np.void(
                 cloudpickle.dumps(self._write_input_funct)
             )
         if self._collect_output_funct is not None:
-            self.project_hdf5["collect_output_function"] = np.void(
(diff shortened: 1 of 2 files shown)


pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -86,24 +86,25 @@ def collect_output(self):
...
             self._write_input_funct = cloudpickle.loads(
-                self.project_hdf5["write_input_function"]
+                job_dict["write_input_function"]
             )
-        if "write_input_function" in self.project_hdf5.list_nodes():
+        if "write_input_function" in job_dict.keys():
             self._collect_output_funct = cloudpickle.loads(
-                self.project_hdf5["collect_output_function"]
+                job_dict["collect_output_function"]
             )
```
pmrv, line 110: If I understand correctly `GenericJob.from_hdf` calls `hdf.read_dict_from_hdf` with `recursive=False`, hence "write_input_function" would never be in `job_dict`.  Same below for the python function container.

jan-janssen: There are two parts in the `from_hdf()` method on `GenericJob`:
```python
        """
        Restore the GenericJob from an HDF5 file

        Args:
            hdf (ProjectHDFio): HDF5 group object - optional
            group_name (str): HDF5 subgroup name - optional
        """
        self._set_hdf(hdf=hdf, group_name=group_name)
        job_dict = self._hdf5.read_dict_from_hdf()
        with self._hdf5.open("input") as hdf5_input:
            job_dict["input"] = hdf5_input.read_dict_from_hdf(recursive=True)
        self.from_dict(job_dict=job_dict)
```
The first `read_dict_from_hdf()` call reads the nodes at the root level of the job group in the HDF5 file and then the second `read_dict_from_hdf(recursive=True)` recursively reads the nodes in the group `input`.

PR #1380, opened 2024-03-16
job.server: remove to_hdf() and from_hdf()

jan-janssen:
The to_hdf() and from_hdf() function are replaced by the to_dict() and from_dict() function.

### DIFF
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -409,7 +409,9 @@ def run_job_with_runmode_queue(job):
     )
     if que_id is not None:
         job.server.queue_id = que_id
-        job._server.to_hdf(job._hdf5)
+        job.project_hdf5.write_dict(
+            data_dict={"server": job.server.to_dict()}
+        )
         print("Queue system id: ", que_id)
     else:
         job._logger.warning("Job aborted")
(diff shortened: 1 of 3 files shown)


jan-janssen:
The `run_job_with_runmode_queue(job)` in `runfunction.py` still requires the `to_hdf()` from the `job.server` object: 
https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/jobs/job/runfunction.py#L412
This was not included in the tests, as it only appears in interaction with queuing systems.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -409,7 +409,7 @@ def run_job_with_runmode_queue(job):
     )
     if que_id is not None:
         job.server.queue_id = que_id
-        job._server.to_hdf(job._hdf5)
```
pmrv, line 412: This is what I worry about, maybe we can keep the hdf methods around for a bit and simply issue warnings in case they are called, since they are using the dict methods internally already anyway.

jan-janssen: From my perspective the `job.server` object is internal to `pyiron_base` so nobody should modify it. Still I am fine with adding a deprecation warning from `0.8` to `0.9`.

pmrv: I'm thinking more in case there are other places where we missed that this is still called.

PR #1381, opened 2024-03-16
ScriptJob: from to_hdf() and from_hdf()

jan-janssen:


pyiron_base/jobs/script.py
```diff
@@ -274,54 +274,22 @@ def set_input_to_read_only(self):
...
-            with self.project_hdf5.open("input") as hdf5_input:
-                if "parallel" in hdf5_input.list_nodes():
-                    self._enable_mpi4py = hdf5_input["parallel"]
-                try:
-                    self.script_path = hdf5_input["path"]
-                except TypeError:
-                    pass
-                self.input.from_hdf(hdf5_input)
-        else:
-            raise ValueError("Cannot handle hdf version: {}".format(version))
```
pmrv, line 324: Will this still load older GenericParameters then?  I don't know off hand when then 0.1.0 -> 0.2.0 bump for GenericParameters was, though.

jan-janssen: You introduced these changes back in October 2020 https://github.com/pyiron/pyiron_base/pull/34 this corresponds to `pyiron_base=0.2.X` . So I am fine with removing this functionality with `pyiron_base=0.8`

pmrv: Yup, that's long enough ago that we don't need to care.

PR #1391, opened 2024-03-19
Move recursive loading of DataContainer to HDF5Content

pmrv:
This will allow users and internal code to use job.content[...] if they what they are looking for is saved in HDF5 storage and obviate the need to rely on job[...] which may be slow because it is also searching the compressed job files.

I will need some logic like this to fix the slow loading of pyiron_atomistics jobs.

### DIFF
--- a/pyiron_base/jobs/job/core.py
+++ b/pyiron_base/jobs/job/core.py
@@ -964,6 +964,7 @@ class JobCore(HasGroups):
             )
             return _job_read_file(self, item)
 
+        name_lst = item.split("/")
         item_obj = name_lst[0]
         if item_obj in self._list_ext_childs():
             # ToDo: Murn['strain_0.9'] - sucht im HDF5 file, dort gibt es aber die entsprechenden Gruppen noch nicht.
--- a/pyiron_base/jobs/job/core.py
+++ b/pyiron_base/jobs/job/core.py
@@ -953,7 +953,7 @@ class JobCore(HasGroups):
             dict, list, float, int, :class:`.DataContainer`, None: data or data object; if nothing is found None is returned
         """
         # first try to access HDF5 directly to make the common case fast
-        value = recursive_load_from_hdf(self._project_hdf5, item)
+        value = recursive_load_from_hdf(self._hdf5, item)
         if value is not None:
             return value
 
--- a/pyiron_base/jobs/job/core.py
+++ b/pyiron_base/jobs/job/core.py
@@ -108,6 +108,52 @@ _doc_str_job_core_attr = """\
 """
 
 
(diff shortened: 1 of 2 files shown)


pyiron_base/jobs/job/core.py
```diff
@@ -108,6 +108,52 @@
...
+        ):
+            return group.to_object(lazy=True)
+        else:
+            return group
+    except ValueError:
+        pass
+
+    name_lst = item.split("/")
+
+    def successive_path_splits(name_lst):
```
jan-janssen, line 141: Can you move the internal function to the beginning of the function it is defined in? In this case line 112? That would help a lot with the readability.

pmrv:
The last commit should actually fix https://github.com/pyiron/pyiron_atomistics/pull/1348 and https://github.com/pyiron/pyiron_atomistics/pull/1350, but I'll probably add them in anyway for cleanliness. 

As of this PR one can access various things in a job like this:

- `job.project_hdf5` access to the raw HDF5 layout
- `job.content` access to HDF, but try to load any DataContainer along the given path (this was previously exclusive to `job[...]`) this could be extended to other objects that can be `to_object()`ed.
- `job.files` access to files 
- `job[...]` will try to act first like `content`, then like `files` if a string without slashes was passed and finally will try to look up child jobs.

pyiron_base/jobs/job/core.py
```diff
@@ -906,60 +952,26 @@ def __getitem__(self, item):
...
+            return value
+
+        # only try to read files when no slashes are present:
+        # downstream code will often do something like job['path/to/output'] to check if certain values exist and branch
+        # on that.  In cases where they don't exists this would then trigger us to decompress the job files in memory on
+        # every check which slows down things a lot.  Generally these value checks will be of the form output/.../...
+        # i.e. contain slashes and file access tend to be just the file name without slashes, so I separate those cases
+        # here like this.  In those cases where we actually have sub directories in the job folders we can beef up the
+        # file browser.
+        if "/" not in item and item in self.files.list():
```
jan-janssen, line 996: I like the idea of using the `/` to identify items which are not file. At the same time this makes me wonder: Can the output files have a hierarchy? Like in principle there is no reason not to have subfolders in the working directory, even though we do not handle this case so far.

pmrv: Yeah, I think it's something we should handle, though it is rare to happen so far.  I think the correct location would be in the `FileBrowser` class.

jan-janssen (approved):
Looks good to me. Still we should think about adding a benchmark which times typical `job[item]` queries and make sure we do not merge changes in the future which slow down.

PR #1392, opened 2024-03-22
Misc FileBrowser features

pmrv:
- support sub directories
- support iteration over file content

@jan-janssen I've removed the str sub classing again, because it pollutes the tab completion a bit and didn't seem to be used anywhere anymore, but I don't have too strong feelings.

### DIFF
--- a/tests/flex/test_executablecontainer.py
+++ b/tests/flex/test_executablecontainer.py
@@ -100,7 +100,7 @@ class TestExecutableContainer(TestWithProject):
             self.assertTrue(file in dir(job.files))
         output_file_path = os.path.abspath(os.path.join(__file__, "..", "test_executablecontainer", "job_output_files_hdf5", "job_output_files", "error.out"))
         if os.name != "nt":
-            self.assertEqual(job.files.error_out, output_file_path)
+            self.assertEqual(str(job.files.error_out), output_file_path)
         else:
             self.assertEqual(job.files.error_out, output_file_path.replace("\\", "/"))
 
--- a/pyiron_base/jobs/job/extension/files.py
+++ b/pyiron_base/jobs/job/extension/files.py
@@ -89,6 +89,9 @@ class FileBrowser:
         return self[file].tail(lines=lines)
 
     def __getitem__(self, item):
+        sub = os.path.join(self._working_directory, item)
+        if os.path.isdir(sub):
+            return FileBrowser(sub)
         if item in _working_directory_list_files(
             working_directory=self._working_directory,
             include_archive=False,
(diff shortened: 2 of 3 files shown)


tests/flex/test_executablecontainer.py
```diff
@@ -100,9 +100,9 @@ def test_job_files(self):
             self.assertTrue(file in dir(job.files))
         output_file_path = os.path.abspath(os.path.join(__file__, "..", "test_executablecontainer", "job_output_files_hdf5", "job_output_files", "error.out"))
         if os.name != "nt":
-            self.assertEqual(job.files.error_out, output_file_path)
+            self.assertEqual(str(job.files.error_out), output_file_path)
```
jan-janssen, line 103: This makes it very confusing for the user. For example when we take a look at the NFDI example we want to be able to access the path of the files directly: https://github.com/BAMresearch/NFDI4IngScientificWorkflowRequirements/blob/main/exemplary_workflow/pyiron/workflow.py#L34

jan-janssen: We can fix this by supporting the `File` object as input for restart_files and convert it to a string there.

pmrv: `input_file_lst=[gmsh.files.square_msh.path]` would be an option. I don't think it would be very confusing, just extra typing.  I've tried to look through the places where `restart_file_lst/dict` are used and we could force a conversion there somewhere, but I'd be afraid I'll miss places.  I'll revert the commits that remove it sub class if you don't see an easy path in between as well.

jan-janssen: I provided a suggestion in https://github.com/pyiron/pyiron_base/pull/1397

jan-janssen:
@pmrv Should we merge this and then release a `pyiron_base=0.7.12` before we start with the `to_dict()` and `from_dict()` functions and the changes to the database https://github.com/pyiron/pyiron_base/pull/1396 for `pyiron_base=0.8.0`?

pmrv:
Can do once were back from lunch.

pmrv:
```
FAIL: test_with_executor_wait (flex.test_pythonfunctioncontainer.TestPythonFunctionContainer.test_with_executor_wait)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/flex/test_pythonfunctioncontainer.py", line 86, in test_with_executor_wait
```

jan-janssen (approved):
Looks good to me

jan-janssen:
@pmrv Is there any other pull request you would like to merge in `0.7.12` otherwise I am going to release it now and then start merging the `to_dict()` pull requests as well as the database changes.

pmrv:
> @pmrv Is there any other pull request you would like to merge in `0.7.12` otherwise I am going to release it now and then start merging the `to_dict()` pull requests as well as the database changes.

No, all fine with me.

PR #1394, opened 2024-03-23
Files: first check working directory then archive

jan-janssen:
I see the restarting of jobs getting very slow as we always check the archive if the file is located in there, still commonly restart files are stored outside of the archive. This pull request introduced an additions `os.listdir()` for the case when the file is inside the archive, but only calls this `os.listdir()` for the case the file is outside the archive.


pmrv (approved):
lgtm.

PR #1397, opened 2024-03-25
no longer derive from string

jan-janssen:


pmrv:
Yeah, this multiple fixes is what I was worried about in my comment in #1392.  It's not worth our time to debug minor issues like this for the whole day, I'll revert the sub class changes in #1392 and we can forgo the changes here, I'm not that into it that we need to invest both our times into it.

pyiron_base/jobs/job/generic.py
```diff
@@ -1043,8 +1048,10 @@ def to_dict(self):
         data_dict = self._type_to_dict()
         data_dict["status"] = self.status.string
         data_dict["input/generic_dict"] = {
-            "restart_file_list": self._restart_file_list,
-            "restart_file_dict": self._restart_file_dict,
+            "restart_file_list": [str(f) for f in self._restart_file_list],
+            "restart_file_dict": {
+                str(f): n for f, n in self._restart_file_dict.items()
+            },
```
jan-janssen: I do not think these changes are necessary, because the parameters are already converted to strings when setting the property.

pmrv: If I see the example correctly `wrap_executable` appends directly to the `restart_file_list` here

https://github.com/pyiron/pyiron_base/blob/c1aa43070dbc825788a4392785db22c4e8eb3eb8/pyiron_base/project/generic.py#L434

In the setters for `restart_file_list` I see no logic to coerce values to string as well

https://github.com/pyiron/pyiron_base/blob/c1aa43070dbc825788a4392785db22c4e8eb3eb8/pyiron_base/jobs/job/generic.py#L281

jan-janssen: Good catch

jan-janssen:
> Yeah, this multiple fixes is what I was worried about in my comment in #1392. It's not worth our time to debug minor issues like this for the whole day, I'll revert the sub class changes in #1392 and we can forgo the changes here, I'm not that into it that we need to invest both our times into it.

I think it is fixed now.

pyiron_base/jobs/job/generic.py
```diff
@@ -295,7 +298,9 @@ def restart_file_dict(self):
         """
         A dictionary of the new name of the copied restart files
         """
-        for actual_name in [os.path.basename(f) for f in self._restart_file_list]:
+        for actual_name in [os.path.basename(f) for f in self.restart_file_list]:
+            if isinstance(actual_name, File):
+                actual_name = str(actual_name)
```
pmrv:
```suggestion
```
Already handled now by the getter.

pmrv (approved):
otherwise lgtm.

jan-janssen:
> otherwise lgtm.

I noticed that we missed the case when the user sets a restart_dictionary directly, this is not frequently used still I fixed the issue now. As we are currently a bit asynchronous, I am fine if you merge this pull request at any time.

PR #1398, opened 2024-03-26
Implement HasDict for HasStorage

pmrv:
Hopefully this would make #1377 unnecessary and also convert a lot more objects at the same time.

### DIFF
--- a/pyiron_base/interfaces/object.py
+++ b/pyiron_base/interfaces/object.py
@@ -53,11 +53,14 @@ class HasStorage(HasDict, HasHDF, ABC):
         return self._group_name
 
     def to_dict(self):
-        return self.storage.to_builtin()
+        return {
+            "data": self.storage.to_builtin(),
+            **self._type_to_dict()
+        }
 
     def from_dict(self, obj_dict: dict, version: str = None):
         self.storage.clear()
-        self.storage.update(obj_dict)
+        self.storage.update(obj_dict["data"])
 
     def _to_hdf(self, hdf: ProjectHDFio):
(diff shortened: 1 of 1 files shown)


pyiron_base/interfaces/object.py
```diff
@@ -49,11 +52,22 @@ def storage(self) -> DataContainer:
     def _get_hdf_group_name(self) -> str:
         return self._group_name
 
+    def to_dict(self):
+        return self.storage.to_builtin()
```
jan-janssen: At this point we loose the type definition in the dictionary like the `__hdf_version__` and so on. In the same way if we just execute a naive `self.stroage.update()` then these special keys also appear in the storage.

pmrv: This (last commit) way it should now save the data into a separate entry.  I will need to test though, if this works with more complicated objects in the storage.

pmrv: Yeah, as I expected saving dict/list with complicated objects inside still breaks h5io, so this is a deal breaker for HasStorage.

jan-janssen: Maybe you can check if the object has a `to_dict()` method? So I would like to have this generalised interface, still at the moment it was easier for me to adjust the server class in https://github.com/pyiron/pyiron_base/pull/1380 then coming up with a universal solution.

pmrv: Yeah, the general solution will have to be to recurse through the dict and call `to_dict` if available and the reverse on loading, but I'd like to implement this more generally as part of `HasDict` after I figure out #1364

PR #1400, opened 2024-03-27
GenericJob: extend the compress() function

jan-janssen:
For codes like VASP we typically remove certain files before we compress the working directory and other files we leave out of the archive. This is now possible using `job.files_to_compress` and `job.files_to_remove`. Both properties return a list, so the user can append additional files.

### DIFF
--- a/pyiron_base/jobs/job/util.py
+++ b/pyiron_base/jobs/job/util.py
@@ -302,7 +302,7 @@ def _job_compress(job, files_to_compress=[], files_to_remove=[]):
                     if "tar" not in name and not stat.S_ISFIFO(os.stat(name).st_mode):
                         tar.add(name)
             for name in files_to_compress:
-                if "tar" not in name:
+                if name != _job_compressed_name(job):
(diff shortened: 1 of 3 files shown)


pyiron_base/jobs/job/util.py
```diff
@@ -295,11 +303,9 @@ def _job_compress(job, files_to_compress=None):
                         tar.add(name)
             for name in files_to_compress:
                 if "tar" not in name:
```
pmrv: Not introduced here, but would be nice to change ` if "tar" not in name:` to `name != _job_compressed_name(job)`.  Otherwise lgtm.

pmrv (approved):

pyiron_base/jobs/job/util.py
jan-janssen:
```suggestion
                if name != _job_compressed_name(job):
```

jan-janssen: Thanks - I updated the pull request

PR #1411, opened 2024-04-15
Use _ensure_str_array_size in FlattenedStorage.extend

### LINKED ISSUE #1407: Flattened storage extend cutting off str
When using the extend method of a FlattenedStorage identifier str values in the extend array are cut of after a len of 20 due to the way the FlattenedStorage initializes it with len 20 .
```
"identifier": np.empty(self._num_chunks_alloc, dtype=np.dtype("U20")),
```
 @pmrv do you have an idea how to check for this?

pmrv:
Should fix #1407.

@Leimeroth Can you give this a try and add your example as a test case?

### DIFF
--- a/pyiron_base/storage/flattenedstorage.py
+++ b/pyiron_base/storage/flattenedstorage.py
@@ -31,6 +31,7 @@ from pyiron_base.interfaces.lockable import Lockable, sentinel
 
 _CHARSIZE = np.dtype("U1").itemsize
 
+
 def _ensure_str_array_size(array, strlen):
     """
     Ensures that the given array can store at least string of length `strlen`.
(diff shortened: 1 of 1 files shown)


Leimeroth:
It works, but I am actually confused why. I am adding the EAM example for the publication notebook and the trainingcontainer has a byte encoded str (dtype.char S). I guess it is accessed in a way that decodes it when reading it from the hdf5 container. However, I am not sure whether this would always be the case, as right now the problem is only the 'identifier' which is defined as a property also.

pmrv:
HDF5 wants to store unicode as UTF-8, but numpy does UCS-32 (dtype 'U') internally, so the `to_hdf`/`from_hdf` convert raw UTF-8 encoded bytestrings (dtype 'S') for the file.

jan-janssen (approved):
Looks good to me

jan-janssen:
Does it make sense to submit this fix to `h5io`?

Leimeroth (approved):

jan-janssen:
@pmrv Can this be merged? I would like to release a new `pyiron_base` version and would like to include this pull request.

pmrv:
> Does it make sense to submit this fix to `h5io`?

Not sure what you mean, this way of storing unicode string arrays in HDF in general should be in h5io?

PR #1416, opened 2024-04-26
Unlock when calling from_hdf for DataContainer

raynol-dsouza:
Reloading a `DataContainer` object threw a warning that the object was locked when `self.clear()` was called in `from_hdf`.

Change 1: Run `from_hdf` while unlocked.

Change 2: Remove additional warning that the object is being unlocked.


pmrv:
The unittest failure was because the original commit also removed some code that triggered the warning the test was locking for.

PR #1419, opened 2024-05-02
meaningful error messages

XzzX:
Why do you replace the real error message with a custom message that does not know what happened?  
https://github.com/pyiron/pyiron_base/blob/0a6c4bb166908380761daa494d9250322c631cb1/pyiron_base/utils/error.py#L148

Example:
`FATAL:  password authentication failed for user "seibll"` is reported as `WARNING - Database busy with too many connections. Trying again in 0.1s. Tried 1 times so far...`

The real error message is only shown after all retries have failed.

Suggesting to print it immediately.

### DIFF
--- a/pyiron_base/utils/error.py
+++ b/pyiron_base/utils/error.py
@@ -145,7 +145,10 @@ def retry(
             return func()
         except error as e:
             logger.warn(
-                f"{msg} Trying again in {delay}s. Tried {i + 1} times so far..."
+                f"{e}"
+            )
+            logger.warn(
+                f"Trying again in {delay}s. Tried {i + 1} times so far..."
             )
             time.sleep(delay)
             delay *= delay_factor


pyiron_base/utils/error.py
```diff
@@ -145,7 +145,10 @@ def retry(
             return func()
         except error as e:
             logger.warn(
-                f"{msg} Trying again in {delay}s. Tried {i + 1} times so far..."
+                f"{e}"
+            )
+            logger.warn(
+                f"Trying again in {delay}s. Tried {i + 1} times so far..."
```
niklassiemer, line 151: What about 
```suggestion
                f"{msg} - Caused by {e.__class__.__name__}: {e}. Trying again in {delay}s. Tried {i + 1} times so far..."
```
to include the message?

XzzX: I think showing a general `msg` for a catch all errors will be misleading more often than it is right. Therefore, I think `e` should be printed. If you also want `msg` go ahead, I am ok with that.

niklassiemer: The point could be to inform the user what is currently not working - sometimes the 'normal' error messages are not helpful and it might be good to say something about the funcion which did not work like ' Connecting to database failed: {e.__class__.__name__: {e}. Will retry ....'. I.e. I also would change the message to not be a error message :)

pmrv:
It sounds more likely that `error` passed to the function is a too general error class.  @XzzX which exception is thrown by sqlalchemy in your case?  Can you post a full stack trace?

pyiron_base/utils/error.py
pmrv: The (not explicitly documented) intent for the function is to silence a specific, known exception temporarily. That's why `msg` is printed and not the exception itself, because the assumption is that the calling code expects this exception to happen sometimes.  It sounds like in your case, we catch too broadly, so I think a better fix is to narrow `errors` where ever the error originates in your case.

XzzX:
> It sounds more likely that `error` passed to the function is a too general error class. @XzzX which exception is thrown by sqlalchemy in your case? Can you post a full stack trace?

The error class captured and thrown is `sqlalchemy.exc.OperationalError`. The error can be easily reproduced by changing the password in `.pyiron` to something invalid.

PR #1429, opened 2024-05-13
Centralized ci

liamhuber:


liamhuber:
Build notebooks failure:

```
Input Notebook:  notebooks/import_example.ipynb
Output Notebook: notebooks/import_example-out.ipynb

Executing:   0%|          | 0/5 [00:00<?, ?cell/s]Executing notebook with kernel: python3

Executing:  20%|██        | 1/5 [00:01<00:05,  1.34s/cell]
Executing:  60%|██████    | 3/5 [00:01<00:00,  2.11cell/s]
Executing:  60%|██████    | 3/5 [00:01<00:01,  1.50cell/s]
  File "/home/runner/miniconda3/envs/my-env/bin/papermill", line 10, in <module>
    sys.exit(papermill())
             ^^^^^^^^^^^
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/papermill/cli.py", line 235, in papermill
    execute_notebook(
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/papermill/execute.py", line 131, in execute_notebook
    raise_for_execution_errors(nb, output_path)
  File "/home/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/papermill/execute.py", line 251, in raise_for_execution_errors
    raise error
papermill.exceptions.PapermillExecutionError: 
Exception encountered at "In [3]":
FileNotFoundError                         Traceback (most recent call last)
Cell In[3], line 1
----> 1 pr.unpack(origin_path="archive_folder",compress=True)
```

I don't know what the main branch CI is doing to stop this from failing, but I totally agree -- there is no such "archive_folder" in the repo to unpack. My best guess is that it's some sort of execution-order-dependence, in which case I think this version of the CI is rightly failing things -- the notebooks should simply run independently and this one doesn't seem to.

liamhuber:
Coverage failure:

```
Run coverage combine
  coverage combine
  coverage xml
  shell: /usr/bin/bash -l {0}
  env:
    WRITE_ENVIRONMENT_OUTPUT_ENV_FILE: ./environment.yml
    INPUT_RUN_POST: true
    CONDA: /home/runner/miniconda3
    CONDA_PKGS_DIR: /home/runner/conda_pkgs_dir
    GITHUB_TOKEN: ***
No data to combine
Error: Process completed with exit code 1.
```

This one confuses me. I don't understand why it can't find the coverage.

liamhuber:
@jan-janssen @niklassiemer @mbruns91 

Here's a demo of applying the centralized CI to another repo. I deleted all the original CI just to save time -- some of it we still need because the centralized guy doesn't do it -- atomistics-compat.yml, backwards.yml, and contrib-compat.yml, but I may be missing something. To get some others (e.g. black-on-tag) we need to bring in the rest of the reusable workflows, [like in pyiron_ontology](https://github.com/pyiron/pyiron_ontology/tree/main/.github/workflows).

At the moment, this targets a particular branch on pyiron/actions, so I don't want to merge it in as-is anyhow. Rather I mean it as a foundation for conversation.

Other than the two failure cases in the comments above, the one outstanding thing that really perplexes me is that I had to make the following modification to the env setup action:

```yaml
  - name: Install local code
    shell: bash -l {0}
    run: | 
      pip install --no-deps ${{ inputs.local-code-directory }}
```

to

```yaml
  - name: Install local code
    shell: bash -l {0}
    run: |
      pip install versioneer[toml]==0.29
      pip install --no-deps ${{ inputs.local-code-directory }} --no-build-isolation
```

I need _both_ changes, or the pip install hits a "malformed node" error. What's bizarre to me here is that neither `pyiron_ontology` (using pyproject.toml) nor `pyiron_workflow` (still using setup.py) seem to care about the difference. Those repos both seem to work with or without the versioneer and/or no-build-isolation changes. Do any of you have any insight into why pyiron_base would be picky about this while those repos are both completely sanguine to either pip install attack?

Once we get this demo the rest of the way ironed out, I'll release a new version of pyiron/actions and we can make a real PR targeting that.

.github/workflows/push-pull.yml
```diff
@@ -0,0 +1,20 @@
+name: Push-Pull
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  pyiron:
+    uses: pyiron/actions/.github/workflows/push-pull.yml@pyiron_base_lessons
```
jan-janssen, line 10: Is there an option to test a new python version? like 3.13 ?

.github/workflows/push-pull.yml
```diff
@@ -0,0 +1,20 @@
...
+    secrets: inherit
+    with:
+      do-commit-updated-env: false
+      do-build-docs: false  # the build-docs action tries to use sphinx, not jupyterbook
+      unit-test-dir: tests
+      coveralls-and-codacy-test-dir: tests
```
jan-janssen, line 20: Can we select `coveralls` and `codacy` independently?

.github/workflows/unittests.yml
```diff
@@ -1,75 +0,0 @@
...
-        python-version: ${{ matrix.python-version }}
-        miniforge-variant: Mambaforge
-        channels: conda-forge
-        channel-priority: strict
-        activate-environment: my-env
-        environment-file: .ci_support/environment.yml
-        use-mamba: true
-    - name: Test
-      shell: bash -l {0}
-      timeout-minutes: 5
```
jan-janssen, line 61: I am missing the option to set the `timeout-minutes` in the new actions.

.github/workflows/push-pull.yml
```diff
@@ -0,0 +1,20 @@
...
+      benchmark-test-dir: test_benchmarks
+      alternate-tests-env-files: .ci_support/environment-old.yml
+      alternate-tests-python-version: '3.10'
```
jan-janssen:
```suggestion
      alternate-tests-python-version: '3.9'
```

jan-janssen:
The Windows and Mac OS tests are still executed with Python 3.11 rather than 3.12.

jan-janssen:
The actions are approximately 15% slower - comparing cached environments to installing the environment directly. For example Python 3.12 now takes 2m 18s vs 2m 1s. This is the reason why I removed the cached environments on `pyiron_base` and `pyiron_atomistics`. Is there an option to disable the caching? At the same time it would be great to have a `conda list` command included again to see the installed versions, especially when a change in dependencies causes our tests to fail while all the dependencies we track remained the same.

jan-janssen:
> I don't know what the main branch CI is doing to stop this from failing, but I totally agree -- there is no such "archive_folder" in the repo to unpack. My best guess is that it's some sort of execution-order-dependence, in which case I think this version of the CI is rightly failing things -- the notebooks should simply run independently and this one doesn't seem to.

In which directory do you execute the notebooks? I feel with the centralised actions it is much harder to see what exactly is happening and how the jupyter notebooks are executed, making it harder to debug.

.github/workflows/push-pull.yml
liamhuber: Yes, of course. One can always use the `unit-tests` action individually in a separate workflow to get a new tests with minimal lines, but the reusable `push-pull` workflow already accommodates running the "main" version of pyiron (default 3.11) across up to three runners (default linux, windows, and macos) and the "main" runner (default linux) across up to three versions of python (default 3.10, 3.9, 3.12). I say "up to" because you can always disable one of these by setting it to "exclude".

Until recently it supported only two alternate versions of python. I am not aware of nice loop-making tools in github's yaml syntax, so new entries need to be manually added. If it becomes helpful to have more runners or python versions covered by default in the reusable workflow, it's an easy extension and I have no objection to adding them for you.

I confess this is not clearly explained at a high level, but all this knowledge is available in the workflow argument descriptions:
https://github.com/pyiron/actions/blob/4bac4483cd063937b27871f5eade0e68a546c7ed/.github/workflows/push-pull.yml#L88-L122

.github/workflows/push-pull.yml
liamhuber: Yeah, sure. Nice idea.

I just pushed a commit to the branch we're currently targeting that slaps feature flags on these steps so they can be disabled independently: https://github.com/pyiron/actions/pull/113/commits/feeecf66caa8c91b7cb66a3a107109c2608e75f9

And then broke the current `do-coveralls-and-codacy` flag down to its constituent parts in the main `push-pull` workflow: https://github.com/pyiron/actions/pull/113/commits/823fee013dcd0c68734b6d890ff403d5ae31088b

.github/workflows/unittests.yml
liamhuber: This is a bit trickier. [Github currently does not support passing `timeout-minutes` to composite workflows](https://github.com/orgs/community/discussions/50481). What we _can_ do is leverage it in `pyiron/actions/unit-tests` in the same step you do here, then expose it as an input variable. For the workhorse `pyiron/.github/workflows/push-pull.yml` I wouldn't want an explosion of input options, but am comfortable with a single timeout that gets applied to all invocations of the unit tests.

liamhuber:
> The Windows and Mac OS tests are still executed with Python 3.11 rather than 3.12.

Sure, super easy. Just relates to my earlier comment about the version and runner flags. Changed in [this commit](https://github.com/pyiron/pyiron_base/pull/1429/commits/344ffd44d946b61c13a221dc6ae52251b5581417). If we have a later version where '3.12' is the default "main" version, these flags will be able to be removed.

liamhuber:
> At the same time it would be great to have a conda list command included again to see the installed versions, especially when a change in dependencies causes our tests to fail while all the dependencies we track remained the same.

That's an excellent idea. I added a `conda list` step to the environment setup action, so we should see it now. https://github.com/pyiron/actions/pull/113/commits/cde2c032d5c0074276ddb5c9a78dcec0fc746142

liamhuber:
> Is there an option to disable the caching?

There is an option to disable caching in the `pyiron/actions/cached-miniforge` action, but I don't expose that input in the downstream reusable workflows.

> The actions are approximately 15% slower - comparing cached environments to installing the environment directly. For example Python 3.12 now takes 2m 18s vs 2m 1s. This is the reason why I removed the cached environments on `pyiron_base` and `pyiron_atomistics`. 

When I was testing the cache flags in `pyiron_ontology`, I saw >2x speedup by using the cache. Because I was testing both the `use-cache` flag and the `cache-number` flag, I saw this result at least twice. Running _without_ the cache, the 17s difference you quote above is comparable to the variance I saw between different runs.

Further, if I dig into the logs, I see that here the unit tests themselves took 88s:

Screen Shot 2024-05-14 at 09 04 55

While in the [other PR](https://github.com/pyiron/pyiron_base/actions/runs/9082054103/job/24957454222?pr=1430) using the existing workflows they took 73s:

Screen Shot 2024-05-14 at 09 05 11

So 15s of the 18s difference has nothing to do with the environment setup.

I can accept that caching may not provide significant speedup, but setting up the environment from scratch involves both downloading and solving, whereas using the cache only involves unpacking. I can imagine situations where the speedup from caching is minimal (rather than the 2x+ observed in `pyiron_ontology`) or even non-existent, but I don't see a mechanism for it to be substantially slower. Your claim is counter-intuitive and IMO the burden of proof lies with you on this one.

jan-janssen (changes requested):
I am fine with the deploy action as discussed in https://github.com/pyiron/pyiron_base/pull/1430 . For merging all actions into one CI workflow, I feel we loose in flexibility by increasing the complexity making the CI environment harder to maintain. Based on this I am not convinced about the benefits of this pull request

liamhuber:
> > I don't know what the main branch CI is doing to stop this from failing, but I totally agree -- there is no such "archive_folder" in the repo to unpack. My best guess is that it's some sort of execution-order-dependence, in which case I think this version of the CI is rightly failing things -- the notebooks should simply run independently and this one doesn't seem to.
> 
> In which directory do you execute the notebooks? I feel with the centralised actions it is much harder to see what exactly is happening and how the jupyter notebooks are executed, making it harder to debug.

You're correct that the logs are less nice -- github provides nice drop-down tabs separating steps of a workflow, but when those steps are nested inside some other action or reusable workflow this nice tab structure is lost. There are ways of formatting step output, so I suspect we could recover this... but ultimately the log data is still all there, still the same, and I'm not willing to invest that much time just to recover the formatting. So, I agree, we could probably fix it, but I won't.

Now to this problem, we can complain about the log formatting, and throw fingers at the centralized CI, but this just ignores the core of my comment -- where's the damned file??? My point was that I couldn't figure out from looking the current main branch CI how it was passing, because one of the notebooks references a file that doesn't exist.

This morning I now have time to dig deeper into it, and, unsurprisingly, it's execution-order-dependence. The current CI runs the notebooks in alphabetical order, and so the fact that `import_example.ipynb` explicitly depends on the `export_example.ipynb` just happens to be OK. Or at least appear OK, of course it's terrible and we should fix it.

The centralized CI runs the notebooks just like the existing CI does, but is in general more powerful because the [existing CI](https://github.com/pyiron/pyiron_base/blob/main/.ci_support/build_notebooks.sh) only hits notebooks in the `notebooks/` dir, while the [centralized CI](https://github.com/pyiron/actions/blob/main/.support/build_notebooks.sh) (a) allows the notebook-containing directory to be specified as a variable, and (b) allows the directory to have internal structure and will recursively search for notebooks inside it.

It is possible to change the centralized CI to sort alphabetically like `ls` does, but I have no intention of doing this. It is right and good that the tests fail here, and the problem is with `pyiron_base` itself.

liamhuber:
> I am fine with the deploy action as discussed in #1430 . For merging all actions into one CI workflow, I feel we loose in flexibility by increasing the complexity making the CI environment harder to maintain. Based on this I am not convinced about the benefits of this pull request

You're the maintainer for this repo, so that's absolutely your prerogative. For me, I can only say that the complexity of having extra layer (workflow/reusable workflow/actions+steps vs workflow/actions+steps) is significantly preferable to the complexity of code duplication (+subtle differences) across many repos.

PR #1430, opened 2024-05-13
Update just the deloy action

jan-janssen:


.github/workflows/deploy.yml
```diff
@@ -28,10 +28,10 @@ jobs:
         channel-priority: strict
         auto-update-conda: true
         environment-file: .ci_support/environment.yml
-    - name: Convert dependencies
-      run: |
-        cp .ci_support/environment-old.yml environment.yml
-        python .ci_support/release.py; cat pyproject.toml
+    - uses: pyiron/actions/update-pyproject-dependencies@main
```
liamhuber:
```suggestion
    - uses: pyiron/actions/update-pyproject-dependencies@actions-2.0.7
```

A particular tag should always be targeted with actions (both ours and everyone else's) to avoid the behaviour changing under our feet.

liamhuber (changes requested):
Other than the tag target, looks good and I expect it to work smoothly.

liamhuber (approved):

PR #1437, opened 2024-05-17
Force user to use `bool` for `delete_existing_job`

samwaseda:
I constantly fall into the trap to set a job name like:

```python
job = pr.create.job.Lammps("lmp", strain)
```

without realising I had to write more parentheses. As the first argument after the job name is `delete_existing_job`, this often erases the job, which just shouldn't happen for a small typo.

### DIFF
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -759,7 +759,8 @@ class GenericJob(JobCore, HasDict):
         """
         if not isinstance(delete_existing_job, bool):
             raise ValueError(
-                "We got delete_existing_job = " + str(delete_existing_job)
+                "We got delete_existing_job = "
+                + str(delete_existing_job)
                 + ". If you meant to delete the job, set delete_existing_job"
                 + " = True"
             )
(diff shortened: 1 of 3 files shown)


liamhuber (commented):
I like fixing this problem, but I actually totally disagree with the attack. Let's rather go back to whatever `__init__` is responsible and just add `*` to [force the subsequent values to be keyword-only](https://peps.python.org/pep-3102/).

samwaseda:
> I like fixing this problem, but I actually totally disagree with the attack. Let's rather go back to whatever `__init__` is responsible and just add `*` to [force the subsequent values to be keyword-only](https://peps.python.org/pep-3102/).

That’s actually more or less what I suggested when the possibility to put multiple keywords was introduced. I cannot remember why it got rejected, but it got rejected at that time. Anyway wouldn’t it be a bit problematic concerning the backward compatibility?

liamhuber:
> That’s actually more or less what I suggested when the possibility to put multiple keywords was introduced. I cannot remember why it got rejected, but it got rejected at that time.

That's unfortunate, it sure seems like the right solution now...

> Anyway wouldn’t it be a bit problematic concerning the backward compatibility?

🤦‍♂️ Ugh, yes. Ok, I'll review this as-is. Stupid backwards compatibility.

liamhuber (approved):
So the the caveat of our preceding conversation about how I wish we were in a different timeline, lgtm.

The only thing I would recommend is to make a custom error class so that when you're running the behaviour tests you can be 100% sure you're catching exactly the value error you're trying to catch and not accidentally passing because some other, unrelated value error is getting raised.

Issue #775, opened 2022-06-21
Suggestions and problems of `pack` and `unpack`


niklassiemer:
List of problems/suggestions regarding `pack` and `unpack`:

1. The following line deletes the original project folder.

```python
pr.pack(pr.name)
```

2. The following code imports jobs multiple times without warnings/errors:

```python
pr.pack('TEST')
pr.unpack('TEST')
pr.unpack('TEST')
```

3. csv-file should be included inside the tar-file.

4. Definition of `destination path`:

Currently, `destination_path` has to be specified, while we think mostly people want to quickly pack and unpack project to download it. -> Sam's suggestion:

```python
class Project(...):

    def pack(self, destination_path=None, csv_file_name=None, compress=True):
        if destination_path is None:
            destination_path = self.name
        if csv_file_name is None: # Only if csv has to be exported separately
            csv_file_name = self.name
    ...
```

5. `unpack` as `@classmethod`:

```python
class Project(...):
    @classmethod
    def unpack(origin_path, project_name=None, csv_file_name=None):
        if project_name is None and not origin_path.endswith('tar.gz'):
            raise ValueError('`project_name` must be specified for uncompressed projects')
        if project_name == origin_path:
            raise ValueError('`project_name` must be different from `origin_path`')
        ...
```

So that the user can do:

```python
from pyiron import Project
pr = Project.unpack('PACKED_PROJECT.tar.gz')
```

6. `import_project` to import jobs into the current project:

```python
class Project(...):
    def import_project(self, origin_path, csv_file_name=None):
```

So that the user can do:

```python
from pyiron import Project
pr = Project('MY_PROJECT')
pr.import_project('PACKED_PROJECT.tar.gz') # This line imports all jobs inside PACKED_PROJECT.tar.gz
```

samwaseda:
Comment on the last one (`import_project`):

In principle the project should be able to have other jobs such as:

```python
from pyiron import Project
pr = Project('TEST')
vasp = pr.create.job.vasp('test')
vasp.structure = pr.create.structure.bulk('Al', cubic=True)
vasp.run()
pr.import_project('PACKED_PROJECT.tar.gz')
```

When there is a conflict of job names: In the first step, pyiron should raise an error. At some point we should think about how to import new stuff intelligently

pmrv:
It's currently not possible to pack project in a different directory from where it's located, i.e.
```python
pr = Project('/a/b/c')
pr.pack('my_archive')
```
[...]

PR #1401, opened 2024-03-28
Fixes for the `pack()` and `unpack()` method

jan-janssen:
@jyang2009 based on our discussion I created a couple of fixes for the `pack()` and `unpack()` method. Can you check if these resolve the issues you had with archiving on `ptmp`?

PR #1440, opened 2024-05-23
[minor] Update pack and unpack methods

samwaseda:
Update of the `pack` and `unpack` methods. https://github.com/pyiron/pyiron_base/issues/775

- [x] `pack` includes the csv file inside the archive
- [x] `pack` filename is optional, uses "<projectname>.tar.gz"
- [x] `pack` with same name as project should not delete the project
- [ ] `pack` selected jobs by id
- [x] `pack` all files in a job
- [x] `pack` from a different directory than where project is located
- [ ] `unpack` method can be called as `pr = Project(filename.tar.gz, unpack=True)` or `pr = Project(filename, unpack=True)`
- [x] `unpack` should not nest project automatically
- [x] `unpack` jobs into existing Project
- [x] Update tests
- [x] Update docstrings
- [ ] Update workflow template

Potentially backward-incompatible changes:

- I removed the argument `compress` from `unpack`. You have to put the full file name so that pyiron knows whether it has to decompress or not.
- Now `export.csv` is **inside** the compressed folder. The algorithm could fail if there is another `export.csv` in the project.


samwaseda:
I am going to take over @srmnitc's work, so I copied the PR from his forked branch

samwaseda:
ok I just realized that I had a different PR open and I opened this one on top. Let me correct this first XD

jan-janssen:
Thanks @samwaseda for picking this up. There are a couple of fixes also included in https://github.com/pyiron/pyiron_base/pull/1401 so maybe it makes sense to merge those as well.

samwaseda:
Ok this thing is now feature-complete. I might want to do some cleaning but otherwise is ready for a review.

samwaseda:
@pmrv @jan-janssen @srmnitc I guess it's now feature complete and this is probably going to be the last PR of the series of pack/unpack. After this one a minor release can be made and it's hopefully settled.

pyiron_base/project/archiving/export_archive.py
```diff
@@ -4,15 +4,14 @@
...
-from pyiron_base.project.archiving.shared import getdir
-
 
 def copy_files_to_archive(
     directory_to_transfer: str,
     archive_directory: str,
     compress: bool = True,
     copy_all_files: bool = False,
     arcname: Optional[str] = None,
+    df: Optional["DataFrame"] = None,
```
jan-janssen, line 14: Is there a specific reason to use `"DataFrame"` as string rather than importing the class from `pandas`?

pyiron_base/project/archiving/export_archive.py
```diff
@@ -77,26 +81,26 @@ def copy_h5_files(src, dst):
...
+            if file.endswith(".h5") or file == "export.csv":
                 src_file = os.path.join(root, file)
                 rel_path = os.path.relpath(root, src)
                 dst_dir = os.path.join(dst, rel_path)
                 os.makedirs(dst_dir, exist_ok=True)
                 shutil.copy2(src_file, os.path.join(dst_dir, file))
 
 
-def export_database(pr):
+def export_database(df: "DataFrame"):
```
jan-janssen, line 92: string data type - see comment above

pyiron_base/project/archiving/import_archive.py
```diff
@@ -91,6 +81,7 @@ def import_jobs(project_instance, archive_directory, df, compressed=True):
         job_id = pr_import.db.add_item_dict(par_dict=entry, check_duplicates=True)
         job_id_lst.append(job_id)
 
+    # print(job_id_lst)
```
jan-janssen:
```suggestion
```

jan-janssen: Alternatively, if this is commonly used for debugging, then it might make sense to add it to the log.

pyiron_base/project/archiving/import_archive.py
```diff
@@ -101,3 +92,49 @@ def import_jobs(project_instance, archive_directory, df, compressed=True):
...
+    """
+    df = get_dataframe(origin_path=origin_path, project_path=project_path)
+    common_path = os.path.commonpath(list(df["project"]))
+    copytree(os.path.join(origin_path, common_path), project_path, dirs_exist_ok=True)
+    return df, common_path
+
+
+def get_dataframe(
+    origin_path: str, project_path: str, csv_file_name: str = "export.csv"
+):
```
jan-janssen:
```suggestion 
) -> "DataFrame":
```

jan-janssen:
As far as I understand the csv file is now included in the tar archive, correct? How do you handle the backwards compatibility for old archives which do not include the csv file? Maybe it makes sense to have a short example either in the Docstring or the jupyter notebook to handle this backwards compatibility. Finally, I liked the option to take a look at the csv file to see which jobs are included in the archive before importing the corresponding jobs. Previously, I did this by loading the csv file with pandas. I can still load the csv file manually from the tar archive, but for other users it would be great to take a look at the archive and get the job table of the contained project from the python side.

pyiron_base/project/archiving/export_archive.py
samwaseda: Because `pandas` is not imported in this file.

samwaseda: Or actually I just assumed that it's better to not import extra classes just for type hinting.

pyiron_base/project/archiving/import_archive.py
samwaseda: nice catch I don't know how it appeared

samwaseda:
> As far as I understand the csv file is now included in the tar archive, correct? How do you handle the backwards compatibility for old archives which do not include the csv file? Maybe it makes sense to have a short example either in the Docstring or the jupyter notebook to handle this backwards compatibility.

That's a good point that I should mention in the code. I guess it helps the future generation to understand the origin of some of the code.

> Finally, I liked the option to take a look at the csv file to see which jobs are included in the archive before importing the corresponding jobs. Previously, I did this by loading the csv file with pandas. I can still load the csv file manually from the tar archive, but for other users it would be great to take a look at the archive and get the job table of the contained project from the python side.

That sounds good but I guess it's a different PR.

samwaseda:
Can I merge this one or should I still wait for a review?

pmrv (approved):

srmnitc (approved):

PR #1443, opened 2024-05-23
Combine import export example notebooks

### LINKED ISSUE #1431: Demo notebooks are execution order dependent
You need to execute `export_example.ipynb` before executing `import_example.ipynb`, or the latter will look for a file that does not exist. IMO each demo notebook should be fully independent.

https://github.com/pyiron/pyiron_base/pull/1429#issuecomment-2110657069

samwaseda:
Closes #1431


jan-janssen (approved):
Looks good to me

Issue #1444, opened 2024-05-23
`test_import.py` cannot run twice


samwaseda:
[test_import.py](https://github.com/pyiron/pyiron_base/blob/main/tests/archiving/test_import.py) cannot run twice because somehow pyiron fails to delete jobs in the first run when the tests finish. In the second run these remaining jobs from the first run apparently make a problem.

samwaseda:
It looks like there are multiple entries created when the jobs are imported with the same job name. ~~On top of this, it looks like the job has itself as a child job, so that when it tries to delete the child job, it cannot identify the job.~~

samwaseda:
So I'm not getting any comment here but I'm gonna keep posting my findings to use it as my log book.

- The problem does not occur with the newly updated [import and export example](https://github.com/pyiron/pyiron_base/blob/main/notebooks/import_export_example.ipynb). That's a good sign.
- Also, when each test is executed individually, there's no error/problem
- Based on this, I thought moving everything from setUpClass to setUp (same for tearDownClass) could solve the problem, but this created other problems.

pmrv:
Ah, I remember this issue when running tests locally.  It might have been that the project that the test imports into is not the same as the one that gets cleared after the test, but I'm not sure anymore and I never quite fixed it.

liamhuber:
> Ah, I remember this issue when running tests locally. It might have been that the project that the test imports into is not the same as the one that gets cleared after the test, but I'm not sure anymore and I never quite fixed it.

Ditto. I even still have a `fix_archive_tests` branch on my local repo that I'm pretty sure was trying to address exactly this, but I got frustrated and quit before I got it solved.

samwaseda:
Yeah it's super frustrating, especially the errors are really not helpful.

I kept digging down, and now I got the feeling that there's a problem with nested projects. It occurs namely even when there are only these two tests: `test_unpack_to_nested_project` and `test_unpack_from_other_dir_uncompress`, but somehow if each of them is run individually the problem does not occur.

samwaseda:
It looks like the opposite is also the case: if I run all tests except for these two, they pass as well.

samwaseda:
Interestingly, there’s also no problem if only one of them is disabled. So probably something in `tearDownClass` prevents the problem.

PR #1448, opened 2024-05-24
Replace copy_tree from distutils by copytree from shutil because dist…

samwaseda:
…utils is going to be removed from python 3.12

I stumbled upon [this article](https://stackoverflow.com/questions/10047521/how-to-copy-directory-with-all-file-from-c-xxx-yyy-to-c-zzz-in-python while trying to fix the [archiving problem](https://github.com/pyiron/pyiron_base/issues/1444) so I did the change.

### DIFF
--- a/pyiron_base/project/archiving/import_archive.py
+++ b/pyiron_base/project/archiving/import_archive.py
@@ -1,8 +1,7 @@
 import os
 import pandas
 import numpy as np
-from shutil import rmtree
-from distutils.dir_util import copy_tree
+from shutil import rmtree, copytree
 import tarfile
 from pyiron_base.project.archiving.shared import getdir
 from pyiron_base.utils.instance import static_isinstance
@@ -57,7 +56,7 @@ def import_jobs(project_instance, archive_directory, df, compressed=True):
     des = project_instance.path
     # source folder; archive folder
     src = os.path.abspath(archive_directory)
-    copy_tree(src, des)
+    copytree(src, des, dirs_exist_ok=True)
     if compressed:
         rmtree(src)
 
--- a/pyiron_base/state/settings.py
+++ b/pyiron_base/state/settings.py
@@ -41,7 +41,6 @@ from pyiron_base.state.publications import publications
 from pathlib import Path
 from pyiron_base.interfaces.singleton import Singleton
 from typing import Union, Dict, List
-from distutils.util import strtobool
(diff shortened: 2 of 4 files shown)


liamhuber (approved):

liamhuber:
Gah, mobile trouble getting my message attached to the review, sorry. Here it is:

The PR itself looks good to me, both in terms of motivation and execution. My only concern is that the archive tests are currently kind of spotty, yeah? Since this is such a straightforward PR, integrating it on top of any other changes to the archiving should be super straightforward, so maybe we want to hold off until those issues are closed, then merge in main and rerun newly improved tests? Just to make sure there's no side effects in the switch that we're oblivious too.

I'm only keeping half an eye on the archive/import-export conversation though, so I leave it your judgement whether this concern is overblown.

samwaseda:
@liamhuber I’m asking for a review again because I decided to replace all occurrences of `distutils`. The one that I replaced, `strtobool` does not seem to have a proper replacement. On the other hand it’s an independent function of only a few lines, so I copied and pasted it from the source code, which I think is vertretbar.

Coming back to your concern, now I don’t think it’s related to files, but purely database entries. Besides the two functions `copy_tree`  and `copytree` are nearly identical, so I think it’s ok.

pmrv (commented):
Please check under what license distutils is distributed and whether its compatible with our MIT, otherwise looks good to me.

liamhuber:
> Please check under what license distutils is distributed and whether its compatible with our MIT, otherwise looks good to me.

BSD, but good point, before copy-pasting you should check that you can/what citations are needed.

The target library of shutils is just a standard library package, right? So almost certainly no worries there.

liamhuber (approved):

liamhuber:
Dammit, I need to figure out how to approve with a message on mobile.

Anyhow, modulo double checking licensing, lgtm.

I do think this is exactly the sort of thing I would rather have in snippets than a particular pyiron module, but since snippets isn't published yet and moving it there will be trivial with my new gist, I have no problem putting it here for now if you don't want to wait.

samwaseda:
> I do think this is exactly the sort of thing I would rather have in snippets than a particular pyiron module, but since snippets isn't published yet and moving it there will be trivial with my new gist, I have no problem putting it here for now if you don't want to wait.

I agree and exactly for the reason that it’s not published I kept it here.

PR #1449, opened 2024-05-24
[minor] refactor super minor things

samwaseda:
I realized that this refactoring part is only confusing so I open another PR for fixing the archiving problem [here](https://github.com/pyiron/pyiron_base/pull/1453)


samwaseda:
Ah and of course the tests will fail because I didn’t update them

PR #1452, opened 2024-05-29
[minor] update docstring for `TemplateJob`

samwaseda:
There was a huge confusion yesterday as it was not clear how to use a custom job defined in `TemplateJob` so I updated the doc string.

### DIFF
--- a/pyiron_base/jobs/job/template.py
+++ b/pyiron_base/jobs/job/template.py
@@ -26,7 +26,7 @@ class TemplateJob(GenericJob, HasStorage):
 
     Example:
 
-    >>> from pyiron_base import TemplateJob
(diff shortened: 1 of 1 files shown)


jan-janssen:
I like the update to the docstring, I am just a bit surprised that people still use the `TemplateJob` class directly. I thought we replaced that by `wrap_python_function()` and `create_job_class()` https://pyiron-base.readthedocs.io/en/latest/example.html

samwaseda:
One case where people want to use `TemplateJob` is when they want to use a variable for both input and output. In that case it makes more sense to define it in `__init__` and use it in `write_input` and `collect_output`.

But one way or other, as long as the class is there we should make it user-friendly.

PR #1453, opened 2024-05-29
[patch] restore forbidding duplicate imports

samwaseda:
This PR partially solves [this problem](https://github.com/pyiron/pyiron_base/issues/1444), as it forbids to import a job whose name already exists in the project, but I make it a draft because it does not clarify why the problem is occurring in the first place.

### DIFF
--- a/pyiron_base/project/archiving/import_archive.py
+++ b/pyiron_base/project/archiving/import_archive.py
@@ -27,8 +27,6 @@ def extract_archive(archive_directory):
 
 
 def import_jobs(project_instance, archive_directory, df, compressed=True):
-    if len(set(df["job"]) & set(project_instance.job_table().job)) > 0:
-        raise ValueError("Overlapping", df["job"], project_instance.job_table().job)
     # Copy HDF5 files
     # if the archive_directory is a path(string)/name of the compressed file
     if static_isinstance(
--- a/pyiron_base/project/archiving/import_archive.py
+++ b/pyiron_base/project/archiving/import_archive.py
@@ -27,6 +27,8 @@ def extract_archive(archive_directory):
 
 
 def import_jobs(project_instance, archive_directory, df, compressed=True):
+    if len(set(df["job"]) & set(project_instance.job_table().job)) > 0:
+        raise ValueError("Overlapping", df["job"], project_instance.job_table().job)
     # Copy HDF5 files
     # if the archive_directory is a path(string)/name of the compressed file
     if static_isinstance(
--- a/tests/archiving/test_import.py
+++ b/tests/archiving/test_import.py
@@ -72,6 +72,7 @@ class TestUnpacking(PyironTestCase):
         compare_obj = dircmp(path_original, path_import)
         self.assertEqual(len(compare_obj.diff_files), 0)
 
+    @unittest.skip("Imports the same name fails")
     def test_unpack_to_nested_project(self):
         pr = self.pr.open("nested")
         pr_imp = pr.open("imported")
--- a/tests/archiving/test_import.py
+++ b/tests/archiving/test_import.py
@@ -72,7 +72,6 @@ class TestUnpacking(PyironTestCase):
         compare_obj = dircmp(path_original, path_import)
         self.assertEqual(len(compare_obj.diff_files), 0)
 
-    @unittest.skip("Imports the same name fails")
     def test_unpack_to_nested_project(self):
         pr = self.pr.open("nested")
         pr_imp = pr.open("imported")
@@ -105,6 +104,7 @@ class TestUnpacking(PyironTestCase):
             rmtree(pack_path)
         except Exception as err_msg:
             print(f"deleting unsuccessful: {err_msg}")
+        pr.remove(enable=True)
 
     def test_import_uncompress(self):
         self.pr.pack(destination_path=self.arch_dir, compress=False)


samwaseda:
So I guess for now this is the most pragmatic solution. I still do not perfectly understand what's happening, but it sounds reasonable to me to forbid importing jobs with an existing job name. I put the error in the importing function and not deeper in base, because I thought maybe it is not fundamentally forbidden that there are jobs with the same name in pyiron? Maybe @jan-janssen can make a comment? One way or other, I think this is more like a temporary fix, and we need an overhaul in [this PR](https://github.com/pyiron/pyiron_base/pull/1440). What do you think?

liamhuber:
> because I thought maybe it is not fundamentally forbidden that there are jobs with the same name in pyiron?

I actually wasn't sure what the behaviour would be here, so I checked. The executive summary is:
- Project doesn't care at all about names until something has been run
- Project won't run a job whose name is already a run job in its table
- Creating a new job with an existing job name loads the existing job
- The `job` column of the project table is not unique, but the combination of `project` and `job` is

Demo:
```python
pr = Project("test")
pr.remove_jobs(silently=True)
j1 = pr.create.job.Lammps("lmp")
j1.structure = pr.create.structure.bulk("Al")

j2 = pr.create.job.Lammps("lmp")
j2.structure = pr.create.structure.bulk("Cu")

print(j1.structure.get_chemical_formula(), j2.structure.get_chemical_formula())
# Al Cu
print(j1.name, j2.name)
# lmp lmp
print(pr.job_table().job)
# Series([], Name: job, dtype: object)

j1.run()
# Runs as expected

j2.run()
# Prints a warning and doesn't run `job exists already and therefore was not created!`

j3 = pr.create.job.Lammps("lmp")
# Silently (really, I set `warnings.simplefilter("always")`) reloads the job

print(j1.structure.get_chemical_formula(), j2.structure.get_chemical_formula(), j3.structure.get_chemical_formula())
# Al Cu Al
print(pr.job_table().job)
# 0    lmp; Name: job, dtype: object
```

Only the project+job column gives a unique name, as it's fine to create and run a job in a subproject:

```python
sub_pr = pr.create_group("sub_pr")
k1 = sub_pr.create.job.Lammps("lmp")
k1.structure = sub_pr.create.structure.bulk("Ni")
k1.run()
print(pr.job_table().job)
# 0    lmp; 1    lmp; Name: job, dtype: object
```

IMO there's nothing faulty with any of this, although it is annoying to find out "late" that your job name is non-unique and won't run -- would be nicer to learn that at creation time. This is impossible in the current setup though, as project is not truly aware of a job until after it has run.

I've attacked this general topic in `pyiron_workflow` using a `Semantic` class, where the most relevant bit is probably [`SemanticParent.add_child`](https://github.com/pyiron/pyiron_workflow/blob/cc5bb7dadbc0ab096eeb9250d81554c4f2dd6b01/pyiron_workflow/semantics.py#L168). This fundamentally differs as it's all happening at the object level in-memory, rather than referring to the database as (I think) `Project` is doing for these checks. I'm not suggesting we need to pull anything from there in here, it's just an FYI for attacking the topic from another angle.

pyiron_base/project/archiving/import_archive.py
```diff
@@ -27,6 +27,8 @@ def extract_archive(archive_directory):
 
 
 def import_jobs(project_instance, archive_directory, df, compressed=True):
+    if len(set(df["job"]) & set(project_instance.job_table().job)) > 0:
+        raise ValueError("Overlapping", df["job"], project_instance.job_table().job)
```
liamhuber: It seems very reasonable to me to test and fail when there is a conflict, but given my comment in the main thread I guess this check will return false-positives for overlap since `project_instance.job_table().job` can be non-unique when jobs of the same name have different (sub)projects.

I like the idea of introducing this, but I guess the tests need to be expanded to make sure we _do_ fail when we want to (direct overlap in the main project, direct overlap in any subproject) and _don't_ fail when we shouldn't (no overlap, or overlap in job column but with different sub-projects)

tests/archiving/test_import.py
```diff
@@ -72,6 +72,7 @@ def test_import_compressed(self):
         compare_obj = dircmp(path_original, path_import)
         self.assertEqual(len(compare_obj.diff_files), 0)
 
+    @unittest.skip("Imports the same name fails")
```
liamhuber: This seems sketchy as hell.

I guess given your changes above, this test now cleanly raises a `ValueError`? But, like, it is still broken, I still want it to fail when it's failing, I don't want to just sweep it under the rug.

pyiron_base/project/archiving/import_archive.py
liamhuber: nit: I've taken to introducing custom errors for most things, so that if I ever need to test it by setting it up to intentionally fail I can make sure it's failing for the exact reason I am wanting and not raising a similar error early for some other reason. E.g.

```python
class OverlappingJobsError(ValueError):
    """Raise when a project tries to import a job whose name it already has"""
```

liamhuber: nit: You can pass a tuple in and the message comes out as a tuple, but I guess it would be prettier to f-string these pieces together?

liamhuber (dismissed):
I am completely behind the idea of having a check here, but I don't think this does what you want and it needs tests to prove it's doing the desired thing.

samwaseda:
> > because I thought maybe it is not fundamentally forbidden that there are jobs with the same name in pyiron?
> 
> I actually wasn't sure what the behaviour would be here, so I checked. The executive summary is:
> 
> * Project doesn't care at all about names until something has been run
> * Project won't run a job whose name is already a run job in its table
> * Creating a new job with an existing job name loads the existing job
> * The `job` column of the project table is not unique, but the combination of `project` and `job` is


Yep so far that was more or less what I had expected. What I don't know is whether it can happen that there are jobs with the same name that have run. I super vaguely remember that this can happen when there are two child jobs which share the same name. In that case it's the same project and the same job name but with different parents, but as I said I don't clearly remember it.

tests/archiving/test_import.py
samwaseda: I agree but I would love to re-introduce it when pack-unpack is fully functional. It's therefore a bit intentional to make you feel uncomfortable when you see `skip`. I can change the message to make sure that people know that this test shouldn't be skipped for a long time.

liamhuber:
> I super vaguely remember that this can happen when there are two child jobs which share the same name. In that case it's the same project and the same job name but with different parents, but as I said I don't clearly remember it.

Oh yeah, I wasn't thinking about parent jobs. I don't intend to dig in and make an example for that though 😝

tests/archiving/test_import.py
liamhuber: Hmm, making me feel uncomfortable now so that future us can feel comfortable and lazy and just trusting future us to come back, get off our lazy butts, and change something seems unwise. If pack-unpack is not fully functional, that should be clear when and by test failure.

liamhuber: Alternatively: the test could `assertRaises(OverlappingJobsError` (or similar), which then _forces_ future-us to come back and fix it once pack-unpack is working as then the test-that-it's-broken test will start failing.

I still would rather leave the broken thing broken as a reminder of its brokenness, but I can tolerate silencing it in a way that will force action later.

pyiron_base/project/archiving/import_archive.py
niklassiemer: > nit: I've taken to introducing custom errors for most things, so that if I ever need to test it by setting it up to intentionally fail I can make sure it's failing for the exact reason I am wanting and not raising a similar error early for some other reason. E.g.
> 
> ```python
> class OverlappingJobsError(ValueError):
>     """Raise when a project tries to import a job whose name it already has"""
> ```

I do not have time to check. Do these errors get captured by a `except ValueError` clause?

tests/archiving/test_import.py
samwaseda: I mean I'm now actively working on it [here](https://github.com/pyiron/pyiron_base/pull/1440) (although I have to admit that because of the error caused by `test_import` I cannot work on it right now), so it's not like Biden or Trump promising something in their next presidency?

pyiron_base/project/archiving/import_archive.py
liamhuber: @niklassiemer, yes, it also counts as the parent class for the purposes of assert raises

tests/archiving/test_import.py
liamhuber: This still sounds like selling a bad thing because "it won't be _that_ bad" rather than any actual argument that it's good. This lets me rerun the tests to on my local machine and fakes that everything is ok (a bad thing IMO) -- is there some other benefit I'm just being clueless about?

pmrv (changes requested):
Agree with @liamhuber's diagnosis.  Probably you want to join `job` and `project` columns in the given `df` first, before making this check.

samwaseda:
I guess now I found the workaround and hopefully the tests will pass this time...

samwaseda:
In the end, it boiled down to `pr.remove(enable=True)`, so I removed all the other lines that I had added. Since there'll be an overhaul [in this PR](https://github.com/pyiron/pyiron_base/pull/1440), I don't think the error that I had added will be particularly helpful.

pmrv (approved):

jan-janssen (approved):
Looks good to me. @samwaseda Can you merge this?

samwaseda:
It's blocked because @liamhuber's approval is missing

PR #1456, opened 2024-05-31
[patch] use retry from pyiron_snippets

liamhuber:


liamhuber:
```
File "/usr/share/miniconda3/envs/my-env/lib/python3.9/site-packages/pyiron_snippets/retry.py", line 20, in <module>
          log: bool | object = True,
      TypeError: unsupported operand type(s) for |: 'type' and 'type'
      [end of output]
```

Naively this looks like `retry` is not 3.9 compatible, but the test runs fine on 3.9 over on the snippets repo, so I'm going to dig a bit deeper.

liamhuber:
I think somehow the test on the snippets repo must have access to `from __future__ import annotations` while it doesn't here. I'll add it explicitly and patch snippets.

liamhuber:
Ok, this is just waiting on `pyiron_snippets-0.1.1` being live.

pyiron_base/__init__.py
```diff
@@ -46,6 +45,9 @@
 
 from pyiron_base.jobs.job.toolkit import Toolkit, BaseTools
 
+# Expose snippets references in base API for backwards compatibility
+from pyiron_snippets.import_alarm import ImportAlarm
+
```
jan-janssen:
```suggestion
```

liamhuber:
Import alarm should not be showing up in the diff here -- something needs to be merged in or the target needs to be changed

PR #1457, opened 2024-05-31
[patch] use deprecate from pyiron_snippets

liamhuber:


liamhuber:
@jan-janssen I left it with the backwards compatibility you found confusing -- I'll stack an additional PR on top doing a [minor] bump that removes the compatibility and clarifies the situation

pyiron_base/__init__.py
```diff
@@ -47,6 +46,7 @@
 
 # Expose snippets references in base API for backwards compatibility
 from pyiron_snippets.import_alarm import ImportAlarm
+from pyiron_snippets.deprecate import Deprecator, deprecate, deprecate_soon
```
jan-janssen:
```suggestion
```

liamhuber:
@jan-janssen going forward I recommend merging stacks from the bottom up instead of the top down -- now this PR is big and it's not immediately clear to me where all the problems are coming from. Diff markup got committed directly though, so that's definitely part of it.

PR #1465, opened 2024-06-03
[patch] replace logger with snippets version

liamhuber:
i.e. pyiron_base.state.logger.logger with pyiron_snippets.logger.logger. The original location leaves a re-direct for backwards compatibility.


liamhuber:
The linter got rid of my redirect. I am going to get rid of it in the next PR anyhow, but I just want to point out that the robot is obliviously wrecking my software engineering choices. Sometimes an unused import is intentional.

Screen Shot 2024-06-03 at 10 26 07

liamhuber:
FFS, is there a flag to stop the damned bot?

liamhuber:
Ok, I tried bypassing the linter by adding `--no-verify` on my local machine before pushing the commit, but this doesn't work. I don't see how to bypass the linter when it's operating on the remote machine, so I'm just going to give in and "trick" it. 👎 👎

jan-janssen:
> i.e. pyiron_base.state.logger.logger with pyiron_snippets.logger.logger. The original location leaves a re-direct for backwards compatibility.

I thought we agreed to completely remove the classes replaced by `pyiron_snippets`.

liamhuber:
> > i.e. pyiron_base.state.logger.logger with pyiron_snippets.logger.logger. The original location leaves a re-direct for backwards compatibility.
> 
> 
> 
> I thought we agreed to completely remove the classes replaced by `pyiron_snippets`. 

Yes, but instead of having three compatibility breaking commits, I opted to go for three backwards-compatible patches and then a single PR to remove all the compatibility shortcuts.

PR #1473, opened 2024-06-10
Remove job._calculate_successor() and job.._calculate_predecessor()

jan-janssen:
I have the feeling nobody is using this functionality, so I removed it.


pmrv (approved):
lgtm, I think I had removed ListMaster already anyway, right?

PR #1474, opened 2024-06-10
Reorder return values from handle_failed_job

pmrv:
At some point this got mixed up in a different PR.


jan-janssen (approved):
Looks good to me

PR #1477, opened 2024-06-11
Introduce calculate() function 

jan-janssen:
In the previous interface a new class could be implemented in pyiron by defining two functions `write_input()` and `collect_output()`. The example class would look like this:
```python
from pyiron_base import GenericJob

class NewJobClass(GenericJob):

    def write_input(self):
        ...

    def collect_output(self):
        ...
```
The new interface replaces the `write_input()` function with an `get_input_parameter_dict()` function which returns the `files_to_write` and the `files_to_copy` as a dictionary. In addition, the `collect_output()` function is placed outside the job class. Both of these changes are designed to support a more functional approach in `pyiron_base`. Only the `get_input_parameter_dict()` and `get_output_parameter_dict()` functions are now part of the new job class. The `get_input_parameter_dict()` function takes the input defined by the user on different properties like `job.input.structure` or functions like `job.set_kpoints()` and merges everything into a dictionary with input file content. This dictionary of input file content allows expert users to overwrite the pyiron generated input with their own input, if the `get_input_parameter_dict()` function defined in the new job class supports this. In the same way the `get_output_parameter_dict()` function allows the developer to provide additional parameters to the `collect_output()` function: 
```python
from pyiron_base import GenericJob

def collect_output(working_directory, output_parameter_1, output_parameter_2):
    ...
    return output_dict

class NewJobClass(GenericJob):
        self._job_with_calculate_function = True
        self._collect_output_funct = collect_output

    def get_input_parameter_dict(self):
        ...
        return {
            "files_to_write": {"test.py": "if __name__ == '__main__':\n    print('Hello')\n"}, 
            "files_to_copy": {"my_bashrc": "~/.bashrc"}
        }

    def get_output_parameter_dict(self):
        return {"output_parameter_1": ..., "output_parameter_2": ...}

```
For those users who only want define the `write_input()` and `collect_output()` function, `pyiron_base` still supports the `wrap_executable()` function: 
```python
def write_input(input_dict, working_directory="."):
    ...

def collect_output(working_directory="."):
    ...
    return output_dict

job = pr.wrap_executable(
    job_name="Cat_Job_energy_1_0",
    write_input_funct=write_input,
    collect_output_funct=collect_output,
    input_dict={"energy": 1.0},
    executable_str="cat input_file > output_file",
    execute_job=True,
)
```
The limitations of the `wrap_executable()` function are: MPI parallelism is not supported and the `input_dictionary` is directly communicated to the `write_input()` function. So it is not possible to have a generic interface and an expert user interface with this approach. Internally the `wrap_executable()` already uses the new job class interface. In the same way the `collect_output()` function only accepts the `working_directory` as input parameter and it is currently not possible to provide additional parameters for the parsing of the output.


jan-janssen:
I start by switching the `ExecutableContainerJob` to using the new `calculate()` function. Once this pull request is merged and the new `pyiron_base` version is released the same changes can be applied to LAMMPS and VASP in https://github.com/pyiron/pyiron_atomistics/pull/1446 and https://github.com/pyiron/pyiron_atomistics/pull/1447

pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -43,6 +49,7 @@ def __init__(self, project, job_name):
         super().__init__(project, job_name)
         self._write_input_funct = None
         self._collect_output_funct = None
+        self._python_only_job = True
```
pmrv: `ExecutableContainerJob` is also used for `wrap_executable` on the project level, right?  Is this flag then necessary force certain behavior?  Would be nice to mention this here in a comment.

pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -72,22 +79,72 @@ def set_job_type(
...
+            "executable_dict": {
+                "executable": self.executable.executable_path,
+                "shell": True,
+                "working_directory": self.working_directory,
+                "conda_environment_name": self.server.conda_environment_name,
+                "conda_environment_path": self.server.conda_environment_path,
+            },
+        }
+
+    def generate_calculate_function(self) -> callable:
```
pmrv: Can be inlined at the only call site, unless you see this being called elsewhere in the future.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,69 @@ def multiprocess_wrapper(
...
+                working_directory=working_directory,
+            )
+        try:
+            shell_output = execute_subprocess(
+                **{
+                    k: v
+                    for k, v in executable_dict.items()
+                    if k not in ["accepted_return_codes", "accept_crash"]
+                }
+            )
```
pmrv:
```suggestion
        accept_crash = executable_dict.pop("accept_crash", False)
        accepted_return_codes = executable_dict.pop("accepted_return_codes", [])
        try:
            shell_output = execute_subprocess(**executable_dict)
```

pmrv (commented):
It's a bit annoying that this now duplicates a lot of the logic present in GenericJob.run_static and `execute_job_with_external_executable`.  Do you see an option to unify this again (in the future)?

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: The general motivation is to replace the previous `run_static()` functionality which calls `write_input()`, the executable and afterwards `collect_output()` with a more functional approach which generates a `calculate()` function with the corresponding arguments using `generate_calculate_function()` and `generate_calculate_function_kwargs()`. This calculate function is then called, at the moment in the same Python process but in future this can be submitted using via an `concurrent.futures.Executor`. After the successful execution the output is stored using the `job._store_output()` function. 

At the moment we use the existing `self._python_only_job` flag to only call `run_static()` rather than the full `run_static()` stack.

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: I want to provide the user with a simple way to convert a pyiron job object to a python function call. At the current stage the user can create the job and interact with it and then call `generate_calculate_function()` and `generate_calculate_function_kwargs()` to have the corresponding `calculate()` function which executes the calculation of the specific job. So yes, I consider the user to call the functions externally, still I am not sure if it might make more sense to rename them or have them in a combined function.

jan-janssen:
> It's a bit annoying that this now duplicates a lot of the logic present in GenericJob.run_static and `execute_job_with_external_executable`. Do you see an option to unify this again (in the future)?

The goal is to replace the current `GenericJob.run_static` logic with this new functional approach. Still at the moment we need the `GenericJob.run_static` logic for backwards compatibility.

jan-janssen:
The motivation for this change is discussed in the [meeting notes](https://github.com/orgs/pyiron/discussions/231#discussioncomment-9751091) .

pyiron_base/jobs/flex/executablecontainer.py
pmrv: > At the moment we use the existing `self._python_only_job` flag to only call `run_static()` rather than the full `run_static()` stack.

A-ok with me, but this is not clear from the code, so it'd be nice to have a comment there.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+                working_directory=working_directory,
+            )
+        else:
+            parsed_output = None
+        return shell_output, parsed_output, job_crashed
+
+    return calculate
+
+
+def execute_calculate_function(job):
```
pmrv: I like the direction of this, but as long as it's not general I'd like a warning or a comment, when `job` is not `ExecutableContainerJob`, because it's not obvious from the function itself.  Comment is probably fine, because no user will call this accidentally, but developers other than you will have a hard time without.

pyiron_base/jobs/flex/executablecontainer.py
pmrv: If it's user facing, than it's ok to have it.  But then it needs docstring that explains how to use it, like your example in the [meeting notes](https://github.com/orgs/pyiron/discussions/231#discussioncomment-9751091). 

Couldn't `generate_calculate_function` return a closure that already has the arguments baked in?  It's a bit awkward to get the function and the kwargs separately when you are really expected to call and use them together.

pyiron_base/jobs/job/runfunction.py
jan-janssen: The same function is used in the `ScriptJob` in https://github.com/pyiron/pyiron_base/pull/1480 and I plan to use it in LAMMPS and VASP once these changes are merged and a new version of `pyiron_base` is released.

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: At the moment this is correct, but at the same time I would like to have the option to submit these functions to an `concurrent.futures.Executor` so I guess we need the function and the keyword arguments separated. Still I am open to have both merged into one function.

jan-janssen:
@pmrv Can you take another look? I would like to merge this to release a new `pyiron_base` version.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
         job_wrap.job.run_static()
 
 
+def generate_calculate_function(write_input_funct=None, collect_output_funct=None):
+    def calculate(input_dict, executable_dict, output_dict={}):
+        working_directory = executable_dict["working_directory"]
+        accept_crash = executable_dict.pop("accept_crash", False)
+        accepted_return_codes = executable_dict.pop("accepted_return_codes", [])
```
samwaseda: I'm not really sure if I understand the strategy correctly here. Doesn't it make more sense to define `accept_crash` and `accepted_return_codes` in `generate_calculate_function_kwargs`, especially if there are pre-defined arguments? Otherwise how does anyone know that these arguments are available?

pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -72,22 +82,76 @@ def set_job_type(
...
+        >>> job._store_output(output_dict=parsed_output, shell_output=shell_output)
+
+        Returns:
+            dict: keyword arguments for the calculate() function
+        """
+        return {
+            "input_dict": self.input.to_builtin(),
+            "executable_dict": {
+                "executable": self.executable.executable_path,
+                "shell": True,
```
samwaseda: Again, I'm not really sure if I understand it correctly, but is there a possibility to unify `shell` and `_python_only_job`? I frankly have the feeling that this should even be automatic according to whether `executable` is set or not, i.e. if `executable` is set -> `shell = True` otherwise `shell = False`.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+        os.makedirs(working_directory, exist_ok=True)
+        if write_input_funct is not None:
+            write_input_funct(
+                input_dict=input_dict,
+                working_directory=working_directory,
+            )
+        try:
+            shell_output = execute_subprocess(**executable_dict)
+            error = None
```
samwaseda: What kind of effect is expected from `error = None`?

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+            job_crashed = False
```
```suggestion
```

Bundled change 2/4

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+        except (subprocess.CalledProcessError, FileNotFoundError) as error:
+            if hasattr(error, "output"):
+                shell_output = error.output
+                if error.returncode in accepted_return_codes:
+                    job_crashed = False
+                elif not accept_crash:
```
```suggestion
                if not accept_crash:
```

Bundled change 3/4

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+                    error_file = posixpath.join(working_directory, "error.msg")
+                    with open(error_file, "w") as f:
+                        f.write(error.output)
+                    raise RuntimeError("External executable failed")
+                else:
```
```suggestion
                elif error.returncode not in accepted_return_codes:
```

Bundled change 4/4

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+        working_directory = executable_dict["working_directory"]
+        accept_crash = executable_dict.pop("accept_crash", False)
+        accepted_return_codes = executable_dict.pop("accepted_return_codes", [])
+        os.makedirs(working_directory, exist_ok=True)
+        if write_input_funct is not None:
+            write_input_funct(
+                input_dict=input_dict,
+                working_directory=working_directory,
+            )
+        try:
```
```suggestion
        job_crashed = False
        try:
```

Bundled change 1/4

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+            parsed_output = collect_output_funct(
+                working_directory=working_directory,
+                **output_dict,
+            )
+        elif not job_crashed and collect_output_funct is not None:
+            parsed_output = collect_output_funct(
+                working_directory=working_directory,
+            )
+        else:
+            parsed_output = None
```
```suggestion
```

Bundled change 2/2

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
...
+                    with open(error_file, "w") as f:
+                        f.write(error.output)
+                    raise RuntimeError("External executable failed")
+                else:
+                    job_crashed = True
+            else:
+                raise RuntimeError("Job aborted")
+        with open(posixpath.join(working_directory, "error.out"), mode="w") as f_err:
+            f_err.write(shell_output)
+        if (
```
```suggestion
        parsed_output = None
        if (
```

Bundled change 1/2

pyiron_base/jobs/job/runfunction.py
```diff
@@ -782,6 +782,86 @@ def multiprocess_wrapper(
         job_wrap.job.run_static()
 
 
+def generate_calculate_function(write_input_funct=None, collect_output_funct=None):
+    def calculate(input_dict, executable_dict, output_dict={}):
```
samwaseda: I'm unhappy about the fact that these args, especially the functions must fulfil certain requirements (`input_dict`, `working_directory` as arguments etc.). I see two options:

- Write all requirements in the doc string (my less favourite option)
- Write where the args of `generate_calculate_function` should come from (I guess `GenericJob._write_input_funct` and `GenericJob._collect_output_funct`?)

samwaseda (changes requested):
Since I've got something to do, I finish the review here for now. I'm just generally unhappy with the fact that the connections between the methods are fairly weak. For example I see this free standing function `generate_calculate_function`, which doesn't straightforwardly allow me to see what kind of component it is in the new scheme. I think there should be some indication in the doc string in the form of e.g. `write_input_funct (dict): input function cf. SomeClass.some_method`.

Btw. `generate_calculate_function` is kind of a bit long (I tried to make it shorter)

pyiron_base/jobs/job/runfunction.py
jan-janssen: The logic behind this is that the execution of an external executable can be divided into three steps: (1) write input files, (2) call executable and (3) parse output files. So I wanted to have three dictionaries as input one for each step. The arguments `accept_crash` and `accepted_return_codes` belong to the execution, so they are included in the `executable_dict`.

pyiron_base/jobs/job/runfunction.py
jan-janssen: The error is set to `none` when the execution was successful, if the execution was not successful the `error` is returned.

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: done.

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: `_python_only_job` is primarily used to overwrite the default behaviour of calling `write_input()` and `collect_output()`. This is a workaround until all codes support the new scheme. The `shell` parameter is part of the `subprocess.Popen()` https://docs.python.org/3/library/subprocess.html#popen-constructor

pyiron_base/jobs/flex/executablecontainer.py
pmrv: I've put together a small snippet that I mention in the meeting.  It defines a picklable `Caller` object that would hold onto the write and collect function as well as their arguments.  I think it clears up the logic quite a bit and would also immediately be able to be used with the executors (assuming the passed write and collect function are (cloud-)picklable, which should be no issue at all.

```python
import os

from pickle import loads, dumps

def write(input_dict):
    print("Im writing to disk!")
    print(input_dict)

def collect():
    print("I'm reading this from disk, honest!")
    return 42

def myfunc(f0, f1, *args, **kwargs):
    print(args)
    r = f0(kwargs)
    return f1()

class Caller:
    __slots__ = ('func', 'args', 'kwargs')

    def __init__(self, func, *args, **kwargs):
        self.func = func
        self.args = args
        self.kwargs = kwargs

    def __call__(self):
        return self.func(*self.args, **self.kwargs)

def calculate(write_function, collect_function, input_dict=None, executable_dict=None):
    if input_dict is not None:
        write_function(input_dict)

    # do something with executable_dict
    os.system("date")

    output = collect_function()
    return output


def generate_calculate(job, write=None, collect=None):
    return Caller(
            calculate,
            # presumably write/collect would be defined per job/simulation code
            # they could live as static methods on the job classes themselves,
            # then we'd be able to use them here as job.write, job.collect
            write,
            collect,
            **job.generate_calculate_function_kwargs()
    )

if __name__ == "__main__":
    loads(dumps(Caller(myfunc, write, collect, 3, [3,5], a='More kwargs')))()
```

pmrv:
I've posted my idea I mentioned yesterday above.  I think this is absolute the right direction to go for, but I would hold off on this PR and implementing it in the other codes in atomistics for a little bit longer.  As it is the logic is difficult to follow (@samwaseda's comments apply).  What I would like see also is a connection to the `get_input_file_dict()` changes.  With those it should be possible to have generic `write_input_funct`.

pyiron_base/jobs/job/runfunction.py
samwaseda: I don't think this answered my question. I'm gonna ask it again: Is there a particular reason that these terms are not given in the dictionary returned by `generate_calculate_function_kwargs`?

pyiron_base/jobs/flex/executablecontainer.py
samwaseda: Are you saying `_python_only_job` is going to be removed?

pyiron_base/jobs/job/runfunction.py
samwaseda: And what does it bring to set it to `None`?

samwaseda:
Since @jan-janssen pm'ed saying he didn't really understand what I was trying to say, I try to formulate it in a very basic way, which applies in general when someone tries to map a concept to actual coding.

I take the article 1 of the Universal Declaration of Human Rights as an example:

> "All human beings are born free and equal in dignity and rights. They are endowed with reason and conscience and should act towards one another in a spirit of brotherhood."

So in this text, "human" follows "All" and then we have "beings" and so on. The logic of the phrase is understood by following each word, which is a bit like a workflow, where we follow functions.

Now, the problem is that, when someone later on tries to look at it, they typically see a fragment of the text. So they might for example see the expression "should act" and wonders what it means. If that's someone who already knows that it must be part of the article 1 of the Universal Declaration of Human Rights, then they can try to figure out where "should act" should belong to, but even then it's not immediately clear where. On top of this, we keep getting new people and they usually do not know anything about it, and therefore it's nearly impossible to figure out that there's the article 1 of the Universal Declaration of Human Rights behind "should act".

I hope what I'm trying to say it clear with this simple example, but I'm gonna put it in words: When you implement a concept, it is easy to map your thoughts into a code, but other people must do the mapping **backwards**, i.e. they must look at the fragments of the code and figure out what the concept behind them is. And that's something which is excruciatingly difficult in pyiron in the current version.

From my perspective there are two ways to resolve the problem:

1. You write the full logic in the doc string or make comments in the code and indicate where the function belongs to (in the example it would be like writing the full article of the Universal Declaration of Human Rights and say "should act" belongs where). This is what @liamhuber and @pmrv often do
2. You write the immediate steps before and after the function in the doc string or in the code comment (it's like saying `"should act" is the term that follows "with reason and conscience" and is followed by "one another"). With this, the user has the possibility to understand the full workflow by following the components recursively.

jan-janssen:
After extended discussions with @samwaseda and @pmrv we agreed that the current syntax is too complex. I thought about different options and my new suggestion would be to have a more explicit `calculate()` function:
```python
def calculate(
    working_directory,
    input_parameter_dict,
    executable_script,
    shell_parameter,
    conda_environment_name,
    conda_environment_path,
    accept_crash,
    accepted_return_codes,
    output_parameter_dict,
)
```

jan-janssen:
> It's a bit annoying that this now duplicates a lot of the logic present in GenericJob.run_static and `execute_job_with_external_executable`. Do you see an option to unify this again (in the future)?

I updated the `run_static()` function to either call `execute_job_with_calculate_function()` or `execute_job_with_external_executable()`.

pyiron_base/jobs/job/runfunction.py
```diff
@@ -783,13 +783,189 @@ def multiprocess_wrapper(
...
+            conda_environment_name=conda_environment_name,
+            conda_environment_path=conda_environment_path,
+            accepted_return_codes=accepted_return_codes,
+            accept_crash=accept_crash,
+        )
+        parsed_output = None
+        if (
+            not job_crashed
+            and collect_output_funct is not None
+            and len(output_parameter_dict) > 0
```
```suggestion
```

pyiron_base/jobs/job/runfunction.py
```diff
@@ -783,13 +783,189 @@ def multiprocess_wrapper(
...
+        ):
+            parsed_output = collect_output_funct(
+                working_directory=working_directory,
+                **output_parameter_dict,
+            )
+        elif not job_crashed and collect_output_funct is not None:
+            parsed_output = collect_output_funct(
+                working_directory=working_directory,
+            )
```
```suggestion
```

I don't think this distinction between `len(output_parameter_dict) > 0` or not is necessary

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: The essential part for me is to be able to execute it with an `Executor`. But with the `Caller` class above both the `ProcesspoolExecutor` and the `pympipool` executor fail - at least in a jupyter notebook. Here is the example code I tried: 
```python
with ProcessPoolExecutor() as exe:
    fs = exe.submit(Caller, myfunc, write, collect, 3, [3,5], a = 'More kwargs')
    print(fs.result())
```

samwaseda (approved):
Except for my minor comment I'm ok with this.

pyiron_base/jobs/job/runfunction.py
jan-janssen: It is as the `collect_output()` typically used with `wrap_executable()` only takes a single input argument the `working_directory` while the more general `collect_output()` can take more arguments. So if `output_parameter_dict` is defined then we call `collect_output(working_directory, **output_parameter_dict)`. This is required for LAMMPS and VASP which need additional parameters to parse the output, like the initial structure.

samwaseda: But you check only the length of `output_parameter_dict`, so it's gonna fail if `collect_output` cannot accept args while `output_parameter_dict` has args. So with or without this `elif` it's the same thing.

pyiron_base/jobs/flex/executablecontainer.py
pmrv: Ah, no, I was not clear enough.  This `Caller(...)` does not execute the function, you need to call the resulting object.

```python
with ProcessPoolExecutor() as exe:
    call = Caller(myfunc, write, collect, 3, [3,5], a = 'More kwargs')
    fs = exe.submit(call)
    print(fs.result())
```

This works for me with the plain `ProcessPoolExecutor` and `pympipool.Executor` when the functions are defined in a file.  When defined in a notebook it has to fail with the normal pickler of course.  I could only test `pympipool.Executor` locally, because on a notebook on the cluster it seems to submit automatically to SLURM(?) and my queue is already full.

EDIT: to quick for myself

In practice my understanding was that `write_input_function` and `collect_function` would be statically defined for each simulation code, so that this would work even with the plain python executor, or not?

pyiron_base/jobs/job/runfunction.py
samwaseda: Maybe to make it clearer: `collect_output(working_directory, **output_parameter_dict)` is the same as `collect_output(working_directory)` if `output_parameter_dict` is an empty dict

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: For `pympipool` you can enforce a specific backend by setting `backend="local"` or basically anything other than `"flux", "slurm", "auto"`. The current default is `"auto"` which causes pympipool to use SLURM when available.

pmrv: Here's an updated example that works on my local machine.  It's not running on the cluster, because we don't seem to have pympipool==0.8 available yet.

```python
import os

from pympipool import Executor
from pickle import loads, dumps

def write(input_dict):
    print("Im writing to disk!")
    print(input_dict)

def collect():
    print("I'm reading this from disk, honest!")
    return 42

def myfunc(f0, f1, *args, **kwargs):
    print(args)
    r = f0(kwargs)
    return f1()

class Caller:
    __slots__ = ('func', 'args', 'kwargs')

    def __init__(self, func, *args, **kwargs):
        self.func = func
        self.args = args
        self.kwargs = kwargs

    def __call__(self):
        return self.func(*self.args, **self.kwargs)

def calculate(write_function, collect_function, input_dict=None, executable_dict=None):
    if input_dict is not None:
        write_function(input_dict)

    # do something with executable_dict
    os.system("date")

    output = collect_function()
    return output


def generate_calculate(job, write=None, collect=None):
    return Caller(
            calculate,
            # presumably write/collect would be defined per job/simulation code
            # they could live as static methods on the job classes themselves,
            # then we'd be able to use them here as job.write, job.collect
            write,
            collect,
            **job.generate_calculate_function_kwargs()
    )

if __name__ == "__main__":
    call = Caller(myfunc, write, collect, 3, [3,5], a='More kwargs')
    loads(dumps(call))()
    with ProcessPoolExecutor() as exe:
        fs = exe.submit(call)
        print(fs.result())
    with Executor() as exe:
        fs = exe.submit(call)
        print(fs.result())
```

pyiron_base/jobs/job/runfunction.py
jan-janssen: Ok, I did not know that `**{}` is just ignored. I updated the code with your suggestion.

pyiron_base/jobs/flex/executablecontainer.py
jan-janssen: I implemented the `Caller` class and it is now working the regular executor, at least for standard jobs like LAMMPS.

pyiron_base/jobs/job/template.py
```diff
@@ -122,5 +122,5 @@ class PythonTemplateJob(TemplateJob):
 
     def __init__(self, project, job_name):
         super().__init__(project, job_name)
-        self._python_only_job = True
+        self._job_with_calculate_function = True
```
pmrv, line 125: I don't this will actually work for `PythonTemplateJob`, because the `CalculateCaller` is now expected to run a script and won't touch `run_static` (which is what implementors are supposed to overload), right?

pmrv (changes requested):
Like I wrote, I don't think this will work with the `TemplateJob`, but it also doesn't need to be unified now.  Keeping `_python_only_job` around and checking for both in `_check_if_input_should_be_written` should be sufficient.

pyiron_base/jobs/job/template.py
jan-janssen: As `run_static()` is defined in `PythonTemplateJob` it overwrites the `run_static()` in `GenericJob`.

pmrv (approved):
Correct, I missed that it's overloaded anyway in case of TemplateJob.

PR #1484, opened 2024-06-18
Export server vars to executable scripts

pmrv:
Sets the variables
PYIRON_SERVER_CORES
PYIRON_SERVER_THREADS
PYIRON_SERVER_GPUs
to the corresponding values from job.server.

Needs https://github.com/pyiron/pyiron_snippets/pull/16 and discussion before merge.

### DIFF
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -676,12 +676,11 @@ def execute_job_with_external_executable(job):
     )
     job_crashed, out = False, None
 
-
     try:
         with TemporaryEnvironment(
-                PYIRON_SERVER_CORES=str(job.server.cores),
-                PYIRON_SERVER_THREADS=str(job.server.threads),
-                PYIRON_SERVER_GPUS=str(job.server.gpus),
+            PYIRON_SERVER_CORES=str(job.server.cores),
+            PYIRON_SERVER_THREADS=str(job.server.threads),
+            PYIRON_SERVER_GPUS=str(job.server.gpus),
         ):
             out = execute_subprocess(
                 executable=executable,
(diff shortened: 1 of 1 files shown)


pyiron_base/jobs/job/runfunction.py
```diff
@@ -674,14 +675,20 @@ def execute_job_with_external_executable(job):
...
-            shell=shell,
-            working_directory=job.working_directory,
-            conda_environment_name=job.server.conda_environment_name,
-            conda_environment_path=job.server.conda_environment_path,
-        )
+        with TemporaryEnvironment(
+            PYIRON_SERVER_CORES=str(job.server.cores),
+            PYIRON_SERVER_THREADS=str(job.server.threads),
+            PYIRON_SERVER_GPUS=str(job.server.gpus),
+        ):
```
jan-janssen, line 684: This also affects the python process pyiron is running in. Maybe what we want instead is setting it directly to the subprocess https://docs.python.org/3/library/subprocess.html#popen-constructor using `Popen(env)`

pmrv: Correct, but the context manager will unset/restore the variables after the block and during the execution of the script nothing will be done in the python process anyway, since we use a blocking call to Popen, so I don't expect any problems from this.

But either way will work and it really doesn't matter to me how it is implemented.

jan-janssen: I have a slight preference to do it in the subprocess directly to stay closer to the Python standard library: https://github.com/pyiron/pyiron_base/pull/1485

liamhuber:
@pmrv  #1485 doesn't seem to need it, but in case you guys come back and settle on this, I just made the necessary snippets release https://github.com/pyiron/pyiron_snippets/releases/tag/pyiron_snippets-0.1.2 -- should be available on conda in the near future.

PR #1485, opened 2024-06-18
Executable: Set environment variables for subprocess

jan-janssen:
Set the environment variables `PYIRON_CORES`, `PYIRON_THREADS` and `PYIRON_GPUS` when pyiron_base calls an external executable. Only set these variables for the `subprocess`.


pmrv (approved):
lgtm

PR #1505, opened 2024-07-02
PythonFunctionContainerJob: Use calculate() function

jan-janssen:
Rather than having a custom `run_static()` function for the `PythonFunctionContainerJob`, it now implements the `get_calculate_function()` function and the `calculate_kwargs` property, just like all the new job classes. This required moving the `_get_executor()` function from the `GenericJob` to a static function.


jan-janssen:
For some reason these changes sometimes causes the continuous integration environment to fail, so this should not be merged at the moment. 

EDIT: Increasing the time out fixed this issue. 
EDT: It is not fixed - so I move this pull request to draft state again.

PR #1506, opened 2024-07-02
wrap_python_function(): Allow calling the function directly

jan-janssen:
Example:
```python

def my_function(a, b):
    return a + b

pr = Project("test")
print(pr.wrap_python_function(my_function, a=1, b=2))
>>> 3
```

Previous syntax: 
```python
print(pr.wrap_python_function(my_function)(a=1, b=2))
>>> 3
```

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -482,18 +482,20 @@ class Project(ProjectPath, HasGroups):
         return table
 
     def wrap_python_function(
-        self, python_function, job_name=None, automatically_rename=True
+        self, python_function, *args, job_name=None, automatically_rename=True, **kwargs
     ):
         """
         Create a pyiron job object from any python function
 
         Args:
             python_function (callable): python function to create a job object from
+            *args: Arguments for the user-defined python function
             job_name (str | None): The name for the created job. (Default is None, use
                 the name of the function.)
             automatically_rename (bool): Whether to automatically rename the job at
                 save-time to append a string based on the input values. (Default is
                 True.)
+            **kwargs: Keyword-arguments for the user-defined python function
 
         Returns:
             pyiron_base.jobs.flex.pythonfunctioncontainer.PythonFunctionContainerJob: pyiron job object
@@ -520,7 +522,7 @@ class Project(ProjectPath, HasGroups):
         )
         job._automatically_rename_on_save_using_input = automatically_rename
         job.python_function = python_function
-        return job
+        return job(*args, **kwargs)
 
     def get_child_ids(self, job_specifier, project=None):
         """
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -482,7 +482,7 @@ class Project(ProjectPath, HasGroups):
         return table
 
     def wrap_python_function(
-        self, python_function, *args, job_name=None, automatically_rename=True, **kwargs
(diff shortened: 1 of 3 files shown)


pmrv (commented):
For `wrap_executable` the same functionality is behind a `execute_job` switch, maybe here we could have the same argument and pass args and kwargs there?  That might also help in case someone passes an argument to `wrap_python_function` accidentally.

In either way the docstring needs to mention that the job is immediately run, if arguments are provided.

pyiron_base/project/generic.py
```diff
@@ -482,18 +482,20 @@ def create_table(self, job_name="table", delete_existing_job=False):
         return table
 
     def wrap_python_function(
-        self, python_function, job_name=None, automatically_rename=True
+        self, python_function, *args, job_name=None, automatically_rename=True, **kwargs
```
```suggestion
        self, python_function, job_name=None, automatically_rename=True, execute_job=None
```

pyiron_base/project/generic.py
```diff
@@ -520,7 +522,10 @@ def wrap_python_function(
         )
         job._automatically_rename_on_save_using_input = automatically_rename
         job.python_function = python_function
-        return job
+        if not args and len(kwargs) == 0:
+            return job
+        else:
+            return job(*args, **kwargs)
```
```suggestion

        if execute_job is not None:
            return job(*execute_job[0], **execute_job[1])
        else:
            return job
```

pmrv:
> For `wrap_executable` the same functionality is behind a `execute_job` switch, maybe here we could have the same argument and pass args and kwargs there? That might also help in case someone passes an argument to `wrap_python_function` accidentally.
> 
> In either way the docstring needs to mention that the job is immediately run, if arguments are provided.

I made the suggestions just to clarify what I mean, it doesn't need to be done this way.

jan-janssen:
> > For `wrap_executable` the same functionality is behind a `execute_job` switch, maybe here we could have the same argument and pass args and kwargs there? That might also help in case someone passes an argument to `wrap_python_function` accidentally.
> > In either way the docstring needs to mention that the job is immediately run, if arguments are provided.
> 
> I made the suggestions just to clarify what I mean, it doesn't need to be done this way.

I am fine with adding the `execute_job=False` parameter to only call the job when this flag is set and otherwise just set the input parameters without calling the job.

pmrv:
> > > For `wrap_executable` the same functionality is behind a `execute_job` switch, maybe here we could have the same argument and pass args and kwargs there? That might also help in case someone passes an argument to `wrap_python_function` accidentally.
> > > In either way the docstring needs to mention that the job is immediately run, if arguments are provided.
> > 
> > 
> > I made the suggestions just to clarify what I mean, it doesn't need to be done this way.
> 
> I am fine with adding the `execute_job=False` parameter to only call the job when this flag is set and otherwise just set the input parameters without calling the job.

Sounds good to me.  In this case the provided args/kwargs would be the defaults in case the wrapped function is called multiple times?

jan-janssen:
> Sounds good to me. In this case the provided args/kwargs would be the defaults in case the wrapped function is called multiple times?

Yes, basically the `job.input` data container typically contains the defaults of the function, but if you call `wrap_python_function(*args, ***kwargs)` you overwrite these defaults with your own `*args, **kwargs`. I guess typically you just want to call `job.run()` afterwards. Still for example to configure additional resources like a GPU this setting might be helpful.

jan-janssen:
@pmrv Again, from my perspective the pull request is now ready to merge, unless you have further questions.

pmrv (approved):

PR #1508, opened 2024-07-02
Delayed execution in pyiron_base - for wrap_python_function() and wrap_executable()

jan-janssen:
Example:
```python

def add_x_and_y(x, y):
    z = x + y
    return z

def add_x_and_y_and_z(x, y, z):
    w = x + y + z
    return w


pr = Project("test")
z = pr.wrap_python_function(python_function=add_x_and_y, x=1, y=2, delayed=True)
w = pr.wrap_python_function(python_function=add_x_and_y_and_z, x=1, y=2, z=z, delayed=True)
w.result()
```

Waiting for:
* https://github.com/pyiron/pyiron_base/pull/1506
* https://github.com/pyiron/pyiron_base/pull/1509
* https://github.com/pyiron/pyiron_base/pull/1514


pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,223 @@
...
+    return {
+        k: None if v.default == inspect._empty else v.default
+        for k, v in inspect.signature(funct).parameters.items()
+    }
+
+
+def get_hash(binary):
+    # Remove specification of jupyter kernel from hash to be deterministic
+    binary_no_ipykernel = re.sub(b"(?<=/ipykernel_)(.*)(?=/)", b"", binary)
+    return str(hashlib.md5(binary_no_ipykernel).hexdigest())
```
samwaseda: I'm ok with this being here for now, but I think this should be moved to `pyiron_snippets` so that there's a unified hashing algorithm for all data.

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def get_graph(obj, obj_name=None, nodes_dict={}, edges_lst=[], link_node=None):
+    if isinstance(obj, DelayedObject):
```
samwaseda:
```suggestion
    if isinstance(obj, DelayedObject) and obj_name is not None:
```

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,223 @@
...
+        try:
+            obj_name = (
+                obj._function.__name__ + "_" + get_hash(binary=cloudpickle.dumps(obj))
+            )
+        except TypeError:
+            obj_name = (
+                str(obj).replace("<", "").replace("<", "").replace(" object at ", "")
```
samwaseda:
```suggestion
                str(obj).replace("<", "").replace(" object at ", "")
```

But to be honest I don't really understand why going through so much hassle...

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,223 @@
...
+    binary_no_ipykernel = re.sub(b"(?<=/ipykernel_)(.*)(?=/)", b"", binary)
+    return str(hashlib.md5(binary_no_ipykernel).hexdigest())
+
+
+def get_graph(obj, obj_name=None, nodes_dict={}, edges_lst=[], link_node=None):
+    if isinstance(obj, DelayedObject):
+        try:
+            obj_name = (
+                obj._function.__name__ + "_" + get_hash(binary=cloudpickle.dumps(obj))
+            )
```
samwaseda:
```suggestion
            obj_name = obj._function.__name__
```

Since hash is added to `obj_name`, it's not necessary here.

pyiron_base/project/delayed.py
samwaseda: Isn't it potentially super dangerous, because what comes after `object at` is just the location of the memory?

pyiron_base/project/delayed.py
```diff
@@ -57,7 +57,7 @@ def get_graph(
     Returns:
         dict, list: dictionary of nodes and list of edges
     """
-    if isinstance(obj, DelayedObject):
+    if isinstance(obj, DelayedObject) and obj_name is not None:
```
samwaseda:
```suggestion
    if isinstance(obj, DelayedObject) and obj_name is None:
```

Sorry I made a mistake XD

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,273 @@
...
+    if obj_name is None:
+        try:
+            obj_name = obj.__name__
+        except AttributeError:
+            obj_name = str(type(obj))
+    try:
+        node_name = obj_name + "_" + get_hash(binary=cloudpickle.dumps(obj))
+    except TypeError:
+        node_name = obj_name
+    nodes_dict.update({node_name: obj})
```
samwaseda: This part should be a separate function. `get_graph` is too long otherwise.

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,273 @@
...
+    from IPython.display import SVG, display
+
+    graph = nx.DiGraph()
+    for k, v in node_dict.items():
+        graph.add_node(k, label=str(k).rsplit("_", 1)[0] + "=" + str(v))
+    for edge in edge_lst:
+        graph.add_edge(edge[1], edge[0])
+    svg = nx.nx_agraph.to_agraph(graph).draw(prog="dot", format="svg")
+    display(SVG(svg))
+    plt.show()
```
samwaseda:
```suggestion
    return display(SVG(svg))
```

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,273 @@
...
+        if self._selector == "files" and name in self._obj._output_file_lst:
+            obj_copy = self._obj.__copy__()
+            obj_copy._output_file = name
+            return obj_copy
+        elif self._selector == "output" and name in self._obj._output_key_lst:
+            obj_copy = self._obj.__copy__()
+            obj_copy._output_key = name
+            return obj_copy
+        else:
+            raise AttributeError()
```
samwaseda:
```suggestion
            return self.__getattribute__(name)
```

It's the same thing but at least it shows an error message.

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,273 @@
...
+        )
+        obj_copy._input = self._input
+        obj_copy._result = self._result
+        return obj_copy
+
+    def __getattr__(self, name):
+        if name in ["files", "output"]:
+            return Selector(obj=self, selector=name)
+        else:
+            raise AttributeError()
```
samwaseda:
```suggestion
            return self.__getattribute__(name)
```

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,273 @@
...
+            else:
+                nodes_dict, edges_lst = get_graph(
+                    obj=v,
+                    obj_name=k,
+                    nodes_dict=nodes_dict,
+                    edges_lst=edges_lst,
+                    link_node=node_name,
+                )
+    elif isinstance(obj, dict) and any(
+        [isinstance(v, DelayedObject) for v in obj.values()]
```
samwaseda, line 111: I don't quite understand the situation here. How is it possible that there's a list that could partially contain `DelayedObject`s?

pyiron_base/project/delayed.py
```diff
@@ -0,0 +1,273 @@
...
+    def result(self):
+        if self._result is None:
+            input_dict = recursive_dict_resolve(input_dict=self._input)
+            if "kwargs" in input_dict.keys():
+                input_dict.update(input_dict.pop("kwargs"))
+            if "args" in input_dict.keys():
+                args = input_dict.pop("args")
+                self._result = self._function(*args, **input_dict)
+            else:
+                self._result = self._function(**input_dict)
```
samwaseda: There are two steps done in one function here. I would suggest to separate `def result` and `def get_result` with

```python
    @cached_property
    def result(self):
        input_dict = recursive_dict_resolve(input_dict=self._input)
        if "kwargs" in input_dict.keys():
            input_dict.update(input_dict.pop("kwargs"))
        if "args" in input_dict.keys():
            args = input_dict.pop("args")
            return self._function(*args, **input_dict)
        else:
            return self._function(**input_dict)
```

and

```python
    def get_result(self):
        if self._output_key is not None:
            return self.get_python_result()
        elif self._output_file is not None:
            return self.get_file_result()
        elif self._list_index is not None:
            return self.result[self._list_index]
        else:
            return self.result
```

And everywhere `self._result` should be replaced by `self.result` and the current `self.result` should be replaced by `self.get_result()`. Not to forget to import `cached_property`

samwaseda: maybe `result` should indeed be a private function. I leave the decision to you.

pyiron_base/project/delayed.py
jan-janssen: The graph in `pyiron_base` is constructed by using one `DelayedObject` as an input for the next `DelayedObject`. If you take a look at the NFDI4Ing example https://github.com/pyiron-dev/NFDI4Ing_pyiron_base/blob/pyiron_base/demo_graph.ipynb then the input files for one calculation consist of a list of files some of these are of type `DelayedObject` but not all of them.

pyiron_base/project/delayed.py
jan-janssen: I moved the evaluation of the function to a separate function.

pyiron_base/project/delayed.py
jan-janssen: It is now separate in an `get_node_name()` function.

jan-janssen:
@samwaseda I would like to include this in the next `pyiron_base` version. Is there anything missing from your side? Do you have a quick moment to review the latest changes?

samwaseda (approved):
There are still some small things I would love to change but they can belong to other PR's.

PR #1509, opened 2024-07-02
ExecutableContainerJob: Fix copying of restart files

jan-janssen:
When the user defines the `write_input()` function then this typically does not contain a call to `super().write_input()` so the restart files are not copied. This is now fixed by extending the user-defined `write_input()` function to also copy the restart files.

### DIFF
--- a/pyiron_base/jobs/flex/executablecontainer.py
+++ b/pyiron_base/jobs/flex/executablecontainer.py
@@ -2,7 +2,10 @@ import cloudpickle
 import numpy as np
 
 from pyiron_base.jobs.job.template import TemplateJob
-from pyiron_base.jobs.job.runfunction import CalculateFunctionCaller, write_input_files_from_input_dict
+from pyiron_base.jobs.job.runfunction import (
+    CalculateFunctionCaller,
+    write_input_files_from_input_dict,
+)
 
 
 class ExecutableContainerJob(TemplateJob):
@@ -115,6 +118,7 @@ class ExecutableContainerJob(TemplateJob):
         Returns:
             callable: calculate() functione
         """
+
         def get_combined_write_input_funct(input_job_dict, write_input_funct):
             def write_input_combo_funct(working_directory, input_dict):
                 write_input_files_from_input_dict(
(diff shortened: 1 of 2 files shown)


pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -115,8 +118,26 @@ def get_calculate_function(self) -> callable:
         Returns:
             callable: calculate() functione
         """
+
+        def get_combined_write_input_funct(input_job_dict, write_input_funct=None):
+            def write_input_combo_funct(working_directory, input_dict):
+                write_input_files_from_input_dict(
+                    input_dict=input_job_dict,
+                    working_directory=working_directory,
+                )
```
pmrv, line 127: It might be easier to follow if we add this call in [CalculateFunctionCaller](https://github.com/pyiron/pyiron_base/blob/51ba6c28a4dddff17bd75ca29971d8659bc652ba/pyiron_base/jobs/job/runfunction.py#L134) directly or would this break something else?

jan-janssen: I looked at it. In the end it comes down to the question: Do developers of `GenericJob` classes remember to add the `super().get_input_parameter_dict()` call to their `get_input_parameter_dict()` function? We could move around this by implementing an `_get_input_parameter_dict()` function but I am not sure if that is more intuitive for scientists. At the same time I like that all input files are now handled via the same interface. 

For the specific case of the `wrap_executable()` function it is harder to add the `super()` call so here we have the internal workaround in the internal class which wraps the executable. 

In summary, I am reluctant to implement this functionality directly in the `CalculateFunctionCaller` as it would break the current scheme in which the content of all files is communicated via the `get_input_parameter_dict()` function.

jan-janssen:
@pmrv If you have a moment to take another look at this pull request. I would like to merge this one to afterwards finish up https://github.com/pyiron/pyiron_base/pull/1508

pyiron_base/jobs/flex/executablecontainer.py
pmrv: Right, because for normal jobs writing of the restart files is part of the passed in write input function anyway. I'm a bit concerned that this makes it harder again to pickle the executable container, but then we may have look at it in the future.

pmrv (approved):

PR #1511, opened 2024-07-03
Add write_input() before transferring files to remote

jan-janssen:
This pull request addresses the issue raised by @Leimeroth in https://github.com/pyiron/pyiron_atomistics/issues/1471 that the remote submission for `LAMMPS` jobs failed after the switch to using the `calculate()` function. While it is not exactly clear why the `VASP` job works fine, the work around in this branch fixes the issue for `LAMMPS`.

### DIFF
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -168,7 +168,6 @@ class GenericJob(JobCore, HasDict):
         self._process = None
         self._compress_by_default = False
         self._job_with_calculate_function = False
-        self._input_is_written = False
         self._write_work_dir_warnings = True
         self.interactive_cache = None
         self.error = GenericError(working_directory=self.project_hdf5.working_directory)
@@ -633,15 +632,10 @@ class GenericJob(JobCore, HasDict):
         Call routines that generate the code specific input files
         Returns:
         """
-        if not self._input_is_written:
-            write_input_files_from_input_dict(
-                input_dict=self.get_input_parameter_dict(),
-                working_directory=self.working_directory,
-            )
-        else:
-            self._logger.info(
-                "write_input was called, but input has already been written."
-            )
+        write_input_files_from_input_dict(
+            input_dict=self.get_input_parameter_dict(),
+            working_directory=self.working_directory,
+        )
 
     def _internal_copy_to(
         self,
@@ -1172,7 +1166,6 @@ class GenericJob(JobCore, HasDict):
             "restart_file_dict": self._restart_file_dict,
             "exclude_nodes_hdf": self._exclude_nodes_hdf,
             "exclude_groups_hdf": self._exclude_groups_hdf,
-            "input_is_written": self._input_is_written,
         }
         data_dict["server"] = self._server.to_dict()
         self._executable_activate_mpi()
@@ -1202,7 +1195,6 @@ class GenericJob(JobCore, HasDict):
             self._restart_file_dict = generic_dict["restart_file_dict"]
             self._exclude_nodes_hdf = generic_dict["exclude_nodes_hdf"]
             self._exclude_groups_hdf = generic_dict["exclude_groups_hdf"]
-            self._input_is_written = generic_dict["input_is_written"]
         # Backwards compatbility
         if "restart_file_list" in input_dict.keys():
             self._restart_file_list = input_dict["restart_file_list"]
--- a/pyiron_base/jobs/job/generic.py
+++ b/pyiron_base/jobs/job/generic.py
@@ -168,6 +168,7 @@ class GenericJob(JobCore, HasDict):
         self._process = None
         self._compress_by_default = False
         self._job_with_calculate_function = False
+        self._input_is_written = False
         self._write_work_dir_warnings = True
         self.interactive_cache = None
         self.error = GenericError(working_directory=self.project_hdf5.working_directory)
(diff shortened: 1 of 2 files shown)


pyiron_base/jobs/job/runfunction.py
```diff
@@ -1,14 +1,14 @@
...
-from datetime import datetime
 import multiprocessing
 import os
 import posixpath
 import shutil
 import subprocess
-from typing import Optional, List, Tuple
+from concurrent.futures import ProcessPoolExecutor
+from datetime import datetime
+from typing import List, Optional, Tuple
```
jan-janssen: @Leimeroth I assume these changes are just related to committing from an older version and not related to this pull request, correct? Then I would revert those for this pull request.

Leimeroth: That is related to trying https://docs.astral.sh/ruff/ for linting. Forgot to disable the automatic sorting

jan-janssen: I adopted the ruff sorting for our continuous integration in https://github.com/pyiron/pyiron_base/pull/1512 so these changes are now already included in the `main` branch.

pyiron_base/jobs/job/generic.py
```diff
@@ -632,10 +633,15 @@ def write_input(self):
         Call routines that generate the code specific input files
         Returns:
         """
-        write_input_files_from_input_dict(
-            input_dict=self.get_input_parameter_dict(),
-            working_directory=self.working_directory,
-        )
+        if not self._input_is_written:
```
jan-janssen: @Leimeroth Is there an issue with files being written twice? Meaning the remote job trying to write the files again and in the worst case overwriting the existing input files.

Leimeroth: I think the write_input_funct is called again somewhere on the cluster. This call tries to copy the potential file, 
so I get the errors when using my own potentials.

jan-janssen: How about checking if the working directory is empty and only write the input files when it is:
```suggestion
        if len(os.listdir(self.working_directory)) != 0:
```

pyiron_base/jobs/job/generic.py
```diff
@@ -168,6 +168,7 @@ def __init__(self, project, job_name):
         self._process = None
         self._compress_by_default = False
         self._job_with_calculate_function = False
+        self._input_is_written = False
```
```suggestion
```

pyiron_base/jobs/job/generic.py
```diff
@@ -1166,6 +1172,7 @@ def to_dict(self):
             "restart_file_dict": self._restart_file_dict,
             "exclude_nodes_hdf": self._exclude_nodes_hdf,
             "exclude_groups_hdf": self._exclude_groups_hdf,
+            "input_is_written": self._input_is_written,
```
```suggestion
```

pyiron_base/jobs/job/generic.py
```diff
@@ -1195,6 +1202,7 @@ def from_dict(self, job_dict):
             self._restart_file_dict = generic_dict["restart_file_dict"]
             self._exclude_nodes_hdf = generic_dict["exclude_nodes_hdf"]
             self._exclude_groups_hdf = generic_dict["exclude_groups_hdf"]
+            self._input_is_written = generic_dict["input_is_written"]
```
```suggestion
```

pyiron_base/jobs/job/generic.py
```diff
@@ -632,10 +633,15 @@ def write_input(self):
...
-            working_directory=self.working_directory,
-        )
+        if not self._input_is_written:
+            write_input_files_from_input_dict(
+                input_dict=self.get_input_parameter_dict(),
+                working_directory=self.working_directory,
+            )
+        else:
+            self._logger.info(
+                "write_input was called, but input has already been written."
```
```suggestion
                "write_input was called, but working_directory is not empty."
```

Leimeroth: added a check to the write_input_from_input_dict function

Leimeroth:
Ok, so check if working directory is empty leads to a bunch of test failures and some horrible side effects, because I created the job with delete_existing_job=True. However the contents of the working directory are not deleted on time,
so on my local computer no input is written with the check.

jan-janssen:
> Ok, so check if working directory is empty leads to a bunch of test failures and some horrible side effects, because I created the job with delete_existing_job=True. However the contents of the working directory are not deleted on time, so on my local computer no input is written with the check.

The `!=` and `==` sign were mixed up. You want to write the input files when the directory is empty and do not write them when it is not empty.

jan-janssen:
@Leimeroth Can you run your tests again, like submitting a job with a custom potential to a remote cluster? Once this works and you are fine with the changes as they are in this pull request, can you approve this pull request? Then I can make sure it is included in the release of the next version. I aim to update both `pyiron_base` and `pyiron_atomistics` before the pyiron meeting on Monday.

Leimeroth:
Now it works with my own potential and the EAM.
However, only the check prevents it from rewriting input, which could cause trouble when trying to rerun jobs.
When I run them again using delete_existing_job=True on my local computer the old files are still present on the remote cluster, so they will be used instead of my updated inputs.

jan-janssen:
Are you sure about this? I thought the files are written locally and then transferred. So as the job is first deleted locally before the new input files are written the working directory is removed and a new empty working directory is created. Then at the copying stage we do not use the `write_input()` function so it should still overwrite the files on the cluster.

Leimeroth:
> Are you sure about this? I thought the files are written locally and then transferred. So as the job is first deleted locally before the new input files are written the working directory is removed and a new empty working directory is created. Then at the copying stage we do not use the `write_input()` function so it should still overwrite the files on the cluster.

You are right, I just tested it. Fine to merge from my side.

PR #1513, opened 2024-07-04
Executable: drop old code argument

pmrv:
Was superceded by codename/module and last touched 4 years ago.  I have looked around and could not find any call sites in base/atomistics/contrib/potentialfit that still used this argument.

### DIFF
--- a/pyiron_base/jobs/job/extension/executable.py
+++ b/pyiron_base/jobs/job/extension/executable.py
@@ -31,7 +31,6 @@ class Executable(HasDict):
         path_binary_codes=None,
         codename=None,
         module=None,
-        code=None,
         overwrite_nt_flag=False,
     ):
         """
@@ -49,11 +48,6 @@ class Executable(HasDict):
         if path_binary_codes is None:
             path_binary_codes = state.settings.resource_paths
         self.storage.version = None
-        if code is not None:  # Backwards compatibility
-            if not isinstance(code.__name__, str):
-                raise TypeError("The codename should be a string.")
-            codename = code.__name__
-            module = code.__module__.split(".")[1]
         if codename is not None and module is not None:
             self.storage.name = codename.lower()
             code_path_lst = [


pyiron_base/jobs/job/extension/executable.py
```diff
@@ -289,6 +283,10 @@ def _executable_versions_list(self):
         Returns:
             dict: list of the available version
         """
+        from pyiron_snippets.resources import ResourceResolver
+        return dict(ResourceResolver(
+            self.path_bin,
+
```
jan-janssen: This part seems to break the unit tests.

jan-janssen:
```suggestion
```

pmrv: Ah, upps. That wasn't supposed to be there yet.

pmrv:
On this note, do you know if `overwrite_nt_flag` is used by anything at all?  Otherwise I would also remove it in another PR.

jan-janssen (approved):
Looks good to me

jan-janssen:
> On this note, do you know if `overwrite_nt_flag` is used by anything at all? Otherwise I would also remove it in another PR.

I thought the use case for the `overwrite_nt_flag` was to connect from a windows client to a linux cluster. But I am not sure if this configuration is still available or how to configure the windows client if you would want to use pyiron in this configuration. I think with the ability of installing similar conda environments on both systems we again get to the point where it should be possible to configure such a system https://github.com/jan-janssen/shared-conda-environment . But in the age of the windows subsystem for linux I am not even sure how many people install pyiron natively on windows after all.

pmrv:
> > On this note, do you know if `overwrite_nt_flag` is used by anything at all? Otherwise I would also remove it in another PR.
> 
> I thought the use case for the `overwrite_nt_flag` was to connect from a windows client to a linux cluster. But I am not sure if this configuration is still available or how to configure the windows client if you would want to use pyiron in this configuration. I think with the ability of installing similar conda environments on both systems we again get to the point where it should be possible to configure such a system https://github.com/jan-janssen/shared-conda-environment . But in the age of the windows subsystem for linux I am not even sure how many people install pyiron natively on windows after all.

Ah, right this is a use case I had not thought about.  It's not hurting right now, so I can leave it in.

PR #1514, opened 2024-07-04
PythonFunctionContainerJob: Skip the run() call for jobs which were reloaded during save()

jan-janssen:


pyiron_base/jobs/job/runfunction.py
```diff
@@ -185,7 +185,11 @@ def run_job_with_status_initialized(job, debug=False):
         print("job exists already and therefore was not created!")
     else:
         job.save()
-        job.run()
+        # The PythonFunctionContainerJob can generate the job_name during job.save(). In particular, this can lead to
+        # the job being reloaded from the database resulting in the job status to change from initialized to finished.
+        # Skipping the job.run() prevents raising a warning by calling job.run() on an already finished job.
+        if not job.status.finished:
```
```suggestion
        if not job.status.finished and not job.status.aborted and not job.status.not_converged:
```

samwaseda (approved):
I'm glad that you put a very extensive doc string (thanks to which I'm ready to accept the current code), but this is again a kind of place where I have to say it lacks an underlying concept. From my point of view, the step we should make is to allow any job to not have any job name, and the step to generate a job name should be properly included in the workflow. I think the handiest place to do it in the current implementation would be in `validate_ready_to_run`, which would also avoid `check_setup` which may or may not cause a problem if we allow it to run. But my gut feeling tells me that it should be yet another step before that just for the generation of the job name.

All this being said, since it's well documented, I'm also ok with merging it in the current format.

PR #1520, opened 2024-07-10
Refactor tests

jan-janssen:


pmrv:
What is running the backwards tests now?

jan-janssen:
> What is running the backwards tests now?

The backwards tests were not running. I mean the action was executed but no tests were run.

pmrv:
[It seems] the problem was versioneer, maybe an old incompatible version.  Imo go ahead and remove the files completely then, since they serve no purpose in this form.

PR #1521, opened 2024-07-10
Add usecases to the tests

jan-janssen:


jan-janssen:
@srmnitc and @samwaseda I added use cases for `pyiron_base` directly to the test suite to guarantee we maintain API compatibility. If you have use cases for Material Digital and NFDI Matwerk you would like to include here as well feel free to add them. The goal is to use these use cases to define a stable API of `pyiron_base` so that we can converge to a version 1.0, where we guarantee that at least the commands used in those use cases remain backwards compatible. 

EDIT: The use cases consist of a jupyter notebook a corresponding GitHub action and additional resources.

PR #1528, opened 2024-07-12
Use resolver classes from pyiron_snippets

pmrv:
Requires a version bump on snippets before merging.


jan-janssen:
Waiting for https://github.com/pyiron/pyiron_base/pull/1533

pmrv:
> Waiting for #1533

Seems like we'll need Liam for it to make the conda release (since I don't have rights on the feedstock).  Since this PR is not critical, it doesn't need to be in the next release.

jan-janssen:
> > Waiting for #1533
> 
> Seems like we'll need Liam for it to make the conda release (since I don't have rights on the feedstock). Since this PR is not critical, it doesn't need to be in the next release.

It should be automated - just give it a bit of time 

EDIT: No Liam does not use the auto merge functionality. I requested access over a month ago but that pull request is still pending https://github.com/conda-forge/pyiron_snippets-feedstock/pull/2

tests/unit/state/test_queue_adapter.py
```diff
@@ -57,7 +57,7 @@ def tearDownClass(cls) -> None:
 
     def test_construction(self):
         self.assertEqual(
-            2,
+            3,
```
jan-janssen: @pmrv Do you understand why the number of matches increased? I just want to prevent that the same files are read twice.

pmrv: The new code finds these three files
```python
['/home/ponder/science/phd/dev/pyiron_base/testqueueadapters/res1/queues/queue.yaml',
 '/home/ponder/science/phd/dev/pyiron_base/testqueueadapters/res2/queues/queue.yaml', 
'/home/ponder/science/phd/dev/pyiron_base/testqueueadapters/res2/queues/clusters.yaml']
```
which are the ones that are created in the `setUpClass`.  It then creates a queue adapter for each of the files. But the way QueueAdapter works is that it wants to be passed the directory that the queue/cluster configuration, so the last two files give the same queue adapter.  It's an easy fix, but going forward do we even want to search for `clusters.yaml` anymore?

jan-janssen: Yes, the `clusters.yaml` is used in the case when pyiron connects to multiple HPC clusters with different queuing systems. But I think it is fine, with the recent changes it only looks for directories and then `pysqa` is handling the rest.

jan-janssen (approved):
Looks good to me

PR #1529, opened 2024-07-13
wrap_executable(): no longer require job_name - enable automatic renaming

jan-janssen:
Coming up with `job_name`s for pyiron job objects is challenging, so in addition to the `wrap_python_function()` function now also the `wrap_executable()` function supports generating job names based on the input provided.


pyiron_base/project/generic.py
```diff
@@ -367,14 +367,15 @@ def create_job_class(
 
     def wrap_executable(
         self,
-        job_name,
         executable_str,
+        job_name="exe",
```
samwaseda:
```suggestion
        job_name=None,
```

pyiron_base/project/generic.py
```diff
@@ -452,6 +459,8 @@ def create_exeuctable_job(
             if internal_file_lst is not None and len(internal_file_lst) > 0:
                 for file in internal_file_lst:
                     job.restart_file_list.append(file)
+            if automatically_rename:
+                job._automatically_rename_on_save_using_input = automatically_rename
```
samwaseda:
```suggestion
```

tests/unit/flex/test_executablecontainer.py
```diff
@@ -215,11 +215,40 @@ def test_series_of_jobs(self):
             input_dict={"x": 1, "y": 2},
             conda_environment_path=None,
             conda_environment_name=None,
+            automatically_rename=False,
```
samwaseda:
```suggestion
```

tests/unit/flex/test_executablecontainer.py
```diff
@@ -215,11 +215,40 @@ def test_series_of_jobs(self):
...
         )
         w = self.project.wrap_executable(
             job_name="job_xyz",
+            executable_str="x=$(cat x.txt); y=$(cat y.txt); z=$(cat result.txt); echo $(($x + $y + $z)) > result.txt",
+            write_input_funct=write_input_series,
+            collect_output_funct=collect_output_series,
+            input_dict={"x": 1, "y": z.output.result},
+            conda_environment_path=None,
+            conda_environment_name=None,
+            automatically_rename=False,
```
samwaseda:
```suggestion
```

pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -52,6 +53,12 @@ def __init__(self, project, job_name):
         # interface with write_input() and collect_output(). Finally, the output dictionary is stored in the HDF5 file
         # using self.save_output(output_dict, shell_output)
         self._job_with_calculate_function = True
+        # Automatically rename job using function and input values at save time
+        # This is useful for the edge case where these jobs are created from a wrapper
+        # and automatically assigned a name based on the function name, but multiple
+        # jobs are created from the same function (and thus distinguished only by their
+        # input)
+        self._automatically_rename_on_save_using_input = False
```
samwaseda:
```suggestion
```

pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -77,6 +84,28 @@ def calculate_kwargs(self) -> dict:
         )
         return kwargs
 
+    def save(self):
+        if self._automatically_rename_on_save_using_input:
```
samwaseda:
```suggestion
        if self.job_name is None:
```

pyiron_base/jobs/flex/executablecontainer.py
```diff
@@ -77,6 +84,28 @@ def calculate_kwargs(self) -> dict:
...
+            self.job_name = (
+                self.job_name
```
samwaseda:
```suggestion
                self.__class__.__name__
```

pyiron_base/project/generic.py
```diff
@@ -367,14 +367,15 @@ def create_job_class(
...
-        job_name,
         executable_str,
+        job_name="exe",
         write_input_funct=None,
         collect_output_funct=None,
         input_dict=None,
         conda_environment_path=None,
         conda_environment_name=None,
         input_file_lst=None,
+        automatically_rename=True,
```
samwaseda:
```suggestion
```

pyiron_base/project/generic.py
```diff
@@ -384,15 +385,21 @@ def wrap_executable(
...
             write_input_funct (callable): The write input function write_input(input_dict, working_directory)
             collect_output_funct (callable): The collect output function collect_output(working_directory)
             input_dict (dict): Default input for the newly created job class
             conda_environment_path (str): path of the conda environment
             conda_environment_name (str): name of the conda environment
             input_file_lst (list): list of files to be copied to the working directory before executing it\
             execute_job (boolean): automatically call run() on the job object - default false
+            automatically_rename (bool): Whether to automatically rename the job at
+                save-time to append a string based on the input values. (Default is
+                True.)
```
samwaseda:
```suggestion
```

samwaseda (commented):
For the sake of backward compatibility, when the job name is set it should be used by default. Instead of defining it, I would set the default to `None`, in which case it will be replaced. In addition to this, the current implementation keeps changing the job name multiple times if `save` is called by mistake, which shouldn't happen but could. Besides I don't think it's clear to new users what `automatically_rename` means beyond just renaming the job.

samwaseda:
Sorry my change suggestions were really difficult to read. They have all the same purpose of fulfilling what I wrote in my comment above.

jan-janssen:
> For the sake of backward compatibility, when the job name is set it should be used by default. Instead of defining it, I would set the default to `None`, in which case it will be replaced. 

I can see the importance of backwards compatibility, still my goal was consistency between `wrap_python_function()` and `wrap_executable()`. As both functions were introduced rather recently, I would prefer to break the backwards compatibility to have both functions consistent. 

> In addition to this, the current implementation keeps changing the job name multiple times if `save` is called by mistake, which shouldn't happen but could. 

In principle a user should never call `save()` directly. I agree that we had those cases in the past, so maybe it should be more hidden. 

> Besides I don't think it's clear to new users what `automatically_rename` means beyond just renaming the job.

Fair, in that case we should rename it for both `wrap_python_function()` and `wrap_executable()`

samwaseda:
> I can see the importance of backwards compatibility, still my goal was consistency between `wrap_python_function()` and `wrap_executable()`. As both functions were introduced rather recently, I would prefer to break the backwards compatibility to have both functions consistent.

But with my suggestion you can keep the backward compatibility while allowing the user to not define the job name. Why not take this option?

jan-janssen:
@samwaseda I implemented the solution @pmrv suggested in the pyiron meeting.

samwaseda:
It was @pmrv and **my** suggestion!!

samwaseda:
Anyway, you haven't addressed the problem that by calling `save` twice you can change the job name twice. So the conversion of `job_name = None` to a real job name must take place also inside `save` to solve this problem

pyiron_base/project/generic.py
```diff
@@ -452,13 +464,15 @@ def create_exeuctable_job(
             if internal_file_lst is not None and len(internal_file_lst) > 0:
                 for file in internal_file_lst:
                     job.restart_file_list.append(file)
+            if automatically_rename:
```
samwaseda: Do we need this if statement?

pyiron_base/project/generic.py
samwaseda:
```suggestion
                False.)
```

pyiron_base/project/generic.py
jan-janssen: I agree - I removed it.

samwaseda:
Now I thought about it a little bit more, but wouldn't it be the safest choice to change the job name right at the beginning of `wrap_executable`? I'm saying it because it is too much of a hack to do it inside `save` as it wasn't made for this purpose. Of course, this shouldn't be the final picture, but I just don't find changing the job name inside `save` fits the current pyiron paradigm and if so, then we should do it in the safest place possible.

jan-janssen:
> Now I thought about it a little bit more, but wouldn't it be the safest choice to change the job name right at the beginning of `wrap_executable`? 

I really like this suggestion and implemented it in the latest changes.

samwaseda (approved):

PR #1537, opened 2024-07-16
Add optional dependencies: gitpython, jinja2 and pint

jan-janssen:
The goal is to reduce the dependency footprint for installing `pyiron_base`. This is tested with a new unit test called `Minimal` which is based only on a minimal conda environment without the optional requirements.


samwaseda:
I don't really think I'm the right person to review this PR. Maybe @pmrv or @niklassiemer ?

niklassiemer (approved):
LGTM!

PR #1542, opened 2024-07-19
Bugfix: Interactive calculation array comparison

jan-janssen:
This seems to related to the recent numpy updates

### DIFF
--- a/pyiron_base/jobs/job/interactive.py
+++ b/pyiron_base/jobs/job/interactive.py
@@ -280,7 +280,9 @@ class InteractiveBase(GenericJob):
                     else:
                         self._extend_hdf(h5=h5, path=path, key=key, data=np.array(data))
                 except ValueError:
-                    self._extend_hdf(h5=h5, path=path, key=key, data=np.array(data, dtype="object"))
+                    self._extend_hdf(
+                        h5=h5, path=path, key=key, data=np.array(data, dtype="object")
+                    )
                 self.interactive_cache[key] = []
 
     def interactive_open(self):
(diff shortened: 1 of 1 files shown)


pyiron_base/jobs/job/interactive.py
```diff
@@ -268,16 +268,21 @@ def interactive_flush(self, path="interactive", include_last_step=False):
...
+                    if (
+                        len(data) > 0
+                        and isinstance(data[0], list)
+                        and len(np.shape(data)) == 1
+                    ):
+                        self._extend_hdf(h5=h5, path=path, key=key, data=data)
+                    elif np.array(data).dtype == np.dtype("O"):
+                        self._extend_hdf(h5=h5, path=path, key=key, data=data)
+                    else:
+                        self._extend_hdf(h5=h5, path=path, key=key, data=np.array(data))
```
niklassiemer, line 281: All these calls to the same function are very similar. Actually, why can the stuff from the except clause not be done here? If not a dtype=O, then construct such an array instead of a regular one?

jan-janssen: With the recent numpy version `np.array(data)` fails, unless `np.array(data, dtype="object")` is set.

pmrv: This probably means that this clause `elif np.array(data).dtype == np.dtype("O")` is never trigger, because it raises an exception.

jan-janssen: I agree, but I am not sure. The clean way would require the user to specify if the number of atoms is going to change in their simulation or not and then choose the corresponding approach for storing the structures. So this is really just a quick fix to use interactive jobs to evaluate machine learned interatomic potentials to calculate the energies and forces for the test and training set.

PR #1549, opened 2024-07-23
Use `copytree` in `pack`

samwaseda:
I largely simplified the step to create directories and copy files. Hopefully the functionality is the same.

PR #1551, opened 2024-07-24
Restore sub directories

samwaseda:
In the end it looks like @pmrv was right about [this statement](https://github.com/pyiron/pyiron_base/pull/1549#discussion_r1688597070). I asked ChatGPT to rewrite it and also to make tests. I hope now it's fine.

### DIFF
--- a/pyiron_base/project/archiving/export_archive.py
+++ b/pyiron_base/project/archiving/export_archive.py
@@ -63,15 +63,20 @@ def copy_files_to_archive(
     if copy_all_files:
         copytree(directory_to_transfer, dst, dirs_exist_ok=True)
     else:
-        copytree(
-            directory_to_transfer, dst, ignore=ignore_non_h5_files, dirs_exist_ok=True
-        )
+        copy_h5_files(directory_to_transfer, dst)
     if compressed:
         compress_dir(archive_directory)
 
 
-def ignore_non_h5_files(dir, files):
-    return [f for f in files if not f.endswith(".h5")]
+def copy_h5_files(src, dst):
+    for root, dirs, files in os.walk(src):
+        for file in files:
+            if file.endswith(".h5"):
+                src_file = os.path.join(root, file)
+                rel_path = os.path.relpath(root, src)
+                dst_dir = os.path.join(dst, rel_path)
+                os.makedirs(dst_dir, exist_ok=True)
+                shutil.copy2(src_file, os.path.join(dst_dir, file))
 
 
 def export_database(pr, directory_to_transfer, archive_directory):
--- a/pyiron_base/project/archiving/export_archive.py
+++ b/pyiron_base/project/archiving/export_archive.py
@@ -1,12 +1,10 @@
 import os
 import tarfile
-from shutil import copytree, rmtree
+import shutil
 
 import numpy as np
-from pyfileindex import PyFileIndex
 
 from pyiron_base.project.archiving.shared import getdir
-from pyiron_base.utils.instance import static_isinstance
 
 
 def new_job_id(job_id, job_translate_dict):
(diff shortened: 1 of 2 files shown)


jan-janssen:
@samwaseda I have the feeling the current unit tests were not sufficient to find this error. Can you extend the unit tests so we at least get an error the next time somebody tries something like this?

samwaseda:
> @samwaseda I have the feeling the current unit tests were not sufficient to find this error. Can you extend the unit tests so we at least get an error the next time somebody tries something like this?

If you are talking about `export_archive.py` in general, I agree, but I guess that's a different PR. For the problem that I encountered here, I guess the test I added here should be good enough.

samwaseda:
Speaking of which, I realized a huge advantage of the function approach. You can copy and paste the code snippet to ChatGPT, state the problem, ask it to rewrite the code, and write unit tests accordingly. This should be more streamlined.

jan-janssen:
> Speaking of which, I realized a huge advantage of the function approach. You can copy and paste the code snippet to ChatGPT, state the problem, ask it to rewrite the code, and write unit tests accordingly. This should be more streamlined.

I guess the streamlined solution is Github Co-Pilot - maybe @srmnitc can comment on this.

jan-janssen:
The windows tests fail:
```
FAIL: test_copy_h5_files (archiving.test_export.TestPack.test_copy_h5_files)
  File "C:\Users\runneradmin\miniconda3\envs\my-env\Lib\unittest\mock.py", line 1390, in patched
    return func(*newargs, **newkeywargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\pyiron_base\pyiron_base\tests\unit\archiving\test_export.py", line 118, in test_copy_h5_files
    mock_makedirs.assert_has_calls(expected_makedirs_calls, any_order=True)
  File "C:\Users\runneradmin\miniconda3\envs\my-env\Lib\unittest\mock.py", line 997, in assert_has_calls
    raise AssertionError(
AssertionError: 'makedirs' does not contain all of (call('/mock/dst/.', exist_ok=True), call('/mock/dst/subdir1', exist_ok=True), call('/mock/dst/subdir2', exist_ok=True)) in its call list, found [call('/mock/dst\\.', exist_ok=True), call('/mock/dst\\subdir1', exist_ok=True), call('/mock/dst\\subdir2', exist_ok=True)] instead

```

samwaseda:
> > Speaking of which, I realized a huge advantage of the function approach. You can copy and paste the code snippet to ChatGPT, state the problem, ask it to rewrite the code, and write unit tests accordingly. This should be more streamlined.
> 
> I guess the streamlined solution is Github Co-Pilot - maybe @srmnitc can comment on this.

Ah right. I used Co-Pilot for the actual writing of the snippets (and I didn't particularly like it at that time), but I haven't tried it for doc string or unit tests.

srmnitc:
> > Speaking of which, I realized a huge advantage of the function approach. You can copy and paste the code snippet to ChatGPT, state the problem, ask it to rewrite the code, and write unit tests accordingly. This should be more streamlined.
> 
> I guess the streamlined solution is Github Co-Pilot - maybe @srmnitc can comment on this.

I use CoPilot extension on VScode. This way it already gets the whole code context and therefore can make better suggestions.

pmrv (approved):
Good to see I am not yet replaced by the machine.  Took me a bit to understand the mocking, but it seems neat.

I don't follow the dot business.  Is it because normpath drops `/.` on unix, but not on windows?

samwaseda:
> I don't follow the dot business. Is it because normpath drops `/.` on unix, but not on windows?

Yes. It's just horrible

PR #1555, opened 2024-07-29
[patch] Use `tempfile` instead of creating a directory

samwaseda:
This one seems to cause a problem for the new pack & unpack implementation as it leaves the directory `tmp` after the test.


samwaseda:
Since it does not change the behaviour, I'm gonna merge it as soon as the tests pass.

samwaseda:
NFDI use case failed but I don't think it's related to this PR.

PR #1560, opened 2024-07-29
Change wrap function status when job starts running

samwaseda:
Currently it looks like it goes directly from `submitted` to `finished` which I find is super confusing.


pyiron_base/jobs/flex/pythonfunctioncontainer.py
```diff
@@ -88,6 +88,7 @@ def save(self):
         super().save()
 
     def run_static(self):
+        self.status.running = True
```
jan-janssen, line 91: The additional database request make it slightly slower, but I am fine with having this part added.

jan-janssen (approved):
Looks good to me

PR #1570, opened 2024-07-31
Use tempfile

samwaseda:
With this PR, pyiron uses the functionality of `tempfile` to create a temporary directory instead of taking care of it by itself (thanks @srmnitc for the idea!). This one makes an important step because it is going to be possible to use the same name as the project in `pr.pack`.


pmrv (approved):
lgtm, sounds like it would already substantially easier to use with this.

For the future: In case of `compress=True` we might even skip the copy to the temp dir completely and just add directly to the tarfile.  I would even argue that `compress=False` is not useful.

samwaseda:
> For the future: In case of `compress=True` we might even skip the copy to the temp dir completely and just add directly to the tarfile. I would even argue that `compress=False` is not useful.

Yes! And I even did the change locally, but then realized that it makes it more difficult to read the changes so I dropped it for this PR, but I will do it over the course of next week.

PR #1573, opened 2024-07-31
Set job.status.running again

pmrv:
Both lammps and vasp jobs currently do not set their status to running anymore, likely because of calculate function refactors.  This seemed like the quickest way to fix it.


samwaseda (approved):
This is actually something I have always been thinking about. I haven't done it because I thought there was an important reason for @jan-janssen to not do it. If that's not the case, frankly I think we should do the status updates everywhere in `generic.py` instead of doing it in the child classes.

PR #1574, opened 2024-07-31
Always write input files

pmrv:
This check always fails for submitted jobs and currently prevents lammps jobs to be submitted to the queue.  For some reason vasp jobs still work, but I'll leave that for someone else to figure out.

### DIFF
--- a/pyiron_base/jobs/job/runfunction.py
+++ b/pyiron_base/jobs/job/runfunction.py
@@ -71,16 +71,11 @@ def write_input_files_from_input_dict(input_dict: dict, working_directory: str):
         input_dict (dict): hierarchical input dictionary with files_to_create and files_to_copy.
         working_directory (str): path to the working directory
     """
-    if len(os.listdir(working_directory)) == 0:
-        for file_name, content in input_dict["files_to_create"].items():
-            with open(os.path.join(working_directory, file_name), "w") as f:
-                f.writelines(content)
-        for file_name, source in input_dict["files_to_copy"].items():
-            shutil.copy(source, os.path.join(working_directory, file_name))
-    else:
-        state.logger.info(
-            "The working directory is not empty, assuming that input has already been written."
-        )
+    for file_name, content in input_dict["files_to_create"].items():
+        with open(os.path.join(working_directory, file_name), "w") as f:
+            f.writelines(content)
+    for file_name, source in input_dict["files_to_copy"].items():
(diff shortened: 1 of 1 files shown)


pyiron_base/jobs/job/runfunction.py
```diff
@@ -71,16 +71,11 @@ def write_input_files_from_input_dict(input_dict: dict, working_directory: str):
...
-        for file_name, content in input_dict["files_to_create"].items():
-            with open(os.path.join(working_directory, file_name), "w") as f:
-                f.writelines(content)
-        for file_name, source in input_dict["files_to_copy"].items():
-            shutil.copy(source, os.path.join(working_directory, file_name))
-    else:
-        state.logger.info(
-            "The working directory is not empty, assuming that input has already been written."
-        )
+    for file_name, content in input_dict["files_to_create"].items():
```
samwaseda, line 83:
```suggestion
    if len(os.listdir(working_directory)) > 0:
        state.logger.info(
            "The working directory is not empty, assuming that input has already been written."
        )
    for file_name, content in input_dict["files_to_create"].items():
```

samwaseda: Formatting looks weird, but I wanted to keep the warning.

jan-janssen:
We added this to fix the remote jobs, so we have to be careful we do not break those. https://github.com/pyiron/pyiron_base/pull/1511

pyiron_base/jobs/job/runfunction.py
pmrv: This would be logged for every job sent to the cluster, i.e. most.  That's a bit excessive.

pmrv:
I did also observe that VASP jobs still work somehow similar to that you wrote in the linked issue.  I'm guessing that somewhere vasp still writes the input files outside of the calculate function control flow.

pyiron_base/jobs/job/runfunction.py
samwaseda: For me this PR is a workaround until the cause is found and fixed. It **should** be excessive in this regard.

pmrv: No, the previous PR was a workaround, this one restores core functionality.  I've anyway updated the check to only emit the warning and skip the function body again when any of the files about to be created already exist.

jan-janssen (approved):
Looks good to me

PR #1578, opened 2024-08-02
Server Dataclass

jan-janssen:
Use a Dataclass inside the Server object. This is the same Dataclass which can be directly read from the HDF5 file.


jan-janssen:
The removal of `send_to_db` is taken from https://github.com/pyiron/pyiron_base/pull/1579

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -552,58 +562,28 @@ def view_queues():
...
-        if len(self.additional_arguments) > 0:
-            server_dict["additional_arguments"] = self.additional_arguments
-        if self._gpus is not None:
-            server_dict["gpus"] = self._gpus
-        if self._environment_name is not None:
-            server_dict["conda_environment_name"] = self._environment_name
-        if self._environment_path is not None:
-            server_dict["conda_environment_path"] = self._environment_path
+        self._data.run_mode = self._run_mode.mode
+        server_dict.update(self._data.__dict__)
```
jan-janssen:
```suggestion
        server_dict.update(asdict(self._data))
```
This requires `from dataclasses import asdict` in the beginning of the file.

pyiron_base/jobs/job/extension/server/generic.py
```diff
@@ -12,6 +12,7 @@
 
 from pyiron_snippets.deprecate import deprecate
 
+from pyiron_base.dataclasses.job import Server as ServerDataClass
```
jan-janssen, line 16:
```suggestion
from dataclasses import asdict
from pyiron_base.dataclasses.job import Server as ServerDataClass
```

pyiron_base/dataclasses/job.py
```diff
@@ -0,0 +1,21 @@
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class Server:
```
pmrv, line 16: Ah, we're not packaging `pyiron_io`, yet right? Otherwise would be good to import directly.

pmrv (approved):
lgtm.

pyiron_base/dataclasses/job.py
jan-janssen: Not yet. At the moment I simply copied the dataclass. It is also not clear to me how to handle the versioning of the dataclasses efficiently, for example we might want to have multiple version of the `Server` dataclass in `pyiron_io` for backwards compatibility while `pyiron_base` should use only the latest one. In addition, a number of dataclasses in `pyiron_io` are very specific to atomistic simulation, so we might want to separate those from the general dataclasses for the job management and so on. Still once we have a central repository it should be quick to update the imports here.

pmrv: For the versioning we might set up a simple dictionary in pyiron_io 
```python
@dataclass
class ServerVersion0_9_12:
  ...
@dataclass
class Server:
  ...
ServerClasses = {
  '0.9.2': ServerVersion0_9_12,
  'latest': Server
}
```
then reading code could grab the correct one
```python
import pyiron_base
ServerClasses.get(pyiron_base.__version__, Server)
```
or so, but am totally fine with delaying this.

PR #1579, opened 2024-08-02
Remove send_to_db

jan-janssen:


pmrv (approved):
I don't think this was ever used, so lgtm.

jan-janssen:
> I don't think this was ever used, so lgtm.

I guess Yury was using it initially for http://www.atomistictools.org but as they switched to their own ASE based tools this no longer applies.

PR #1582, opened 2024-08-03
Only create temp_dir when compress=True

samwaseda:
@pmrv pointed it out [here](https://github.com/pyiron/pyiron_base/pull/1570#pullrequestreview-2216272679)


pmrv:
Would be nice to also avoid the copy when compressing, since `tarfile.add` shouldn't care where the contents come from, but can be done later, so lgtm.

jan-janssen (approved):
Looks good to me

PR #1586, opened 2024-08-05
Introduce `pyiron_mini`

jan-janssen:
For users who just want to use pyiron as a workflow manager for any kind of python function or executable, `pyiron_mini` provides a minimal set of functions. You can import it as:
```python
import pyiron_base.mini as pyiron_mini
```
Then you have access to:
* `pyiron_mini.wrap_exe()` to wrap external executables.
* `pyiron_mini.wrap_py()` to wrap python functions.
* `pyiron_mini.cache_list()` to list the currently cached tasks.
* `pyiron_mini.cache_clear()` to clear the cache.
* `pyiron_mini.conda_create_env()` to create new conda environments to execute tasks in.

If `pyiron_mini` is not short enough you can also import it as `import pyiron_base.mini as pm` in that case it is as short as the `pr` we commonly use to reference a `Project` object.

### DIFF
--- a/pyiron_base/mini.py
+++ b/pyiron_base/mini.py
@@ -157,7 +157,12 @@ def cache_clear(silently: bool = False):
     Project(path=".").remove_jobs(recursive=True, silently=silently)
 
 
-def conda_create_env(env_name: str, env_file: str, use_mamba: bool = False, global_installation: bool = True):
+def conda_create_env(
+    env_name: str,
+    env_file: str,
+    use_mamba: bool = False,
+    global_installation: bool = True,
+):
     """
     Create conda environment to execute selected pyiron tasks in a separate conda environment
 
@@ -173,5 +178,5 @@ def conda_create_env(env_name: str, env_file: str, use_mamba: bool = False, glob
         env_name=env_name,
         env_file=env_file,
         global_installation=global_installation,
-        use_mamba=use_mamba
+        use_mamba=use_mamba,
     )
--- a/pyiron_base/mini.py
+++ b/pyiron_base/mini.py
@@ -1,7 +1,7 @@
 from pyiron_base.project.generic import Project
 
 
-def warp_exe(
+def wrap_exe(
     executable_str,
     job_name=None,
     write_input_funct=None,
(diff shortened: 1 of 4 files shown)


samwaseda (changes requested):
I see where you want to go, but we need an overall consensus before we open too many fronts to the public.

jan-janssen:
> I see where you want to go, but we need an overall consensus before we open too many fronts to the public.

I added it to the agenda of today's meeting. Currently, I feel we have too many features in pyiron that makes it hard for us to communicate what pyiron is and hard for others to understand what they can do with pyiron. So I want to start the discussion about the minimal amount of features which make pyiron and then provide a minimalistic interface to those features. The advantage of this restricted interface is in addition that we can change the backend more easily, for example replacing `pyiron_base` with `executorlib` while maintaining the same user facing interface.

samwaseda:
> I added it to the agenda of today's meeting. Currently, I feel we have too many features in pyiron that makes it hard for us to communicate what pyiron is and hard for others to understand what they can do with pyiron. So I want to start the discussion about the minimal amount of features which make pyiron and then provide a minimalistic interface to those features. The advantage of this restricted interface is in addition that we can change the backend more easily, for example replacing `pyiron_base` with `executorlib` while maintaining the same user facing interface.

I see your perspective and I fully agree. It's true that this one is a step towards a functional approach which makes it clear what are the most fundamental pieces to compose a workflow component of. **BUT** I'm against offering it as a new interface to the public, because it is already difficult to explain `ScriptJob` vs. `TemplateJob` vs. `pyiron_workflow` vs. `pyiron_base` to the people who need a workflow manager, because those are often the ones who are not extremely familiar with programming in general. Since we are overall agreeing on the functional approach, we should sketch all the specs in order to be able to say "guys use just this interface and you'll be fine in the foreseeable future".

On the other hand, I'm ok with offering this as a playground feature or pro feature to the developers, although in that case I also don't really see why it should require a new interface.

jan-janssen:
> On the other hand, I'm ok with offering this as a playground feature or pro feature to the developers, although in that case I also don't really see why it should require a new interface.

The changes in this pull request compared to just using `wrap_executable()` and `wrap_python_function()` is that the `Project()` is no longer required. The calculation are always executed in the same directory as the jupyter notebook. We could in principle provide the option to create a sub-folder but I like the idea of just making the Project creation optional.

PR #1589, opened 2024-08-07
[patch] Not create tmp file when copying everything

samwaseda:
@pmrv pointed this case out, and somehow the PR got closed so I open another one.

### DIFF
--- a/pyiron_base/project/archiving/export_archive.py
+++ b/pyiron_base/project/archiving/export_archive.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import tarfile
 import tempfile
+from typing import Optional
 
 from pyiron_base.project.archiving.shared import getdir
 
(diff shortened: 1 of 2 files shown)


pyiron_base/project/archiving/export_archive.py
```diff
@@ -38,7 +38,11 @@ def update_project(project_instance, directory_to_transfer, archive_directory, d
 
 
 def copy_files_to_archive(
-    directory_to_transfer, archive_directory, compress=True, copy_all_files=False, arcname=None
+    directory_to_transfer,
+    archive_directory,
+    compress=True,
+    copy_all_files=False,
+    arcname=None,
```
jan-janssen: It would be great to add type hints - now that we use co-pilot this should be easy

jan-janssen (approved):
Looks good to me

samwaseda:
apparently the new case is not at all covered by unit tests.

samwaseda:
I don't understand why it's failing. I convert it to draft for now.

pmrv:
The error comes from `.git/objects/pack/...` files which are read-only set by git.  So I'm guessing the import tries to pack some git repository (base itself?) by accident and when the export comes around it tries to restore the same packed object files again in place, which causes the error.

samwaseda:
I don't know why but it looks like the changes made in #1595 seem to have solved the problems.

PR #1558, opened 2024-07-29
[patch] Use tempfile instead of creating a directory

samwaseda:
Another try for [this PR](https://github.com/pyiron/pyiron_base/pull/1555)

PR #1590, opened 2024-08-07
not use relpath by default

samwaseda:
This is mainly because windows tests fail when I try to determine `arcname` using `relpath`. I guess this PR will fix the problems that I encountered in [this PR](https://github.com/pyiron/pyiron_base/pull/1589) and [this PR](https://github.com/pyiron/pyiron_base/pull/1558)


jan-janssen (approved):
Looks good to me

pmrv (approved):
lgtm

PR #1592, opened 2024-08-09
Move import tests folder locations

pmrv:
Doesn't fix the tests itself though


samwaseda:
I guess we can close this one because the original problem got somehow resolved

pmrv:
I think it's still nice to not have the tests clutter my local checkout, but I can apply this after all your changes are through, so it doesn't get too messy.

PR #1596, opened 2024-08-12
[patch] Add warning when the same job is imported multiple times with `unpack`

samwaseda:
In the current format the following code would import the same job `test` twice.

```python


pr = Project("TEST")
job = pr.create_job(ToyJob, "test")
job.run()
pr.pack("archive")
pr = Project("import")
pr.unpack("archive")
pr.unpack("archive")
```

And with this PR it does not import the same job and issues a warning.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -7,11 +7,11 @@ DatabaseAccess class deals with accessing the database
 
 import os
 import re
+import warnings
 from datetime import datetime
 from queue import Empty as QueueEmpty
 from queue import SimpleQueue
 from threading import Lock, Thread
-import warnings
 
 import numpy as np
 import pandas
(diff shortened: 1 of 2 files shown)


samwaseda:
I asked review also from @liamhuber because this was also disturbing him before.

pmrv (approved):

pyiron_base/database/generic.py
```diff
@@ -631,6 +648,9 @@ def add_item_dict(self, par_dict):
         """
         if not self._view_mode:
             try:
+                if check_duplicates and self._check_duplidates(par_dict):
+                    warnings.warn(f"Duplicate entry found in database: {par_dict}")
```
liamhuber, line 652: Only nit is that we might want the logger warning instead of `warning.warn`.

liamhuber (approved):
Post-merge review: A nit to think about changing later, but wanted to stamp my approval. Nicely backwards compatible with the flag for the feature 👍

PR #1598, opened 2024-08-13
[patch] Unify `copy_to` and `move_to`

samwaseda:
I realized that `project.copy_to` and `project.move_to` are almost identical so I put them together.


pmrv (approved):
lgtm

PR #1599, opened 2024-08-13
[patch] Append tag when patch, minor or major is added

samwaseda:
I got this from ChatGPT.

### DIFF
--- a/.github/workflows/label-to-title.yml
+++ b/.github/workflows/label-to-title.yml
@@ -40,8 +40,9 @@ jobs:
         if: github.event.action == 'unlabeled' && steps.check_label.outputs.label != ''
         run: |
           TITLE="${{ github.event.pull_request.title || github.event.issue.title }}"
-          NEW_TITLE="${TITLE/${{ steps.check_label.outputs.label }}/}"
-          NEW_TITLE="$(echo $NEW_TITLE | sed 's/  / /g' | sed 's/^ *//;s/ *$//')" # Remove extra spaces
+          LABEL_TO_REMOVE="${{ steps.check_label.outputs.label }}"
+          NEW_TITLE=$(echo "$TITLE" | sed "s/$LABEL_TO_REMOVE //g" | sed "s/$LABEL_TO_REMOVE//g")
+          NEW_TITLE=$(echo "$NEW_TITLE" | sed 's/  / /g' | sed 's/^ *//;s/ *$//') # Remove extra spaces
           if [[ $TITLE != ${NEW_TITLE} ]]; then
             curl -s -X PATCH \
               -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
--- a/.github/workflows/label-to-title.yml
+++ b/.github/workflows/label-to-title.yml
@@ -1,0 +1,38 @@
+name: Label to Title
+
+on:
+  issues:
+    types: [labeled]
+  pull_request:
+    types: [labeled]
+
+jobs:
+  update-title:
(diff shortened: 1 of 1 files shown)


samwaseda:
ok it seems to work

pmrv:
Seems like it go to pyiron/actions, but if it works I'm good with it.

samwaseda:
> Seems like it go to pyiron/actions, but if it works I'm good with it.

Yeah I agree. I leave the decision to @liamhuber

samwaseda:
ok clearly it's not working when I try to remove the tag

samwaseda:
ok the removal is still not working. I don't find it particularly critical because it's rather rare that you would change the level. If you guys see how to do it feel free to change it.

samwaseda:
Finally it worked. My attempts made this PR one of the most colourful PRs in this repository.

liamhuber:
> > Seems like it go to pyiron/actions, but if it works I'm good with it.
> 
> Yeah I agree. I leave the decision to @liamhuber

This was a rollercoaster and then we got here:

> Finally it worked. My attempts made this PR one of the most colourful PRs in this repository.

And then yes, I would very much like this to be in pyiron/actions!

samwaseda:
Now does it mean I should merge it here or I shouldn't?

liamhuber:
> Now does it mean I should merge it here or I shouldn't?

Jan wants base independent of pyiron/actions, so I think this is a "yes and" rather than an "either-or"

samwaseda:
> Jan wants base independent of pyiron/actions, so I think this is a "yes and" rather than an "either-or"

ok are you going to do it? I don't know how things are organised in `pyiron/actions`.

liamhuber:
> > Jan wants base independent of pyiron/actions, so I think this is a "yes and" rather than an "either-or"
> 
> ok are you going to do it? I don't know how things are organised in `pyiron/actions`.

It doesn't have to be you, but someone other than me should do this. It's low-hanging fruit and people not-me need to know how that repo runs.

This looks completely stand-alone to me, which means it's a great candidate for putting as an action (as opposed to a reusable workflow). That means you can even ignore [the directions for how to add stuff to pyiron/actions](https://github.com/pyiron/actions?tab=readme-ov-file#developing-these-actions) because there's no need to re-direct internal tag references. Just:
- Check out a branch
- Make a new folder named after the action and put a `action.yml` file in it
- Write the action:
  - Add the `name` field as usual
  - add a `description` field beneath it
  - This doesn't seem like it needs any input, so everything here between `name` and `steps` can be replaced with `runs:\n\tusing: 'composite'`
  - Copy the `steps` from here into the file
  - [Here](https://github.com/pyiron/actions/blob/main/add-to-python-path/action.yml) is an almost-as-simple example to follow
- Make a PR in some other repo and test your action out with some new `.github/workflow` that leverages the `on:` stuff you _didn't_ copy into the action and leverage the action like:

```
name: Testing Sam's Action

on:
  pull_request:
    types: [labeled, unlabeled]

jobs:
  whatever-you-call-the-job:
    runs-on: ubuntu-latest
    steps:
    - uses: pyiron/actions/whatever-you-called-the-action@whatever-you-called-your-branch
```

We could incorporate it into the `.github/workflows/pr-labeled.yml` reusable workflow later, but you can see in that demo script that just having the meat of the stuff encapsulated in an action already makes it a lot easier to leverage in other places.

This second PR should nicely leverage the new action, at which point you can ping me for review, reference the second PR in the actions PR, then we can merge the actions PR and delete the PR and branch that was being used elsewhere to test it.

PR #1601, opened 2024-08-14
[patch] Use tempfile to open compressed data

samwaseda:


samwaseda:
@liamhuber it looks like as soon as the jobs get cancelled `label-to-title` also gets cancelled. Do you think there's a good way to make it run anyway?

liamhuber:
> @liamhuber it looks like as soon as the jobs get cancelled `label-to-title` also gets cancelled. Do you think there's a good way to make it run anyway?

It's not clear what you mean -- if you manually cancel jobs, it should get cancelled. If some other workflows fail, e.g. unit tests, `label-to-title` should be unaffected because it's in its own separate workflow file. Can you give more detail for the undesirable symptom?

liamhuber:
Ahhh, nevermind, I think I can see by looking at the page history: you pushed, added the label, then the bot pushed over top of you and that literally restarted the entire CI pipeline.

...nope, don't know how to fix that. I can't rule out our ability to power through here, but my gut tells me it's unlikely GitHub gives us that power.

liamhuber:
There is a work-around where we trigger `on: push` and do a check whether the title matches expectations given the label, and if not then update it. But that is a slightly different action than the one you wrote.

samwaseda:
ok at the end of the day it's not so much work to add [patch] in the title myself, so for now I'm gonna keep it like this.

PR #1602, opened 2024-08-16
[minor] Recursive to/from_dict

pmrv:
Adds a create_from_dict function that takes the output of HasDict.to_dict and turns it into a live object, similar to our earlier to_object() method on ProjectHDFio.
Adds an instantiate class method to HasDict to support this, this allows HasDictfromHDF to work and will be useful for dataclasses in the future. HasDict.to_dict now goes over the contents of what is returned from _to_dict and automatically converts any HasDict/HasHDF objects it finds. I haven't used this in downstream code yet to keep the change small, but in principle this will allow GenericJob/DataContainer to stop calling to_dict on their children explicitly and let the generic interface handle it.

The rest of the changes are renaming everything to _from_dict/_to_dict and normalizing the argument name to obj_dict.

This allows to store "abitrary" objects again in the input/outputs of TemplateJobs.

Separate PR for easier review.  I will add extra unit tests, but it worked already in my local notebook tests.

### DIFF
--- a/pyiron_base/interfaces/has_dict.py
+++ b/pyiron_base/interfaces/has_dict.py
@@ -112,6 +112,7 @@ class HasDict(ABC):
             obj_dict (dict): data previously returned from :meth:`.to_dict`
             version (str): version tag written together with the data
         """
+
         def load(inner_dict):
             if not isinstance(inner_dict, dict):
                 return inner_dict
--- a/pyiron_base/interfaces/has_dict.py
+++ b/pyiron_base/interfaces/has_dict.py
@@ -19,7 +19,11 @@ from abc import ABC, abstractmethod
 from typing import Any
 
 from pyiron_base.interfaces.has_hdf import HasHDF
-from pyiron_base.storage.hdfio import DummyHDFio
+from pyiron_base.storage.hdfio import (
+    DummyHDFio,
+    _extract_module_class_name,
+    _import_class,
+)
 
 __author__ = "Jan Janssen"
 __copyright__ = (
@@ -33,11 +37,50 @@ __status__ = "production"
 __date__ = "Dec 20, 2023"
 
 
+def create_from_dict(obj_dict):
+    """
+    Create and restores an object previously written as a dictionary.
+
+    Args:
+        obj_dict (dict): must be the output of HasDict.to_dict()
+
+    Returns:
+        object: restored object
+    """
+    if "TYPE" not in obj_dict:
+        raise ValueError(
+            "invalid obj_dict! must contain type information and be the output of HasDict.to_dict!"
+        )
+    type_field = obj_dict["TYPE"]
+    module_path, class_name = _extract_module_class_name(type_field)
+    class_object = _import_class(module_path, class_name)
+    version = obj_dict.get("VERSION", None)
+    obj = class_object.instantiate(obj_dict, version)
+    obj.from_dict(obj_dict, version)
+    return obj
+
+
 class HasDict(ABC):
     __dict_version__ = "0.1.0"
 
-    @abstractmethod
+    @classmethod
+    def instantiate(cls, obj_dict: dict, version: str = None) -> "Self":
+        return cls()
+
     def from_dict(self, obj_dict: dict, version: str = None):
+        def load(inner_dict):
+            if not isinstance(inner_dict, dict):
+                return inner_dict
+            if not all(
+                k in inner_dict for k in ("NAME", "TYPE", "OBJECT", "DICT_VERSION")
+            ):
+                return {k: load(v) for k, v in inner_dict.items()}
+            return create_from_dict(inner_dict)
+
+        self._from_dict({k: load(v) for k, v in obj_dict.items()}, version)
+
+    @abstractmethod
+    def _from_dict(self, obj_dict: dict, version: str = None):
         pass
 
     @abstractmethod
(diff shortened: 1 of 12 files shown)


pyiron_base/jobs/job/extension/executable.py
```diff
@@ -51,6 +51,8 @@ def __init__(
         else:
             operation_system_nt = os.name == "nt"
 
+        if codename is None:
+            breakpoint()
```
pmrv:
```suggestion
```

pyiron_base/interfaces/has_dict.py
```diff
@@ -114,12 +167,21 @@ class HasDictfromHDF(HasDict, HasHDF):
...
+
+    def _from_dict(self, obj_dict: dict, version: str = None):
         # DummyHDFio(project=None) looks a bit weird, but it was added there
         # only to support saving/loading jobs which already use the HasDict
         # interface
-        hdf = DummyHDFio(None, "/", obj_dict)
-        self.from_hdf(hdf, group_name=self._get_hdf_group_name())
+        group_name = self._get_hdf_group_name()
+        if group_name is not None:
+            hdf = DummyHDFio(None, "/", {self._get_hdf_group_name(): obj_dict})
```
```suggestion
            hdf = DummyHDFio(None, "/", {group_name: obj_dict})
```

pyiron_base/jobs/job/generic.py
```diff
@@ -1174,14 +1174,14 @@ def _to_dict(self):
...
-        if "executable" in job_dict.keys() and job_dict["executable"] is not None:
-            self.executable.from_dict(job_dict["executable"])
-        input_dict = job_dict["input"]
+    def _from_dict(self, obj_dict, version=None):
+        self._type_from_dict(type_dict=obj_dict)
+        if "import_directory" in obj_dict.keys():
+            self._import_directory = obj_dict["import_directory"]
+        self._server = obj_dict["server"]
+        if "executable" in obj_dict.keys() and obj_dict["executable"] is not None:
+            self._executable = obj_dict["executable"]
```
jan-janssen, line 1183: I am a bit surprised about this part, when is the dictionary of the executable converted to the executable object? Is this already happening when reading the executable from the HDF5 file?

pyiron_base/interfaces/has_dict.py
```diff
@@ -33,11 +37,50 @@
...
+
     def from_dict(self, obj_dict: dict, version: str = None):
+        def load(inner_dict):
+            if not isinstance(inner_dict, dict):
+                return inner_dict
+            if not all(
+                k in inner_dict for k in ("NAME", "TYPE", "OBJECT", "DICT_VERSION")
+            ):
+                return {k: load(v) for k, v in inner_dict.items()}
+            return create_from_dict(inner_dict)
```
jan-janssen, line 123: Ok, I understand that the conversion is handled here, still I am wondering if this is what we want. I thought the intention was to get towards using `__getstate__()` and `__setstate__()`. Now with the executable and the server object using datacalsses for data storage internally, I was hoping that we could use a simple dictionary representation there where the class is stored by storing the attached dataclass.

pyiron_base/jobs/job/generic.py
pmrv: In this new setup HasDict.from_dict converts nested objects back from their dictionary form before the implementations HasDict._from_dict is called.  So by the time GenericJob._from_dict runs and looks for the executable Executable._from_dict already ran.

pmrv: In principle the reverse also works, i.e. it's not always necessary anymore to call to_dict on ones children that are returned in the obj_dict.

pyiron_base/interfaces/has_dict.py
```diff
@@ -66,7 +109,17 @@ def _type_to_dict(self):
...
+        type_dict = self._type_to_dict()
+        data_dict = {}
+        child_dict = {}
+        for k, v in self._to_dict().items():
+            if isinstance(v, HasDict):
+                child_dict[k] = v.to_dict()
+            elif isinstance(v, HasHDF):
+                child_dict[k] = HasDictfromHDF.to_dict(v)
+            else:
+                data_dict[k] = v
```
jan-janssen: Just to clarify this, this part currently does not work for the `Atoms` object in `pyiron_atomistics`, as it is neither derived from `HasDict` nor from `HasHDF`, correct? It only works when the `Atoms` object was stored as part of a `DataContainer`, correct?

pyiron_base/jobs/job/generic.py
jan-janssen: I think it makes sense to document this at least in the `HasDict` class, so developers understand how to derive their own classes from `HasDict` and that they only have to implement the `_to_dict()` and `_from_dict()` interface. Finally, I am wondering if it makes sense to overload the `__getstate__()` and `__setstate__()` function in the `HasDict` class. Then we can slowly transition from calling `to_dict()` and `from_dict()` to using the `__getstate__()` and `__setstate__()` interface to reload the pyiron objects. This would simplify the interface for developers who just want to integrate a new code. As long as their classes can be pickled, then we can store their classes in our jobs. If they want to optimise the performance and benefit from the hierarchical nature of the HDF5 file, then they can overload the `__getstate__()` and `__setstate__()` method, for example by attaching a dataclass to their class which they use for data storage or by attaching a data container.

PR #1610, opened 2024-08-20
Catch all exception in pyiron table

pmrv:
Previously we caught ValueError/TypeError, but other exceptions may occur in the table functions and it's somewhat annoying if this kills a long table aggregation because a single job fails.

This also interacts a bit with [executorlib](https://github.com/pyiron/executorlib/issues/389) as the uncaught exception are apparently causing trouble with some executors.


jan-janssen (approved):
Looks good to me

PR #1611, opened 2024-08-20
Backwards compatible loading of server/executable

pmrv:
The issue in https://github.com/pyiron/pyiron_atomistics/pull/1531 is caused by the fact that the dictionaries stored for servers/executables in HDF5 changed slightly and I assumed in #1602 that they would always be automatically loaded by the new HasDict functions.  But that only works when the newer dictionaries are written, the older representation doesn't have the type information that allows HasDict to automatically load them, so `GenericJob._from_dict` assigned the plain dictionaries which then triggered the actual errors seen in CI.

With these changes I can load older Calphy and Vasp jobs again.


samwaseda:
I don’t merge #1617 until this one is merged for the sake of readability. Feel free to merge #1617 when you merge this one.

PR #1620, opened 2024-08-23
Revert changes to DataContainer storage format

pmrv:
PR #1364 changed saved the contents of the container in a data subgroup, but that breaks backwards compat and is not actually necessary for the TemplateJob, so revert it here.


jan-janssen:
It is only the new test that fails:
```
FAIL: test_dict_empty_list (generic.test_datacontainer.TestDataContainer.test_dict_empty_list) [to/from_dict should give back the same list as written.]
HasDict interface should work
  File "/Users/runner/work/pyiron_base/pyiron_base/tests/unit/generic/test_datacontainer.py", line 655, in test_dict_empty_list
    self.assertEqual(self.pl, l)
AssertionError: DataContainer({0: DataContainer({'foo': 'bar'}), 1: 2, 2: 4[110 chars]8])}) != DataContainer({0: {'foo': 'bar'}, 1: 2, 2: 42, 3: {'next': [35 chars] 8]})

```

jan-janssen:
@pmrv While the tests in this pull request fail, this fixes the issue in the notebook tests of `pyiron_atomistics`.

pmrv:
I have a fix for the tests that I'll polish and push on Monday.

PR #1613, opened 2024-08-21
Update integration tests

jan-janssen:
This pull request fixes the issues with the integration tests. In addition, this pull request adds a compatibility test for notebooks in `pyiron_atomistics`.

PR #1637, opened 2024-09-04
Update integration tests and a new one for notebooks

pmrv:
Previously we tried to install the deps of atomistics/contrib on top of the base environment, now we install those dependencies first and then add pyiron_base HEAD without dependencies.
The former breaks when there are incompatible versions, the latter can break if base actually needs features or whole modules from its own environment, but we're betting that that's rarer.

Squashed and rebased from #1613


jan-janssen (approved):
Looks good to me

PR #1639, opened 2024-09-07
Always use absolute path for restart files

### LINKED ISSUE #1633: Jupyter notebooks submitted to the Queuing System using ScriptJob does not write files
This worked until `0.9.1` but started failing afterwards - in addition no restart files are written.

jan-janssen:


niklassiemer:
How does this work with remote submission?

jan-janssen:
For remote we call `write_input()` before we transfer the job to the remote host:
https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/jobs/job/runfunction.py#L504
The `write_input()` function then calls `write_input_files_from_input_dict()`: 
https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/jobs/job/runfunction.py#L63
the `input_dict` for this function is generated using `get_input_parameter_dict()`:
https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/jobs/job/generic.py#L527
which internally copies the restart files `_get_restart_copy_dict()`:
https://github.com/pyiron/pyiron_base/blob/main/pyiron_base/jobs/job/util.py#L221

samwaseda:
It appears to me that it’s `File`’s task to show the absolute path, like `f.abspath()`.

jan-janssen:
> It appears to me that it’s `File`’s task to show the absolute path, like `f.abspath()`.

I like the suggestion and implemented a `abspath()` function for the `File` class.

pyiron_base/jobs/job/generic.py
```diff
@@ -283,7 +283,8 @@ def restart_file_list(self) -> list:
             list: list of files
         """
         self._restart_file_list = [
-            str(f) if isinstance(f, File) else f for f in self._restart_file_list
+            str(f) if isinstance(f, File) else os.path.abspath(f)
```
samwaseda:
```suggestion
            f.abspath() if isinstance(f, File) else os.path.abspath(f)
```

pyiron_base/jobs/job/generic.py
```diff
@@ -322,7 +323,9 @@ def restart_file_dict(self, val: str) -> None:
                 if isinstance(k, File):
                     k = str(k)
                 if isinstance(v, File):
-                    v = str(v)
+                    v = v.abspath()
+                else:
+                    v = os.path.abspath(v)
```
samwaseda: It might be my personal taste but I would extend `File` in a way that you don't need an if-statement here, i.e. I would simply put `v = File(v).abspath()`. You just need to check the type in `__init__` inside `File` to make that happen. I'm dying this also because the same thing can be applied elsewhere in this file.

jan-janssen: Why do you want to initialise the `File` class just to call `abspath()` directly afterwards? This does not make sense to me. If it is a string then we can directly call `os.path.abspath()`.

samwaseda: So that the if statement won't be necessary. But I can live with the if statement as well.

jan-janssen: ok, can I merge it as is?

samwaseda (approved):

PR #1640, opened 2024-09-08
Add Docstrings and type hints with Github Co-pilot

jan-janssen:

PR #1641, opened 2024-09-08
Add more docstrings and type hints

jan-janssen:
This is the second part following https://github.com/pyiron/pyiron_base/pull/1640

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -267,7 +267,7 @@ class DatabaseAccess(IsDatabase):
         job: Optional[str] = None,
         sub_job_name: str = "%",
         element_lst: List[str] = None,
-    ) -> Union[List[dict]]:
+    ) -> List[dict]:
         """
         Internal function to access the database from the project directly.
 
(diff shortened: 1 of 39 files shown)


samwaseda:
I’m not sure if I’d import `argparse` just for the type hinting. Don’t you want to do a string type hinting?

jan-janssen:
> I’m not sure if I’d import `argparse` just for the type hinting. Don’t you want to do a string type hinting?

In `pyiron_base.cli.control` the `argparse` package is required to parse the command line arguments. The other command line interfaces in `pyiron_base.cli.*` are only imported in `pyiron_base.cli.control`, so it does not matter if we already import `argparse` in the command line interfaces of in `pyiron_base.cli.control`, the number of modules imported when using one of the command line utilities remains the same.

pyiron_base/database/generic.py
```diff
@@ -253,19 +255,19 @@ def view_mode(self, value):
...
-        element_lst=None,
-    ):
+        sql_query: str,
+        user: str,
+        project_path: str,
+        recursive: bool,
+        job: Optional[str] = None,
+        sub_job_name: str = "%",
+        element_lst: List[str] = None,
+    ) -> Union[List[dict]]:
```
samwaseda: What does this mean

samwaseda (approved):
Just a nit otherwise it looks good to me

pyiron_base/database/generic.py
samwaseda: I mean `Union[List[dict]]`

jan-janssen: Good point - I fixed it.

samwaseda: Did `copilot` make this mistake?

PR #1649, opened 2024-09-24
Add wrap() function which combines wrap_executable() and wrap_python_function()

jan-janssen:
@samwaseda and I had a discussion about simplifying the user interface of `executorlib` and `pyiron_base`. The primary question is: Should we provide two separate functions, one to wrap external executables which have to be executed in a separate subprocess and one for python functions, or should both be combined in a single function. The argument for combining both is that it simplifies the interface for the user, because from their perspective it makes no difference what they execute, it should behave in the same way for external executables and python functions. The argument for keeping both separated is that they return technically different objects depending on whether the input is a string for the external executable command or a python function.

In this pull request I implement a general `wrap()` function which combines `wrap_executable()` and `wrap_python_function()`. If this pull request gets merged, then we could move `wrap_executable()` and `wrap_python_function()` to hidden functions.

### DIFF
--- a/pyiron_base/project/generic.py
+++ b/pyiron_base/project/generic.py
@@ -11,7 +11,7 @@ import os
 import posixpath
 import shutil
 import stat
-from typing import TYPE_CHECKING, Any, Dict, Generator, List, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, Generator, List, Literal, Optional, Union, Callable
 
 import cloudpickle
 import numpy as np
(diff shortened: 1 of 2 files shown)


pmrv:
I don't have a strong opinion, but purely from a place of API hygiene I feel the cleanest solution would be to offer one method to wrap an arbitrary function and another that constructs a callable object from the parameters used currently in the `wrap_executable` method.  This would avoid all the `TypeError` throws in this PR.  It could be done with fairly little code as well

```python
@dataclass
class ExecutableWrapper:
  exectuable: str
  conda_env: str
  files: list[str]
  ...
  def __call__(self):
    ... # call our existing subprocess wrapper here
```

Users can either do `pr.wrap(ExecutableWrapper(...))` or we still offer a `wrap_executable` that does this under the hood.  Having two disjoint sets of arguments on one function is very confusing and makes refactoring later much more complicated.

jan-janssen:
> Having two disjoint sets of arguments on one function is very confusing and makes refactoring later much more complicated.

Yes, I agree and I am also in favour of using two separate functions. The argument from @samwaseda was that having a single function makes it more easy for users, because it reduces the number of options in auto-completion.

samwaseda (approved):
I'm not so sure about the name `wrap`, but I'm ok with the suggestion otherwise

PR #1663, opened 2024-10-11
Fix support for functions without arguments

jan-janssen:
Example:
```python
import subprocess

def get_python_version():
    return subprocess.check_output("python --version", shell=True, universal_newlines=True)

pr = Project("test")
python_version = pr.wrap_python_function(
    python_function=get_python_version,
    delayed=True,
)
python_version.pull()
>>> 'Python 3.12.5\n'
```

### DIFF
--- a/pyiron_base/jobs/flex/pythonfunctioncontainer.py
+++ b/pyiron_base/jobs/flex/pythonfunctioncontainer.py
@@ -147,7 +147,10 @@ class PythonFunctionContainerJob(PythonTemplateJob):
             with self._get_executor(max_workers=self.server.cores) as exe:
                 output = self._function(**input_dict, executor=exe)
         else:
-            output = self._function(**self.input.to_builtin())
+            if len(self.input.to_builtin()) > 0:
+                output = self._function(**self.input.to_builtin())
+            else:
+                output = self._function()
         self.output.update({"result": output})
         self.to_hdf()
         self.status.finished = True
--- a/pyiron_base/jobs/flex/pythonfunctioncontainer.py
+++ b/pyiron_base/jobs/flex/pythonfunctioncontainer.py
@@ -150,6 +150,10 @@ class PythonFunctionContainerJob(PythonTemplateJob):
             if len(self.input.to_builtin()) > 0:
                 output = self._function(**self.input.to_builtin())
             else:
+                # An empty DataContainer returns a list rather than a dict, so **self.input.to_builtin()
+                # results in "TypeError: argument after ** must be a mapping, not list".
+                # As a workaround, when the input DataContainer is empty the function is called without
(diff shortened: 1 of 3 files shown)


samwaseda (commented):
I generally don't really understand why these changes are needed. The changed lines (not tests) merit short comments.

pyiron_base/project/generic.py
```diff
@@ -666,10 +666,7 @@ def create_function_job(*args, **kwargs):
             )
             job._automatically_rename_on_save_using_input = automatically_rename
             job.python_function = python_function
-            if not args and len(kwargs) == 0:
-                return job
-            else:
-                return job(*args, **kwargs)
```
jan-janssen, line 679: This code lead to a bug of python functions which do not require any arguments not being executed. It is now removed.

jan-janssen:
> I generally don't really understand why these changes are needed. The changed lines (not tests) merit short comments.

@samwaseda the pull request primarily fixes a bug introduced previously.

pyiron_base/jobs/flex/pythonfunctioncontainer.py
```diff
@@ -147,7 +147,14 @@ def run_static(self) -> None:
             with self._get_executor(max_workers=self.server.cores) as exe:
                 output = self._function(**input_dict, executor=exe)
         else:
-            output = self._function(**self.input.to_builtin())
+            if len(self.input.to_builtin()) > 0:
+                output = self._function(**self.input.to_builtin())
```
pmrv:
```suggestion
                output = self._function(**self.input)
```
works even with empty data containers or functions that take no arguments, so there is no need to special case when you leave out the `to_builtin` call.

jan-janssen: I adopted your suggestion and removed the if-statement. I have to admit that I was not aware of the powerful implications we gain from deriving the `DataContainer` class from `MutableMapping`.

pyiron_base/jobs/flex/pythonfunctioncontainer.py
```diff
@@ -147,7 +147,7 @@ def run_static(self) -> None:
             with self._get_executor(max_workers=self.server.cores) as exe:
                 output = self._function(**input_dict, executor=exe)
         else:
-            output = self._function(**self.input.to_builtin())
+            output = self._function(**self.input)
```
jan-janssen, line 150: Calling `to_builtin()` on an empty DataContainer results in an empty list rather than an empty dictionary. As @pmrv pointed out, it is not necessary to call `to_builtin()` as DataContainer is derived from MutableMapping. This is fixed now.

samwaseda (approved):

PR #1664, opened 2024-10-12
Add draw() function to delayed object

jan-janssen:


samwaseda (approved):
I'm fine with this, but in order for the user to have full flexibility, we should also think about returning the SVG object at some point.

jan-janssen:
> we should also think about returning the SVG object at some point.

I real like the idea - I add it to the list and address it in a separate pull request.

PR #1667, opened 2024-10-12
Implement decorators in `pyiron_base`

jan-janssen:
First implementation with decorator arguments:
```python
from pyiron_base import pyiron_job, Project

pr = Project("test")

@pyiron_job(project=pr, cores=2)
def my_function(a, b=8):
    return a + b

c = my_function(a=1, b=2)
d = my_function(a=c, b=3)
d.pull()
>>> 6
```

The disadvantage of this approach is that the decorator requires the `()` even without parameters: 
```python
@pyiron_job()
def my_function(a, b=8):
    return a + b
```

To address this limitation I created a second implementation: 
```python
from pyiron_base import pyiron_job_simple, Project

@pyiron_job_simple
def my_function(a, b=8):
    return a + b

pr = Project("test")
c = my_function(a=1, b=2, project=pr)
d = my_function(a=c, b=3, project=pr, resource_dict={"cores": 2})
d.pull()
>>> 6
```

### DIFF
--- a/pyiron_base/project/decorator.py
+++ b/pyiron_base/project/decorator.py
@@ -53,6 +53,7 @@ def pyiron_job(
         >>> d = my_function_b(a=c, b=3, pyiron_project=pr)
         >>> print(d.pull())
 
+        Output: 6
     """
 
     def get_delayed_object(
--- a/pyiron_base/project/decorator.py
+++ b/pyiron_base/project/decorator.py
@@ -1,7 +1,8 @@
 import inspect
 from typing import Optional
-from pyiron_base.project.generic import Project
+
 from pyiron_base.jobs.job.extension.server.generic import Server
+from pyiron_base.project.generic import Project
 
 
 def pyiron_job(
@@ -39,7 +40,9 @@ def pyiron_job(
                 new_hdf=new_hdf,
             )
             return delayed_job_object
+
         return function
+
     return pyiron_job_function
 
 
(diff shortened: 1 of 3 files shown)


samwaseda:
@jan-janssen I unified `pyiron_job_simple` and `pyiron_job` so that `pyiron_job` works with or without parentheses. If you don't like it you can revert it.

jan-janssen:
> @jan-janssen I unified `pyiron_job_simple` and `pyiron_job` so that `pyiron_job` works with or without parentheses. If you don't like it you can revert it.

Nice catch - I really like the suggestion. From my perspective it is ready to be merged.

samwaseda:
Can you copy and paste your first comment in this PR in the doc string?

jan-janssen:
I updated the pull request based on the feedback from the pyiron meeting today.

jan-janssen:
> Can you copy and paste your first comment in this PR in the doc string?

@samwaseda I added the doc strings as requested.

pyiron_base/project/decorator.py
```diff
@@ -0,0 +1,137 @@
...
+        >>>
+        >>> @pyiron_job(cores=2)
+        >>> def my_function_b(a, b=8):
+        >>>     return a + b
+        >>>
+        >>> pr = Project("test")
+        >>> c = my_function_a(a=1, b=2, pyiron_project=pr)
+        >>> d = my_function_b(a=c, b=3, pyiron_project=pr)
+        >>> print(d.pull())
+
```
samwaseda, line 55:
```suggestion

        Output: 6
```

In this particular case it's super obvious, but I find it generally super important to have an example in the doc string **with** what to expect as an output.

pyiron_base/project/decorator.py
```diff
@@ -0,0 +1,137 @@
...
+        resource_default_dict: dict,
+        **kwargs,
+    ):
+        pyiron_resource_dict.update(
+            {
+                k: v
+                for k, v in resource_default_dict.items()
+                if k not in pyiron_resource_dict.keys()
+            }
+        )
```
samwaseda:
```suggestion
        for k, v in resource_default_dict.items():
            if k not in pyiron_resource_dict:
                pyiron_resource_dict[k] = v
```

Less lines

samwaseda (approved):

jan-janssen:
For future reference, the decorators in `pyiron_base` are inspired by the decorators in `desk` with the added ability to control the computing resources assigned to a given function https://docs.dask.org/en/stable/delayed.html#decorator .

PR #1676, opened 2024-10-22
Always write status to HDF5 file

### LINKED ISSUE #1675: Job status of Lammps jobs is not updated in h5 files
**Summary**

The status of successful lammps jobs (did not test for other codes) is changed to finished in job_table(), but the content of the corresponding h5 file is not updated.
This causes more problems downstream, f.e. remote jobs are are not updated in the database after fetching from the cluster and ovito can not open the h5 file.

**Steps to Reproduce**

```
pr  = Project('status_in_h5')
job = pr.create.job.Lammps('H5Status')
job.structure = pr.create.structure.ase.bulk('Cu', cubic=True)
job.potential = job.list_potentials()[0]
job.calc_md(temperature=50, pressure=[0,0,0,], n_ionic_steps=5)
job.run()

print(job.content['status'])
pr.job_table()
```
** Edit **
After looking through the code with the functional approach I don't really have an Idea where this should be set, i.e. on the individual code interface level or in the general run functions? Also the amount of ways the status is set and the interaction between job.project_hdf5, job._hdf5 and job.status.string is somewhat confusing. Futhermore, I assume the issue  applies to other code interfaces as well.
Maybe you have a plan where it should be done @jan-janssen ?

jan-janssen:


Leimeroth (approved):
works now, thanks

PR #1680, opened 2024-10-28
Use properties for files_to_remove/compress in JobCore.compress

pmrv:
Also fixes a typo in to_dict that overwrote files_to_compress with files_to_remove.


pmrv:
contrib failure is env related, I support contrib is out of date again.

jan-janssen (approved):
Looks good to me

PR #1702, opened 2024-11-26
Fix: Setting Queue ID for remote submission

jan-janssen:
For remote submission when a workstation is connected via SSH to an HPC cluster to submit jobs, a `Locked` error is raised when the job is submitted to the queue on the remote HPC.


pmrv (approved):
lgtm

PR #1735, opened 2025-01-15
[patch] Escape database queries

pmrv:
Queries containing underscores take significantly longer when used in LIKE sql queries, often this is the result of filenames or directory paths containing them and not intended.

If the sqlite backend is not used, no escaping is done.

### DIFF
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -900,10 +900,10 @@ class DatabaseAccess(IsDatabase):
                 self.conn.connection.create_function("like", 2, self.regexp)
 
             result = self.conn.execute(query)
-        row = result.fetchall()
+        results = [row._asdict() for row in result.fetchall()]
         if not self._keep_connection:
             self.conn.close()
-        return [dict(zip(col._mapping.keys(), col._mapping.values())) for col in row]
+        return results
 
     def get_job_status(self, job_id: int) -> Union[str, None]:
         try:
--- a/pyiron_base/database/generic.py
+++ b/pyiron_base/database/generic.py
@@ -313,7 +313,8 @@ class DatabaseAccess(IsDatabase):
 
         """
         if not self._sql_lite:
-            def escape(s, escape_char='\\', special_chars="_%"):
+
+            def escape(s, escape_char="\\", special_chars="_%"):
                 """Insert escape_char in front of special_chars, unless present.
 
                 Handles the cases where s already contains escaped characters,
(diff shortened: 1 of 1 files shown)


pyiron_base/database/generic.py
```diff
@@ -880,10 +900,10 @@ def get_items_dict(
                 self.conn.connection.create_function("like", 2, self.regexp)
 
             result = self.conn.execute(query)
-        row = result.fetchall()
+        results = [row._asdict() for row in result.fetchall()]
```
pmrv, line 903: A slightly faster (~10%) alternative would be
```suggestion
        results = result.mappings().all()
```
but the result is then a list of sqlalchemy objects that behave like dicts, not standard dicts.  One can also feed `result.mappings()` directly into a DataFrame directly, but that would change api more dramatically.

pmrv:
Fixes #1725, but the performance gain is mainly eaten by the need to transform the query result into a list of dicts and then to a DataFrame.

pyiron_base/database/generic.py
pmrv: Just to add some numbers for a query on a project with ~2M jobs:
1. ~10s for the plain sqlalchemy query
2. ~20s for a call to get_items_dict
3. ~30s for a call to job_table

XzzX: database reports around 3s for your `2024%` queries. The rest seems to be sqlalchemy and pyiron.

pyiron_base/database/generic.py
```diff
@@ -312,10 +312,30 @@ def _job_dict(
...
+            def escape(s, escape_char="\\", special_chars="_%"):
+                """Insert escape_char in front of special_chars, unless present.
+
+                Handles the cases where s already contains escaped characters,
+                including the escape character itself.
+
+                Defaults for LIKE in SQL statements."""
+                for c in special_chars:
+                    if c in s:
+                        s = s.replace(escape_char + c, c)
```
jan-janssen, line 326: I understand why you are doing it, but I wonder if this case ever happens. Do you have an example when this is the case?

jan-janssen (approved):
Looks good to me

pyiron_base/database/generic.py
pmrv: No, I had just asked ChatGPT to make sure it handles that case and it seemed to work in my testing.

jan-janssen:
@pmrv Can this be merged? I would like to release a `pyiron_base=0.11` version.

XzzX (commented):
Seems good to me. However, I cannot judge if it is complete.

PR #1744, opened 2025-01-30
Fix copy_to and move_to

niklassiemer:
@pmrv and I found a bug in `pr.move_to`. It only copies jobs for sub-projects and never copies moves the project data. This should address these issues.


niklassiemer:
tests missing

pmrv (approved):
Test would be nice, but code lgtm.

PR #1754, opened 2025-02-13
Preserve resource order when loading queue config

pmrv:
Previously the call to set destroyed the order, when multiple queue.yamls are defined in a users resources.


jan-janssen (approved):
Looks good to me

PR #1763, opened 2025-02-17
[bug]: Fix database issue

jan-janssen:


pmrv:
Yes, this definitely needed. I must have lost this in the earlier PR. But when I tested this in the afternoon with a non modal job, it still crashed. I'm waiting for myjob submission to test it on the queue.

pmrv:
@jan-janssen So the test running on the queue ended up passing, so I think this is all that was needed after all.

PR #1900, opened 2025-09-19
PythonFunctionContainer: Convert dictionary to data container before updating input

jan-janssen:


jan-janssen:
```
FAIL: test_with_internal_executor (flex.test_pythonfunctioncontainer.TestPythonFunctionContainer.test_with_internal_executor)
  File "/home/runner/work/pyiron_base/pyiron_base/tests/unit/flex/test_pythonfunctioncontainer.py", line 152, in test_with_internal_executor
    self.assertEqual(job.output["result"], [6, 8, 10, 12])
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: DataContainer([6, 8, 10, 12]) != [6, 8, 10, 12]

```

pyiron_base/jobs/flex/pythonfunctioncontainer.py
```diff
@@ -65,7 +66,9 @@ def set_input(self, *args, **kwargs):
             **kwargs: Keyword arguments to be passed to the function.
         """
         self.input.update(
-            inspect.signature(self._function).bind(*args, **kwargs).arguments
+            DataContainer(
+                inspect.signature(self._function).bind(*args, **kwargs).arguments
+            ).to_builtin()
```
jan-janssen, line 71: @pmrv This is a quick workaround to check if the object can be stored in a data container. The previous code - just calling `update()` - did not fail when we set a dictionary with integer keys, while this workaround fails with a nice error message. Still it might be nice to have a higher level function to do this check instead.

pmrv: The problem is not quite clear to me, so I don't know if it could be checked before hand.
In principle DataContainer can handle integer keys, but it expects them to be in ascending order, if that was the issue here.

jan-janssen: I guess the difference is between `update(wrap=False)` and `update(wrap=True)` - with the wrap option is raises a `ValueError` when I provide a dict like `{1: 33, 2: 151, 3: 6}`.

PR #1906, opened 2025-09-25
Rename job.core -> job.base in static strings

pmrv:
This was missing in #1862 and broke job.refresh_job_status for job paths.


jan-janssen (approved):

jan-janssen:
The renaming happened in https://github.com/pyiron/pyiron_base/pull/1862 - is there a way to check the type strings automatically as part of the unit tests?

pmrv:
Not sure how one would automatically find the locations where strings are hardcoded to type names in a unit test setup.  Actually covering the static instance checks would help, but is tricky in this case as the functionally underneath requires SLURM.

jan-janssen:
I guess it is better to merge this one, as the resolution of the other challenges takes a little longer.

PR #1907, opened 2025-09-25
GenericJob: compare to class string rather than static string

jan-janssen:


pmrv:
Is this not now equivalent to `isinstance(item, JobCore)`? and doesn't this defeat the purpose of using `static_isinstance` in the first place?  I thought the point had been to avoid importing the classes through out the code base.

jan-janssen:
@pmrv the difference is that `static_isinstance()` compares the class equality while `isinstance()` is also true for derived classes, is this case for example `GenericJob` is `isinstance()` of `JobCore` while `static_isinstance()` would return false. In addition, `JobCore` is already imported in the file. Finally, the second reason to use `static_isinstance()` in other cases is to prevent circular imports.

pmrv:
`static_isinstance` also traverses the MRO to check whether the first argument is a subclass of the given static string and it's necessary to do this because in this code path the `item` is a `JobPath` (that is a subclass of `JobCore`).  What I mean is that we can here just use `isinstance` directly, since as you said `JobCore` is already imported anyway and the normal considerations why we have `static_isinstance` do not apply.

jan-janssen:
I would disagree, if the object is of type Jobcore then no reload is necessary, if it instead is from a derived type like GenericJob or classes derived from GenericJob then it is necessary that the job object is reloaded.

pmrv:
I don't disagree on the intended functionality, but you can quickly check that `static_isinstance(GenericJob, 'pyiron_base.jobs.job.base.JobCore') == True`, so it really is exactly the same as `isinstance(GenericJob, JobCore)`.
I actually think the line (already before this PR) has a bug, as it compares the class of `project` with `JobCore`, which is always false, leading to all objects being reloaded after a copy here.
It does not really matter to me which resolution to take here, if you want to stick with the static check, at least it has to be
```python
obj_type=JobCore.__module__ + '.' + JobCore.__name__
```
because `str()` adds the funny `<class ...>` stuff.

jan-janssen:
@pmrv I finally got around to fix this, can you take another look?

pyiron_base/jobs/job/generic.py
```diff
@@ -663,9 +663,7 @@ def _internal_copy_to(
             return new_job_core, file_project, hdf5_project, reloaded
 
         # Reload object from HDF5 file
-        if not static_isinstance(
-            obj=project.__class__, obj_type="pyiron_base.jobs.job.base.JobCore"
-        ):
+        if str(project.__class__) != str(JobCore):
```
pmrv, line 666: This check still compares the class of *project* with *JobCore*, so it's always `True`, I suppose `new_job_core` was initially intended.  I've followed the code paths a bit and `_job_reload_after_copy` ends up calling `new_job_core.from_hdf`, I'm guessing the simplest change would actually be to drop the if here, always call `_job_reload_after_copy` and then change [the implementation in JobCore](https://github.com/pyiron/pyiron_base/blob/b348ef29cb558740bc03217c092c93375201027a/pyiron_base/jobs/job/base.py#L1030) to `pass`.

jan-janssen: I agree, it remains unclear when the code is actually used. So let's close the pull request right now.

PR #1909, opened 2025-09-29
Update numpy to 2.3.3

jan-janssen:


jan-janssen:
Wait for `pyiron_dataclasses`:
```
      The following packages are incompatible
      ├─ numpy =2.3.3 * is requested and can be installed;
      └─ pyiron_dataclasses =0.0.4 * is not installable because it requires
         └─ numpy >=1.23.5,<=2.3.2 * but there are no viable options
            ├─ numpy [1.23.5|1.24.0|...|2.3.2] conflicts with any installable versions previously reported;
            └─ numpy 2.3.2 would require
               └─ python >=3.14.0rc2,<3.15.0a0 *, which requires
                  └─ _python_rc =* *, which does not exist (perhaps a missing channel).
```

PR #1915, opened 2025-10-07
Delayed Object raise error for failing jobs

jan-janssen:
Example:
```python
from pyiron_base import Project, job

@job
def return_dict(i):
    raise ValueError()

pr = Project("test")
j = return_dict(i=1, pyiron_project=pr)
result = j.pull()
result
```
Previously this example just returned a `JobFuture` object, now it raises the error as expected.

### DIFF
--- a/pyiron_base/project/delayed.py
+++ b/pyiron_base/project/delayed.py
@@ -322,6 +322,8 @@ class DelayedObject:
                 self._job.run()
                 if self._job.status.finished:
                     self._result = self._job.output["result"]
+                elif self._job.status.aborted:
+                    raise self._job.output.error
                 else:
                     return JobFuture(job=self._job)
             else:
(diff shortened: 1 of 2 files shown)


ltalirz:
Great - I find this behavior much more intuitive from a user perspective.

My only remaining question is: under what circumstance should it still return a `JobFuture`? Is this `else` clause required?

jan-janssen:
@ltalirz The job future is used when ever the job is executed in the background, for example when we set `j.server.run_mode.non_modal = True` or when we specify a queuing system by setting `j.server.queue = "my_queue"`. An example for using the `@job` decorator and submitting the resulting job to the queuing system is available in: 
https://github.com/pyiron-dev/pyiron_base_job_flux/blob/main/07_lammps_queue.ipynb

ltalirz:
I see, thanks. That then also answers my second question (how do we implement the API for use with the queueing system).

PR #1954, opened 2026-01-16
Ramp up priority of the sql_connection_string, if set.

### LINKED ISSUE #1904: `PYIRONSQLCONNECTIONSTRING` documentation misleading
The documentation on the `PYIRONSQLCONNECTIONSTRING` env var states

https://github.com/pyiron/pyiron_base/blob/542a8deb4b2b1ffda01339443880c16119489422/docs/installation.md?plain=1#L258-L260

This suggests that you can just set the `PYIRONSQLCONNECTIONSTRING` and be done with it.

However, setting the `PYIRONSQLCONNECTIONSTRING` actually has no effect.
Looking at 
https://github.com/pyiron/pyiron_base/blob/542a8deb4b2b1ffda01339443880c16119489422/pyiron_base/database/manager.py#L83-L98

we see that it is only considered, if we also change the sql type from the default `'SQlite'`  to `'SQLalchemy'`.

While adapting the documentation would be an option, in my opinion the implementation should be changed.
It is absolutely possible to provide sql connection strings for postgres, mysql as well, so if pyiron gives the user the opportunity to provide a connection string, the natural assumption will be that it works in all those cases.

niklassiemer:
closes #1904

### DIFF
--- a/pyiron_base/state/settings.py
+++ b/pyiron_base/state/settings.py
@@ -75,7 +75,7 @@ class Settings(metaclass=Singleton):
         resource_paths / RESOURCE_PATHS / PYIRONRESOURCEPATHS (list):
         project_paths / PROJECT_PATHS / PYIRONPROJECTPATHS (list):
         connection_timeout / CONNECTION_TIMEOUT / PYIRONCONNECTIONTIMEOUT (int):
-        sql_connection_string / CONNECTION / PYIRONSQLCONNECTIONSTRING (str|None): If set, this is used to connect to the 
+        sql_connection_string / CONNECTION / PYIRONSQLCONNECTIONSTRING (str|None): If set, this is used to connect to the
            database, potentially ignoring other configurations related to the connection to the database.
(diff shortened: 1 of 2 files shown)


pyiron_base/state/settings.py
```diff
@@ -298,32 +299,37 @@ def _valid_sql_types(self) -> List[str]:
...
+                            "For sql_type SQLite, the sql_file must not be None"
+                        )
+                    elif os.path.dirname(sql_file) != "":
+                        os.makedirs(os.path.dirname(sql_file), exist_ok=True)
+                elif sql_type == "SQLalchemy":
+                    raise ValueError(
+                        "sql_type was SQLalchemy but did not find a sql_connection_string setting."
+                    )
+            elif sql_type != "SQLalchemy":
+                if any([k in config.keys() for k in ["sql_user_key", "sql_host", "sql_database", "sql_file"]])
```
niklassiemer:
```suggestion
                if any([k in config.keys() for k in ["sql_user_key", "sql_host", "sql_database", "sql_file"]]):
```

PR #2032, opened 2026-06-22
Add more tests

jan-janssen:


jan-janssen:
@copilot Fix the Test that fails on windows

